PR #2, opened 2023-04-03
Add JobSet types

danielvegamyhre:
The purpose of these changes is to define the types and data structures the JobSet API is composed of.


api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +24,54 @@ import (
...
+	SequentialStartup *bool `json:"sequentialStartup,omitempty"`
 }
 
 // JobSetStatus defines the observed state of JobSet
 type JobSetStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// A list of pointers to currently running jobs.
+	// +optional
+	Active []corev1.ObjectReference `json:"active,omitempty"`
```
ahg-g, line 40: This shouldn't be necessary, the jobs created by the jobset should have ownerReference set that points to the jobset instance they belong to; we should also add a selector as well that we automatically populate, but we can leave that for later.

danielvegamyhre: Ok, I removed it.

api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +24,54 @@ import (
...
+	// Template defines the template of the Job that will be created.
+	Template *batchv1.JobTemplateSpec `json:"template,omitempty"`
+	// Network defines the networking options for the job.
+	Network *Network `json:"network"`
+}
+type Network struct {
+	// HeadlessService determines if a headless service will be created for
+	// the associated job. The service name is the same as the JobSet name,
+	// which will also be set in the field spec.jobs[*].template.spec.subdomain.
+	HeadlessService *bool `json:"headlessService"`
```
ahg-g, line 69: lets name it: enableDNSHostnames

api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +23,51 @@ import (
...
-	// Important: Run "make" to regenerate code after modifying this file
+	// Jobs is the group of jobs that will form the set.
+	Jobs []JobTemplate `json:"jobs"`
 
-	// Foo is an example field of JobSet. Edit jobset_types.go to remove/update
-	Foo string `json:"foo,omitempty"`
+	// If sequentialStartup if set, then spec.jobs will be created one at a time
+	// according to their order in the list. A job is created only after
+	// the pods of the previous one are Ready.
+	SequentialStartup *bool `json:"sequentialStartup,omitempty"`
```
ahg-g, line 32: I wouldn't add this yet, too early, the semantics of this can be tricky

api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +23,51 @@ import (
 
 // JobSetSpec defines the desired state of JobSet
 type JobSetSpec struct {
-	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
-	// Important: Run "make" to regenerate code after modifying this file
+	// Jobs is the group of jobs that will form the set.
```
ahg-g, line 26: add the following
```
// +listType=map
// +listMapKey=name
```

so that this is treated as a map with key name

danielvegamyhre:
@ahg-g Thanks for the review, I've addressed your comments and this is ready for another look. I'll squash the commits before submitting as well.

api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +23,48 @@ import (
...
+	// Name is the name of the entry and will be used as a suffix
+	// for the Job name.
+	Name string `json:"name"`
+	// Template defines the template of the Job that will be created.
+	Template *batchv1.JobTemplateSpec `json:"template,omitempty"`
+	// Network defines the networking options for the job.
+	Network *Network `json:"network"`
+}
+type Network struct {
+	// HeadlessService determines if a headless service will be created for
```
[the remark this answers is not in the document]
danielvegamyhre: Done, sorry I forgot to update the comment before, I'll try to be less hasty.

api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +23,46 @@ import (
...
 	metav1.TypeMeta `json:",inline"`
 	metav1.ListMeta `json:"metadata,omitempty"`
 	Items           []JobSet `json:"items"`
 }
+type JobTemplate struct {
+	// Name is the name of the entry and will be used as a suffix
+	// for the Job name.
+	Name string `json:"name"`
+	// Template defines the template of the Job that will be created.
+	Template *batchv1.JobTemplateSpec `json:"template,omitempty"`
```
ahg-g, line 58: sorry, I missed it, this doesn't need to be a reference.

api/v1alpha1/jobset_types.go
```diff
@@ -22,39 +23,46 @@ import (
...
+	// Network defines the networking options for the job.
+	Network *Network `json:"network"`
+}
+type Network struct {
+	// EnableDNSHostnames allows pods to be reached via their hostnames.
+	EnableDNSHostnames *bool `json:"enableDNSHostnames"`
```
ahg-g, line 64: ditto, this doesn't need to be a reference.

PR #8, opened 2023-04-07
Add integration tests

danielvegamyhre:
Integration test cases:

1. **JobSet with pod DNS hostnames disabled - success scenario**: create JobSet with pod DNS hostnames disabled. Ensure all jobs are created, and ensure JobSet status is `JobSetCompleted` after all jobs have completed.
2. **JobSet with pod DNS hostnames disabled - failure scenario**: create JobSet with pod DNS hostnames disabled. Ensure all jobs are created, and ensure JobSet status is `JobSetFailed` after 1 job fails.
3. **JobSet with pod DNS hostnames enabled - success scenario**: create JobSet with pod DNS hostnames enabled. Ensure all jobs are created, ensure all headless services are created, then ensure JobSet status is `JobSetCompleted` after all jobs have completed.

**Notes**:
- I designed the test cases to each run in their own namespace, in order to avoid weird race conditions I was running into where resources from a previous test case were somehow interfering with the subsequent test case, despite successfully deleting the resources in the AfterEach() block.


test/integration/jobset_controller_test.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	for _, c := range fetchedJS.Status.Conditions {
+		if c.Type == string(condition) {
+			return true, nil
+		}
+	}
+	return false, nil
+}
+
+func constructJobSet(args *jobSetArgs) *jobset.JobSet {
```
ahg-g, line 220: lets create a testing pkg that includes wrappers; see example here: https://github.com/kubernetes-sigs/kueue/blob/main/pkg/util/testing/wrappers.go

test/integration/jobset_controller_test.go
```diff
@@ -0,0 +1,264 @@
...
+			js := constructJobSet(&jobSetArgs{
+				name:      "js-simple",
+				namespace: ns.Name,
+				numJobs:   3,
+			})
+			Expect(k8sClient.Create(ctx, js)).Should(Succeed())
+
+			// We'll need to retry getting this newly created JobSet, given that creation may not immediately happen.
+			By("checking JobSet was created successfully")
+			Eventually(jobSetExists, timeout, interval).WithArguments(js).Should(BeTrue())
```
ahg-g, line 88:
```suggestion
			Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).Should(Succeed())
```

and remove the jobSetExists func

ahg-g: ditto other places below

test/integration/jobset_controller_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+func jobSetExists(js *jobset.JobSet) bool {
+	err := k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{})
+	if err != nil {
+		return false
+	}
+	return true
+}
+
+func assertJobSetStatus(js *jobset.JobSet, condition jobset.JobSetConditionType) (bool, error) {
```
ahg-g, line 207:
```suggestion
func checkJobSetCondition(js *jobset.JobSet, condition jobset.JobSetConditionType) (bool, error) {
```

"assert" is usually tied to the semantics of  "failing when the condition is not met"; so "check" feels more appropriate here.

test/integration/jobset_controller_test.go
danielvegamyhre: Ok I created the testing package with wrappers for JobSet and ReplicatedJob, under /pkg/util/testing.

danielvegamyhre:
@ahg-g Thanks for the review, I finished addressing your comments and this is ready for another look.

pkg/util/testing/core.go
```diff
@@ -0,0 +1,47 @@
...
+
+func IndexedJob(name, ns string) batchv1.JobTemplateSpec {
+	indexedCompletionMode := batchv1.IndexedCompletion
+	return batchv1.JobTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: ns,
+		},
+		Spec: batchv1.JobSpec{
+			CompletionMode: &indexedCompletionMode,
```
ahg-g, line 32: use the "k8s.io/utils/pointer" pkg instead

pkg/util/testing/wrappers.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// AddReplicatedJobs adds `count` number of copies of a ReplicatedJob to the JobSet.
+// Suffixes are added to the names to distinguish them.
+// Used for easily adding multiple similar ReplicatedJobs to a JobSet for testing.
+func (j *JobSetWrapper) AddReplicatedJobs(job jobset.ReplicatedJob, count int) *JobSetWrapper {
+	for i := 0; i < count; i++ {
+		// Add suffixes to distinguish jobs.
+		job.Name += fmt.Sprintf("-%d", i)
+		job.Template.Name += fmt.Sprintf("-%d", i)
```
ahg-g, line 64: This should be set by the JobSet controller

test/integration/jobset_controller_test.go
```diff
@@ -0,0 +1,208 @@
...
+	ginkgo.When("a jobset is created without DNS hostnames enabled", func() {
+		ginkgo.It("should create all jobs and complete successfully once all jobs are completed", func() {
+			ginkgo.By("creating a new JobSet")
+			ctx := context.Background()
+			jsw := testing.MakeJobSet("js-succeed", ns.Name)
+
+			// Add 3 replicated jobs.
+			rj := testing.MakeReplicatedJob("test-job").SetJob(testing.IndexedJob("test-job", ns.Name)).Obj()
+			jsw.AddReplicatedJobs(rj, 3)
+			js := jsw.Obj()
```
ahg-g, line 76: The pattern we have been using is something like this:

```
js := testing.MakeJobSet("js-fail", ns.Name).
  AddReplicatedJobs(testing.MakeReplicatedJob("test-job").
                                SetJob(testing.IndexedJob("test-job", ns.Name)
).Obj(), 3).Obj()
```

test/integration/jobset_controller_test.go
```diff
@@ -0,0 +1,208 @@
...
+			ginkgo.By("creating a new JobSet")
+			ctx := context.Background()
+
+			// Construct JobSet.
+			jsw := testing.MakeJobSet("js-simple", ns.Name)
+
+			// Add 3 replicated jobs.
+			rj := testing.MakeReplicatedJob("test-job").SetJob(testing.IndexedJob("test-job", ns.Name)).SetEnableDNSHostnames(true).Obj()
+			jsw.AddReplicatedJobs(rj, 3)
+			js := jsw.Obj()
```
ahg-g, line 155: ditto:

```
js := testing.MakeJobSet("js-fail", ns.Name).
  AddReplicatedJobs(testing.MakeReplicatedJob("test-job").
                                      SetJob(testing.IndexedJob("test-job", ns.Name)
).Obj(), 3).Obj()
```

test/integration/jobset_controller_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+		ginkgo.It("should create all jobs and fail if any job fails", func() {
+			ginkgo.By("creating a new JobSet")
+			jsw := testing.MakeJobSet("js-fail", ns.Name)
+
+			// Add 3 replicated jobs.
+			rj := testing.MakeReplicatedJob("test-job").SetJob(testing.IndexedJob("test-job", ns.Name)).Obj()
+			jsw.AddReplicatedJobs(rj, 3)
+			js := jsw.Obj()
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
```
ahg-g, line 115: ditto

```
js := testing.MakeJobSet("js-fail", ns.Name).
  AddReplicatedJobs(testing.MakeReplicatedJob("test-job").
                                      SetJob(testing.IndexedJob("test-job", ns.Name)
).Obj(), 3).Obj()
```

test/integration/suite_test.go
```diff
@@ -0,0 +1,114 @@
...
+
+var _ = AfterSuite(func() {
+	// https://github.com/kubernetes-sigs/controller-runtime/issues/1571
+	cancel()
+	By("tearing down the test environment,but I do nothing here.")
+	err := testEnv.Stop()
+	// Set 4 with random
+	if err != nil {
+		time.Sleep(4 * time.Second)
+	}
```
ahg-g, line 111: why this?

pkg/util/testing/wrappers.go
danielvegamyhre: can you clarify? the JobSet controller currently sets the job names to `<jobset name>-<job name>`, so if the job names are exactly the same here (no suffixes), it will result in an error creating the 2nd job since it already exists.

test/integration/suite_test.go
danielvegamyhre: I got this snippet [here](https://github.com/kubernetes-sigs/controller-runtime/issues/1571#issuecomment-1437304502) for a workaround to the linked issue (timeout waiting for kube-apiserver to stop). However, I just did some testing and it seems that at least in my local environment, only the context cancellation is necessary, and we don't need to try `testenv.Stop()` more than once, so I've removed the sleep.

pkg/util/testing/core.go
danielvegamyhre: I wasn't able to find an applicable function in the pointer package, but looking at examples it seems usually either the current method I have is used, or a `completionModePtr` helper function is implemented to make it a bit clean, so I've updated it to use this approach. Let me know if you have any thoughts.

pkg/util/testing/wrappers.go
ahg-g: I am referring to `job.Template.Name`, this field is set by the jobset controller when the job is created. 

I just noticed that this is wrong: https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L309

this should be `fmt.Sprintf("%s-%s", js.Name, jobTemplate.Name)`

danielvegamyhre: Ah I see, that makes more sense. I included that change in the latest commit.

pkg/util/testing/core.go
ahg-g: sg

PR #14, opened 2023-04-10
Use contextual logging

danielvegamyhre:
Use contextual logging in JobSet controller.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -250,16 +250,14 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
 
 		// Set controller owner reference for garbage collection and reconcilation.
 		if err := ctrl.SetControllerReference(js, &headlessSvc, r.Scheme); err != nil {
-			log.Error(err, "setting controller owner reference for headless service", "service", klog.KObj(&headlessSvc))
 			return err
 		}
 
 		// Create headless service.
 		if err := r.Create(ctx, &headlessSvc); err != nil {
-			log.Error(err, "creating headless service", "service", klog.KObj(&headlessSvc))
 			return err
 		}
-		log.V(2).Info("successfully created headless service", "service", klog.KObj(&headlessSvc))
+		log.V(2).Info("successfully created headless service", "service", klog.KObj(job))
 	}
 	return nil
 }
@@ -290,9 +288,6 @@ func (r *JobSetReconciler) shouldCreateJob(ctx context.Context, job *batchv1.Job
 // updateStatus updates the status of a JobSet by appending
 // the new condition to jobset.status.conditions.
 func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) error {
-	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js))
-	ctx = ctrl.LoggerInto(ctx, log)
-
 	js.Status.Conditions = append(js.Status.Conditions, condition)
 	if err := r.Status().Update(ctx, js); err != nil {
 		return err
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -22,7 +22,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/klog"
+	"k8s.io/klog/v2"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -207,18 +210,18 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 
 		// Create headless service if specified for this job.
 		if rjob.Network.EnableDNSHostnames != nil && *rjob.Network.EnableDNSHostnames {
-			klog.Infof("creating headless service: %s", job.Name)
+			log.Info("creating headless service", "service", job.Name)
```
ahg-g, line 213: v(2)

pkg/controllers/jobset_controller.go
```diff
@@ -207,18 +210,18 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
...
-				klog.Errorf("error creating headless service: %v", err)
+				log.Error(err, "error creating headless service", "service", job.Name)
 				return err
 			}
 		}
 
 		// Create the job.
 		// TODO: Deal with the case where the job exists but is not owned by the jobset.
-		klog.Infof("creating job %s", job.Name)
+		log.Info("creating job", "job", job.Name)
```
ahg-g, line 222: Lets use v(2) for the logs that we want to show in a production run, and something higher for debugging

pkg/controllers/jobset_controller.go
```diff
@@ -86,7 +87,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			Reason:             "FailedJobs",
 			Message:            "jobset failed due to one or more failed jobs",
 		}); err != nil {
-			klog.Errorf("error updating jobset status: %s", err)
+			log.Error(err, "error updating jobset status", "jobset", js.Name)
```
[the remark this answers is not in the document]
ahg-g: all other places below as well

pkg/controllers/jobset_controller.go
```diff
@@ -73,7 +74,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// Get Jobs owned by JobSet.
 	ownedJobs, err := r.getChildJobs(ctx, &js, req)
 	if err != nil {
-		klog.Errorf("error getting jobs owned by jobset %s: %v", js.Name, err)
+		log.Error(err, "error getting jobs owned by jobset", "jobset", js.Name)
```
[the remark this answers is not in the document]
ahg-g: ditto other places below

pkg/controllers/jobset_controller.go
```diff
@@ -192,10 +192,13 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet,
...
 
 func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js))
+	ctx = ctrl.LoggerInto(ctx, log)
+
 	for _, rjob := range js.Spec.Jobs {
 		job, err := r.constructJobFromTemplate(js, &rjob)
 		if err != nil {
-			klog.Errorf("error constructing job from template %s: %v", rjob.Name, err)
+			log.Error(err, "error constructing job from template", "job", rjob.Name)
```
ahg-g, line 201: the caller is logging the error as well, lets not do it again here.

pkg/controllers/jobset_controller.go
ahg-g: I would also move it after the call and say: "successfully created headless service"

danielvegamyhre:
@ahg-g Thanks for the review, I addressed your comments and this is ready for another look.

pkg/controllers/jobset_controller.go
```diff
@@ -287,9 +290,11 @@ func (r *JobSetReconciler) shouldCreateJob(ctx context.Context, job *batchv1.Job
 // updateStatus updates the status of a JobSet by appending
 // the new condition to jobset.status.conditions.
 func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) error {
+	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js))
+	ctx = ctrl.LoggerInto(ctx, log)
+
```
ahg-g, line 295: remove

pkg/controllers/jobset_controller.go
```diff
@@ -248,15 +250,16 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
 
 		// Set controller owner reference for garbage collection and reconcilation.
 		if err := ctrl.SetControllerReference(js, &headlessSvc, r.Scheme); err != nil {
-			klog.Errorf("error setting controller owner reference for headless service: %s", headlessSvc.Name)
+			log.Error(err, "setting controller owner reference for headless service", "service", klog.KObj(&headlessSvc))
```
ahg-g, line 253: if the error is being returned, then we expect the caller to log it, which is the case here, I think we should remove this.

pkg/controllers/jobset_controller.go
```diff
@@ -248,15 +250,16 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
 			return err
 		}
 
 		// Create headless service.
 		if err := r.Create(ctx, &headlessSvc); err != nil {
-			klog.Errorf("error creating headless service: %s", headlessSvc.Name)
+			log.Error(err, "creating headless service", "service", klog.KObj(&headlessSvc))
```
ahg-g, line 259: ditto

pkg/controllers/jobset_controller.go
danielvegamyhre: Make sense, I removed all logging in this function except the "successfully created headless service" log line which needs access to the `headlessSvc` object to use the object reference in the log function.

danielvegamyhre:
@ahg-g I've addressed the most recent comments and this is ready for another look

Issue #11, opened 2023-04-09
Support creating multiple instances of the same job template


ahg-g:
The jobs should be indexed, and the job index should be added as a label and annotation.

PR #15, opened 2023-04-10
Add support for creating multiple instances of the same job template

danielvegamyhre:
Fixes https://github.com/kubernetes-sigs/jobset/issues/11

**Reviewer Notes**:
- Job names are now generated using the format `<jobset name>-<job name>-<job index>`
- Job index label and job index annotation are both the same: `jobset.sigs.k8s.io/job-index` should they be distinct?
- Currently, when `enableDNSHostnames` is true, we create 1 headless service per job. So for a ReplicatedJob with 100 replicas, we will create 100 headless services. Can we simplify this by adding a label with the JobSet name to all its jobs, then simply creating 1 headless service that selects that label?


pkg/controllers/jobset_controller.go
```diff
@@ -142,28 +143,54 @@ func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-	// job name (a headless service with same name as job will be created later).
-	if rjob.Network.EnableDNSHostnames != nil && *rjob.Network.EnableDNSHostnames {
-		job.Spec.Template.Spec.Subdomain = job.Name
+	// Defaulting and validation.
+	// TODO: Do defaulting and validation in webhook instead of here (https://github.com/kubernetes-sigs/jobset/issues/6)
+	replicas := 1
+	if rjob.Replicas != nil && *rjob.Replicas > 0 {
+		replicas = *rjob.Replicas
+	}
+	labels := rjob.Template.Labels
```
ahg-g, line 155: We should always copy the labels and annotations to avoid updating the template itself. 

Also, In case we do the job creates in parallel in the future, we should do the copy inside the loop below so that the maps are shared between all instances.

test/integration/jobset_controller_test.go
```diff
@@ -187,6 +189,64 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			gomega.Eventually(checkJobSetStatus, timeout, interval).WithArguments(js, jobset.JobSetCompleted).Should(gomega.Equal(true))
 		})
 	})
+
+	ginkgo.When("a jobset is created with 2 replicated jobs with 3 replicas each and pod DNS hostnames enabled", func() {
```
ahg-g, line 193: Lets add a test for failure as well.

pkg/controllers/jobset_controller.go
```diff
@@ -196,30 +223,32 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 	ctx = ctrl.LoggerInto(ctx, log)
 
 	for _, rjob := range js.Spec.Jobs {
-		job, err := r.constructJobFromTemplate(js, &rjob)
+		jobs, err := r.constructJobsFromTemplate(js, &rjob)
```
ahg-g, line 226: I find this a bit strange that we construct the full object of all jobs and then decide which ones to persist on the api server. I think we should have logic that decides which jobs to create (only job names is sufficient), and then for each we construct the object and create it on the api-server.

pkg/controllers/jobset_controller.go
ahg-g: oh, are not* shared

api/v1alpha1/jobset_types.go
```diff
@@ -18,7 +18,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-const JobLabelKey string = "jobset.sigs.k8s.io/job-name"
+const (
+	JobNameLabel  string = "jobset.sigs.k8s.io/job-name"
```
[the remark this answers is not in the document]
danielvegamyhre: Oh I forgot we switched back to using the auto-populated `job-name` label. I removed this for now.

pkg/controllers/jobset_controller.go
danielvegamyhre: Makes sense, done.

pkg/controllers/jobset_controller.go
danielvegamyhre: Makes sense, I moved the `shouldCreateJob` call into `constructJobsFromTemplate` so we don't unnecessarily construct jobs we aren't going to create.

test/integration/jobset_controller_test.go
danielvegamyhre: Added.

danielvegamyhre:
@ahg-g Thanks for the review, I've addressed your comments and this is ready for another look.

ahg-g:
> Currently, when enableDNSHostnames is true, we create 1 headless service per job. So for a ReplicatedJob with 100 replicas, we will create 100 headless services. Can we simplify this by adding a label with the JobSet name to all its jobs, then simply creating 1 headless service that selects that label?

The part that will be part of our api is the fqdns name of the pod, which will depend on whether or not we create a service per job... we need to decide on whether the fully qualified name will be `service-<jobIndex>.jobset-job-<jobIndex>-<podIndex>` or `service.jobset-job-<jobIndex>-<podIndex>`, I guess we can punt this decision to when we graduate to beta, so for now lets create a service per job.

pkg/controllers/jobset_controller.go
```diff
@@ -142,28 +143,62 @@ func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	// Construct jobs.
+	for i := 0; i < replicas; i++ {
+		// Check if we need to create this job. If not, skip it.
+		jobName := genJobName(js, rjob, i)
+		if create := r.shouldCreateJob(jobName, ownedJobs); !create {
+			continue
+		}
+
+		// Copy labels/annotations to avoid modifying the template itself.
+		labels := rjob.Template.Labels
```
ahg-g, line 165: This is just copying the reference, we need to do a proper copy:

```
labels = make(map[string]string)
for k, v := rjob.Template.Labels {
  labels[k] = v
}
```

danielvegamyhre: Woops, can't believe I overlooked that. Fixed!

api/v1alpha1/jobset_types.go
```diff
@@ -71,6 +71,8 @@ type ReplicatedJob struct {
 	Template batchv1.JobTemplateSpec `json:"template,omitempty"`
 	// Network defines the networking options for the job.
 	Network *Network `json:"network"`
+	// Replicas is the number of jobs that will be created from this ReplicatedJob's template.
+	Replicas *int `json:"replicas,omitempty"`
 }
 type Network struct {
 	// EnableDNSHostnames allows pods to be reached via their hostnames.
```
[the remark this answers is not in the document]
danielvegamyhre: Done. I also added a short line of documentation for "EnableDNSHostnames" explaining pods will be reachable via their fully qualified hostname.

pkg/controllers/jobset_controller.go
```diff
@@ -196,30 +231,26 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
...
+				if err := r.createHeadlessSvcIfNotExist(ctx, js, job); err != nil {
+					return err
+				}
+			}
 
-		// Create headless service if specified for this job.
-		if rjob.Network.EnableDNSHostnames != nil && *rjob.Network.EnableDNSHostnames {
-			if err := r.createHeadlessSvcIfNotExist(ctx, js, job); err != nil {
+			// Create the job.
+			// TODO: Deal with the case where the job exists but is not owned by the jobset.
```
ahg-g, line 248:
```suggestion
			// TODO(#18): Deal with the case where the job exists but is not owned by the jobset.
```

api/v1alpha1/jobset_types.go
```diff
@@ -71,9 +71,15 @@ type ReplicatedJob struct {
...
 	Network *Network `json:"network"`
+	// Replicas is the number of jobs that will be created from this ReplicatedJob's template.
+	// Pod hostnames for pods created by these jobs will be in the format:
+	// <jobsetName>-<jobName>-<jobIndex>-<podIndex>
+	Replicas *int `json:"replicas,omitempty"`
 }
 type Network struct {
 	// EnableDNSHostnames allows pods to be reached via their hostnames.
+	// Pods will be reachable using the fully qualified pod hostname, which is in
+	// the format: <jobsetName>-<jobName>-<jobIndex>-<podIndex>.<jobName>
```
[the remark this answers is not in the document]
danielvegamyhre: The subdomain will include the JobSet name as well, since it is the same as the job name, which we set to `<jobSet.name>-<spec.replicatedJob.name>-<job-index>` as you noted. 

So the final fully qualified pod hostname is: `<jobSet.name>-<spec.replicatedJob.name>-<job-index>-<pod-index>.<jobSet.name>-<spec.replicatedJob.name>-<job-index>`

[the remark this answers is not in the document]
ahg-g: Yes, please add that

api/v1alpha1/jobset_types.go
```diff
@@ -71,9 +71,15 @@ type ReplicatedJob struct {
...
 	// Network defines the networking options for the job.
 	Network *Network `json:"network"`
+	// Replicas is the number of jobs that will be created from this ReplicatedJob's template.
+	// Pod hostnames for pods created by these jobs will be in the format:
+	// <jobsetName>-<jobName>-<jobIndex>-<podIndex>
+	Replicas *int `json:"replicas,omitempty"`
 }
 type Network struct {
 	// EnableDNSHostnames allows pods to be reached via their hostnames.
+	// Pods will be reachable using the fully qualified pod hostname, which is in
```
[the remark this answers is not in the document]
danielvegamyhre: I'm thinking we should also mark the JobSet as failed if it fails to create a headless service for a job since it is not in indexed completion mode. Do you agree? 

I implemented this check in the reconciler, and added a test case for it as well, but haven't pushed the changes yet. Do you mind if we include this new validation and test case in a separate PR?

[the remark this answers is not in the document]
ahg-g: sounds good

pkg/controllers/jobset_controller.go
```diff
@@ -196,30 +231,29 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
...
 
-		// Check if we need to create this job.
-		// If not, skip this job and continue iterating to the next job.
-		if create := r.shouldCreateJob(ctx, job, ownedJobs); !create {
-			continue
-		}
+		for _, job := range jobs {
+			// Create headless service if specified for this job.
+			if dnsHostnamesEnabled(&rjob) {
+				if !isIndexedJob(job) {
```
ahg-g, line 242: I see you already added the check here, I guess the integration test is what is left, right?

PR #16, opened 2023-04-10
add dependabot

### LINKED ISSUE #13: Automate dependencies update
Can be done using dependabot https://github.com/kubernetes-sigs/kueue/issues/435

kannon92:
Fixes #13 

Add dependabot and follow same configuration as Kueue.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+version: 2
+updates:
+  # Maintain dependencies for go
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    labels:
+      - "ok-to-test"
+    ignore:
+      # Ignore major and minor versions for dependencies updates
+      # Allow patches and security updates.
+      - dependency-name: k8s.io/*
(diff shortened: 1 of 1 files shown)


ahg-g:
Thanks Kevin, can you pls explain a bit what will that do? which dependencies will be updated and how the update will be triggered?

kannon92:
So this is the exact same config that we are using in Kueue for updating dependencies.  

```
version: 2
updates:

  # Maintain dependencies for go
  - package-ecosystem: "gomod" // Look at gomodules
    directory: "/"
    schedule:
      interval: "weekly" // Check on a weekly cadence and see if anything needs an update
    labels:
      - "ok-to-test" // Adds the "ok-to-test" label to every PR
    ignore:
      # Ignore major and minor versions for dependencies updates
      # Allow patches and security updates.
// Generally only update major.minor upgrades for K8 dependencies.
      - dependency-name: k8s.io/*
        update-types: ["version-update:semver-major", "version-update:semver-minor"]
```

I figured the Kueue config you all are happy with as I don't believe it has changed much since I added it.

PR #19, opened 2023-04-11
Validation and defaulting using kubebuilder markers

danielvegamyhre:
- Add CEL validation checking that if `replicatedJob.network.enableDNSHostnames` is true, then the job must be in indexed completion mode.
- Add defaulting to `replicatedJob.replicas` to 1
- Add integration test for the validation
- Refactor integration tests to use wrapper pattern for Job objects

**Reviewer Notes**:
- This PR includes the CRD yaml changes to the comment/documentation on `EnableDNSHostnames` because after implementing https://github.com/kubernetes-sigs/jobset/pull/15 I did not run `make` to generate the CRD after modifying the API.


test/integration/jobset_controller_test.go
```diff
@@ -188,6 +189,30 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			// Check JobSet has completed.
 			gomega.Eventually(checkJobSetStatus, timeout, interval).WithArguments(js, jobset.JobSetCompleted).Should(gomega.Equal(true))
 		})
+
+		ginkgo.It("should throw an error if job completion mode is not indexed", func() {
```
ahg-g, line 193: Actually once we have validation, this test will need to change. 

I think we should add this test once we add the validation on the field; ideally using CEL:

```
// +kubebuilder:validation:XValidation:rule
```

config/crd/bases/batch.x-k8s.io_jobsets.yaml
```diff
@@ -48,13 +48,16 @@ spec:
                         job.
                       properties:
                         enableDNSHostnames:
-                          description: EnableDNSHostnames allows pods to be reached
-                            via their hostnames.
+                          description: 'EnableDNSHostnames allows pods to be reached
```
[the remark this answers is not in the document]
danielvegamyhre: Yes, see my notes in the PR description

test/integration/jobset_controller_test.go
danielvegamyhre: Added CEL validation rule and updated integration test case.

pkg/controllers/jobset_controller.go
```diff
@@ -239,9 +239,9 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 		for _, job := range jobs {
 			// Create headless service if specified for this job.
 			if dnsHostnamesEnabled(&rjob) {
-				if !isIndexedJob(job) {
-					return fmt.Errorf("EnableDNSHostnames requires job completion mode to be indexed")
-				}
+				// if !isIndexedJob(job) {
+				// 	return fmt.Errorf("EnableDNSHostnames requires job completion mode to be indexed")
+				// }
```
ahg-g, line 244: pls remove this completely, no need to do this check again.

ahg-g (commented):
Can you also add defaulting to Replicas as well using // +kubebuilder:default


The PR name could be something like add "validation and defaulting using kubebuilder markers"

pkg/util/testing/core.go
```diff
@@ -46,6 +46,29 @@ func IndexedJob(name, ns string) batchv1.JobTemplateSpec {
 	}
 }
 
+// Job constructs a simple job in non-indexed completion mode.
+func Job(name, ns string) batchv1.JobTemplateSpec {
```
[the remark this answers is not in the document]
danielvegamyhre: Added wrapper for Job and updated integration tests accordingly (and deleted core.go).

test/integration/jobset_controller_test.go
```diff
@@ -315,3 +354,7 @@ func checkJobSetStatus(js *jobset.JobSet, condition jobset.JobSetConditionType)
 	}
 	return false, nil
 }
+
+func completionModePtr(m batchv1.CompletionMode) *batchv1.CompletionMode {
```
ahg-g, line 357: remove

test/integration/jobset_controller_test.go
```diff
@@ -37,6 +37,18 @@ const (
...
+	Spec: corev1.PodSpec{
+		RestartPolicy: "Never",
+		Containers: []corev1.Container{
+			{
+				Name:  "test-container",
+				Image: "busybox:latest",
+			},
+		},
+	},
+}
```
ahg-g, line 49: make it part of MakeJob

pkg/util/testing/wrappers.go
```diff
@@ -104,3 +105,50 @@ func (r *ReplicatedJobWrapper) SetReplicas(val int) *ReplicatedJobWrapper {
...
+func (j *JobWrapper) SetParallelism(parallelism int) *JobWrapper {
+	j.Spec.Parallelism = pointer.Int32(int32(parallelism))
+	return j
+}
+
+// SetTEmplate sets the value of the pod template spec (job.spec.template)
+func (j *JobWrapper) SetTemplate(template corev1.PodTemplateSpec) *JobWrapper {
+	j.Spec.Template = template
+	return j
+}
```
ahg-g, line 149: I would remove all those three functions, we don't have any logic that is directly impacted by those values, so just set parallelism and completion to some default value in MakeJob

danielvegamyhre: Makes sense, I wasn't sure if in the wrapper pattern it would be unexpected for the object to have some default values that the user didn't explicitly set. Anyway I finished refactoring this.

ahg-g: sg, I don't think you uploaded a new commit yet.

danielvegamyhre: sorry about that, pushed it now.

PR #20, opened 2023-04-12
feat: add suspend to jobset api

### LINKED ISSUE #9: Add suspend flag
To facilitate integration with Kueue

kannon92:
Fixes #9 

Add suspend to JobSet.


kannon92:
@ahg-g or @danielvegamyhre this is ready for review when you get a chance.

pkg/util/testing/wrappers.go
```diff
@@ -30,15 +30,16 @@ type JobSetWrapper struct {
 }
 
 // MakeJobSet creates a wrapper for a JobSet.
-func MakeJobSet(name, ns string) *JobSetWrapper {
+func MakeJobSet(name, ns string, suspend bool) *JobSetWrapper {
```
ahg-g, line 33: add the SetSuspend func instead to continue with the wrapper pattern

test/integration/jobset_controller_test.go
```diff
@@ -311,6 +308,87 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			gomega.Eventually(checkJobSetStatus, timeout, interval).WithArguments(js, jobset.JobSetFailed).Should(gomega.Equal(true))
 		})
 	})
+	ginkgo.When("a jobset is created with suspend disabled", func() {
+		ginkgo.It("should create all jobs and headless services, then complete successfully once all jobs are completed", func() {
```
ahg-g, line 312: this test is not needed, it is similar to one of the tests above.

test/integration/jobset_controller_test.go
```diff
@@ -311,6 +308,87 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				})
+				gomega.Expect(k8sClient.Status().Update(ctx, &job)).Should(gomega.Succeed())
+			}
+			// Check JobSet has completed.
+			gomega.Eventually(checkJobSetStatus, timeout, interval).WithArguments(js, jobset.JobSetCompleted).Should(gomega.Equal(true))
+		})
+
+	})
+	ginkgo.When("a jobset is created with suspend enabled", func() {
+		ginkgo.It("should create all jobs in a suspended state", func() {
```
ahg-g, line 353: this is good to have, but lets add a test that suspends the job after it started, something like:

1. start -> all jobs should have suspend=false and the suspend condition on jobset is not existent
2. suspend -> all jobs should have suspend=true and the suspend condition on jobset is true
3. restart -> all jobs should have suspend=false and the suspend condition on jobset is false

pkg/util/testing/wrappers.go
kannon92: done.

test/integration/jobset_controller_test.go
kannon92: done.

pkg/controllers/jobset_controller.go
```diff
@@ -328,6 +348,15 @@ func isJobFinished(job *batchv1.Job) (bool, batchv1.JobConditionType) {
 	return false, ""
 }
 
+func IsJobSetSuspended(js *jobset.JobSet) bool {
```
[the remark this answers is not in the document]
kannon92: done.

test/integration/jobset_controller_test.go
kannon92: https://github.com/kubernetes-sigs/jobset/pull/20/files#diff-2c3dfd6ff1abd441d5beb3096d5cbf53d91218680c176d6df3549f835f28a0dcR355

I am working on this but for some reason I am unable to update a JobSet once it is set in the integration tests.  

Do you have an idea of why this would be the case?  

I'm getting:

```
     Operation cannot be fulfilled on jobsets.batch.x-k8s.io "suspend-true": 
the object has been modified
; please apply your changes to the latest version and try again
```

ahg-g: I think this is because the status was updated, so you need to get it again before you do the second update.

kannon92: Yep! Found that out and I think it’s ready for another review.

pkg/controllers/jobset_controller.go
```diff
@@ -126,6 +127,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, nil
 	}
 
+	// If JobSpec has a suspend field of true,
+	// set condition to say jobspec is suspended.
+	if js.Spec.Suspend != nil && *js.Spec.Suspend && !JobSetSuspended(&js) {
```
ahg-g, line 132: create the a variable `jobsetSuspended := js.Spec.Suspend != nil && *js.Spec.Suspend`

pkg/controllers/jobset_controller.go
```diff
@@ -126,6 +127,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		if err := r.updateStatusWithCondition(ctx, &js, corev1.EventTypeNormal, metav1.Condition{
```
ahg-g, line 133: We can depend on the job status because that will cause race conditions: imagine that case where this update succeeded, but then the controller restarts, we will not get to suspending the jobs ever.

Imagine also the case where a child job was unsuspended out of band, we need to suspend it again.

The condition above should only consider the js.spec.suspend field, `!JobSetSuspended(&js)` must not be part of the condition to determine where or not to suspend/unsuspend the child jobs.

pkg/controllers/jobset_controller.go
```diff
@@ -126,6 +127,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			Message:            "jobset is resumed",
+		}); err != nil {
+			log.Error(err, "updating jobset status")
+			return ctrl.Result{}, nil
+		}
+		//TODO: Is active sufficient for all suspend jobs?
+		// They aren't failed, deleted or successful.
+		for _, job := range ownedJobs.active {
+			job.Spec.Suspend = pointer.Bool(false)
+			if r.Update(ctx, job); err != nil {
```
ahg-g, line 172: we should check if the job is already unsuspended before issuing the update

pkg/controllers/jobset_controller.go
```diff
@@ -431,10 +485,19 @@ func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet,
 	return nil
 }
 
-// TODO: update condition in place if it exists.
 func (r *JobSetReconciler) updateStatusWithCondition(ctx context.Context, js *jobset.JobSet, eventType string, condition metav1.Condition) error {
```
ahg-g, line 488: lets call it `ensureCondition` then

pkg/controllers/jobset_controller.go
```diff
@@ -431,10 +485,19 @@ func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet,
...
+		if condition.Type == val.Type {
+			js.Status.Conditions[i] = condition
+			conditionExists = true
+		}
+	}
+	if !conditionExists {
+		js.Status.Conditions = append(js.Status.Conditions, condition)
+	}
 
 	if err := r.Status().Update(ctx, js); err != nil {
```
ahg-g, line 502: We should update only if the condition changed

kannon92:
Thank you for the reviews! I’m going to work on this next week. Had Kubecon but I’m interested in getting this working!

ahg-g:
> Thank you for the reviews! I’m going to work on this next week. Had Kubecon but I’m interested in getting this working!

Thanks, that sounds great!

test/integration/controller/jobset_controller_test.go
```diff
@@ -354,6 +355,20 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("suspended jobset will not start", &testCase{
```
kannon92, line 358: I need to add these three cases here:  

1) Start a JobSet as suspend should have suspend jobs.
2) Resuming a suspend job should start it.
3) Starting a job then suspended should suspend.

pkg/controllers/jobset_controller.go
kannon92: I don't really understand this.  

We check for JobSetFinished or JobSetFailed by checking the condition.

I tried to check suspend via the spec but I realize I can't really distinguish between a JobSet starting or a JobSet being resumed.  I know that most of the tests fail because we are pushing out a condition that says a jobset is resumed even if we are starting the job without a suspend.  

I check the suspend field and then check the status of the jobset to see if the jobset was previously suspended.  If that is the case then we push a condition that says we were resumeded. 

Otherwise I am unsure how to detect between suspend of false and suspend of false when it was original suspend of true.

test/integration/controller/jobset_controller_test.go
kannon92: Added.

pkg/controllers/jobset_controller.go
kannon92: Done.

api/v1alpha1/jobset_types.go
```diff
@@ -45,6 +47,11 @@ type JobSetSpec struct {
 	// The JobSet is always declared failed if all jobs in the set
 	// finished with status failed.
 	FailurePolicy *FailurePolicy `json:"failurePolicy,omitempty"`
+	// Suspend flag assumes that the JobSet will be
+	// used with a queueing framework.
+	// True means that no objects are running so the ReplicatedJobs should be also suspended.
```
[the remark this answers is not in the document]
kannon92: Done.

pkg/util/testing/wrappers.go
```diff
@@ -59,6 +59,12 @@ func (j *JobSetWrapper) AddReplicatedJob(job jobset.ReplicatedJob) *JobSetWrappe
 	return j
 }
 
+// AddSuspend adds a suspend flag to JobSet
+func (j *JobSetWrapper) AddSuspend(suspend bool) *JobSetWrapper {
```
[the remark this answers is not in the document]
kannon92: Done.

pkg/controllers/jobset_controller.go
```diff
@@ -126,6 +127,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			Status:             metav1.ConditionStatus(corev1.ConditionFalse),
+			LastTransitionTime: metav1.Now(),
+			Reason:             "ResumeJobs",
+			Message:            "jobset is resumed",
+		}); err != nil {
+			log.Error(err, "updating jobset status")
+			return ctrl.Result{}, nil
+		}
+		//TODO: Is active sufficient for all suspend jobs?
+		// They aren't failed, deleted or successful.
```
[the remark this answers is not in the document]
kannon92: TY. Done.

pkg/controllers/jobset_controller.go
kannon92: Done.

test/integration/controller/jobset_controller_test.go
```diff
@@ -222,24 +239,31 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			// Perform a series of updates to jobset resources and check resulting jobset state after each update.
 			for _, update := range tc.updates {
 
+				var jobSet jobset.JobSet
+				gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobSet), timeout, interval).Should(gomega.Succeed())
+
+				if update.jobSetUpdateFn != nil {
+					update.jobSetUpdateFn(&jobSet)
+					// Verify a valid jobset update succeeded.
+					gomega.Eventually(k8sClient.Update(ctx, &jobSet), timeout, interval).Should(gomega.Succeed())
```
ahg-g, line 248: just to be consistent with jobUpdateFn, lets call the update inside jobSetUpdateFn

test/integration/controller/jobset_controller_test.go
```diff
@@ -354,6 +378,58 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				return testJobSet(ns).
+					Suspend(true)
+			},
+			updates: []*jobUpdate{
+				{
+					jobUpdateFn: func(jobList *batchv1.JobList) {
+						suspendJob(&jobList.Items[0])
+						suspendJob(&jobList.Items[1])
+					},
+					expectedJobSetCondition: jobset.JobSetSuspended,
```
ahg-g, line 403: as mentioned above, add a checkJobState func that iterates over the jobs and checks that they are suspended.

test/integration/controller/jobset_controller_test.go
```diff
@@ -354,6 +378,58 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				},
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						js.Spec.Suspend = pointer.Bool(false)
+					},
+					checkJobSetState: func(js *jobset.JobSet) {
+						ginkgo.By("checking all jobs are resumed")
+						gomega.Eventually(checkJobsResumed, timeout, interval).WithArguments(js).Should(gomega.Equal(true))
```
ahg-g, line 411: we can do this inside a checkJobsState

test/integration/controller/jobset_controller_test.go
```diff
@@ -354,6 +378,58 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+		ginkgo.Entry("jobset that gets suspended after startup transitions to suspended", &testCase{
+			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
+				return testJobSet(ns)
+			},
+			updates: []*jobUpdate{
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						js.Spec.Suspend = pointer.Bool(true)
+					},
+					expectedJobSetCondition: jobset.JobSetSuspended,
```
ahg-g, line 429: also check that the child jobs are unsuspended

test/integration/controller/jobset_controller_test.go
```diff
@@ -222,24 +239,31 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			// Perform a series of updates to jobset resources and check resulting jobset state after each update.
 			for _, update := range tc.updates {
 
+				var jobSet jobset.JobSet
+				gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobSet), timeout, interval).Should(gomega.Succeed())
+
```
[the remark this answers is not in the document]
kannon92: updated.

test/integration/controller/jobset_controller_test.go
```diff
@@ -222,24 +239,31 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				}
 				// Fetch updated job objects so we always have the latest resource versions to perform mutations on.
 				var jobList batchv1.JobList
-				gomega.Expect(k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace))).Should(gomega.Succeed())
-
+				gomega.Eventually(k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace)), timeout, interval).Should(gomega.Succeed())
 				// Perform mutation on jobset if specified.
 				if update.jobUpdateFn != nil {
 					update.jobUpdateFn(&jobList)
 				}
```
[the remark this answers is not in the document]
kannon92: good call.  added.

test/integration/controller/jobset_controller_test.go
kannon92: Added.

test/integration/controller/jobset_controller_test.go
```diff
@@ -354,6 +378,58 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				{
+					jobUpdateFn: func(jobList *batchv1.JobList) {
+						suspendJob(&jobList.Items[0])
+						suspendJob(&jobList.Items[1])
+					},
+					expectedJobSetCondition: jobset.JobSetSuspended,
+				},
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						js.Spec.Suspend = pointer.Bool(false)
```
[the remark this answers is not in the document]
kannon92: Added a function to suspend and send an update.

test/integration/controller/jobset_controller_test.go
kannon92: added.

kannon92:
@ahg-g and @danielvegamyhre this is ready for another review

test/integration/controller/jobset_controller_test.go
```diff
@@ -349,9 +367,96 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+						ginkgo.By("checking all jobs are suspended")
+						gomega.Eventually(matchJobsSuspendState, timeout, interval).WithArguments(js, true).Should(gomega.Equal(true))
+					},
+				},
+			},
+		}),
 	) // end of DescribeTable
 }) // end of Describe
 
+func checkJobSetStatus(js *jobset.JobSet, condition jobset.JobSetConditionType) (bool, error) {
```
danielvegamyhre, line 439: These 2 functions (`checkJobSetStatus`, `numExpectedJobs`) are now part of a shared `sigs.k8s.io/jobset/test/util` pkg so you can use those shared functions here.

test/integration/controller/jobset_controller_test.go
```diff
@@ -189,8 +202,10 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 	// jobUpdate contains the mutations to perform on the jobs and the
 	// checks to perform afterwards.
 	type jobUpdate struct {
```
danielvegamyhre, line 204: Should we consider renaming this struct to something other than `jobUpdate` now, considering how it also updates  the jobset, the name is somewhat inaccurate/misleading now. I would suggest just calling it `update` but we would need to think of a new name for the var used in `for _, update := range tc.updates { ...` below. Perhaps call this struct `update` and the range var `u`? Any thoughts?

test/integration/controller/jobset_controller_test.go
```diff
@@ -221,18 +236,21 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
-				var jobList batchv1.JobList
-				gomega.Expect(k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace))).Should(gomega.Succeed())
+				var jobSet jobset.JobSet
+				gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobSet), timeout, interval).Should(gomega.Succeed())
 
-				// Perform mutation on jobset if specified.
-				if update.jobUpdateFn != nil {
+				if update.jobSetUpdateFn != nil {
+					update.jobSetUpdateFn(&jobSet)
+				} else if update.jobUpdateFn != nil {
```
danielvegamyhre, line 244: Why is this `else if` instead of just `if`? Wouldn't it be more flexible to allow test cases to perform both a jobset update and then update job(s) in the same iteration? Or is there a reason these must be mutually exclusive?

pkg/controllers/jobset_controller.go
```diff
@@ -184,6 +200,11 @@ func (r *JobSetReconciler) constructJobsFromTemplate(js *jobset.JobSet, rjob *jo
 			job.Spec.Template.Spec.Subdomain = job.Name
 		}
 
+		// If JobSet is suspend we should set Suspend for all jobs
+		if js.Spec.Suspend != nil && *js.Spec.Suspend {
+			job.Spec.Suspend = pointer.Bool(*js.Spec.Suspend)
+		}
```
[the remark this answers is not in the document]
kannon92: Thank you for pointing this out.  I added these test cases.

pkg/controllers/jobset_controller.go
```diff
@@ -129,6 +130,57 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		if err := r.ensureCondition(ctx, &js, corev1.EventTypeNormal, metav1.Condition{
+			Type:               string(jobset.JobSetSuspended),
+			Status:             metav1.ConditionStatus(corev1.ConditionFalse),
+			LastTransitionTime: metav1.Now(),
+			Reason:             "ResumeJobs",
+			Message:            "jobset is resumed",
+		}); err != nil {
+			log.Error(err, "updating jobset status")
+			return ctrl.Result{}, nil
+		}
```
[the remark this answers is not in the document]
kannon92: Updated.

test/integration/controller/jobset_controller_test.go
kannon92: Ah thank you.  I think its rebased but it was stupid.  I'll make sure to use the util one and remove this.

test/integration/controller/jobset_controller_test.go
kannon92: @ahg-g suggested that we should only really do either job or jobset updates.

kannon92: https://github.com/kubernetes-sigs/jobset/pull/20/files#r1176494351

test/integration/controller/jobset_controller_test.go
kannon92: removed.

test/integration/controller/jobset_controller_test.go
kannon92: updated.

pkg/controllers/jobset_controller_test.go
```diff
@@ -550,6 +592,83 @@ func TestConstructJobsFromTemplate(t *testing.T) {
...
+					Job(testutils.MakeJobTemplate(jobName, ns).Obj()).
+					Replicas(1).
+					Obj()).Obj(),
+			eventType:      string(jobset.JobSetSuspended),
+			newCondition:   metav1.Condition{Type: string(jobset.JobSetSuspended), Reason: "JobsResumed", Status: metav1.ConditionStatus(corev1.ConditionFalse)},
+			conditions:     []metav1.Condition{{Type: string(jobset.JobSetSuspended), Reason: "JobsSuspended", Status: metav1.ConditionStatus(corev1.ConditionTrue)}},
+			expectedUpdate: true,
+		},
+		{
+			name: "duplicateComplete",
```
ahg-g, line 648:
```suggestion
			name: "duplicate complete condition",
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -219,32 +234,35 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+
+				var jobSet jobset.JobSet
+				gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobSet), timeout, interval).Should(gomega.Succeed())
+
+				if up.jobSetUpdateFn != nil {
+					up.jobSetUpdateFn(&jobSet)
+				} else if up.jobUpdateFn != nil {
+					// Fetch updated job objects so we always have the latest resource versions to perform mutations on.
+					var jobList batchv1.JobList
+					gomega.Eventually(k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace)), timeout, interval).Should(gomega.Succeed())
```
ahg-g, line 247: we need to check that the number of jobs is the expected one

test/integration/controller/jobset_controller_test.go
```diff
@@ -349,6 +367,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("suspended jobset will not start", &testCase{
```
ahg-g, line 370:
```suggestion
		ginkgo.Entry("jobset created in suspended state", &testCase{
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -349,6 +367,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				{
+					expectedJobSetCondition: jobset.JobSetSuspended,
+					checkJobSetState: func(js *jobset.JobSet) {
+						ginkgo.By("checking all jobs are suspended")
+						gomega.Eventually(matchJobsSuspendState, timeout, interval).WithArguments(js, true).Should(gomega.Equal(true))
+					},
+				},
+			},
+		}),
+		ginkgo.Entry("suspended jobset resumes on setting suspend to false", &testCase{
```
ahg-g, line 385:
```suggestion
		ginkgo.Entry("resume a suspended jobset", &testCase{
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -349,6 +367,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+						gomega.Eventually(matchJobsSuspendState, timeout, interval).WithArguments(js, false).Should(gomega.Equal(true))
+					},
+				},
+				{
+					jobUpdateFn:             completeAllJobs,
+					expectedJobSetCondition: jobset.JobSetCompleted,
+				},
+			},
+		}),
+		ginkgo.Entry("jobset that gets suspended after startup transitions to suspended", &testCase{
```
ahg-g, line 413:
```suggestion
		ginkgo.Entry("suspend a running jobset", &testCase{
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -385,7 +469,30 @@ func failJob(job *batchv1.Job) {
 		Type:   batchv1.JobFailed,
 		Status: corev1.ConditionTrue,
 	})
-	gomega.Expect(k8sClient.Status().Update(ctx, job)).Should(gomega.Succeed())
+	gomega.Eventually(k8sClient.Status().Update(ctx, job), timeout, interval).Should(gomega.Succeed())
+}
+
+func jobSetSuspend(js *jobset.JobSet, suspend bool) {
```
ahg-g, line 475:
```suggestion
func suspendJobSet(js *jobset.JobSet, suspend bool) {
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -349,6 +367,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+						ginkgo.By("checking all jobs are suspended")
+						gomega.Eventually(matchJobsSuspendState, timeout, interval).WithArguments(js, true).Should(gomega.Equal(true))
+					},
+				},
+			},
+		}),
+		ginkgo.Entry("suspended jobset resumes on setting suspend to false", &testCase{
+			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
+				return testJobSet(ns).
+					Suspend(true)
```
ahg-g, line 388:
```suggestion
				return testJobSet(ns).Suspend(true)
```

ahg-g (commented):
Thanks! this is great.

Can you please amend the commits and not merge them, just so I can easily tell the difference.

pkg/controllers/jobset_controller.go
```diff
@@ -129,6 +130,59 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+					return ctrl.Result{}, nil
+				}
+			}
+		}
+
+		// If JobSpec is unsuspended either by removing suspend field or setting to false
+		// and we detect a condition that says it was suspended
+		// we must update all jobs to be suspended and modify condition.
+	} else {
+		if jobSetSuspended(&js) {
```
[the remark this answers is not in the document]
kannon92: ensureCondition does a few things:

1) Replaces a condition if it exists but ConditionStatus is different.
2) Appends if condition doesn't exist
3) if duplicate then does nothing

So either 1 or 2 will then force an update.  So I think we still need this check because this guards against the case that if a job was previously suspended, then we know that we should resume.  Otherwise we don't need to set that condition.

pkg/controllers/jobset_controller.go
```diff
@@ -496,6 +568,15 @@ func isJobFinished(job *batchv1.Job) (bool, batchv1.JobConditionType) {
 	return false, ""
 }
 
+func jobSetSuspended(js *jobset.JobSet) bool {
```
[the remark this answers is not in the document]
kannon92: If we can remove it above then we can remove but currently we use it to detect difference between a normal job and suspended job being resumed.

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
kannon92: But maybe we should change the name then.  

This was UpdateStatusWithCondition but we changed it to only update in case of condition being changed or non existent.

test/integration/controller/jobset_controller_test.go
kannon92: done.

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
danielvegamyhre: Given the current logic I think it is necessary. Checking `jobset.spec.suspend` and the jobset status condition of `JobSetSuspended` are used to determine which point in the suspend/resume logic we are currently in, so I don't see how we could remove it. What were you thinking

Either way I think this is fine for now and we can look into this further in a follow up PR if necessary.

pkg/controllers/jobset_controller.go
```diff
@@ -129,6 +130,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		if err := r.ensureCondition(ctx, &js, corev1.EventTypeNormal, metav1.Condition{
+			Type:               string(jobset.JobSetSuspended),
+			Status:             metav1.ConditionStatus(corev1.ConditionTrue),
+			LastTransitionTime: metav1.Now(),
+			Reason:             "SuspendedJobs",
+			Message:            "jobset is suspended",
+		}); err != nil {
+			log.Error(err, "updating jobset status")
+			return ctrl.Result{}, nil
+		}
```
ahg-g, line 145: pls move this to after the loop below, we should suspend the jobs first and then update the status.

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
ahg-g: I don't get why it is necessary to check? `spec.suspend` is the source of truth that dictates how to reconcile.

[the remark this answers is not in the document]
danielvegamyhre: Oh I thought we were using both. In that case, I think maybe it is not necessary. To confirm, the logic should be:

Suspend:
1. user suspends jobset (`.spec.suspended=true`), reconciliation triggered
2. controller sees `.spec.suspended=true`, suspends all jobs which are not already suspended, then sets jobset status to suspended

Resume:
1. user resumes jobset (`.spec.suspended=false`), reconciliation triggered
2. controller sees `.spec.suspended=false`, then resumes all jobs which are currently suspended, then updates the jobset's `JobSetSuspended` condition to `ConditionStatus=false`.

If the above is correct, then we don't need conditional logic based on `jobSetSuspended()` check, so we can remove it.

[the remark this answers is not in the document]
kannon92: Third case:

JobSet starts without suspend ever set (so not resume).

The third case is why I added the check for condition.  

I thought it would be weird for a JobSet to send a condition saying the job is resumed even if the suspend flag was never set to true.  I was thinking a jobset that starts with no suspend field should not emit a condition saying it was resumed.

pkg/controllers/jobset_controller.go
danielvegamyhre: looks like this was done

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
ahg-g: ok, fair point, then I think ensureCondition should simply be changed to not set the condition if it doesn't exist and the ConditionStatus=false.

pkg/controllers/jobset_controller.go
```diff
@@ -129,6 +130,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+				Status:             metav1.ConditionStatus(corev1.ConditionFalse),
+				LastTransitionTime: metav1.Now(),
+				Reason:             "ResumeJobs",
+				Message:            "jobset is resumed",
+			}); err != nil {
+				log.Error(err, "updating jobset status")
+				return ctrl.Result{}, nil
+			}
+		}
+		for _, job := range ownedJobs.active {
```
ahg-g, line 171: pls move this loop to execute before the call to set the condition

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
kannon92: Not sure what you are asking.  

If the condition doesn't exist, we use ensureCondition to set a condition and then update the status field.  

If a ConditionStatus=True -> ConditionStatus=False I think we would want to also capture the change.  

Or are you saying that if we suspend a condition we emit a suspended condition and then if we resume, we append a new condition that says it was resumed and keep the old suspend condition?

pkg/controllers/jobset_controller.go
kannon92: Thanks!

pkg/controllers/jobset_controller.go
kannon92: done.

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
ahg-g: > If the condition doesn't exist, we use ensureCondition to set a condition and then update the status field.

I am saying if the condition does not exist and the condition we are trying to set is associated with ConditionStatus=False, we don't set the condition at all. The condition not existing is equal to the condition exist with ConditionStatus=False

basically updateCondition will become:

```
func updateCondition(js *jobset.JobSet, condition metav1.Condition) bool {
	condition.LastTransitionTime = metav1.Now()
	for i, val := range js.Status.Conditions {
		if condition.Type == val.Type && condition.Status != val.Status {
			js.Status.Conditions[i] = condition
			// Condition found but different status so we should update
			return true
		} else if condition.Type == val.Type && condition.Status == val.Status {
			// Duplicate condition so no update
			return false
		}
	}
        // condition doesn't exist, update only if the status is false
        if condition.Status == corev1.ConditionTrue {
	  js.Status.Conditions = append(js.Status.Conditions, condition)
	  return true
        }
       return false
}```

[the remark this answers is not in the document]
Comment (kannon92) [reply pkg/controllers/jobset_controller.go:570]: Ah thank you for clearing that up.  I updated it and added a few more test cases.

File: pkg/controllers/jobset_controller.go [hunk 37]
```diff
@@ -129,6 +130,55 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			if pointer.BoolDeref(job.Spec.Suspend, false) != false {
+				job.Spec.Suspend = pointer.Bool(false)
+				if r.Update(ctx, job); err != nil {
+					log.Error(err, "unsuspending job", "job", klog.KObj(job))
+					return ctrl.Result{}, nil
+				}
+			}
+		}
+		if jobSetSuspended(&js) {
```
ahg-g, line 168: you still need to remove this and the jobSetSuspended func :)

kannon92: I'm with you now.  You asked for the refactor on ensureCondition so we can not send updates if its false.

kannon92: done.

PR #23, opened 2023-04-12
Add support for failure policy of target=Any and restartPolicy=RecreateAll

danielvegamyhre:
Add support for failure policy of target=Any and restartPolicy=RecreateAll


api/v1alpha1/jobset_types.go
```diff
@@ -76,7 +92,7 @@ type ReplicatedJob struct {
 	// Replicas is the number of jobs that will be created from this ReplicatedJob's template.
 	// Jobs names will be in the format: <jobSet.name>-<spec.replicatedJob.name>-<job-index>
 	// +kubebuilder:default=1
-	Replicas *int `json:"replicas,omitempty"`
```
danielvegamyhre, line 79: I changed fields with kubebuilder default marker to be concrete data types instead of pointers, since the defaulting didn't seem to be working with pointers.

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +246,197 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+	if err := r.updateStatus(ctx, js, metav1.Condition{
+		Type:               string(jobset.JobSetRestarting),
+		Status:             metav1.ConditionStatus(corev1.ConditionTrue),
+		LastTransitionTime: metav1.Now(),
+		Reason:             "Restarting",
+		Message:            "jobset is currently restarting as part of recreation restart policy",
+	}); err != nil {
+		return fmt.Errorf("updating jobset status to restarting while executing recreation restart policy: %v", err)
+	}
+
```
ahg-g, line 332: Imagine the controller is restarted right here after updating the status but before it gets to delete all jobs below, or in the middle of the loop issuing job deletions inside deleteAllJobs; in this case, the jobset will be stuck since at 316 above we will exit early.

I feel this approach of depending on the condition to determine what to do is brittle, also the semantics of "restarting" condition is not clear what it means from the user perspective (when do we exit it?)

I suggest we label the child jobs with the restart count to determine which jobs should be counted against the status of the jobset and which should not and be deleted. This approach will be more robust against controller restarts and race conditions.

The algorithm may look like this:

1. Child Jobs are created with label `jobset.sigs.k8s.io/restart-attampt` = `jobset.status.restarts`
2. On each reconcile, to determine success/failure, we only evaluate the status of jobs with `jobset.sigs.k8s.io/restart-attempt` == `jobset.status.restarts`
3. If there is failure, then increment `jobset.status.restarts`
4. Jobs where jobset.sigs.k8s.io/restart-attampt < jobset.status.restarts should be deleted.

api/v1alpha1/jobset_types.go
```diff
@@ -86,6 +102,38 @@ type Network struct {
 	EnableDNSHostnames *bool `json:"enableDNSHostnames,omitempty"`
 }
 
+type TerminationPolicyTarget string
+
+const (
+	TerminationPolicyTargetAll TerminationPolicyTarget = "All"
+	TerminationPolicyTargetAny TerminationPolicyTarget = "Any"
```
ahg-g, line 109: add comment:

// Any means any of the jobs managed by the jobset.

Note that AnyOf is the operator that will allow selecting any of a subset of jobs.

pkg/controllers/jobset_controller.go
danielvegamyhre: Good point! I wasn't thinking about the race conditions that could arise as a result of controller restarts mid-execution, though I can see how important that is especially in the k8s world where pods are somewhat ephemeral and may be restarted/moved to another node/etc. at any time.

I've implemented the suggested approach, please take a look and let me know if you have any additional thoughts.

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +246,197 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+	}
+	return nil
+}
+
+func (r *JobSetReconciler) executeRestartPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	switch js.Spec.FailurePolicy.RestartPolicy {
+	case jobset.RestartPolicyRecreate:
+		return r.restartPolicyRecreate(ctx, js, ownedJobs)
+	case jobset.RestartPolicyNone:
+		return fmt.Errorf("unimplemented")
```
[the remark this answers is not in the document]
danielvegamyhre: Updated this to fail the jobset for Falure Policy with RestartPolicyNone

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +246,197 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+	default:
+		return fmt.Errorf("invalid termination policy target %s", js.Spec.FailurePolicy.Target)
+	}
+}
+
+func (r *JobSetReconciler) failureTargetAny(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	// If any job from the target ReplicatedJob has failed, fail the JobSet.
+	// TODO: can we refactor using a different data structure for ownedJobs to avoid these nested for loops?
+	for _, target := range js.Spec.FailurePolicy.Jobs {
```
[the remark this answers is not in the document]
danielvegamyhre: I see, I didn't realize AnyOf and Any were distinct targets - I understand now. I updated this accordingly.

pkg/controllers/jobset_controller.go
```diff
@@ -286,7 +234,175 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+			LastTransitionTime: metav1.Now(),
+			Reason:             "FailedJobs",
+			Message:            "jobset failed due to one or more job failures",
+		})
+	default:
+		return fmt.Errorf("invalid restart policy: %s", js.Spec.FailurePolicy.RestartPolicy)
+	}
+}
+
+func (r *JobSetReconciler) restartPolicyRecreate(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
```
ahg-g, line 281: RecreateAll

pkg/controllers/jobset_controller.go
```diff
@@ -73,29 +78,23 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	// Get Jobs owned by JobSet.
-	ownedJobs, err := r.getChildJobs(ctx, &js, req)
+	ownedJobs, err := r.getChildJobs(ctx, &js)
 	if err != nil {
 		log.Error(err, "getting jobs owned by jobset")
 		return ctrl.Result{}, nil
 	}
 
-	// If any jobs have failed, JobSet has failed.
+	// If any jobs have failed, execute the JobSet failure policy (if any).
```
ahg-g, line 87: here we should call deleteJobs on the list of jobsToDelete identified by getChildJobs

pkg/controllers/jobset_controller.go
```diff
@@ -143,80 +142,29 @@ func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		switch finishedType {
 		case "": // active
-			jobs.active = append(jobs.active, &childJobList.Items[i])
+			ownedJobs.active = append(ownedJobs.active, &childJobList.Items[i])
 		case batchv1.JobFailed:
-			jobs.failed = append(jobs.failed, &childJobList.Items[i])
+			ownedJobs.failed = append(ownedJobs.failed, &childJobList.Items[i])
 		case batchv1.JobComplete:
-			jobs.successful = append(jobs.successful, &childJobList.Items[i])
+			ownedJobs.successful = append(ownedJobs.successful, &childJobList.Items[i])
```
ahg-g, line 163: active, failed and successful should only include jobs with restart-attempt = restarts

pkg/controllers/jobset_controller.go
```diff
@@ -286,7 +234,175 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+		if err != nil {
+			return err
+		}
+		if jobRestarts < js.Status.Restarts {
+			delete = append(delete, job)
+		}
+	}
+
+	// Delete all jobs in parallel.
+	for _, job := range delete {
```
ahg-g, line 328: we need to limit the number of go routines we create, perhaps add a comment to address this in a followup PR. we probably need a common function to delete/create jobs in parallel.

pkg/controllers/jobset_controller.go
```diff
@@ -286,7 +234,175 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+		// Jobs with jobset.sigs.k8s.io/restart-attempt < jobset.status.restarts are marked for
+		// deletion, as they were part of the previous JobSet lifecycle iteration.
+		jobRestarts, err := strconv.Atoi(job.Labels[jobset.RestartsLabel])
+		if err != nil {
+			return err
+		}
+		if jobRestarts < js.Status.Restarts {
+			delete = append(delete, job)
+		}
+	}
```
ahg-g, line 325: this logic better not be here, the function should only be concerned with deleting a list of jobs, the caller is responsible for specifying which exact jobs to delete.

pkg/controllers/jobset_controller.go
ahg-g: those three lists in addition to the jobsToDelete one should be mutually exclusive

pkg/controllers/jobset_controller.go
```diff
@@ -309,9 +425,9 @@ func (r *JobSetReconciler) shouldCreateJob(jobName string, ownedJobs *childJobs)
 	return true
 }
 
-// updateStatus updates the status of a JobSet by appending
+// updateStatusCondition updates the status of a JobSet by appending
 // the new condition to jobset.status.conditions.
-func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) error {
+func (r *JobSetReconciler) updateStatusCondition(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) error {
```
ahg-g, line 430: always appending is fine for now since we only update one condition that is terminal, but when supporting cases like suspend, we need to first check if the condition already exists and update it in place if necessary

danielvegamyhre: Ok, noted.

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +253,155 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+	case jobset.TerminationPolicyTargetAny:
+		// To reach this point a job must have failed, and TerminationPolicyTargetAny applies to any job
+		// in the JobSet, so we can skip directly to executing the restart policy.
+		return r.failureRestartPolicy(ctx, js, ownedJobs)
+	default:
+		return fmt.Errorf("invalid termination policy target %s", js.Spec.FailurePolicy.Operator)
+	}
+}
+
+func (r *JobSetReconciler) failureRestartPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
```
ahg-g, line 283: just to keep with the pattern:

```suggestion
func (r *JobSetReconciler) executeRestartPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
```

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +253,155 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+	case jobset.RestartPolicyNone:
+		return r.updateStatusCondition(ctx, js, metav1.Condition{
+			Type:               string(jobset.JobSetFailed),
+			Status:             metav1.ConditionStatus(corev1.ConditionTrue),
+			LastTransitionTime: metav1.Now(),
+			Reason:             "FailedJobs",
+			Message:            "jobset failed due to one or more job failures",
+		})
+	default:
+		return fmt.Errorf("invalid restart policy: %s", js.Spec.FailurePolicy.RestartPolicy)
```
ahg-g, line 296: log an error but also assume None policy.

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +253,155 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+			defer wg.Done()
+
+			// Delete job. This deletion event will trigger another reconcilliation,
+			// where the jobs are recreated.
+			backgroundPolicy := metav1.DeletePropagationBackground
+			if err := r.Delete(ctx, job, &client.DeleteOptions{PropagationPolicy: &backgroundPolicy}); err != nil {
+				finalErr = err
+				return
+			}
+			log.V(2).Info("successfully deleted job", "job", klog.KObj(job))
```
ahg-g, line 343: log the restart-attempt associated with the job

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +253,155 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+			LastTransitionTime: metav1.Now(),
+			Reason:             "ReachedMaxRestarts",
+			Message:            "jobset failed due to reaching max number of restarts",
+		})
+	}
+
+	// Increment JobSet restarts. This will trigger reconciliation and result in deletions
+	// of old jobs not part of the current jobSet lifecycle.
+	js.Status.Restarts += 1
+	return r.Status().Update(ctx, js)
```
ahg-g, line 318: log info here that the jobset is attempting restart number x after successfully returning from the update call

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +253,155 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+func (r *JobSetReconciler) restartPolicyRecreateAll(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js))
+	ctx = ctrl.LoggerInto(ctx, log)
+
+	// If JobSet has reached max number of restarts, mark it as failed and return.
+	if js.Status.Restarts == js.Spec.FailurePolicy.MaxRestarts {
+		return r.updateStatusCondition(ctx, js, metav1.Condition{
+			Type:               string(jobset.JobSetFailed),
+			Status:             metav1.ConditionStatus(corev1.ConditionTrue),
+			LastTransitionTime: metav1.Now(),
```
ahg-g, line 309: this can be set inside updateStatusCondition

test/integration/jobset_controller_test.go
```diff
@@ -311,6 +329,77 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+					SetJob(testing.MakeJob("test-job-leader", ns.Name).
+						SetCompletionMode(batchv1.IndexedCompletion).Obj()).
+					SetReplicas(3).
+					Obj()).
+				AddReplicatedJob(testing.MakeReplicatedJob("replicated-job-worker").
+					SetJob(testing.MakeJob("test-job-worker", ns.Name).
+						SetCompletionMode(batchv1.IndexedCompletion).Obj()).
+					SetReplicas(3).
+					Obj()).
+				// Set failure policy to "Any" with restart policy "Recreate" with max 1 restart (default).
```
ahg-g, line 350:
```suggestion
				// Set failure policy to "Any" with restart policy "Recreate" with max 1 restart.
```

pkg/controllers/jobset_controller.go
```diff
@@ -286,6 +253,155 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+	for _, job := range jobsForDeletion {
+		job := job
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+
+			// Delete job. This deletion event will trigger another reconcilliation,
+			// where the jobs are recreated.
+			backgroundPolicy := metav1.DeletePropagationBackground
+			if err := r.Delete(ctx, job, &client.DeleteOptions{PropagationPolicy: &backgroundPolicy}); err != nil {
```
ahg-g, line 339: add `IgnoreNotFound` option as well

pkg/controllers/jobset_controller.go
```diff
@@ -143,80 +149,41 @@ func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if err := r.List(ctx, &childJobList, client.InNamespace(js.Namespace), client.MatchingFields{jobOwnerKey: js.Name}); err != nil {
 		return nil, err
 	}
-
-	jobs := childJobs{}
+	// Categorize each job into a bucket: active, successful, failed, or delete.
+	ownedJobs := childJobs{}
 	for i, job := range childJobList.Items {
+		// Jobs with jobset.sigs.k8s.io/restart-attempt < jobset.status.restarts are marked for
+		// deletion, as they were part of the previous JobSet lifecycle.
```
ahg-g, line 164:
```suggestion
		// deletion, as they were part of the previous JobSet run.
```

pkg/controllers/jobset_controller.go
```diff
@@ -152,6 +151,9 @@ func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 // getChildJobs gets jobs owned by the JobSet then categorizes them by status (active, successful, failed).
 // Another list (`delete`) is also added which tracks jobs marked for deletion.
 func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet) (*childJobs, error) {
+	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js))
```
ahg-g, line 154: pass in the logger as a parameter instead of creating it every time, here and in all other places

test/integration/jobset_controller_test.go
```diff
@@ -311,6 +329,77 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			// Make a job fail.
+			ginkgo.By("making 1 job fail")
+			job := originalJobList.Items[0]
+			job.Status.Conditions = append(job.Status.Conditions, batchv1.JobCondition{
+				Type:   batchv1.JobFailed,
+				Status: corev1.ConditionTrue,
+			})
+			gomega.Expect(k8sClient.Status().Update(ctx, &job)).Should(gomega.Succeed())
+
+			// Verify jobs were recreated successfully.
```
ahg-g, line 383: we need to fail again to reach max restarts and check that the jobset is failed, and on another instance of the test we succeed and verify that the jobset is succeeded, you can add the latter on a followup PR using table format.

pkg/controllers/jobset_controller.go
danielvegamyhre: discussed via chat and decided on using `log := ctrl.LoggerFrom(ctx)` to retrieve logger from context, and removing the extraneous WithValues and `LoggerInto` calls since they are already done in the Reconcile() method.

ahg-g:
attempt 2
/test pull-jobset-test-integration-main

ahg-g:
attempt 3
/test pull-jobset-test-integration-main

pkg/controllers/jobset_controller.go
```diff
@@ -316,14 +436,24 @@ func (r *JobSetReconciler) shouldCreateJob(jobName string, ownedJobs *childJobs)
 	return true
 }
 
-// updateStatus updates the status of a JobSet by appending
-// the new condition to jobset.status.conditions.
-func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) error {
-	js.Status.Conditions = append(js.Status.Conditions, condition)
+// updateStatus updates the status of a JobSet.
+// TODO: update condition in place if it exists.
+func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, conditions ...metav1.Condition) error {
```
ahg-g, line 441: we want to log an event when we restart, I suggest:

```
updateStatus(ctx context.Context, js *jobset.JobSet, eventReason, eventMsg string) {
   if err := r.Status().Update(ctx, js); err != nil {
     return err
   }
   r.Record.Eventf(js, corev1.EventTypeNormal, eventReason, eventMsg)
}

updateStatusWithCondition(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) {
  c.LastTransitionTime = metav1.Now()
  js.Status.Conditions = append(js.Status.Conditions, c)
  updateStatus(ctx, js, condition.Reason, condition.Message)
}
```

pkg/controllers/jobset_controller.go
```diff
@@ -80,34 +86,33 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
 	// If all jobs have succeeded, JobSet has succeeded.
 	if len(ownedJobs.successful) == len(js.Spec.Jobs) {
 		if err := r.updateStatus(ctx, &js, metav1.Condition{
-			Type:               string(jobset.JobSetCompleted),
-			Status:             metav1.ConditionStatus(corev1.ConditionTrue),
-			LastTransitionTime: metav1.Now(),
-			Reason:             "AllJobsCompleted",
-			Message:            "jobset completed successfully",
+			Type:    string(jobset.JobSetCompleted),
```
ahg-g, line 112:
```
updateStatusWithCondition(ctx, &js, v1.EventTypeNormal, metav1.Condition{
			Type:    string(jobset.JobSetCompleted),
                         Status:  metav1.ConditionStatus(corev1.ConditionTrue),
			Reason:  "AllJobsCompleted",
			Message: "jobset completed successfully",
		})
```

pkg/controllers/jobset_controller.go
```diff
@@ -293,6 +261,158 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	// If no failure policy is defined, the default failure policy is to mark the JobSet
+	// as failed if any of its jobs have failed.
+	if js.Spec.FailurePolicy == nil {
+		return r.updateStatus(ctx, js, metav1.Condition{
+			Type:    string(jobset.JobSetFailed),
+			Status:  metav1.ConditionStatus(corev1.ConditionTrue),
+			Reason:  "FailedJobs",
+			Message: "jobset failed due to one or more failed jobs",
+		})
```
ahg-g, line 273: r.updateStatusWithCondition(ctx, js, v1.EventTypeWarning, metav1.Condition{
			Type:    string(jobset.JobSetFailed),
			Status:  metav1.ConditionStatus(corev1.ConditionTrue),
			Reason:  "FailedJobs",
			Message: "jobset failed due to one or more failed jobs",
		})

pkg/controllers/jobset_controller.go
```diff
@@ -293,6 +261,158 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+
+func (r *JobSetReconciler) executeRestartPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	failJobSet := func() error {
+		return r.updateStatus(ctx, js, metav1.Condition{
+			Type:    string(jobset.JobSetFailed),
+			Status:  metav1.ConditionStatus(corev1.ConditionTrue),
+			Reason:  "FailedJobs",
+			Message: "jobset failed due to one or more job failures",
```
ahg-g, line 295: define this function outside and use it at 267

pkg/controllers/jobset_controller.go
```diff
@@ -316,13 +436,24 @@ func (r *JobSetReconciler) shouldCreateJob(jobName string, ownedJobs *childJobs)
 	return true
 }
 
-// updateStatus updates the status of a JobSet by appending
-// the new condition to jobset.status.conditions.
-func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, condition metav1.Condition) error {
+// updateStatus updates the status of a JobSet.
+func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, eventReason, eventMsg string) error {
```
ahg-g, line 440: add eventType string
```suggestion
func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, eventType, eventReason, eventMsg string) error {
```

pkg/controllers/jobset_controller.go
```diff
@@ -293,6 +261,158 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
...
+			Status:  metav1.ConditionStatus(corev1.ConditionTrue),
+			Reason:  "ReachedMaxRestarts",
+			Message: "jobset failed due to reaching max number of restarts",
+		})
+	}
+
+	// Increment JobSet restarts. This will trigger reconciliation and result in deletions
+	// of old jobs not part of the current jobSet run.
+	js.Status.Restarts += 1
+	if err := r.updateStatus(ctx, js, "Restarting", fmt.Sprintf("restarting jobset, attempt %d", js.Status.Restarts)); err != nil {
```
ahg-g, line 326: pass event type `corev1.EventTypeWarning`

pkg/controllers/jobset_controller.go
```diff
@@ -293,6 +261,158 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
 	return nil
 }
 
+func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	// If no failure policy is defined, the default failure policy is to mark the JobSet
+	// as failed if any of its jobs have failed.
+	if js.Spec.FailurePolicy == nil {
+		return r.updateStatusCondition(ctx, js, metav1.Condition{
```
ahg-g, line 268: pass event type `corev1.EventTypeWarning`

PR #24, opened 2023-04-13
add events when conditions are set

### LINKED ISSUE #12: Emit events reflecting the jobset status
When marking a job completed, failed, suspended or resumed

kannon92:
Add an EventRecord so we can log events.  

Fixes #12


pkg/controllers/jobset_controller.go
```diff
@@ -91,6 +98,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			log.Error(err, "updating jobset status")
 			return ctrl.Result{}, nil
 		}
+		r.Record.Eventf(&js, corev1.EventTypeNormal, string(jobset.JobSetFailed), "jobset %s failed", js.Name)
```
ahg-g, line 101: can we put this inside updateStatus?

PR #28, opened 2023-04-14
Add more failure policy integration tests and refactor into table driven format

danielvegamyhre:
- Adds more failure policy integration tests and refactor into table driven format.
- Fixes a bug in the jobset controller causing it to try to recreate existing jobs marked for deletion that had not yet been deleted
- Refactor `constructJobsFromTemplate` method of `JobSetReconciler` to make it more concise and readable.

### DIFF
--- a/test/integration/jobset_controller_test.go
+++ b/test/integration/jobset_controller_test.go
@@ -17,6 +17,7 @@ package test
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
@@ -330,8 +331,17 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 		})
 	})
 
-	ginkgo.When("a jobset is created with failure policy 'Any' and restart policy 'RecreateAll'", func() {
-		ginkgo.It("should create all jobs, then restart every job once any job fails", func() {
+	// jobSetUpdate contains the mutations to perform on the jobset and the
+	// checks to perform afterwards.
+	type jobSetUpdate struct {
+		name                    string
+		jobUpdateFn             func(jobList *batchv1.JobList) error
+		checkJobSetState        func(js *jobset.JobSet) (bool, error)
+		expectedJobSetCondition jobset.JobSetConditionType
+	}
+
+	ginkgo.DescribeTable("a jobset is created with failure policy 'Any' and restart policy 'RecreateAll'",
+		func(updates []*jobSetUpdate) {
 			ginkgo.By("creating a new JobSet")
 			ctx := context.Background()
 
@@ -362,6 +372,7 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			ginkgo.By("checking JobSet was created successfully")
 			gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).Should(gomega.Succeed())
 
+			// Check all jobs are created successfully.
 			ginkgo.By("checking JobSet eventually has 6 active jobs")
 			var originalJobList batchv1.JobList
 			gomega.Eventually(func() (int, error) {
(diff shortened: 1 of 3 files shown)


test/integration/jobset_controller_test.go
```diff
@@ -371,48 +382,115 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			},
+			{
+				jobUpdateFn: func(jobList *batchv1.JobList) error {
+					ginkgo.By("failing another job")
+					job := &jobList.Items[0]
+					job.Status.Conditions = append(job.Status.Conditions, batchv1.JobCondition{
+						Type:   batchv1.JobFailed,
+						Status: corev1.ConditionTrue,
+					})
+					gomega.Expect(k8sClient.Status().Update(ctx, job)).Should(gomega.Succeed())
```
ahg-g, line 444: put this in a "failJob" function

test/integration/jobset_controller_test.go
```diff
@@ -371,48 +382,115 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+		ginkgo.Entry("1 job succeeds 1 job fails, all jobs recreated", []*jobUpdate{
+			{
+				jobUpdateFn: func(jobList *batchv1.JobList) error {
+					ginkgo.By("succeeding a job")
+					job := jobList.Items[0]
+					job.Status.Conditions = append(job.Status.Conditions, batchv1.JobCondition{
+						Type:   batchv1.JobComplete,
+						Status: corev1.ConditionTrue,
+					})
+					gomega.Expect(k8sClient.Status().Update(ctx, &job)).Should(gomega.Succeed())
```
ahg-g, line 459: put this in a `completeJob` func

test/integration/jobset_controller_test.go
```diff
@@ -426,3 +247,163 @@ func checkJobSetStatus(js *jobset.JobSet, condition jobset.JobSetConditionType)
...
+			return -1, err
+		}
+		return len(svcList.Items), nil
+	}).Should(gomega.Equal(numExpectedJobs(js)))
+}
+
+// - 3 replicated jobs, 1 replica each
+func simpleJobSet() *jobset.JobSet {
+	return testing.MakeJobSet("js-succeed", ns.Name).
+		AddReplicatedJob(testing.MakeReplicatedJob("replicated-job-a").
```
ahg-g, line 328: to simplify things and make sure we test cases where replicas > 1, lets have two replicatedJobs here, one with replicas = 1 and one with replicas > 1 and use that in all tests.

test/integration/jobset_controller_test.go
```diff
@@ -426,3 +247,163 @@ func checkJobSetStatus(js *jobset.JobSet, condition jobset.JobSetConditionType)
...
+		var svcList corev1.ServiceList
+		if err := k8sClient.List(ctx, &svcList, client.InNamespace(js.Namespace)); err != nil {
+			return -1, err
+		}
+		return len(svcList.Items), nil
+	}).Should(gomega.Equal(numExpectedJobs(js)))
+}
+
+// - 3 replicated jobs, 1 replica each
+func simpleJobSet() *jobset.JobSet {
```
ahg-g, line 326: return a wrapper so we can use it to add FailurePolicy

test/integration/jobset_controller_test.go
```diff
@@ -426,3 +247,163 @@ func checkJobSetStatus(js *jobset.JobSet, condition jobset.JobSetConditionType)
...
+			SetJob(testing.MakeJob("test-job", ns.Name).SetCompletionMode(batchv1.IndexedCompletion).Obj()).
+			SetEnableDNSHostnames(true).
+			Obj()).
+		Obj()
+}
+
+// - 2 replicated jobs, 3 replicas each
+// - Indexed completion mode
+// - Failure Policy of (operator='Any', restartPolicy='RecreateAll', maxRestarts=1)
+func jobSetFailurePolicyAnyRecreateAll() *jobset.JobSet {
```
ahg-g, line 362: we can remove this func, the test entry can call:

```
simpleJobSet().SetFailurePolicy(&jobset.FailurePolicy{
			Operator:      jobset.TerminationPolicyTargetAny,
			RestartPolicy: jobset.RestartPolicyRecreateAll,
			MaxRestarts:   1,
		}).
		Obj()
```

ahg-g:
> Fixes a bug in the jobset controller causing it to try to recreate existing jobs marked for deletion that had not yet been deleted

Is this really a bug? perhaps we will get creation errors, but I would expect the jobs will eventually be created (the reconciler will get the delete event and that should allow recreating the job successfully)?

danielvegamyhre:
> Is this really a bug? perhaps we will get creation errors, but I would expect the jobs will eventually be created (the reconciler will get the delete event and that should allow recreating the job successfully)?

It does seem like a bug to me even though it doesn't break anything and the jobs are eventually recreated correctly. The previous behavior was sending 2 job creation requests per deleted job: once when it is marked for deletion but still exists (this call will fail),  then once after the deletion occurs and another reconciliation is triggered and it sees the job does not exist and recreates it (this call succeeds). 

My thinking was since we know the first call will alway fail, i don't see a reason to spam the apiserver with bad requests and pollute the logs with error messages that may confuse/mislead users. What do you think?

ahg-g:
A bug means the controller is not behaving as expected, which is not the case here (because the job is eventually created at the right time), but that doesn't mean we shouldn't optimize away those unnecessary calls. 

The reason I asked was mainly to double check my assumptions about the events the controller gets and how it reconciles them.

test/integration/jobset_controller_test.go
```diff
@@ -36,383 +38,202 @@ const (
...
+				{
+					jobUpdateFn:             completeAllJobs,
+					expectedJobSetCondition: jobset.JobSetCompleted,
+				},
+			},
+		}),
+		ginkgo.Entry("jobset validation should fail if DNS hostnames is enabled and job completion mode is not indexed", &testCase{
+			makeJobSet:               jobSetWithDNSHostnamesEnabledNonIndexed,
+			jobSetCreationShouldFail: true,
+		}),
```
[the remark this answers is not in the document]
danielvegamyhre: Moved to top of table and added TODO to create separate Describe+DescribeTable for validation tests.

test/integration/jobset_controller_test.go
danielvegamyhre: Ok the params for this base test jobset are now:

- one replicated job with 1 replica
- one replicated job with 3 replicas and pod DNS hostnames enabled

danielvegamyhre:
@ahg-g I see, makes sense. Anyway I have finished addressing your comments and this is ready for another look

test/integration/jobset_controller_test.go
```diff
@@ -113,8 +130,19 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				}
 			}
 		},
+		// TODO: move validation tests to separate Describe + DescribeTable
+		ginkgo.Entry("jobset validation should fail if DNS hostnames is enabled and job completion mode is not indexed", &testCase{
```
ahg-g, line 134: "validate enableDNSHostnames can't be set if job is not Indexed"

ahg-g (commented):
one minor suggestion

test/integration/jobset_controller_test.go
danielvegamyhre: Done

PR #37, opened 2023-04-18
Add additional labels/annotations, and add unit tests for constructJobsFromTemplate

### LINKED ISSUE #30: Add the Job Index to the labels in the pod template within the job
We currently add the label/annotation on the Job itself only, we should add them to the pod template within the job template so that when the job controller creates the pod they get the job index.

https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L383

we should also add:
```
job.Spec.Template.Labels[jobset.JobIndexLabel] = strconv.Itoa(i)
job.Spec.Template.Annotations[jobset.JobIndexLabel] = strconv.Itoa(i)
```

### LINKED ISSUE #35: Additional job label/annotation for the number of job replicas
`jobset.sigs.k8s.io/job-replicas`: number of replicas in a replicatedJob

danielvegamyhre:
Fixes #35 
Fixes #30

- On every job created from a replicated job, add a label storing the total replica count.
- Add job index as a label and annotation to the pod template spec.
- Refactor `constructJobsFromTemplate` so it is more easily testable and add several unit tests for it.

### DIFF
--- a/api/v1alpha1/jobset_types.go
+++ b/api/v1alpha1/jobset_types.go
@@ -19,10 +19,10 @@ import (
 )
 
 const (
-	JobIndexLabel string = "jobset.sigs.k8s.io/job-index"
-	ReplicasLabel string = "jobset.sigs.k8s.io/job-replicas"
-	RestartsLabel string = "jobset.sigs.k8s.io/restart-attempt"
-	JobNameKey    string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
+	JobIndexKey string = "jobset.sigs.k8s.io/job-index"
+	ReplicasKey string = "jobset.sigs.k8s.io/job-replicas"
+	RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
+	JobNameKey  string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 )
 
 type JobSetConditionType string
--- a/api/v1alpha1/jobset_types.go
+++ b/api/v1alpha1/jobset_types.go
@@ -20,6 +20,7 @@ import (
 
 const (
 	JobIndexLabel string = "jobset.sigs.k8s.io/job-index"
+	ReplicasLabel string = "jobset.sigs.k8s.io/job-replicas"
 	RestartsLabel string = "jobset.sigs.k8s.io/restart-attempt"
 	JobNameKey    string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 )
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -383,16 +383,22 @@ func (r *JobSetReconciler) constructJob(js *jobset.JobSet, rjob *jobset.Replicat
 		Spec: *rjob.Template.Spec.DeepCopy(),
 	}
 
-	// Add restart-attempt count label, it should be equal to jobSet restarts
-	// to indicate is part of the current jobSet run.
+	// Add replica count as label to the job.
+	job.Labels[jobset.ReplicasLabel] = strconv.Itoa(rjob.Replicas)
+
+	// Add restart-attempt count label to the job, it should be equal to
+	// jobSet restarts to indicate is part of the current jobSet run.
 	job.Labels[jobset.RestartsLabel] = strconv.Itoa(js.Status.Restarts)
 
-	// Add job index as a label and annotation.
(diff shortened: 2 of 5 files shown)


api/v1alpha1/jobset_types.go
```diff
@@ -20,6 +20,7 @@ import (
 
 const (
 	JobIndexLabel string = "jobset.sigs.k8s.io/job-index"
+	ReplicasLabel string = "jobset.sigs.k8s.io/job-replicas"
 	RestartsLabel string = "jobset.sigs.k8s.io/restart-attempt"
```
ahg-g, line 24: lets call them keys, not labels

pkg/controllers/jobset_controller.go
```diff
@@ -384,14 +384,23 @@ func (r *JobSetReconciler) constructJob(js *jobset.JobSet, rjob *jobset.Replicat
...
-	// Add job index as a label and annotation.
+	// Add job index as a label and annotation to the job.
 	job.Labels[jobset.JobIndexLabel] = strconv.Itoa(jobIdx)
 	job.Annotations[jobset.JobIndexLabel] = strconv.Itoa(jobIdx)
 
+	// Add job index as a label and annotation to the pod template spec.
+	job.Spec.Template.Labels = cloneMap(job.Spec.Template.Labels)
+	job.Spec.Template.Annotations = cloneMap(job.Spec.Template.Annotations)
+	job.Spec.Template.Labels[jobset.JobIndexLabel] = strconv.Itoa(jobIdx)
+	job.Spec.Template.Annotations[jobset.JobIndexLabel] = strconv.Itoa(jobIdx)
```
ahg-g, line 402: add the replicas too; this can be used by initContainers for jobs like MPI

pkg/controllers/jobset_controller.go
```diff
@@ -384,14 +384,23 @@ func (r *JobSetReconciler) constructJob(js *jobset.JobSet, rjob *jobset.Replicat
 		Spec: *rjob.Template.Spec.DeepCopy(),
 	}
 
-	// Add restart-attempt count label, it should be equal to jobSet restarts
-	// to indicate is part of the current jobSet run.
+	// Add replica count as label to the job.
```
ahg-g, line 387: lets add unit tests for `constructJobsFromTemplate`; to make it easy to test, we can remove it from being a member function of `JobSetReconciler`

danielvegamyhre:
@ahg-g Thanks for the review, I addressed your comments and refactored `constructJobsFromTemplate` so it is more easily testable, and added several unit tests for it.

pkg/controllers/jobset_controller.go
```diff
@@ -336,21 +341,52 @@ func (r *JobSetReconciler) deleteJobs(ctx context.Context, js *jobset.JobSet, jo
...
+// updateStatus updates the status of a JobSet.
+func (r *JobSetReconciler) updateStatus(ctx context.Context, js *jobset.JobSet, eventType, eventReason, eventMsg string) error {
+	if err := r.Status().Update(ctx, js); err != nil {
+		return err
+	}
+	r.Record.Eventf(js, eventType, eventReason, eventMsg)
+	return nil
+}
+
+// TODO: update condition in place if it exists.
```
ahg-g, line 360: we should do this in a follow up PR, this is important

pkg/controllers/jobset_controller_test.go
```diff
@@ -263,3 +265,305 @@ func TestSetExclusiveAffinities(t *testing.T) {
...
+	jobName   string
+	ns        string
+	subdomain string
+	replicas  int
+	jobIdx    int
+	restarts  int
+	exclusive *jobset.Exclusive
+}
+
+func makeJob(args *expectedJobArgs) *batchv1.Job {
```
ahg-g, line 507: I would make this return a job wrapper and remove the subdomain and exclusive args since each are used in a single test.

For example, the subdomain test will look like: `makeJob().Subdomain("test-jobset-replicated-job-0")`; same with the exclusive test.

pkg/controllers/jobset_controller.go
danielvegamyhre: Created #39 and referenced it in this TODO.

pkg/controllers/jobset_controller_test.go
danielvegamyhre: Makes sense, done.

PR #45, opened 2023-04-18
Use jobset and replicated job names as selectors for headless svc

danielvegamyhre:
During manual testing with a distributed pytorch training workload I discovered a bug (pods could not reach each other over the network via hostnames). Upon investigation I realize the generated subdomain name assigned to pods and name of the headless service selecting those pods did not match. 

In addition, in order for this headless service to select pods from all jobs within a replicated job (as was the goal of #42) we need to update the service's label selectors to use jobset name and replicated job name labels, rather than just a single `job-name` label.

### DIFF
--- a/api/v1alpha1/jobset_types.go
+++ b/api/v1alpha1/jobset_types.go
@@ -19,10 +19,12 @@ import (
 )
 
 const (
-	JobIndexKey string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey string = "jobset.sigs.k8s.io/job-replicas"
-	RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
-	JobNameKey  string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
+	JobSetKey        string = "jobset.sigs.k8s.io/jobset"
+	JobIndexKey      string = "jobset.sigs.k8s.io/job-index"
+	ReplicasKey      string = "jobset.sigs.k8s.io/job-replicas"
+	RestartsKey      string = "jobset.sigs.k8s.io/restart-attempt"
+	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicated-job"
+	JobNameKey       string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 )
 
 type JobSetConditionType string
--- a/api/v1alpha1/jobset_types.go
+++ b/api/v1alpha1/jobset_types.go
@@ -19,11 +19,11 @@ import (
 )
 
 const (
-	JobSetKey        string = "jobset.sigs.k8s.io/jobset"
+	JobSetKey        string = "jobset.sigs.k8s.io/jobset-name"
 	JobIndexKey      string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey      string = "jobset.sigs.k8s.io/job-replicas"
+	ReplicasKey      string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	RestartsKey      string = "jobset.sigs.k8s.io/restart-attempt"
-	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicated-job"
+	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicatedjob-name"
 	JobNameKey       string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 )
 
--- a/api/v1alpha1/jobset_types.go
+++ b/api/v1alpha1/jobset_types.go
@@ -19,12 +19,11 @@ import (
 )
 
 const (
-	JobSetKey        string = "jobset.sigs.k8s.io/jobset-name"
-	JobIndexKey      string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey      string = "jobset.sigs.k8s.io/replicatedjob-replicas"
-	RestartsKey      string = "jobset.sigs.k8s.io/restart-attempt"
-	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicatedjob-name"
-	JobNameKey       string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
+	JobSetNameKey         string = "jobset.sigs.k8s.io/jobset-name"
+	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
+	ReplicatedJobNameKey  string = "jobset.sigs.k8s.io/replicatedjob-name"
+	JobIndexKey           string = "jobset.sigs.k8s.io/job-index"
+	JobNameKey            string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 )
 
 type JobSetConditionType string
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -32,6 +32,8 @@ import (
 	jobset "sigs.k8s.io/jobset/api/v1alpha1"
 )
 
+const RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
+
 var (
 	jobOwnerKey = ".metadata.controller"
 	apiGVStr    = jobset.GroupVersion.String()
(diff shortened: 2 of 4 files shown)


api/v1alpha1/jobset_types.go
```diff
@@ -19,10 +19,12 @@ import (
 )
 
 const (
-	JobIndexKey string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey string = "jobset.sigs.k8s.io/job-replicas"
-	RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
-	JobNameKey  string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
+	JobSetKey        string = "jobset.sigs.k8s.io/jobset"
+	JobIndexKey      string = "jobset.sigs.k8s.io/job-index"
+	ReplicasKey      string = "jobset.sigs.k8s.io/job-replicas"
```
ahg-g, line 24: should we name this `replicatedjob-replicas`?

api/v1alpha1/jobset_types.go
```diff
@@ -19,10 +19,12 @@ import (
...
+	RestartsKey      string = "jobset.sigs.k8s.io/restart-attempt"
+	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicated-job"
```
ahg-g, line 26:
```suggestion
	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicatedjob-name"
```

api/v1alpha1/jobset_types.go
```diff
@@ -19,10 +19,12 @@ import (
 )
 
 const (
-	JobIndexKey string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey string = "jobset.sigs.k8s.io/job-replicas"
-	RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
-	JobNameKey  string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
+	JobSetKey        string = "jobset.sigs.k8s.io/jobset"
```
ahg-g, line 22:
```suggestion
	JobSetKey        string = "jobset.sigs.k8s.io/jobset-name"
```

pkg/controllers/jobset_controller.go
```diff
@@ -237,16 +237,17 @@ func (r *JobSetReconciler) createHeadlessSvcIfNotExist(ctx context.Context, js *
 	// Check if service already exists. Service name is same as replicatedJob name.
 	// If the service does not exist, create it.
 	var headlessSvc corev1.Service
-	if err := r.Get(ctx, types.NamespacedName{Name: rjob.Name, Namespace: js.Namespace}, &headlessSvc); err != nil {
+	if err := r.Get(ctx, types.NamespacedName{Name: genSubdomain(js, rjob), Namespace: js.Namespace}, &headlessSvc); err != nil {
```
ahg-g, line 240: `subdomain := genSubdomain(js, rjob)`

pkg/controllers/jobset_controller.go
```diff
@@ -406,6 +407,31 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
...
+
+	// If this job should be exclusive per topology, set the pod affinities/anti-affinities accordingly.
+	if rjob.Exclusive != nil {
+		setExclusiveAffinities(job, rjob.Exclusive.TopologyKey, rjob.Exclusive.NamespaceSelector)
+	}
+
+	return job, nil
+}
+
+func jobLabelsAndAnnotations(js *jobset.JobSet, rjob *jobset.ReplicatedJob, job *batchv1.Job, jobIdx int) {
```
ahg-g, line 428: There should be a way to unify those two functions into one, perhaps we can pass `[runtime.Object](https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/meta.go)`? and use Get/SetLabels and Get/SetAnnotations

also, I would call it `labelAndAnnotateObject`

pkg/controllers/jobset_controller.go
danielvegamyhre: Woops I already caught this and fixed it earlier but didn't push the changes I guess, my bad

pkg/controllers/jobset_controller.go
danielvegamyhre: Good idea! Done

api/v1alpha1/jobset_types.go
```diff
@@ -19,11 +19,11 @@ import (
...
 
 const (
-	JobSetKey        string = "jobset.sigs.k8s.io/jobset"
+	JobSetKey        string = "jobset.sigs.k8s.io/jobset-name"
 	JobIndexKey      string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey      string = "jobset.sigs.k8s.io/job-replicas"
+	ReplicasKey      string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	RestartsKey      string = "jobset.sigs.k8s.io/restart-attempt"
-	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicated-job"
+	ReplicatedJobKey string = "jobset.sigs.k8s.io/replicatedjob-name"
```
ahg-g, line 26: name the constants in a consistent manner: 

JobSetNameKey, ReplicatedJobNameKey etc.

api/v1alpha1/jobset_types.go
```diff
@@ -19,11 +19,11 @@ import (
 )
 
 const (
-	JobSetKey        string = "jobset.sigs.k8s.io/jobset"
+	JobSetKey        string = "jobset.sigs.k8s.io/jobset-name"
 	JobIndexKey      string = "jobset.sigs.k8s.io/job-index"
-	ReplicasKey      string = "jobset.sigs.k8s.io/job-replicas"
+	ReplicasKey      string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	RestartsKey      string = "jobset.sigs.k8s.io/restart-attempt"
```
ahg-g, line 25: just noticed this, I don't think we should put it in the api, this is an implementation detail.

pkg/controllers/jobset_controller.go
```diff
@@ -524,6 +486,11 @@ func shouldCreateJob(jobName string, ownedJobs *childJobs) bool {
 	return true
 }
 
+func labelAndAnnotateObject(obj metav1.Object, labels map[string]string, annotations map[string]string) {
+	obj.SetLabels(labels)
+	obj.SetAnnotations(annotations)
```
ahg-g, line 491: does this override or merge with existing maps? we should merge with existing ones.

pkg/controllers/jobset_controller.go
```diff
@@ -541,6 +508,40 @@ func genSubdomain(js *jobset.JobSet, rjob *jobset.ReplicatedJob) string {
 	return fmt.Sprintf("%s-%s", js.Name, rjob.Name)
 }
 
+func genLabels(js *jobset.JobSet, rjob *jobset.ReplicatedJob, job *batchv1.Job, jobIdx int) map[string]string {
+	labels := make(map[string]string)
+
+	// Add jobset name as label to pod template spec.
```
ahg-g, line 514: all comments in this and the genAnnotations func are not really adding value, the code is self explanatory. I would also just put all of this inside labelAndAnnotate func since they are always used together.

api/v1alpha1/jobset_types.go
danielvegamyhre: Ok I moved the RestartsKey constant to the controller.

pkg/controllers/jobset_controller.go
danielvegamyhre: It overrides but I had already cloned the labels/annotations before calling this so it safe to set, however, I refactored this to simpler and do the cloning+merging in one place.

pkg/controllers/jobset_controller.go
ahg-g: not yet addressed

danielvegamyhre: Sorry I misinterpreted the comment to mean the svc name should be the subdomain rather than the replicatedJob name, which was already the case. I've updated the code to store the subdomain in a variable and reuse it.

Issue #6, opened 2023-04-06
Setup validation and defaulting webhook


ahg-g:

ahg-g:
Using CEL would certainly be better, lets start with that and see how far we can go. Does it cover defaulting too?

ahg-g:
Things to validate:
- EnableDNSHostnames is allowed only with completion mode Indexed.

kannon92:
Should we close this issue?

ahg-g:
Yes

ahg-g:
/reopen

we need to enforce immutability of most fields in the JobSetSpec, I think the only field we should allow for now is suspend

ahg-g:
We may need to setup a defaulting webhook if we are to default things within the job template, like Indexed mode.

PR #49, opened 2023-04-19
Validate job spec is immutable

danielvegamyhre:
Part of #6 

Added CEL validation rules enforcing immutability of the JobSet spec, and added some integration test cases for this.

I also moved the validation tests into their own Describe+DescribeTable.

### DIFF
--- a/config/crd/bases/batch.x-k8s.io_jobsets.yaml
+++ b/config/crd/bases/batch.x-k8s.io_jobsets.yaml
@@ -33,7 +33,6 @@ spec:
           metadata:
             type: object
           spec:
-            description: JobSetSpec defines the desired state of JobSet
             properties:
               failurePolicy:
                 description: FailurePolicy, if set, configures when to declare the
@@ -9515,6 +9514,9 @@ spec:
             required:
             - jobs
             type: object
+            x-kubernetes-validations:
+            - message: JobSet spec is immutable
+              rule: self == oldSelf
           status:
             description: JobSetStatus defines the observed state of JobSet
             properties:
--- a/config/crd/bases/batch.x-k8s.io_jobsets.yaml
+++ b/config/crd/bases/batch.x-k8s.io_jobsets.yaml
@@ -33,6 +33,7 @@ spec:
           metadata:
             type: object
           spec:
+            description: JobSetSpec defines the desired state of JobSet
             properties:
               failurePolicy:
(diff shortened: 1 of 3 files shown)


test/integration/jobset_controller_test.go
```diff
@@ -71,9 +70,8 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 	// jobSetUpdate contains the mutations to perform on the jobset and the
 	// checks to perform afterwards.
 	type jobSetUpdate struct {
-		jobUpdateFn             func(*batchv1.JobList)
-		checkJobSetState        func(*jobset.JobSet)
-		expectedJobSetCondition jobset.JobSetConditionType
+		updateFn         func(*jobset.JobSet)
+		updateShouldFail bool
```
ahg-g, line 74: I would flip this to updateShouldSucceed since currently all cases expect failure, and so we can remove all those `updateShouldFail: true` lines

danielvegamyhre: Makes sense, done

config/crd/bases/batch.x-k8s.io_jobsets.yaml
```diff
@@ -33,7 +33,6 @@ spec:
           metadata:
             type: object
           spec:
-            description: JobSetSpec defines the desired state of JobSet
```
ahg-g, line 36: run make again pls

test/integration/jobset_controller_test.go
danielvegamyhre: Ah i misread and thought you were referring to jobset creation not jobset update, sorry about that. Updated.

test/integration/jobset_controller_test.go
```diff
@@ -71,18 +70,17 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 	// jobSetUpdate contains the mutations to perform on the jobset and the
 	// checks to perform afterwards.
 	type jobSetUpdate struct {
-		jobUpdateFn             func(*batchv1.JobList)
-		checkJobSetState        func(*jobset.JobSet)
-		expectedJobSetCondition jobset.JobSetConditionType
+		updateFn            func(*jobset.JobSet)
+		updateShouldSucceed bool
```
ahg-g, line 74: sorry, another nit: we can remove the update prefix for both parameters, the struct name is update already

PR #51, opened 2023-04-20
Set up defaulting webhook

### LINKED ISSUE #6: Setup validation and defaulting webhook

danielvegamyhre:
Fixes #6 

- Adds defaulting and validation webhook
- Defaults job.spec.completionMode to Indexed if it is unset (includes unit tests and integration tests for this)
- Currently the webhook does no validation; all validation is still done via CEL
- Manually tested by deploying on a cluster and creating a JobSet with a job that has completionMode unset, then verifying the job was created in Indexed completion mode.


test/integration/webhook/suite_test.go
```diff
@@ -0,0 +1,132 @@
...
+		},
+	}
+
+	var err error
+	// cfg is defined in this file globally.
+	cfg, err = testEnv.Start()
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	scheme := runtime.NewScheme()
```
ahg-g, line 76: the problem is here, you are not starting from the scheme that includes the core apis, see example: https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/suite_test.go#L72

api/v1alpha1/jobset_webhook_test.go
```diff
@@ -0,0 +1,82 @@
...
+
+func TestJobSetDefaulting(t *testing.T) {
+	testCases := []struct {
+		name string
+		js   *JobSet
+		want *JobSet
+	}{
+		{
+			name: "job completion mode is unset",
+			js: &JobSet{
```
danielvegamyhre, line 18: I was not able to use wrappers here because it would create a circular import dependency (testing package imports v1alpha1, so v1alpha1 cannot import testing). 

We could consider a refactor to solve this (kueue uses a separate [webhooks](https://github.com/kubernetes-sigs/kueue/tree/main/apis/kueue/webhooks) package and sets up a [CustomerDefaulter](https://github.com/kubernetes-sigs/kueue/blob/main/apis/kueue/webhooks/workload_webhook.go#L35-L64), which allows for more flexibility, contextual logging, etc. but this would require creating a separate webhook struct rather than adding methods to the JobSet struct to implement the Defaulter interface, and it becomes more complicated and didn't seem like it was worth it for this very simple defaulting we are doing.

test/integration/webhook/suite_test.go
danielvegamyhre: Ah that makes sense, great find, thank you. I fixed this and added back the BeforeEach/AfterEach logic creating a separate namespace for each test

README.md
```diff
@@ -8,6 +8,7 @@ To install the CRD and deploy the controller on the cluster selected on your `~/
 ```
 git clone https://github.com/kubernetes-sigs/jobset.git
 cd jobset
+kubectl apply -f https://github.com/cert-manager/cert-manager/releases/download/v1.11.0/cert-manager.yaml
```
ahg-g, line 11: add a note that cert-manager is needed to create certificates for the webhook and link to https://cert-manager.io/docs/installation/

Issue #38, opened 2023-04-18
Add documentation


ahg-g:
Add initial documentation discussing parameters semantics and examples.

kannon92:
Can you add some examples to the documentation?

I usually like to install these operators and have yaml examples to apply to just see how it works.

PR #57, opened 2023-04-24
Added initial structure for docs

ahg-g:
Part of #38

### DIFF
--- a/docs/setup/install.md
+++ b/docs/setup/install.md
@@ -1,0 +1,151 @@
+# Installation
+
+## Before you begin
+
+Make sure the following conditions are met:
+
+- A Kubernetes cluster with version 1.21 or newer is running. Learn how to [install the Kubernetes tools](https://kubernetes.io/docs/tasks/tools/).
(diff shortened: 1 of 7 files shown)


docs/setup/install.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+The controller runs in the `jobset-system` namespace.
+
+### Uninstall
+
+To uninstall JobSet, run the following command:
+
+```shell
+kubectl delete -k github.com/kubernetes-sigs/jobset/config/default
```
danielvegamyhre, line 121: Tried manually testing these commands and for some reason, kubectl -k flag always results in a decoding error for me (same as original issue preventing CRD installation we discussed today):

`error: unable to decode "github.com/kubernetes-sigs/jobset/config/default": parsing time "null" as "2006-01-02T15:04:05Z07:00": cannot parse "null" as "2006"`

We can merge this to test the cloudbuild now but we will need to look into this I think.

PR #64, opened 2023-04-25
Adding concepts page to the docs

ahg-g:
Part of #38 

/hold

### DIFF
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -23,7 +23,7 @@ spec:
            spec:
             containers:
             - name: pytorch
-              image: gcr.io/danielvm-gke-dev2/pytorch-resnet:latest
+              image: gcr.io/k8s-staging-jobset/pytorch-resnet:latest
               ports:
               - containerPort: 3389
               env:
@@ -69,8 +69,8 @@ pytorch-workers-0-3-mn8c8   1/1     Running   0          13m
 
 ## JobSet defaults for Jobs and Pods
 
-- Job [completionMode](https://kubernetes.io/docs/concepts/workloads/controllers/job/#completion-mode) is defaulted to `Indexed`. 
-- Pod [restartPolicy](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-template) is defaulted to `OnFailure`.
+- Job [`completionMode`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#completion-mode) is defaulted to `Indexed` 
+- Pod [`restartPolicy`](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-template) is defaulted to `OnFailure`
 
 
 ## JobSet labels
--- a/docs/concepts/README.md
+++ b/docs/concepts/README.md
@@ -1,4 +1,139 @@
 # Concepts
 
-This section of the documentation helps you learn about the JobSet API and its semantics.
+A JobSet creates one or more Jobs. It allows you to create sets of jobs of different or identical templates, and control their lifecycle together. 
+
+## Running an Example JobSet
+
+Here is an example JobSet. It runs a distributed PyTorch training workload.
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha1
+kind: JobSet
+metadata:
+  name: pytorch
+spec:
+  replicatedJobs:
+    - name: workers
+      template:
+        spec:
+          parallelism: 4
+          completions: 4
+          backoffLimit: 0
+          template:
+           spec:
+            containers:
+            restartPolicy: Never
+            - name: pytorch
+              image: gcr.io/danielvm-gke-dev2/pytorch-resnet:latest
+              ports:
+              - containerPort: 3389
+              env:
+              - name: MASTER_ADDRESS
+                value: "pytorch-workers-0.pytorch-workers"
+              - name: MASTER_PORT
+                value: "3389"
+              command:
+              - bash
+              - -xc
+              - |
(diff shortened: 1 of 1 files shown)


docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+        spec:
+          parallelism: 4
+          completions: 4
+          backoffLimit: 0
+          template:
+           spec:
+            containers:
+            restartPolicy: Never
+            - name: pytorch
+              image: gcr.io/danielvm-gke-dev2/pytorch-resnet:latest
```
danielvegamyhre, line 27: I can probably push this image the jobset GCR repo (k8s-staging-jobset or prod one) then we can update the example to use this public image instead.

docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+pytorch-workers-0-2-5m6d6   1/1     Running   0          13m
+pytorch-workers-0-3-mn8c8   1/1     Running   0          13m
+```
+
+## ReplicatedJob
+
+The list `.spec.replicatedJobs` allows to define groups of Jobs of different templates.
+
+Each entry of `.spec.replicatedJobs` defines a Job template in `spec.replicatedJobs[*].template`, 
+and `spec.replicatedJobs[*].replicas` which is the number replicas that should be created. When
```
danielvegamyhre, line 77:
```suggestion
and the number replicas that should be created in `spec.replicatedJobs[*].replicas`. When 
```

docs/concepts/README.md
ahg-g: btw, any chance we make reduce the size to ensure it runs fast somehow?

docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+## ReplicatedJob
+
+The list `.spec.replicatedJobs` allows to define groups of Jobs of different templates.
+
+Each entry of `.spec.replicatedJobs` defines a Job template in `spec.replicatedJobs[*].template`, 
+and `spec.replicatedJobs[*].replicas` which is the number replicas that should be created. When 
+unset, it is defaulted to 1.
+
+Each Job in each `spec.replicatedJobs` gets a different job-index in the range 0 to `.spec.replicatedJob[*].replicas-1`. 
+The Job name will have the following format: `.metadata.name-.spec.replicatedJobs[*].name-$(job-index)`.
```
danielvegamyhre, line 81: Optional: I personally find the syntax `<jobSetName>-<replicatedJobName>-<jobIndex>` more clear since it is closer to plain English, but if k8s users prefer more technical/specific documentation then the current syntax makes more sense.

docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+
+Here is an example JobSet. It runs a distributed PyTorch training workload.
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha1
+kind: JobSet
+metadata:
+  name: pytorch
+spec:
+  replicatedJobs:
```
danielvegamyhre, line 15: This is still `jobs` in the JobSetSpec API definition, we need to change it to `replicatedJobs`

docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+    - name: workers
+      template:
+        spec:
+          parallelism: 4
+          completions: 4
+          backoffLimit: 0
+          template:
+           spec:
+            containers:
+            restartPolicy: Never
```
danielvegamyhre, line 25:
```suggestion
            restartPolicy: Never
            containers:
```

docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+            - name: pytorch
+              image: gcr.io/danielvm-gke-dev2/pytorch-resnet:latest
+              ports:
+              - containerPort: 3389
+              env:
+              - name: MASTER_ADDRESS
+                value: "pytorch-workers-0.pytorch-workers"
```
danielvegamyhre, line 32:
```suggestion
                value: "pytorch-workers-0-0.pytorch-workers"
```

docs/concepts/README.md
```diff
@@ -1,4 +1,139 @@
...
+          template:
+           spec:
+            containers:
+            restartPolicy: Never
+            - name: pytorch
+              image: gcr.io/danielvm-gke-dev2/pytorch-resnet:latest
+              ports:
+              - containerPort: 3389
+              env:
+              - name: MASTER_ADDRESS
```
danielvegamyhre, line 31:
```suggestion
              - name: MASTER_ADDR
```
name of env var consumed by pytorch is `MASTER_ADDR`
https://pytorch.org/docs/stable/elastic/run.html#environment-variables

docs/concepts/README.md
ahg-g: removed it completely since we now default this

docs/concepts/README.md
ahg-g: done.

docs/concepts/README.md
danielvegamyhre: I was referring to the json/yaml representation, it still says `json:"jobs"`, it needs to change to `json:"replicatedJobs"` for this example to work.

ahg-g: that is a bug we should fix :)

sending a PR in a sec

docs/concepts/README.md
ahg-g: updated the image path

danielvegamyhre: sounds good, once the next PR merges and we get the new build in staging I'll confirm the pytorch example works out the box

ahg-g:
Perhaps we can merge this now?

PR #65, opened 2023-04-25
API cleanup

ahg-g:
While working on the concepts page in https://github.com/kubernetes-sigs/jobset/pull/64, I found it hard to explain configurations that only offer one option, we will likely add those later in the future, but we should do so only when other options are implemented; specifically this PR removes the RestartPolicy and the TargetOperator from FailurePolicy.

This PR also renames .spec.Jobs to .spec.ReplicatedJobs, which I find easier to refer to in the docs.


pkg/controllers/jobset_controller.go
```diff
@@ -277,29 +277,16 @@ func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.
...
-	// Handle different types of failure policy targets.
-	switch js.Spec.FailurePolicy.Operator {
-	case jobset.TerminationPolicyTargetAny:
-		// To reach this point a job must have failed, and TerminationPolicyTargetAny applies to any job
-		// in the JobSet, so we can skip directly to executing the restart policy.
-		return r.executeRestartPolicy(ctx, js, ownedJobs)
-	default:
-		return fmt.Errorf("invalid termination policy target %s", js.Spec.FailurePolicy.Operator)
-	}
+	// To reach this point a job must have failed, and TerminationPolicyTargetAny applies to any job
```
danielvegamyhre, line 280: Nit: given `TerminationPolicyTargetAny` is removed here, this comment might be confusing, maybe consider rephrasing?

ahg-g: good catch, done.

pkg/controllers/jobset_controller.go
```diff
@@ -274,29 +274,15 @@ func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.
...
-	log := ctrl.LoggerFrom(ctx)
-
-	switch js.Spec.FailurePolicy.RestartPolicy {
-	case jobset.RestartPolicyRecreateAll:
-		return r.restartPolicyRecreateAll(ctx, js, ownedJobs)
-	case jobset.RestartPolicyNone:
-		return r.failJobSet(ctx, js)
-	default:
-		log.Error(fmt.Errorf("invalid restart policy: %s", js.Spec.FailurePolicy.RestartPolicy), "invalid restart policy, defaulting to None")
+	if js.Spec.FailurePolicy.MaxRestarts == 0 {
```
danielvegamyhre, line 282: Is this necessary? I think maybe we can skip this and call restartPolicyRecreateAll directly because the first thing it does is check if JobSet has reached max number of restarts, mark it as failed and return, so if MaxRestarts == 0 it will do this immediately.

ahg-g: The message in the status and event is different, that is why I kept it.

PR #66, opened 2023-04-25
Renamed api group from batch.x-k8s.io to jobset.x-k8s.io

ahg-g:
To be consistent with how APIs are created and avoid confusion with the upstream batch group.

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -13,7 +13,7 @@ resources:
     namespaced: true
   controller: true
   domain: x-k8s.io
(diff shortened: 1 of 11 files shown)


danielvegamyhre:
@ahg-g Looks good, has this been tested manually? If not I can do that, just to be safe.

ahg-g:
> @ahg-g Looks good, has this been tested manually? If not I can do that, just to be safe.

yes, tested manually

PR #72, opened 2023-04-25
Migrate remaining references to the old api group

ahg-g:


danielvegamyhre:
Confirmed via manual testing this works

PR #73, opened 2023-04-25
Enable dns hostnames by default if job template uses indexed completion mode

### LINKED ISSUE #62: Default enableDNSHostNames to true
This is a common setting for HPC and AI/ML workloads

danielvegamyhre:
Fixes #62 

Enable DNS hostnames by default if unset, if the replicated job template uses indexed completion mode (we cannot enable DNS hostnames for non-indexed completion mode).

### DIFF
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -15,6 +15,7 @@ package v1alpha1
 
 import (
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/jobset_webhook.go
```diff
@@ -33,12 +34,20 @@ var _ webhook.Defaulter = &JobSet{}
...
 func (r *JobSet) Default() {
-	// Default job completion mode to indexed.
-	for i, rjob := range r.Spec.ReplicatedJobs {
-		if rjob.Template.Spec.CompletionMode == nil {
+	for i, _ := range r.Spec.ReplicatedJobs {
+		// Default job completion mode to indexed.
+		if r.Spec.ReplicatedJobs[i].Template.Spec.CompletionMode == nil {
 			r.Spec.ReplicatedJobs[i].Template.Spec.CompletionMode = completionModePtr(batchv1.IndexedCompletion)
 		}
+		// Enable DNS hostnames by default if job template uses indexed completion mode.
```
ahg-g, line 42: We should always default to true, defaulting one parameter based on another is not a good practice.

api/v1alpha1/jobset_webhook_test.go
```diff
@@ -64,6 +68,37 @@ func TestJobSetDefaulting(t *testing.T) {
...
 								},
 							},
+							Network: &Network{EnableDNSHostnames: pointer.Bool(true)},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "enableDNSHostnames is unset",
```
ahg-g, line 78: do we have a test that sets it explicitly to false?

api/v1alpha1/jobset_webhook.go
danielvegamyhre: We cannot default to true unconditionally, because our validation enforces that enableDNSHostnames cannot be true if completionMode == NonIndexed.

api/v1alpha1/jobset_webhook_test.go
danielvegamyhre: added

api/v1alpha1/jobset_webhook.go
ahg-g: I think we can. The user should be setting the field to false explicitly if they set the completion mode to non-indexed.

PR #76, opened 2023-04-26
Default pod restart policy to OnFailure

### LINKED ISSUE #61: Default restartPolicy
Job only accepts only Never or OnFailure restartPolicy. We can default it to Never since most HPC and ML training workloads don't tolerate failures and can only recover by restarting the whole job

danielvegamyhre:
Fixes #61 

I also bumped the e2e test timeout to avoid flakiness.

### DIFF
--- a/api/v1alpha1/jobset_webhook_test.go
+++ b/api/v1alpha1/jobset_webhook_test.go
@@ -9,13 +9,15 @@ import (
 	"k8s.io/utils/pointer"
 )
 
-// TestPodSpec is the default pod spec used for testing.
-var TestPodSpec = corev1.PodSpec{
-	RestartPolicy: "Never",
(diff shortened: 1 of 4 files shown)


api/v1alpha1/jobset_webhook_test.go
```diff
@@ -21,7 +33,11 @@ func TestJobSetDefaulting(t *testing.T) {
 					ReplicatedJobs: []ReplicatedJob{
 						{
 							Template: batchv1.JobTemplateSpec{
```
kannon92, line 35: Could you create this template once and reuse everywhere?

danielvegamyhre: makes sense, done

PR #80, opened 2023-04-27
Add pytorch example

danielvegamyhre:
Create examples folder and add pytorch example.


config/crd/bases/batch.x-k8s.io_jobsets.yaml
```diff
@@ -1,9597 +0,0 @@
----
```
ahg-g, line 1: oh, this file still exists!

examples/pytorch.yaml
```diff
@@ -0,0 +1,33 @@
...
+    network:
+      enableDNSHostnames: true
+    template:
+      spec:
+        parallelism: 4
+        completions: 4
+        backoffLimit: 0
+        template:
+          spec:
+            restartPolicy: Never
```
ahg-g, line 17: remove this just to showcase our defaulting

examples/pytorch.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: jobset.x-k8s.io/v1alpha1
+kind: JobSet
+metadata:
+  name: pytorch
+spec:
+  replicatedJobs:
+  - name: workers
+    network:
+      enableDNSHostnames: true
```
ahg-g, line 9: remove this since we default it already

examples/pytorch.yaml
```diff
@@ -0,0 +1,33 @@
+apiVersion: jobset.x-k8s.io/v1alpha1
```
ahg-g, line 1: can we create a pytorch folder, I think other examples may require more than one artifact, so better to put each example in a different folder. 

so something like:
examples/pytorch/resnet.yaml

danielvegamyhre: done

examples/pytorch.yaml
danielvegamyhre: done

examples/pytorch/resnet.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: jobset.x-k8s.io/v1alpha1
```
ahg-g, line 1: revert all files but this one since they are merged in the other pr

danielvegamyhre: done

PR #63, opened 2023-04-25
remove CEL validation, validate inside the webhook

charles-chenzz:
try to solve #60 

but might be wrong

PR #89, opened 2023-04-27
require templates in replicatejobs

kannon92:
Discussing with @ahg-g it doesn't really seem to make sense to not require job templates in a replicatedjob so we will require those fields to be set.

### DIFF
--- a/config/crd/bases/jobset.x-k8s.io_jobsets.yaml
+++ b/config/crd/bases/jobset.x-k8s.io_jobsets.yaml
@@ -9492,13 +9492,8 @@ spec:
                       type: object
                   required:
                   - name
+                  - template
                   type: object
-                  x-kubernetes-validations:
-                  - message: EnableDNSHostnames requires job to be in indexed completion
-                      mode
-                    rule: '!has(self.network) || !has(self.network.enableDNSHostnames)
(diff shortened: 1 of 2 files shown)


config/crd/bases/jobset.x-k8s.io_jobsets.yaml
```diff
@@ -9492,13 +9492,8 @@ spec:
                       type: object
                   required:
                   - name
+                  - template
                   type: object
-                  x-kubernetes-validations:
```
danielvegamyhre, line 9496: is this validation removal supposed to be part of this PR? I thought that was done in #63

kannon92: Hmm.  it shouldn't be here but all I did was `make generate manifests` and then it added this?

kannon92: https://github.com/kubernetes-sigs/jobset/pull/63/files doesn't seem to have the manifests files committed so maybe thats why they are included in mine?

danielvegamyhre: ah that must be it, ok cool

PR #96, opened 2023-04-28
Add failure policy with max restarts example

danielvegamyhre:
This example runs a JobSet with a failurePolicy with maxRestarts=3. It has 2 replicated jobs (`worker1` and `worker2`), with 1 replica each.

worker1 has a bash script which will fail pod index 0 after 10 seconds, causing the job worker1 to fail. This will trigger the jobset controller to restart all the jobs. This will process will repeat until maxRestarts (3) is hit.

I thought this would be a helpful example for quickly demonstrating how the failure policy with max restarts works.

### DIFF
--- a/examples/simple/max-restarts.yaml
+++ b/examples/simple/max-restarts.yaml
@@ -13,14 +13,13 @@ spec:
       spec:
         # Set backoff limit to 0 so job will immediately fail if any pod fails.
         backoffLimit: 0 
-        completionMode: Indexed
         completions: 2
         parallelism: 2
         template:
           spec:
             restartPolicy: Never
             containers:
-            - name: worker
+            - name: leader
               image: bash:latest
               # Default failure policy is to recreate all jobs if any jobs fails. 
               # The bash script provides a simple demonstration of it by failing
@@ -49,7 +48,6 @@ spec:
     template:
       spec:
         backoffLimit: 0 
-        completionMode: Indexed
         completions: 2
         parallelism: 2
         template:
--- a/examples/simple/max-restarts.yaml
+++ b/examples/simple/max-restarts.yaml
@@ -7,7 +7,7 @@ spec:
   failurePolicy:
     maxRestarts: 3
   replicatedJobs:
-  - name: worker1
+  - name: leader
     replicas: 1
     template:
       spec:
@@ -44,7 +44,7 @@ spec:
                   echo "$i"
                   sleep 1
                 done
-  - name: worker2
+  - name: workers
     replicas: 1
     template:
       spec:
--- a/examples/simple/max-restarts.yaml
+++ b/examples/simple/max-restarts.yaml
@@ -1,0 +1,64 @@
+apiVersion: jobset.x-k8s.io/v1alpha1
+kind: JobSet
+metadata:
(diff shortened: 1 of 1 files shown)


examples/simple/max-restarts.yaml
```diff
@@ -0,0 +1,64 @@
+apiVersion: jobset.x-k8s.io/v1alpha1
+kind: JobSet
+metadata:
+  name: failurepolicy
+spec:
+  # On failure, restart all jobs up to 3 times.
+  failurePolicy:
+    maxRestarts: 3
+  replicatedJobs:
+  - name: worker1
```
ahg-g, line 10: I would avoid naming different replicatedJobs worker1 and worker2, they should have distinct names to indicate that multiple replicatedJobs are needed only when they are sufficiently different in what they do.

danielvegamyhre: makes sense, done

examples/simple/max-restarts.yaml
```diff
@@ -0,0 +1,64 @@
...
+        # Set backoff limit to 0 so job will immediately fail if any pod fails.
+        backoffLimit: 0 
+        completionMode: Indexed
+        completions: 2
+        parallelism: 2
+        template:
+          spec:
+            restartPolicy: Never
+            containers:
+            - name: worker
```
ahg-g, line 23: leader

examples/simple/max-restarts.yaml
```diff
@@ -0,0 +1,64 @@
...
+  failurePolicy:
+    maxRestarts: 3
+  replicatedJobs:
+  - name: leader
+    replicas: 1
+    template:
+      spec:
+        # Set backoff limit to 0 so job will immediately fail if any pod fails.
+        backoffLimit: 0 
+        completionMode: Indexed
```
ahg-g, line 16: no need

examples/simple/max-restarts.yaml
```diff
@@ -0,0 +1,64 @@
...
+                do
+                  echo "$i"
+                  sleep 1
+                done
+  - name: workers
+    replicas: 1
+    template:
+      spec:
+        backoffLimit: 0 
+        completionMode: Indexed
```
ahg-g, line 52: ditto, no need

examples/simple/max-restarts.yaml
```diff
@@ -13,14 +13,13 @@ spec:
       spec:
         # Set backoff limit to 0 so job will immediately fail if any pod fails.
         backoffLimit: 0 
-        completionMode: Indexed
         completions: 2
         parallelism: 2
         template:
           spec:
             restartPolicy: Never
```
ahg-g, line 20: sorry, just saw this, not need for this

danielvegamyhre: yep i removed it after i saw the other comment

ahg-g: not removed yet :)

danielvegamyhre: oh woops sorry I thought this was in reference to the `completionMode: Indexed` - my bad, I removed the restartPolicy so it can be defaulted.

PR #109, opened 2023-05-04
Refactor suspend/resume logic in Reconcile()

danielvegamyhre:
I'd like to keep the top level entry point to the jobset reconciler (i.e. the `Reconcile()` method) as simple and readable as possible, abstracting away the implementation details of how certain logical steps are performed into lower levels of the code.

This is just a suggestion/preference, we don't necessarily need to merge this. @ahg-g @kannon92 any thoughts?

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -137,25 +137,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, nil
 		}
 	} else {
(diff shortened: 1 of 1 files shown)


kannon92 (commented):
I think this is a good change.  I like small functions.

charles-chenzz:
generally looks good.
little suggestion:
I think function like this can move to a util.go or util dir

https://github.com/kubernetes-sigs/jobset/blob/01ebf9df16d9495429d724d3cfc78b8e87317ff1/pkg/controllers/jobset_controller.go#L584-L619

PR #115, opened 2023-05-05
remove namespace deletion as it is not recommended from kubebuilder

### LINKED ISSUE #114: Namespace deletion in testing is not working as intended.
Generally Kubebuilder does not support deletion of namespaces. https://book.kubebuilder.io/reference/envtest.html#namespace-usage-limitation

https://github.com/kubernetes-sigs/controller-runtime/issues/880

Due to this, I think we should update our tests to delete a namespace.  I ran into some oddness with https://github.com/kubernetes-sigs/jobset/pull/101.  

My proposal would be to remove the cleanup of namespaces and make sure the objects that get created are removed correctly.

kannon92:
Fixes #114 

Remove deletion of namespace and just delete the jobset.

### DIFF
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -64,6 +64,10 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 		}, timeout, interval).Should(gomega.BeTrue())
 	})
 
+	ginkgo.AfterEach(func() {
+		gomega.Expect(util.DeleteNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
+	})
+
 	// jobSetUpdate contains the mutations to perform on the jobset and the
 	// checks to perform afterwards.
 	type jobSetUpdate struct {
@@ -115,7 +119,6 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 					gomega.Expect(k8sClient.Update(ctx, js)).Should(gomega.Succeed())
 				}
 			}
-
 			// Let's delete the jobset
 			gomega.Expect(k8sClient.Delete(ctx, js)).Should(gomega.Succeed())
 			gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).ShouldNot(gomega.Succeed())
@@ -189,6 +192,10 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 			}
 			return true
 		}, timeout, interval).Should(gomega.BeTrue())
+	})
+
+	ginkgo.AfterEach(func() {
+		gomega.Expect(util.DeleteNamespace(ctx, k8sClient, ns)).Should(gomega.Succeed())
 	})
 
 	// update contains the mutations to perform on the jobs/jobset and the
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -119,10 +119,6 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 					gomega.Expect(k8sClient.Update(ctx, js)).Should(gomega.Succeed())
 				}
 			}
-			// Let's delete the jobset
-			gomega.Expect(k8sClient.Delete(ctx, js)).Should(gomega.Succeed())
-			gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).ShouldNot(gomega.Succeed())
-
 		},
 		ginkgo.Entry("validate enableDNSHostnames can't be set if job is not Indexed", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -64,11 +64,6 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 		}, timeout, interval).Should(gomega.BeTrue())
 	})
 
-	ginkgo.AfterEach(func() {
-		// Delete test namespace after each test.
-		gomega.Expect(k8sClient.Delete(ctx, ns)).To(gomega.Succeed())
-	})
-
 	// jobSetUpdate contains the mutations to perform on the jobset and the
 	// checks to perform afterwards.
 	type jobSetUpdate struct {
@@ -120,6 +115,11 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 					gomega.Expect(k8sClient.Update(ctx, js)).Should(gomega.Succeed())
 				}
 			}
+
+			// Let's delete the jobset
+			gomega.Expect(k8sClient.Delete(ctx, js)).Should(gomega.Succeed())
+			gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).ShouldNot(gomega.Succeed())
+
 		},
 		ginkgo.Entry("validate enableDNSHostnames can't be set if job is not Indexed", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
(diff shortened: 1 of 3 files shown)


test/integration/controller/jobset_controller_test.go
```diff
@@ -64,11 +64,6 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 		}, timeout, interval).Should(gomega.BeTrue())
 	})
 
-	ginkgo.AfterEach(func() {
-		// Delete test namespace after each test.
-		gomega.Expect(k8sClient.Delete(ctx, ns)).To(gomega.Succeed())
-	})
```
ahg-g, line 70: as I mentioned on the issue, I think it is better to have a DeleteNamespace func that lists the deletes the objects we know get created as part of those integration tests, we did that in kueue: https://github.com/kubernetes-sigs/kueue/blob/main/test/util/util.go#L79

kannon92: Sounds good.  I'll add that.

test/util/util.go
```diff
@@ -51,3 +55,24 @@ func CheckNumJobs(ctx context.Context, k8sClient client.Client, js *jobset.JobSe
 	}
 	return len(jobList.Items), nil
 }
+
+// DeleteNamespace deletes all objects the tests typically create in the namespace.
+func DeleteNamespace(ctx context.Context, c client.Client, ns *corev1.Namespace) error {
```
danielvegamyhre, line 60: Looks good, you need to call this function in AfterEach() though

test/integration/controller/jobset_controller_test.go
```diff
@@ -120,6 +119,11 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 					gomega.Expect(k8sClient.Update(ctx, js)).Should(gomega.Succeed())
 				}
 			}
+
+			// Let's delete the jobset
+			gomega.Expect(k8sClient.Delete(ctx, js)).Should(gomega.Succeed())
```
danielvegamyhre, line 124: Shouldn't the `util.DeleteNamespace()` in `AfterEach()` handle this already? I see it deletes all jobsets.

kannon92: good call. updated.

test/integration/controller/jobset_controller_test.go
```diff
@@ -257,6 +255,10 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 					gomega.Eventually(util.CheckJobSetStatus, timeout, interval).WithArguments(ctx, k8sClient, js, up.expectedJobSetCondition).Should(gomega.Equal(true))
 				}
 			}
+			// Let's delete the jobset
+			gomega.Expect(k8sClient.Delete(ctx, js)).Should(gomega.Succeed())
```
danielvegamyhre, line 259: Same here, the util.DeleteNamespace() in AfterEach() should delete the jobset already, no need for this deletion here.

kannon92:
/hold

I think there is a bug with deletion of jobs as I sometimes see that the jobs are not being cleaned up even though I tell them too.

kannon92:
Since deletion does not really happen I think we may need to use unique names for the objects.  

From envtest:

```
One common example is garbage collection; because there are no controllers monitoring built-in resources, objects do not get deleted, even if an OwnerReference is set up.
```

I am sometimes getting an error saying that a job exists so I don't think deleting the jobs is actually working.

danielvegamyhre:
> Since deletion does not really happen I think we may need to use unique names for the objects.

I set up the integration tests to create a unique namespace for each test in order to get around this problem. I will take a look at this later as well and see if I can figure out what is going on.

kannon92:
> > Since deletion does not really happen I think we may need to use unique names for the objects.
> 
> I set up the integration tests to create a unique namespace for each test in order to get around this problem. I will take a look at this later as well and see if I can figure out what is going on.

Let me push up what I have.  

I don't think deletion of k8 objects works (Jobs, Services).  Once I commented the deletion out, things seem to behave a bit better.  

Its still a WIP but if you want to take a look, I pushed up what I have.  Its EOW for me so I'll tackle this next week.

kannon92:
@danielvegamyhre I'm not quite sure what to do with this PR at the moment.  

I am still dealing with a few errors but not sure they are related to this PR:

```
2023-05-08T19:26:12Z  ERROR   Reconciler error        {"controller": "jobset", "controllerGroup": "jobset.x-k8s.io", "controllerKind": "JobSet", "JobSet": {"name":"test-js","namespace":"test-ns-btc88"}, "namespace": "test-ns-btc88", "name": "test-js", "reconcileID": "1255663e-6e79-4442-93f6-ea6638188ee6", "error": "Operation cannot be fulfilled on jobs.batch \"test-js-replicated-job-b-1\": StorageError: invalid object, Code: 4, Key: /registry/jobs/test-ns-btc88/test-js-replicated-job-b-1, ResourceVersion: 0, AdditionalErrorMsg: Precondition failed: UID in precondition: 4092259b-47f9-499b-b92f-23819f6fa635, UID in object meta: "}
```

This seems to occur now when I try to delete the jobs in `DeleteNamespace`.  It seems to be a known flaky issue but it seems to resolve itself.  I'm not sure why we can have failures in the logs but the tests are successful.  So maybe this is just happening because we are calling it with `Eventually` so it may fail but it overall passes?  

The other issue I see is 

```
 2023-05-08T19:35:58Z  ERROR   creating jobs   {"controller": "jobset", "controllerGroup": "jobset.x-k8s.io", "controllerKind": "JobSet", "JobSet": {"name":"test-js","namespace":"test-ns-fcq8l"}, "namespace": "test-ns-fcq8l", "name": "test-js", "reconcileID": "1693076b-ebbc-455a-b96e-8951518b3e82", "jobset": {"name":"test-js","namespace":"test-ns-fcq8l"}, "error": "jobs.batch \"test-js-replicated-job-b-2\" already exists"}
 ```
 
I think you can see most of these problems if you try running `make test-integration` a few times.

**Comment by danielvegamyhre** [comment]
@kannon92 I haven't seen the first error before, where the object UID in precondition doesn't match the (empty) UID in object meta, that is super weird. I've never gotten `DeleteAllOf` to work for any type of object throughout the process of building this project, so I had given up trying to use it.

The second error causing by redundant job creations is a known issue tracked in #41, but it is just an extra job creation attempt that can be optimized away, it doesn't break the underlying expected behavior.

**Comment by ahg-g** [comment]
> This seems to occur now when I try to delete the jobs in DeleteNamespace.

@kannon92 is this verified? is this error really the result of the delete we are doing inside DeleteNamespace? (it would be great to verify that it is not from the controller)

> The second error caused by redundant job creations is a known issue tracked in https://github.com/kubernetes-sigs/jobset/issues/41, but it is just an extra job creation attempt that can be optimized away, it doesn't break the underlying expected behavior.

@danielvegamyhre do we know why we have that extra call?

**Comment by kannon92** [comment]
> > This seems to occur now when I try to delete the jobs in DeleteNamespace.
> 
> @kannon92 is this verified? is this error really the result of the delete we are doing inside DeleteNamespace? (it would be great to verify that it is not from the controller)
> 

https://github.com/kubernetes-sigs/controller-runtime/issues/1881 seems to be relevant issue but I’m not sure of the solution.

I know that without this change I get a lot of the namespace deletion errors.  Note this really only happens if I try to run the test a bunch so not sure the solution for that case.  Maybe I’ll try with a real Jobset and see if we have some cleanup issues?

> > The second error caused by redundant job creations is a known issue tracked in #41, but it is just an extra job creation attempt that can be optimized away, it doesn't break the underlying expected behavior.
> 
> @danielvegamyhre do we know why we have that extra call?

File: test/util/util.go [hunk 5]
```diff
@@ -110,3 +114,24 @@ func checkJobSetStatus(ctx context.Context, k8sClient client.Client, js *jobset.
...
+	if ns == nil {
+		return nil
+	}
+	if err := c.DeleteAllOf(ctx, &jobset.JobSet{}, client.InNamespace(ns.Name)); err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
```
ahg-g, line 125: Regarding the delete error messages, I think the issue is that the api-server is trying to delete the child jobs in the background and we try to delete them below, so perhaps there is a race here. I suggest that we do foreground delete here so that we are sure no more jobs owned by JobSet are left before continuing to delete the remaining jobs (I don't think we have tests creating jobs directly, but perhaps we can do this just in case we add one in the future).

test/util/util.go
```diff
@@ -110,3 +114,24 @@ func checkJobSetStatus(ctx context.Context, k8sClient client.Client, js *jobset.
...
+	err := c.DeleteAllOf(ctx, &batchv1.Job{}, client.InNamespace(ns.Name), client.PropagationPolicy(metav1.DeletePropagationBackground))
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+	if err := c.DeleteAllOf(ctx, &corev1.Service{}, client.InNamespace(ns.Name)); err != nil && !apierrors.IsNotFound(err) {
```
ahg-g, line 130: lets change to foreground delete here and for jobs above too since we are expecting to be able to delete the namespace at the end.

kannon92:
/hold cancel

I think the `Foreground` change really helped here.

PR #119, opened 2023-05-05
Fix default success policy

### LINKED ISSUE #117: Default jobset success policy doesn't apply to replicatedJobs with more than 1 replica
Currently the success policy is configured to mark a jobset as succeeded if the number of completed jobs == number of replicatedJobs (see [here](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L114)). This only works if the replicatedJobs only have 1 replica each. We need to update this to handle the case where each replicatedJob has an arbitrary number of replicas and we want all jobs to complete before marking the jobset as succeeded.

/king bug

danielvegamyhre:
Fixes #117

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -112,6 +112,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	// If all jobs have succeeded, JobSet has succeeded.
+	// TODO(#120): ensure that the successful jobs counted are actually the ones that we expect JobSet to have
 	if len(ownedJobs.successful) == totalJobs(&js) {
 		if err := r.ensureCondition(ctx, &js, corev1.EventTypeNormal, metav1.Condition{
 			Type:    string(jobset.JobSetCompleted),
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -112,7 +112,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	// If all jobs have succeeded, JobSet has succeeded.
-	if len(ownedJobs.successful) == len(js.Spec.ReplicatedJobs) {
+	if len(ownedJobs.successful) == totalJobs(&js) {
 		if err := r.ensureCondition(ctx, &js, corev1.EventTypeNormal, metav1.Condition{
 			Type:    string(jobset.JobSetCompleted),
 			Status:  metav1.ConditionStatus(corev1.ConditionTrue),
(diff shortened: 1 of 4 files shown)


ahg-g:
I am ok with this fix for now just so we backport a smaller patch, but i think we need to ensure that the successful jobs we counted are actually the ones that we expect JobSet to have (e.g., we shouldn't count jobs marked as owner by the JobSet, but not part of any ReplicatedJob, this could happen if someone creates a job out-of-band and manually sets the owner reference to an existing jobset)

danielvegamyhre:
> I am ok with this fix for now just so we backport a smaller patch, but i think we need to ensure that the successful jobs we counted are actually the ones that we expect JobSet to have (e.g., we shouldn't count jobs marked as owned by the JobSet, but not part of any ReplicatedJob, this could happen if someone creates a job out-of-band and manually sets the owner reference to an existing jobset)

I'll add a TODO for this

pkg/controllers/jobset_controller.go
```diff
@@ -112,7 +112,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	// If all jobs have succeeded, JobSet has succeeded.
-	if len(ownedJobs.successful) == len(js.Spec.ReplicatedJobs) {
+	if len(ownedJobs.successful) == totalJobs(&js) {
```
ahg-g, line 115: Add an integration test pls

danielvegamyhre: Added

test/integration/controller/jobset_controller_test.go
```diff
@@ -267,6 +267,20 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			makeJobSet: testJobSet,
+			updates: []*update{
+				{
+					jobUpdateFn: func(jobList *batchv1.JobList) {
+						ginkgo.By("completing all but 1 job")
+						for i := 0; i < len(jobList.Items)-1; i++ {
+							completeJob(&jobList.Items[i])
+						}
+					},
+					expectedJobSetCondition: "", // active
```
ahg-g, line 280: The test skips doing the check when the condition is not set. Also, the check currently does "Eventually", in this case it needs to be "Consistently" for few seconds to ensure that the controller indeed didn't attempt to complete the jobset (i.e., the test didn't just check the condition before the controller tried to set it)

Side note: Please ensure that the test fails without the fix to verify that we are properly covering the bug.

danielvegamyhre: I addressed this issue and also refactored the way our integration+e2e tests check JobSet statuses to be more flexible and use shared logic.

Also updated e2e bash script to reduce flakiness.

Let me know if you have any thoughts.

Issue #55, opened 2023-04-24
Provide the component config


tenzen-y:
I would like to provide the component config like kueue: https://github.com/kubernetes-sigs/kueue/blob/9ca57c86cf06c11a94a2d5b7badf60233a51a2f2/config/components/manager/controller_manager_config.yaml

Maybe, as a first step, we can provide the functionality for users to select whether to use the internal cert or the cert-manager.

/kind feature

ref: #52

tenzen-y:
The component-config is deprecated and will be removed. So we should hold similar schemes as component-config in this repository.

ref: https://github.com/kubernetes-sigs/kueue/issues/725

tenzen-y:
Maybe, I don't have enough time to work on this.

/unassign

charles-chenzz:
I can take it up. do you have any idea on how to implement it or any examples that I can refer to? @tenzen-y 
/assign

charles-chenzz:
can you be more specify about the ref and implement? I think adding this feature will require us to use code-gen to gen the file and implement the function. then there comes a the key question is: do we actually need this? and are there any more scenes that might need this api/feature?(beside internal-cert and cert manager) @tenzen-y 

also /cc @ahg-g  for advice

ahg-g:
> an you be more specify about the ref and implement? I think adding this feature will require us to use code-gen to gen the file and implement the function.

see https://book.kubebuilder.io/component-config-tutorial/tutorial.html

> and are there any more scenes that might need this api/feature?(beside internal-cert and cert manager)
kube connection configuration is important to have to set qps limits, similar to https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta1/configuration_types.go#L57.

tenzen-y:
> do we actually need this? and are there any more scenes that might need this api/feature?

If we introduce the component config, we can set up many parameters for the controller besides the cert.

Note: ComponentConfig is deprecated. So we must implement all features of the https://github.com/kubernetes-sigs/controller-runtime/tree/30dea34848b0e0b30690dd461a0eadb74d4869f3/pkg/config in this repo.

charles-chenzz:
> Note: ComponentConfig is deprecated. So we must implement all features of the https://github.com/kubernetes-sigs/controller-runtime/tree/30dea34848b0e0b30690dd461a0eadb74d4869f3/pkg/config in this repo.

thanks for the heads up, then I need to check this link and see how to implement it.

[...]

PR #123, opened 2023-05-07
Limit the number of goroutines used when deleting jobs

tenzen-y:
I limited the number of goroutines used when deleting Jobs.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -15,6 +15,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"sync"
@@ -25,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -33,7 +35,10 @@ import (
 	jobset "sigs.k8s.io/jobset/api/v1alpha1"
 )
 
-const RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
+const (
+	RestartsKey       string = "jobset.sigs.k8s.io/restart-attempt"
+	parallelDeletions int    = 50
+)
 
 var (
 	jobOwnerKey = ".metadata.controller"
@@ -97,7 +102,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	// Delete any jobs marked for deletion.
-	if err := r.deleteJobs(ctx, &js, ownedJobs.delete); err != nil {
+	if err := r.deleteJobs(ctx, ownedJobs.delete); err != nil {
 		log.Error(err, "deleting jobs")
 		return ctrl.Result{}, err
 	}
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -364,31 +369,23 @@ func (r *JobSetReconciler) restartPolicyRecreateAll(ctx context.Context, js *job
...
-		job := job
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
-
-			// Delete job. This deletion event will trigger another reconcilliation,
-			// where the jobs are recreated.
-			backgroundPolicy := metav1.DeletePropagationBackground
-			if err := r.Delete(ctx, job, &client.DeleteOptions{PropagationPolicy: &backgroundPolicy}); client.IgnoreNotFound(err) != nil {
-				finalErr = err
```
tenzen-y, line 385: As a potential issue, a race condition may occur.

pkg/controllers/jobset_controller.go
```diff
@@ -364,31 +369,23 @@ func (r *JobSetReconciler) restartPolicyRecreateAll(ctx context.Context, js *job
 	return nil
 }
 
-func (r *JobSetReconciler) deleteJobs(ctx context.Context, js *jobset.JobSet, jobsForDeletion []*batchv1.Job) error {
```
tenzen-y, line 367: This function doesn't use the `JobSet`.

pkg/controllers/jobset_controller.go
```diff
@@ -364,31 +369,23 @@ func (r *JobSetReconciler) restartPolicyRecreateAll(ctx context.Context, js *job
...
-		}()
-	}
-	wg.Wait()
+	workqueue.ParallelizeUntil(ctx, parallelDeletions, len(jobsForDeletion), func(i int) {
+		targetJob := jobsForDeletion[i]
+		// Delete job. This deletion event will trigger another reconciliation,
+		// where the jobs are recreated.
+		backgroundPolicy := metav1.DeletePropagationBackground
+		if err := r.Delete(ctx, targetJob, &client.DeleteOptions{PropagationPolicy: &backgroundPolicy}); client.IgnoreNotFound(err) != nil {
+			lock.Lock()
```
danielvegamyhre, line 382: Why are you only acquiring the lock if an error occurs deleting the job (ignoring 'not found' errors)?

tenzen-y: Because this lock is only for `finalErr`.

Let me know if I'm missing anything.

pkg/controllers/jobset_controller.go
```diff
@@ -364,31 +369,23 @@ func (r *JobSetReconciler) restartPolicyRecreateAll(ctx context.Context, js *job
...
+			defer lock.Unlock()
+			finalErr = errors.Join(finalErr, err)
```
tenzen-y, line 384: Oh, `errors.Join()` is introduced since Go 1.20, although this project uses Go 1.19.

tenzen-y: @ahg-g @danielvegamyhre I'd like to update the Go version to 1.20 in another PR before merging this one into the main. 

Does bumping the Go version to v1.20 sound good?

danielvegamyhre: Sounds good to me, let's wait for Abdullah to confirm as well though.

pkg/controllers/jobset_controller.go
danielvegamyhre: Ah I see now. Thanks for clarifying.

pkg/controllers/jobset_controller.go
tenzen-y: Sure :)

kannon92: I will say that I remember upgrading the e2e tests in test infra to 1.20

tenzen-y: > I will say that I remember upgrading the e2e tests in test infra to 1.20

That makes sense.

charles-chenzz: > @ahg-g @danielvegamyhre I'd like to update the Go version to 1.20 in another PR before merging this one into the main.
> 
> Does bumping the Go version to v1.20 sound good?

@tenzen-y  We can do update it like k/k repos, adding a .goversion file? or we have other ways? just curious

tenzen-y: @charles-chenzz We can update it by just bumping in go.mod. And then, we need to update it in k/test-infra as mentioned by @kannon92. However, we don't decide to bump it yet.

https://github.com/kubernetes-sigs/jobset/blob/1ef251cb5e72a289e540883559f2c2bc28191178/go.mod#L3

kannon92: I meant that I think we are already using a go1.20 for the e2e tests

tenzen-y: Thanks!

kannon92: https://github.com/kubernetes-sigs/jobset/pull/129

tenzen-y:
Squashed into one and rebased.

pkg/controllers/jobset_controller.go
```diff
@@ -33,7 +35,10 @@ import (
 	jobset "sigs.k8s.io/jobset/api/v1alpha1"
 )
 
-const RestartsKey string = "jobset.sigs.k8s.io/restart-attempt"
+const (
+	RestartsKey       string = "jobset.sigs.k8s.io/restart-attempt"
+	parallelDeletions int    = 8
```
ahg-g, line 40: isn't this too low? any idea what do we set this number to in other places like the Job API?

ahg-g (commented):
Thanks Yuki, I think we need to parallelize job creation too, I created an issue: https://github.com/kubernetes-sigs/jobset/issues/130

pkg/controllers/jobset_controller.go
tenzen-y: > isn't this too low?

It makes sense. But I'm not sure appropriate value. How about 100 as a default value? Or, do you have any good idea?

> any idea what do we set this number to in other places like the Job API?

How about setting it via config: #55?

tenzen-y: Or, does that mean like https://github.com/kubernetes/kubernetes/blob/5fd29596ef699a8658c941c1c7cf7a45e796c1c6/pkg/controller/job/job_controller.go#L1458-L1466C63?

danielvegamyhre: I like the concept of a "slow start" in the linked code from the job controller. Maybe it would be simpler to start with some constant value like 100 then implement slow start separately?

ahg-g: I guess for deletions we can continue to use a constant (I would set it to 50); but adopt a slow start approach for parallel creation.

tenzen-y: > I guess for deletions we can continue to use a constant (I would set it to 50); but adopt a slow start approach for parallel creation.

Great suggestions! I agree.

tenzen-y:
@ahg-g @danielvegamyhre I have updated.

ahg-g:
> /test pull-jobset-test-e2e-main-1-24

any idea why this failed?

tenzen-y:
> > /test pull-jobset-test-e2e-main-1-24
> 
> any idea why this failed?

Actually, I created #135.

PR #125, opened 2023-05-07
Pass a context from controller to the IndexerFunc

tenzen-y:
I modified the func `SetupIndexes` to pass a context from the controller to the IndexerFunc.

### DIFF
--- a/main.go
+++ b/main.go
@@ -96,7 +96,8 @@ func main() {
 		os.Exit(1)
 	}
 
-	if err := controllers.SetupIndexes(mgr.GetFieldIndexer()); err != nil {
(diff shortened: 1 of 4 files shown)


danielvegamyhre:
Is the purpose of this to have a shared context for contextual logging?

tenzen-y:
> Is the purpose of this to have a shared context for contextual logging?

In general, I think we should pass a context generated in the parent to the child function instead of using an empty context (another parent context).

So, we can propagate canceling to the child.

danielvegamyhre:
/lgtm

Will leave approval to Abdullah

tenzen-y:
Rebased.

Issue #59, opened 2023-04-24
Run Job validation on the templates as part of JobSet validation


ahg-g:
This is a better experience vs catching such problems at runtime when jobset attempts to create the job.

kannon92:
Do you know how you should do validation?  I know k/k has a validation package but not as clear what users of api libraries have.

kannon92:
@ahg-g AFAIK we tried this but found that there isn’t really a way to use job code for validation. I think we can close this issue.

/close

PR #128, opened 2023-05-08
WIP: run job validation on the templates as part of jobset validation

charles-chenzz:
a solution for #59 
need to discuss if this is what we want
also do some tiny refactor(for code readability)


api/v1alpha1/jobset_webhook.go
```diff
@@ -36,21 +37,22 @@ var _ webhook.Defaulter = &JobSet{}
 
 // Default implements webhook.Defaulter so a webhook will be registered for the type
 func (r *JobSet) Default() {
-	for i, _ := range r.Spec.ReplicatedJobs {
+	replicatedJob := r.Spec.ReplicatedJobs
+	for i := range replicatedJob {
```
danielvegamyhre, line 41: I don't think we should put this in a separate variable, `r.Spec.ReplicatedJobs` (plural) is sufficiently concise and descriptive. It is more clear than `replicatedJob` (singular).

danielvegamyhre:
@charles-chenzz Is it possible to simply use this exported function to do the validation https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/validation/validation.go#L147

charles-chenzz:
> @charles-chenzz Is it possible to simply use this exported function to validate the JobTemplateSpec https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/validation/validation.go#L546

yes, I think we can try this exported function, will update later.

kannon92:
> > @charles-chenzz Is it possible to simply use this exported function to validate the JobTemplateSpec https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/validation/validation.go#L546
> 
> yes, I think we can try this exported function, but we need to pass a `field path` params to it.

Is that a good idea?  I thought when you want to use kubernetes APIs as a library you should use versioned APIs.  

It would probably be worth seeing what options are available for validation of JobTemplates and maybe we should move that to a `staging` as I can see it being useful but I think we would want a versioned library rather than a function at head of k/k.

ahg-g:
Right, we can't replicate the validation logic here, if k8s doesn't export a validation func, then unfortunately we are out of luck, and best we can do is log an event when job creation fails.

charles-chenzz:
yes, that's what I was concerned and why I didn't use that exported function at first, cus it need to pass some parameters that we currently didn't implement in jobset. I think there's still have options
- We create ``pkg/util/validation``, we implement validation logic here (all base on jobset validate logic, and don't rely on k/k function)
- discuss when and how we log an event when job creations fails.

charles-chenzz:
do we still want to run job validation on the template?

ahg-g:
> do we still want to run job validation on the template?

I think we can close this for now until upstream k8s provide an external library for validation functions.

PR #139, opened 2023-05-11
Relax validation on enableDNSHostnames

### LINKED ISSUE #138: Allow setting enableDNSHostnames for non-indexed jobs

Allow setting enableDNSHostnames for non-indexed jobs

**Why is this needed**:

See discussion https://github.com/kubernetes-sigs/jobset/issues/132#issuecomment-1543348063

danielvegamyhre:
Fixes #138

### DIFF
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -14,7 +14,6 @@ limitations under the License.
 package v1alpha1
 
 import (
-	"fmt"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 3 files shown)


danielvegamyhre:
> I think there were a few test cases in jobset_controller_test.go that we should also remove.
> 
> https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L123

Am I crazy or does this test case not do what the test description says it does? If so, how did this happen...?

kannon92:
Github blame says it was me!  https://github.com/kubernetes-sigs/jobset/pull/20

kannon92:
Let me know how you want to handle this test case.

> The jobset is created with a job in NonIndexed completion mode and EnableDNSHostnames=true, but it is configured to expected the jobset creation to succeed.

I don't think we ever do the validation in the jobset code as it is enforced for webhooks.  With this PR I think you'd want to remove this test case anyway as we now allow `enableDNSHostnames` for both.

danielvegamyhre:
> Github blame says it was me! #20

No worries! I am un-merging the test cases in this PR.

kannon92 (approved):
lgtm

Issue #134, opened 2023-05-10
Setting NodeAffinity foreach Job during unsuspend


mcariatm:
**What would you like to be added**:
 Node selectors should be updated on each of jobs based on ReplicatedJob template during unsuspend.
**Why is this needed**:
On suspend and unsuspend process the Node Selectors can be replaced by Kueue Operator.
The JobSet Operator should knows and updates each job before starting them.

alculquicondor:
wdyt @ahg-g @danielvegamyhre?

alculquicondor:
ref https://github.com/kubernetes-sigs/kueue/issues/738

ahg-g:
I think it makes sense, note that currently the template is immutable, so we need to relax that.

danielvegamyhre:
Sounds good to me. On the JobSet side, I think the work can be divided into two tasks:

1. Relax immutability validation for `.spec.replicatedJobs[*].template`
2. Add logic to jobset operator to reconcile the nodeSelectors on the replicatedJob template and the jobs created from that template.

mcariatm:
BTW I propose to add StartTime and PodsReady fields in status.  Any concerns?

alculquicondor:
Can you use conditions and their timestamps?

mcariatm:
Ok, but in on  ResetStatus() should I reset timestamp from condition? how about PodsReady?

alculquicondor:
The API practice is that the timestamp should be reset when there is a transition from true to false or false to true. Is this possible?

alculquicondor:
@kannon92 can you provide some guidance?

kannon92:
> @kannon92 can you provide some guidance?

So my concern is that JobSets control ReplicatedJobs so I don't really understand the `PodsReady`.  We have an issue in k/k to discuss adding a condition when Jobs are Ready (https://github.com/kubernetes/kubernetes/issues/117758).  I think it isn't clear me what this means with replicatedjobs.  For example, you could have 1 replicatedjob fully ready while another one is still starting? 

I would defer to @ahg-g or @danielvegamyhre though.  

> The API practice is that the timestamp should be reset when there is a transition from true to false or false to true. Is this possible?

For Suspend condition for Jobset we do this:

```
func (r *JobSetReconciler) suspendJobSet(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
	for _, job := range ownedJobs.active {
[...]

PR #143, opened 2023-05-16
Set node affinity during unsuspend

mcariatm:
Node selectors should be updated on each of jobs based on ReplicatedJob template during unsuspend.
### Why is this needed:
On suspend and unsuspend process the Node Selectors can be replaced by Kueue Operator.

### Issue Number:
#134

### DIFF
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -16,6 +16,7 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
+	"reflect"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/pointer"
@@ -81,7 +82,15 @@ func (js *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (js *JobSet) ValidateUpdate(old runtime.Object) error {
-	return nil
+	oldObj := old.(*JobSet)
+	var allErr []error
+	for index := range js.Spec.ReplicatedJobs {
+		oldObj.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = r.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
+		if !reflect.DeepEqual(oldObj.Spec.ReplicatedJobs, js.Spec.ReplicatedJobs) {
+			allErr = append(allErr, fmt.Errorf("the spec.ReplicatedJobs value is imutable excepting Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector"))
+		}
+	}
+	return errors.Join(allErr...)
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -18,8 +18,10 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/google/go-cmp/cmp"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
@@ -83,23 +85,22 @@ func (js *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (js *JobSet) ValidateUpdate(old runtime.Object) error {
-	oldObjCopy := old.DeepCopyObject().(*JobSet)
-	if pointer.BoolDeref(oldObjCopy.Spec.Suspend, false) {
-		return nil
-	}
-	var allErr []error
+	mungedSpec := js.Spec.DeepCopy()
+	oldSpec := old.(*JobSet).Spec
 	var updatableFields = []string{
-		"`template.spec.template.spec.nodeSelector`",
+		"`template.spec.template.spec.nodeSelector` when jobset is suspended",
 	}
-	for index := range js.Spec.ReplicatedJobs {
-		oldObjCopy.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = js.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
-		if !apiequality.Semantic.DeepEqual(oldObjCopy.Spec.ReplicatedJobs, js.Spec.ReplicatedJobs) {
-			allErr = append(allErr, fmt.Errorf(
-				fmt.Sprintf("%s update validation: spec.ReplicatedJobs updates may not change fields other than %s", js.Spec.ReplicatedJobs[index].Name, strings.Join(updatableFields, ", ")),
-			))
+	if pointer.BoolDeref(oldSpec.Suspend, false) {
+		for index := range js.Spec.ReplicatedJobs {
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = oldSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
 		}
 	}
-	return errors.Join(allErr...)
+
+	if !apiequality.Semantic.DeepEqual(oldSpec, mungedSpec) {
+		specDiff := cmp.Diff(oldSpec, mungedSpec)
+		return field.Forbidden(field.NewPath("spec"), fmt.Sprintf("jobset updates may not change fields other than %s\n%v", strings.Join(updatableFields, ","), specDiff))
+	}
+	return nil
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+	"strings"
 
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/pointer"
@@ -82,12 +83,17 @@ func (js *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (js *JobSet) ValidateUpdate(old runtime.Object) error {
-	oldObj := old.(*JobSet)
+	oldObjCopy := old.DeepCopyObject().(*JobSet)
 	var allErr []error
+	var updatableFields = []string{
+		"`template.spec.template.spec.nodeSelector`",
+	}
 	for index := range js.Spec.ReplicatedJobs {
-		oldObj.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = r.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
-		if !reflect.DeepEqual(oldObj.Spec.ReplicatedJobs, js.Spec.ReplicatedJobs) {
-			allErr = append(allErr, fmt.Errorf("the spec.ReplicatedJobs value is imutable excepting Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector"))
+		oldObjCopy.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = js.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
+		if !reflect.DeepEqual(oldObjCopy.Spec.ReplicatedJobs, js.Spec.ReplicatedJobs) {
+			allErr = append(allErr, fmt.Errorf(
+				fmt.Sprintf("%s update validation: spec.ReplicatedJobs updates may not change fields other than %s", r.Spec.ReplicatedJobs[index].Name, strings.Join(updatableFields, ", ")),
+			))
 		}
 	}
 	return errors.Join(allErr...)
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -96,8 +96,8 @@ func (js *JobSet) ValidateUpdate(old runtime.Object) error {
 		}
 	}
 
-	if !apiequality.Semantic.DeepEqual(oldSpec, mungedSpec) {
-		specDiff := cmp.Diff(oldSpec, mungedSpec)
+	if !apiequality.Semantic.DeepEqual(oldSpec, *mungedSpec) {
+		specDiff := cmp.Diff(oldSpec, *mungedSpec)
 		return field.Forbidden(field.NewPath("spec"), fmt.Sprintf("jobset updates may not change fields other than %s\n%v", strings.Join(updatableFields, ","), specDiff))
 	}
 	return nil
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -16,9 +16,9 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
-	"reflect"
 	"strings"
 
+	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/pointer"
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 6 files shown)


ahg-g:
Thanks @mcariatm pls let us know how it goes; @danielvegamyhre and I are happy to help and very interested to get this done as soon as possible to unblock JobSet integration with Kueue.

mcariatm:
> Thanks @mcariatm pls let us know how it goes; @danielvegamyhre and I are happy to help and very interested to get this done as soon as possible to unblock JobSet integration with Kueue.

Hi @ahg-g, it goes fine. Now i'm trying to find an optimal way to update PodsReady status.

mcariatm:
I decided to create a new PR for PodsReady status part. 
This part is ready for review.

api/v1alpha1/jobset_webhook.go
```diff
@@ -66,7 +70,15 @@ func (r *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *JobSet) ValidateUpdate(old runtime.Object) error {
-	return nil
+	oldObj := old.(*JobSet)
```
ahg-g, line 73: I think we need a deep copy here, we shouldn't change the object itself

api/v1alpha1/jobset_webhook.go
```diff
@@ -66,7 +70,15 @@ func (r *JobSet) ValidateCreate() error {
...
+	var allErr []error
+	for index := range r.Spec.ReplicatedJobs {
+		oldObj.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = r.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
+		if !reflect.DeepEqual(oldObj.Spec.ReplicatedJobs, r.Spec.ReplicatedJobs) {
+			allErr = append(allErr, fmt.Errorf("the spec.ReplicatedJobs value is imutable excepting Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector"))
```
ahg-g, line 78:
```suggestion
			allErr = append(allErr, fmt.Errorf(fmt.Sprintf("spec.ReplicatedJobs updates may not change fields other than %s\n%v", strings.Join(updatableFields, ","))
```

and define this above:
```
var updatableFields = []string{
	"`template.spec.template.spec.nodeSelector`",
}
```

pkg/controllers/jobset_controller.go
```diff
@@ -238,10 +239,17 @@ func (r *JobSetReconciler) suspendJobSet(ctx context.Context, js *jobset.JobSet,
 }
 
 func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
+	nodeAffinities := map[string]map[string]string{}
+	for _, replicatedJob := range js.Spec.ReplicatedJobs {
+		nodeAffinities[fmt.Sprintf("%s-%s", js.Name, replicatedJob.Name)] = replicatedJob.Template.Spec.Template.Spec.NodeSelector
+	}
```
ahg-g, line 245: just use replicatedJob.Name as the map key; each job has a label with the replicatedJob name under the key jobset.ReplicatedJobNameKey

pkg/controllers/jobset_controller.go
```diff
@@ -238,10 +239,17 @@ func (r *JobSetReconciler) suspendJobSet(ctx context.Context, js *jobset.JobSet,
...
+
 	// If JobSpec is unsuspended, ensure all active child Jobs are also
 	// unsuspended and update the suspend condition to true.
 	for _, job := range ownedJobs.active {
 		if pointer.BoolDeref(job.Spec.Suspend, false) != false {
+			i := strings.LastIndex(job.Name, "-")
+			job.Spec.Template.Spec.NodeSelector = nodeAffinities[job.Name[:i]]
```
ahg-g, line 252:
```suggestion
                   if job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] != "" {
			job.Spec.Template.Spec.NodeSelector = nodeAffinities[job.Labels[jobset.ReplicatedJobNameKey]]
		   } else {
			  log.Error(nil, job missing ReplicatedJobName label)
		  }
```

pkg/controllers/jobset_controller.go
```diff
@@ -238,10 +239,17 @@ func (r *JobSetReconciler) suspendJobSet(ctx context.Context, js *jobset.JobSet,
...
 			job.Spec.Suspend = pointer.Bool(false)
 			if err := r.Update(ctx, job); err != nil {
```
ahg-g, line 254: we also need to update the job start time: `Status.StartTime = nil` 

otherwise the update will be rejected.

test/integration/controller/jobset_controller_test.go
```diff
@@ -516,6 +527,27 @@ func matchJobsSuspendState(js *jobset.JobSet, suspend bool) (bool, error) {
...
+func matchJobsNodeSelectors(js *jobset.JobSet, nodeSelectors map[string]map[string]string) (bool, error) {
+	var jobList batchv1.JobList
+	if err := k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace)); err != nil {
+		return false, err
+	}
+	jobsChanged := 0
+	// Check if all jobs with name prefix js.Name+replicatedJobName have updated node selectors.
+	for _, job := range jobList.Items {
+		for index, nodeSelector := range nodeSelectors {
+			if strings.HasPrefix(job.Name, index) {
```
ahg-g, line 539: we don't need two for loops, we can lookup the replciatedJobName from the labels and use it to lookup the expected node selectors


```
for _, job := range jobList.Items {
  wantSelectors := nodeSelectors[job.Labels[jobset.ReplicatedJobNameKey]]
 ...
}

```

test/integration/controller/jobset_controller_test.go
```diff
@@ -516,6 +527,27 @@ func matchJobsSuspendState(js *jobset.JobSet, suspend bool) (bool, error) {
...
+				if reflect.DeepEqual(job.Spec.Template.Spec.NodeSelector, nodeSelector) {
+					jobsChanged++
+					continue
+				}
+				return false, nil
+			}
+		}
+	}
+	return len(nodeSelectors) <= jobsChanged, nil
```
ahg-g, line 548: probably better to add up replica counts of each replicatedJob to calculate the number of jobs

api/v1alpha1/jobset_webhook.go
```diff
@@ -66,7 +71,20 @@ func (r *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (r *JobSet) ValidateUpdate(old runtime.Object) error {
-	return nil
+	oldObjCopy := old.DeepCopyObject().(*JobSet)
+	var allErr []error
+	var updatableFields = []string{
+		"`template.spec.template.spec.nodeSelector`",
+	}
+	for index := range r.Spec.ReplicatedJobs {
```
ahg-g, line 79: I forgot, we should allow updating nodeSelectors only if the jobSet is suspended.

ahg-g: pls add an integration test to the webhook as well to verify that updating nodeSelectors is allowed only when JobSet is suspended

test/integration/controller/jobset_controller_test.go
ahg-g: yes, but again you are assuming 1 job per replicatedJob, better to add up the replicatedJob.replicas of the replicatedJobs that we are updating. The name of those replicatedJobs exist as the `nodeSelectors` key:

```
wantJobsChanged := 0
for rj := js.ReplicatedJobs {
  if _, exists := nodeSelectors[rj.name]; exists {
    wantJobsChaned += rj.Replicas
  }
}
return wantJobsChanged == jobsChanged
```

api/v1alpha1/jobset_webhook.go
```diff
@@ -81,7 +83,23 @@ func (js *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (js *JobSet) ValidateUpdate(old runtime.Object) error {
-	return nil
+	oldObjCopy := old.DeepCopyObject().(*JobSet)
+	if pointer.BoolDeref(oldObjCopy.Spec.Suspend, false) {
+		return nil
```
ahg-g, line 88: we need to return an error if there was an attempt to do any update

api/v1alpha1/jobset_webhook.go
```diff
@@ -81,7 +83,23 @@ func (js *JobSet) ValidateCreate() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (js *JobSet) ValidateUpdate(old runtime.Object) error {
-	return nil
+	oldObjCopy := old.DeepCopyObject().(*JobSet)
```
ahg-g, line 86: this should actually be the new object, we can call it `mungedCopy`

api/v1alpha1/jobset_webhook.go
```diff
@@ -81,7 +83,23 @@ func (js *JobSet) ValidateCreate() error {
...
+	}
+	for index := range js.Spec.ReplicatedJobs {
+		oldObjCopy.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = js.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
+		if !apiequality.Semantic.DeepEqual(oldObjCopy.Spec.ReplicatedJobs, js.Spec.ReplicatedJobs) {
+			allErr = append(allErr, fmt.Errorf(
+				fmt.Sprintf("%s update validation: spec.ReplicatedJobs updates may not change fields other than %s", js.Spec.ReplicatedJobs[index].Name, strings.Join(updatableFields, ", ")),
+			))
+		}
+	}
+	return errors.Join(allErr...)
```
ahg-g, line 102:
```suggestion
	mungedSpec := js.Spec.DeepCopy()
	oldSpec := old.(*JobSet).Spec
	var updatableFields = []string{
		"`template.spec.template.spec.nodeSelector` when jobset is suspended",
	}
	if pointer.BoolDeref(oldObjCopy.Spec.Suspend, false) {
		for index := range js.Spec.ReplicatedJobs {
			mungedCopy.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = oldSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
			}
		}
	}
	if !apiequality.Semantic.DeepEqual(oldSpec, mungedSpec) {
	  specDiff := cmp.Diff(oldSpec, mungedSpec)
		return field.Forbidden(specPath, fmt.Sprintf("jobset updates may not change fields other than %s\n%v", strings.Join(updatableFields, ","), specDiff))
	}
	return nil
```

ahg-g: you need to import:
	- "github.com/google/go-cmp/cmp"
	- "k8s.io/apimachinery/pkg/util/validation/field"

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -155,6 +177,59 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 				return js.Spec.ReplicatedJobs[0].Template.Spec.Template.Spec.RestartPolicy == corev1.RestartPolicyOnFailure
 			},
 		}),
+		ginkgo.Entry("validate jobSet should not fail on NodeSelectors Update", &testCase{
+			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
```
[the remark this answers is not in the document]
mcariatm: Yes

api/v1alpha1/jobset_webhook.go
ahg-g: for `specPath`, just replace it with `field.NewPath("spec")`

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -155,6 +168,40 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 				return js.Spec.ReplicatedJobs[0].Template.Spec.Template.Spec.RestartPolicy == corev1.RestartPolicyOnFailure
 			},
 		}),
+		ginkgo.Entry("validate jobSet should fail on NodeSelectors Update", &testCase{
```
ahg-g, line 171: we need a test that creates a suspended jobset and the update succeeds.

mcariatm: Should I remove "validate jobSet immutable for fields over than NodeSelector when jobSet is not suspended" test from line 202 in this case?

ahg-g: no, that should stay because it validates a different case. The test case I am asking for is exactly like this one, but the jobset is created suspended and so `updateShouldFail=false`

ahg-g: sg, thanks

test/integration/controller/jobset_controller_test.go
```diff
@@ -522,6 +505,11 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 					},
 					checkJobSetCondition: testutil.JobSetSuspended,
 				},
+				{
+					jobSetUpdateFn: func(set *jobset.JobSet) {
```
danielvegamyhre, line 509: nit for consistency with other test cases:

```suggestion
					jobSetUpdateFn: func(js *jobset.JobSet) {
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -522,6 +505,11 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 					},
 					checkJobSetCondition: testutil.JobSetSuspended,
 				},
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						updateJobSetNodeSelectors(set, nodeSelectors)
```
danielvegamyhre, line 510:
```suggestion
						updateJobSetNodeSelectors(js, nodeSelectors)
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -643,6 +644,34 @@ func matchJobsSuspendState(js *jobset.JobSet, suspend bool) (bool, error) {
...
+		return false, err
+	}
+	// Count number of changed jobs
+	jobsChanged := 0
+	for _, job := range jobList.Items {
+		nodeSelectorKey, ok := job.Labels[jobset.ReplicatedJobNameKey]
+		if !ok {
+			return false, fmt.Errorf(fmt.Sprintf("%s job missing ReplicatedJobName label", job.Name))
+		}
+		if apiequality.Semantic.DeepEqual(job.Spec.Template.Spec.NodeSelector, nodeSelectors[nodeSelectorKey]) {
```
danielvegamyhre, line 659: Should this logic be inverted (i.e., if node selectors are NOT equal, then increment `jobsChanged`)?

test/integration/controller/jobset_controller_test.go
```diff
@@ -643,6 +644,34 @@ func matchJobsSuspendState(js *jobset.JobSet, suspend bool) (bool, error) {
...
+			jobsChanged++
+			continue
+		}
+		return false, nil
```
danielvegamyhre, line 663: Question, why would we return early here if one of the jobs has not changed? Don't we still want to compare `jobsChanged` with `wantJobsChanged` below?

test/integration/controller/jobset_controller_test.go
```diff
@@ -625,6 +616,16 @@ func suspendJobSet(js *jobset.JobSet, suspend bool) {
 	gomega.Eventually(k8sClient.Update(ctx, js), timeout, interval).Should(gomega.Succeed())
 }
 
+func updateJobSetNodeSelectors(js *jobset.JobSet, nodeSelectors map[string]map[string]string) {
+	for index := range js.Spec.ReplicatedJobs {
+		js.Spec.ReplicatedJobs[index].
+			Template.Spec.Template.Spec.Hostname = "test"
```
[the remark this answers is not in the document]
mcariatm: my mistake

test/integration/controller/jobset_controller_test.go
ahg-g: this is tracking the jobs that got the expected change

test/integration/controller/jobset_controller_test.go
mcariatm: I think It can be removed.

ahg-g: we are trying to check if the child jobs got the updated nodeSelector (which is set in nodeSelectors)

@mcariatm can you switch the condition like this:

```
if !apiequality.Semantic.DeepEqual(job.Spec.Template.Spec.NodeSelector, nodeSelectors[nodeSelectorKey]) {
  return false, nil
}
jobsChanged++
```

ahg-g: also, please use substitute the word `change` with `update` in all places in this function

PR #144, opened 2023-05-18
Ready replicates jobs status

mcariatm:
ReplicatedJobsStatus.ReadyJobs should be added and updated in JobSet statu..

### Why is this needed:
We need to store and update numbers of JobsReady for each ReplicatedJob in Jobset status to operate with them in Kueue.

### Issue Number:
https://github.com/kubernetes-sigs/jobset/issues/134


api/v1alpha1/jobset_types.go
```diff
@@ -74,6 +74,20 @@ type JobSetStatus struct {
 
 	// Restarts tracks the number of times the JobSet has restarted (i.e. recreated in case of RecreateAll policy).
 	Restarts int `json:"restarts,omitempty"`
+
+	// ReplicatedJobsStatus track the number of PodsReady for each replicatedJob.
```
kannon92, line 78:
```suggestion
	// ReplicatedJobsStatus track the number of JobsReady for each replicatedJob.
```

You are getting the number of jobs that are ready.

pkg/controllers/jobset_controller.go
```diff
@@ -149,6 +150,8 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 			return ctrl.Result{}, err
 		}
 	}
+	// Calculate PodsReady and update statuses for each ReplicatedJob
+	r.calculateAndUpdateReplicatedJobsStatuses(ctx, &js, ownedJobs)
```
kannon92, line 154: I think we probably want to allow for returning an error and trigger a reconcile.  Following a similar pattern as we do in this function.

pkg/controllers/jobset_controller.go
```diff
@@ -219,6 +222,63 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
 	return &ownedJobs, nil
 }
 
+func (r *JobSetReconciler) calculateAndUpdateReplicatedJobsStatuses(ctx context.Context, js *jobset.JobSet, jobs *childJobs) {
```
kannon92, line 225: Could you refactor it so you can unit test a lot of the logic aside from the Status Update?

I see some cases I think you should target in a unit test but may be difficult to do as an integration test.  

I think modifying this function so it could return a (ReplicaedJobStatus, error) and you update the status.  That way you can at least unit test the majority of this logic.

danielvegamyhre: Yes, please do this. You could extract out everything up until the `r.Status().Update(...)` call into a separate function so that you can easily unit test the logic being used to set the values we are updating the status with. You should also return an error to trigger reconciliation as @kannon92 noted.

Pseudocode example:

```
func calculateReplicatedJobStatuses(...) {
  // all your logic for getting the statuses and add unit tests for this function
}

func (r* JobSetReconciler) calculateAndUpdateReplicatedJobsStatuses(...) (jobset.ReplicatedJobsStatus, error){
    status, err := calculateReplicatedJobStatuses(...)
    if err != nil {
        return nil, err
    }
    return r.Status().Update(...)
}
```

api/v1alpha1/jobset_types.go
```diff
@@ -74,6 +74,20 @@ type JobSetStatus struct {
...
+	// +listMapKey=name
+	ReplicatedJobsStatus []ReplicatedJobStatus `json:"ReplicatedJobsStatus,omitempty"`
+}
+
+// ReplicatedJobStatus defines the observed ReplicatedJobs Readiness.
+type ReplicatedJobStatus struct {
+	Name string `json:"name"`
+
+	// +kubebuilder:default=0
+	ReadyJobs int32 `json:"readyJobs"`
```
danielvegamyhre, line 90: On the Kueue side how are you comparing the number of ready jobs to the number of expected jobs for the replicatedJob? I assume you must be iterating through the `jobSet.spec.replicatedJobs` and to check the `replicas` value then comparing it with the `ReadyJobs` value in this status struct?

pkg/controllers/jobset_controller.go
```diff
@@ -219,6 +222,63 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+
+	// Calculate ReadyPods for each Replicated Job
+	for _, job := range append(jobs.active, append(jobs.successful, jobs.failed...)...) {
+		ready := pointer.Int32Deref(job.Status.Ready, 0)
+		// parallelism is always set as it is otherwise defaulted by k8s to 1
+		podsCount := *(job.Spec.Parallelism)
+		if job.Spec.Completions != nil && *job.Spec.Completions < podsCount {
+			podsCount = *job.Spec.Completions
+		}
+		if job.Status.Succeeded+ready >= podsCount {
```
danielvegamyhre, line 242:
```suggestion
		if job.Status.Succeeded + ready >= podsCount {
```

pkg/controllers/jobset_controller.go
```diff
@@ -219,6 +222,63 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+	// Update current ReplicatedJobsStatuses
+	for index := range status {
+		val, ok := replicatedJobsReady[status[index].Name]
+		if !ok {
+			log.Info(fmt.Sprintf("the %s replicated job is not anymore relevant", status[index].Name))
+			status = append(status[:index], status[index+1:]...)
+		} else {
+			status[index].ReadyJobs = val
+		}
+		delete(replicatedJobsReady, status[index].Name)
```
danielvegamyhre, line 261: Why do we need to delete the entry in `replicatedJobsReady` after updating the status?

pkg/controllers/jobset_controller.go
```diff
@@ -219,6 +222,63 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+	}
+
+	// Add new ReplicatedJobsStatuses
+	for name, readyJobs := range replicatedJobsReady {
+		status = append(status, jobset.ReplicatedJobStatus{
```
danielvegamyhre, line 266: If we only ever append to status, won't it just grow infinitely? Shouldn't we be updating each ReplicatedJobStatus in place?

pkg/controllers/jobset_controller.go
mcariatm: image

It is a gofmt standard in if statements.

pkg/controllers/jobset_controller.go
mcariatm: On line 251 I'v created a variable to keep all ReplicatedJobStatuses positions using current status DeepCopy, I update them, and on 266 I append the rest.

pkg/controllers/jobset_controller.go
mcariatm: On line 251 I'v created a variable to keep all ReplicatedJobStatuses positions using current status DeepCopy, I update them, and on 266 I append the rest.
I use this mode so that the positions stay the same and it is smooth to read.

pkg/controllers/jobset_controller.go
```diff
@@ -219,6 +222,63 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		})
+	}
+
+	// Check if status ReplicatedJobsStatus has changed
+	if apiequality.Semantic.DeepEqual(js.Status.ReplicatedJobsStatus, status) {
+		return
+	}
+	js.Status.ReplicatedJobsStatus = status
+	if err := r.Status().Update(ctx, js); err != nil {
+		log.Error(err, fmt.Sprintf("jobset update error"))
```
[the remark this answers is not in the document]
mcariatm: I've decided that an error like this is not vital and anyway will be updated in the next iteration.
But Ok, I can return all errors and catch them in reconciler.

pkg/controllers/jobset_controller.go
danielvegamyhre: Oh interesting I didn't know that, I find it more readable with the spaces.. Oh well, you can disregard this comment then :)

pkg/controllers/jobset_controller.go
```diff
@@ -218,6 +224,67 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+	}
+	// Check if status ReplicatedJobsStatus has changed
+	if apiequality.Semantic.DeepEqual(js.Status.ReplicatedJobsStatus, status) {
+		return nil
+	}
+	js.Status.ReplicatedJobsStatus = status
+	return r.Status().Update(ctx, js)
+}
+
+func calculateReplicatedJobStatuses(js *jobset.JobSet, jobs *childJobs) ([]jobset.ReplicatedJobStatus, error) {
```
kannon92, line 240: Please add a unit test in job_set_controller_test.go in pkg/controllers.

pkg/controllers/jobset_controller.go
```diff
@@ -218,6 +224,67 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		// parallelism is always set as it is otherwise defaulted by k8s to 1
+		podsCount := *(job.Spec.Parallelism)
+		if job.Spec.Completions != nil && *job.Spec.Completions < podsCount {
+			podsCount = *job.Spec.Completions
+		}
+		if job.Status.Succeeded+ready >= podsCount {
+			if job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] != "" {
+				replicatedJobsReady[job.Labels[jobset.ReplicatedJobNameKey]]++
+			} else {
+				return nil, fmt.Errorf(fmt.Sprintf("job %s missing ReplicatedJobName label", job.Name))
```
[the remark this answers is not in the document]
mcariatm: As I remember @ahg-g recomanded the same approach on an older PR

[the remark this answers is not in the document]
mcariatm: Ah.. I'm aborting the loop with return. Ok, Will use log.Error

api/v1alpha1/jobset_types.go
```diff
@@ -74,6 +74,20 @@ type JobSetStatus struct {
...
```
ahg-g, line 90: we should also report the number of succeeded jobs, Kueue will need to add up ready + succeeded to determine if the jobset successfully started.

pkg/controllers/jobset_controller.go
```diff
@@ -218,6 +224,69 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		if !ok {
+			// the current replicated job is not anymore relevant
+			rjStatus = append(rjStatus[:index], rjStatus[index+1:]...)
+		} else {
+			rjStatus[index].ReadyJobs = val
+		}
+		// Delete updated ReplicatedJobsReady from replicatedJobsReady
+		// it will be used to append non-existent ones
+		delete(replicatedJobsReady, rjStatus[index].Name)
+	}
```
ahg-g, line 279: The set of replicatedJobs is immutable, so this loop is not necessary, just calculate ReplicatedJobsStatus from scratch below like this:

``
rjStatus := []ReplicatedJobStatus{}
...
return rjStatus
```

File: pkg/controllers/jobset_controller.go [hunk 13]
```diff
@@ -218,6 +224,69 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+	log := ctrl.LoggerFrom(ctx)
+
+	// Prepare replicatedJobsReady for optimal iteration
+	replicatedJobsReady := map[string]int32{}
+	for _, replicatedJob := range js.Spec.ReplicatedJobs {
+		replicatedJobsReady[replicatedJob.Name] = 0
+	}
+
+	// Calculate jobsReady for each Replicated Job
+	for _, job := range append(jobs.active, append(jobs.successful, jobs.failed...)...) {
```
ahg-g, line 250: successful and failed jobs should not be counted as "ready", we should only look at active.

but we also need to iterate over successful in a second loop to calculate the number of Succeeded jobs

api/v1alpha1/jobset_types.go
mcariatm: Job Ready means: 
`podsCount := *(job.Spec.Parallelism)
if job.Spec.Completions != nil && *job.Spec.Completions < podsCount {
	podsCount = *job.Spec.Completions
}
if job.Status.Succeeded+ready >= podsCount {
       Job Is READY
}`
And if Job is ready we just increment the ReadyJobs

ahg-g: I am not sure how this is related to the comment, I am saying we need to report the number of succeeded jobs in the status; for example, if a ReplicatedJob have 20 jobs, 10 active and 10 succeeded, out of the 10 active jobs only 5 are ready, then the ReplicatedJobsStatus will be:

Ready: 5
Succeeded: 10

api/v1alpha1/jobset_types.go
```diff
@@ -74,6 +74,19 @@ type JobSetStatus struct {
...
+	// +optional
+	// +listType=map
+	// +listMapKey=name
+	ReplicatedJobsStatus []ReplicatedJobStatus `json:"ReplicatedJobsStatus,omitempty"`
+}
+
+// ReplicatedJobStatus defines the observed ReplicatedJobs Readiness.
+type ReplicatedJobStatus struct {
+	Name          string `json:"name"`
+	ReadyJobs     int32  `json:"readyJobs"`
```
ahg-g, line 88: pls remove the `Jobs` suffix, Ready/Succeeded is enough.

pkg/controllers/jobset_controller.go
```diff
@@ -237,17 +234,17 @@ func (r *JobSetReconciler) calculateAndUpdateReplicatedJobsStatuses(ctx context.
 	return r.Status().Update(ctx, js)
 }
 
-func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, js *jobset.JobSet, jobs *childJobs) ([]jobset.ReplicatedJobStatus, error) {
+func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, js *jobset.JobSet, jobs *childJobs) []jobset.ReplicatedJobStatus {
 	log := ctrl.LoggerFrom(ctx)
 
 	// Prepare replicatedJobsReady for optimal iteration
-	replicatedJobsReady := map[string]int32{}
+	replicatedJobsReady := map[string]*[2]int32{}
```
ahg-g, line 241: This is harder to follow and makes it easy to get the indices of each status wrong, just use two maps pls.

pkg/util/testing/wrappers.go
```diff
@@ -235,6 +235,18 @@ func (j *JobWrapper) Subdomain(subdomain string) *JobWrapper {
...
+func (j *JobWrapper) JobSpec(jobSpec batchv1.JobSpec) *JobWrapper {
+	j.Spec = jobSpec
+	return j
+}
+
+// JobStatus sets the job spec.
+func (j *JobWrapper) JobStatus(jobSpec batchv1.JobStatus) *JobWrapper {
+	j.Status = jobSpec
+	return j
+}
```
ahg-g, line 248: instead of those, pls add wrapper functions to set specific fields in spec and status:

```
Parallelism(value int)
Completions(value int)
Ready(value int)
Succeeded(value int)
```

pkg/controllers/jobset_controller_test.go
```diff
@@ -708,10 +929,23 @@ type makeJobArgs struct {
 	replicas          int
 	jobIdx            int
 	restarts          int
+	parallelism       *int32
+	completions       *int32
+	succeeded         int
+	ready             *int32
```
ahg-g, line 935: remove those pls, we can set them directly on the wrapperJob returned by makeJob

pkg/controllers/jobset_controller_test.go
```diff
@@ -700,6 +704,223 @@ func TestUpdateConditions(t *testing.T) {
...
+						jobSetName:        jobSetName,
+						replicatedJobName: "replicated-job-1",
+						jobName:           "test-jobset-replicated-job-1-test-job-0",
+						ns:                ns,
+						replicas:          1,
+						jobIdx:            0,
+						parallelism:       pointer.Int32(1),
+						completions:       pointer.Int32(1),
+						ready:             pointer.Int32(1),
+						succeeded:         1}).Suspend(false).Obj(),
```
ahg-g, line 741:
```suggestion
						jobSetName:        jobSetName,
						replicatedJobName: "replicated-job-1",
						jobName:           "test-jobset-replicated-job-1-test-job-0",
						ns:                ns,
						replicas:          1,
						jobIdx:            0}).
						Parallelism(1).
						Completions(1).
						Ready(1),
						Succeeded(1).Obj(),
```

same below

pkg/controllers/jobset_controller_test.go
```diff
@@ -700,6 +704,223 @@ func TestUpdateConditions(t *testing.T) {
...
+						jobSetName:        jobSetName,
+						replicatedJobName: "replicated-job-2",
+						jobName:           "test-jobset-replicated-job-2-test-job-2",
+						ns:                ns,
+						replicas:          3,
+						jobIdx:            0,
+						parallelism:       pointer.Int32(2),
+						completions:       pointer.Int32(3),
+						ready:             pointer.Int32(2),
+						succeeded:         1}).Obj(),
```
ahg-g, line 773: can you add another job that is not ready (ready+succeeded less than completions/parallelism)

pkg/controllers/jobset_controller_test.go
```diff
@@ -700,6 +704,223 @@ func TestUpdateConditions(t *testing.T) {
...
+						jobSetName:        jobSetName,
+						replicatedJobName: "replicated-job-2",
+						jobName:           "test-jobset-replicated-job-2-test-job-0",
+						ns:                ns,
+						replicas:          3,
+						jobIdx:            0,
+						parallelism:       pointer.Int32(5),
+						ready:             pointer.Int32(2),
+						succeeded:         3}).Obj(),
+					makeJob(&makeJobArgs{
```
ahg-g, line 836: just to reduce the length of the file, I think one job here is enough (instead of 3)

pkg/controllers/jobset_controller.go
mcariatm: I was thinking about memory optimization. ok.

ahg-g: in the larger scheme of things, this is not significant

Issue #157, opened 2023-05-23
Send a status update when resetting StartTime


ahg-g:
In https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L258 we reset StartTime to allow updating the nodeSelectors on the job. However StartTime is in status, which we can update via a status update call only: `client.Status().Update(..)`

@mcariatm

/kind bug

mcariatm:
/close 
#160

PR #158, opened 2023-05-23
Set node affinity during unsuspend

mcariatm:
In #143  we reverted the mutability of suspend.

### Why is this needed:
Suspend should be mutable

### Issue Number:
#157

/kind-bug


mcariatm:
replaced by #159

Issue #156, opened 2023-05-23
Make suspend mutable


ahg-g:
In https://github.com/kubernetes-sigs/jobset/pull/143 we reverted the mutability of suspend.

Please also add a corresponding integration test.

/assign @mcariatm

ahg-g:
@mcariatm

kannon92:
https://github.com/kubernetes-sigs/jobset/pull/159

/close

PR #159, opened 2023-05-23
Suspend field should be mutable

mcariatm:
we reset StartTime to allow updating the nodeSelectors on the job. However StartTime is in status, which we can update via a status update call only

### Why is this needed:
Update start time using client.Status().Update(..)

### Issue Number:
#156 

/kind bug

### DIFF
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -96,8 +96,8 @@ func (js *JobSet) ValidateUpdate(old runtime.Object) error {
 		}
 	}
 
-	if !apiequality.Semantic.DeepEqual(oldSpec, *mungedSpec) {
-		specDiff := cmp.Diff(oldSpec, *mungedSpec)
+	if !apiequality.Semantic.DeepEqual(oldSpec.ReplicatedJobs, mungedSpec.ReplicatedJobs) {
+		specDiff := cmp.Diff(oldSpec.ReplicatedJobs, mungedSpec.ReplicatedJobs)
 		return field.Forbidden(field.NewPath("spec"), fmt.Sprintf("jobset updates may not change fields other than %s\n%v", strings.Join(updatableFields, ","), specDiff))
 	}
 	return nil
--- a/api/v1alpha1/jobset_webhook.go
+++ b/api/v1alpha1/jobset_webhook.go
@@ -16,10 +16,8 @@ package v1alpha1
 import (
 	"errors"
 	"fmt"
-	"strings"
 
-	"github.com/google/go-cmp/cmp"
-	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	"k8s.io/utils/pointer"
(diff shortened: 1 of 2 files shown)


test/integration/webhook/jobset_webhook_test.go
```diff
@@ -270,5 +271,42 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 			},
 			jobSetCreationShouldFail: true,
 		}),
+		ginkgo.Entry("jobSet should suspended", &testCase{
```
danielvegamyhre, line 274:
```suggestion
		ginkgo.Entry("suspend jobset", &testCase{
```

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -270,5 +271,42 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+			},
+			defaultsApplied: func(js *jobset.JobSet) bool {
+				return true
+			},
+			updateJobSet: func(js *jobset.JobSet) {
+				js.Spec.Suspend = pointer.Bool(true)
+			},
+			updateShouldFail: false,
+		}),
+		ginkgo.Entry("jobSet should unsuspended", &testCase{
```
danielvegamyhre, line 292:
```suggestion
		ginkgo.Entry("resume jobset", &testCase{
```

danielvegamyhre (changes requested):

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -270,5 +271,42 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+				return testing.MakeJobSet("js-hostnames-non-indexed", ns.Name).
+					ReplicatedJob(testing.MakeReplicatedJob("rjob").
+						Job(testing.MakeJobTemplate("job", ns.Name).
+							PodSpec(testing.TestPodSpec).
+							CompletionMode(batchv1.IndexedCompletion).Obj()).
+						EnableDNSHostnames(true).
+						Obj())
+			},
+			defaultsApplied: func(js *jobset.JobSet) bool {
+				return true
```
danielvegamyhre, line 285: No need to define this if it isn't validating defaults were applied. The `defaultsApplied` function is only executed if it is not nil. In fact, can we remove this from other test cases which do this as well?

api/v1alpha1/jobset_webhook.go
```diff
@@ -96,8 +96,8 @@ func (js *JobSet) ValidateUpdate(old runtime.Object) error {
 		}
 	}
 
-	if !apiequality.Semantic.DeepEqual(oldSpec, *mungedSpec) {
-		specDiff := cmp.Diff(oldSpec, *mungedSpec)
+	if !apiequality.Semantic.DeepEqual(oldSpec.ReplicatedJobs, mungedSpec.ReplicatedJobs) {
+		specDiff := cmp.Diff(oldSpec.ReplicatedJobs, mungedSpec.ReplicatedJobs)
 		return field.Forbidden(field.NewPath("spec"), fmt.Sprintf("jobset updates may not change fields other than %s\n%v", strings.Join(updatableFields, ","), specDiff))
 	}
 	return nil
```
ahg-g, line 103:
```suggestion
  allErrs = append(allErrs, apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldSpec.ReplicatedJobs, specPath.Child("replicatedJobs"))...)
  allErrs = append(allErrs, apivalidation.ValidateImmutableField(mungedSpec.SuccessPolicy, oldSpec.SuccessPolicy, specPath.Child("successPolicy"))...)
  allErrs = append(allErrs, apivalidation.ValidateImmutableField(mungedSpec.FailurePolicy, oldSpec.FailurePolicy, specPath.Child("failurePolicy"))...)
return allErrs   
```

ahg-g: and import `apivalidation "k8s.io/apimachinery/pkg/api/validation"`

ahg-g (commented):
success and failure policy fields should remain immutable, see the suggested fix

api/v1alpha1/jobset_webhook.go
mcariatm: they are immutable. See on jobset_types.go line 53 and 60

mcariatm: I can add some tests to demonstrate that....

mcariatm: Conclusion:
After some tests, FailurePolicy is immutable only if it was set before. Should I do something else on this side?

ahg-g: ah, right, I updated the suggestion, lets still use `apivalidation.ValidateImmutableField` to check the replicatedJobs field and have a comment indicating that success and failure policy are made immutable via CEL

mcariatm: done

PR #160, opened 2023-05-23
fixed resetting start time bug

mcariatm:
Update status using client.Status().Update(..)

### Why is this needed:
we reset StartTime to allow updating the nodeSelectors on the job. However StartTime is in status, which we can update via a status update call only

### Issue Number:
#157

/kind bug

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -258,6 +258,9 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
 			if err := r.Update(ctx, job); err != nil {
 				return err
 			}
+			if job.Status.StartTime == nil {
+				continue
+			}
 			job.Status.StartTime = nil
 			if err := r.Status().Update(ctx, job); err != nil {
 				return err
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -255,8 +255,11 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
 				log.Error(nil, "job missing ReplicatedJobName label")
 			}
 			job.Spec.Suspend = pointer.Bool(false)
-			job.Status.StartTime = nil
 			if err := r.Update(ctx, job); err != nil {
+				return err
+			}
+			job.Status.StartTime = nil
+			if err := r.Status().Update(ctx, job); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -255,10 +255,13 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
 				log.Error(nil, "job missing ReplicatedJobName label")
 			}
 			job.Spec.Suspend = pointer.Bool(false)
-			job.Status.StartTime = nil
 			if err := r.Update(ctx, job); err != nil {
 				return err
 			}
+			job.Status.StartTime = nil
```
ahg-g, line 261: please check if StartTime is already nil and avoid the update if it is

mcariatm: In reconcile we call this function only if the job is suspended. But ok.

pkg/controllers/jobset_controller.go
```diff
@@ -255,10 +255,16 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
...
 			if err := r.Update(ctx, job); err != nil {
 				return err
 			}
+			if job.Status.StartTime == nil {
+				continue
+			}
+			job.Status.StartTime = nil
+			if err := r.Status().Update(ctx, job); err != nil {
+				return err
+			}
```
ahg-g, line 267: this is less intuitive than simply encapsulate the logic under the condition:
```suggestion
			if job.Status.StartTime != nil {
			  job.Status.StartTime = nil
			  if err := r.Status().Update(ctx, job); err != nil {
			 	return err
			  }
			}
```

pkg/controllers/jobset_controller.go
```diff
@@ -255,10 +255,15 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
...
-			job.Status.StartTime = nil
 			if err := r.Update(ctx, job); err != nil {
 				return err
 			}
+			if job.Status.StartTime != nil {
+			  job.Status.StartTime = nil
+			  if err := r.Status().Update(ctx, job); err != nil {
+			 	return err
+			  }
+			}
```
ahg-g, line 266: @mcariatm it seems that there is a missing bracket

mcariatm: no, it was due to the indentation being too short.

mcariatm: should be ok now....

Issue #161, opened 2023-05-25
Jobset status table


kannon92:
**What would you like to be added**:

When you do kubectl get jobsets you only get a list of jobsets.  It’s be nice for this to be similar to a job at a minimum ie completions and/or failures. 

Maybe we can include how many replicated jobs have completed or failed? 

**Why is this needed**:

better debug ability

ahg-g:
This should be doable using some kubebuilder tags: https://github.com/kubernetes-sigs/kueue/pull/156/files

do you want to take a stab at this @kannon92 ?

kannon92:
I’ll be out of town the next couple weeks so I’ll leave it open.  I think with the recent changes for ReplicatedJobStatus this should be pretty accessible.

A-Hilaly:
I'll take a quick stab at this if you haven't already @ahg-g :)
 
/assign

ahg-g:
@A-Hilaly please go ahead! Welcome to the project :)

A-Hilaly:
@ahg-g @kannon92 what do you think about adding `Completions`, `Failures` (similar to status.Restart) to JobSetStatus and make sure to update those during the reconciliation? We could add those as printcolumn tags.

ahg-g:
Can we surface the replicatedjob status? https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha1/jobset_types.go#L87-L89

A-Hilaly:
Not sure i fully understand @ahg-g. The idea would be to add `Completions/Succeded` and `Failures` status fields to `JobSetStatus` and those will be calculated from replicatedJob status

ahg-g:
I am not sure if there is a way to meaningfully calculate an aggregated status. I think users would expect a breakdown by replicatedJob

A-Hilaly:
Makes sense! We can close this as it's not possible to show all the replicatedJob statuses in kubectl table response

kannon92:
Thank you for leading this effort @A-Hilaly!!

/close

PR #165, opened 2023-05-29
Add `Completed` and `Restarts` printcolumns to JobSet

A-Hilaly:
Addresses part of https://github.com/kubernetes-sigs/jobset/issues/161

This PR adds new printcolumns, `Completed` and `Restarts` that will be part of `kubectl get` response table.

example:
```bash
> k get jobset
NAME    RESTARTS   COMPLETED   AGE
sleep              True        3m48s
```

### DIFF
--- a/api/v1alpha1/jobset_types.go
+++ b/api/v1alpha1/jobset_types.go
@@ -90,6 +90,9 @@ type ReplicatedJobStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
+// +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
+// +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type=date,description="Time this JobSet was created"
 // JobSet is the Schema for the jobsets API
 type JobSet struct {
 	metav1.TypeMeta   `json:",inline"`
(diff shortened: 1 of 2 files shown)


api/v1alpha1/jobset_types.go
```diff
@@ -90,6 +90,9 @@ type ReplicatedJobStatus struct {
 
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
+// +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
+// +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type=date,description="Time this JobSet was created"
```
A-Hilaly, line 95: As mentioned in #161 I like the idea of also adding number of completions/failures - but i couldn't find a way to do it using Kubernetes [JSONPath](https://kubernetes.io/docs/reference/kubectl/jsonpath/).

Maybe we could add new fields `Completions`, `Failures` to `JobSetStatus` and make sure to update those during the reconciliation? That will make it easier for us to support completion/failure in printcolumn. Thoughts @ahg-g ?

kannon92: Could you use ReplicatedJobStatus which was added recently? I think they have succeeded but not failed so you could add that

ahg-g: We added Ready and Succeeded to ReplicatedJobStatus (see https://github.com/kubernetes-sigs/jobset/blob/main/api/v1alpha1/jobset_types.go#L81), I am not sure though how we can print this per ReplicatedJob?

PR #167, opened 2023-05-30
Fix bug with setting node selectors when resuming suspended jobset

### LINKED ISSUE #166: `resume suspended jobset` controller integration test is failing at HEAD when testing logic bug is fixed
[checkJobState](https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L173) test case struct field is only specified in 1 test case [here](https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L512)  and is not called in the [testing logic](https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L207-L232) itself, so this [matchNodeSelectors](https://github.com/kubernetes-sigs/jobset/pull/143/files#diff-17c52df3cfe1010d354b94a59938b85133c0120c960c38854ec529546f3e8ea6R524-R525) check was never actually being called when we merged #134. It turns out this check actually fails when this bug is fixed and the call is executed:

```
  [FAILED] in [It] - 
  .../jobset/test/integration/controller/jobset_controller_test.go:512 @ 05/30/23 23:05:39.03
• [FAILED] [5.593 seconds]
JobSet controller jobset is created and its jobs go through a series of updates [It] resume a suspended jobset
.../jobset/test/integration/controller/jobset_controller_test.go:483

  [FAILED] Timed out after 5.001s.
  Expected
      <bool>: false
  to equal
      <bool>: true
  In [It] at: .../jobset/test/integration/controller/jobset_controller_test.go:512 @ 05/30/23 23:05:39.03
------------------------------
```

/kind bug

danielvegamyhre:
Fixes #166 

The issue was the nodeSelector map key was using the replicatedJob name as the key, when it should be in the format `<jobSetName>-<replicatedJobName>`, so the key didn't exist.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -250,6 +250,8 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
 	for _, job := range ownedJobs.active {
 		if pointer.BoolDeref(job.Spec.Suspend, false) != false {
 			if job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] != "" {
+				// When resuming a job, its nodeSelectors should match that of the replicatedJob template
+				// that it was created from, which may have been updated while it was suspended.
 				job.Spec.Template.Spec.NodeSelector = nodeAffinities[job.Labels[jobset.ReplicatedJobNameKey]]
 			} else {
 				log.Error(nil, "job missing ReplicatedJobName label")
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -170,7 +170,6 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 		jobSetUpdateFn       func(*jobset.JobSet)
 		jobUpdateFn          func(*batchv1.JobList)
 		checkJobSetState     func(*jobset.JobSet)
-		checkJobState        func(*jobset.JobSet)
 		checkJobSetCondition func(context.Context, client.Client, *jobset.JobSet, time.Duration)
 	}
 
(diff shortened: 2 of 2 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -250,6 +250,8 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
 	for _, job := range ownedJobs.active {
 		if pointer.BoolDeref(job.Spec.Suspend, false) != false {
 			if job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] != "" {
+				// When resuming a job, its nodeSelectors should match that of the replicatedJob template
+				// that is was created from, which may have been updated while it was suspended.
```
ahg-g, line 254:
```suggestion
				// that it was created from, which may have been updated while it was suspended.
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -638,20 +637,22 @@ func matchJobsNodeSelectors(js *jobset.JobSet, nodeSelectors map[string]map[stri
 	// Count number of updated jobs
 	jobsUpdated := 0
 	for _, job := range jobList.Items {
-		nodeSelectorKey, ok := job.Labels[jobset.ReplicatedJobNameKey]
+		rjobName, ok := job.Labels[jobset.ReplicatedJobNameKey]
 		if !ok {
 			return false, fmt.Errorf(fmt.Sprintf("%s job missing ReplicatedJobName label", job.Name))
 		}
-		if !apiequality.Semantic.DeepEqual(job.Spec.Template.Spec.NodeSelector, nodeSelectors[nodeSelectorKey]) {
+		key := nodeSelectorKey(js.Name, rjobName)
```
ahg-g, line 644: instead of this, just change the key to use the replicatedJob name: https://github.com/kubernetes-sigs/jobset/blob/11db48caf22438a98807091a4fb1cc711b7d6805/test/integration/controller/jobset_controller_test.go#L182

you also need to change: https://github.com/kubernetes-sigs/jobset/blob/11db48caf22438a98807091a4fb1cc711b7d6805/test/integration/controller/jobset_controller_test.go#L609 to use only the replicatedjob name

danielvegamyhre: done

PR #168, opened 2023-05-31
Delete active jobs for finished JobSets

### LINKED ISSUE #162: Delete active child jobs when jobset is considered failed
Currently we leave the jobs running to complete on their own, but this may not be desirable since we want the jobs to immediately release the resources they hold; so to be consistent with the Job API, we should delete all active jobs once the job succeeds or fails.

/kind bug

danielvegamyhre:
Fixes #162 

Also fixes a bug in the `success policy 'all' with empty replicated jobs list` test case parameters (success policy target should be empty) which had somehow slipped our test logic until I added these changes to the controller.

### DIFF
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -257,7 +257,7 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				return testJobSet(ns).
 					SuccessPolicy(&jobset.SuccessPolicy{
 						Operator:             jobset.OperatorAll,
-						TargetReplicatedJobs: []string{"replicated-job-a"},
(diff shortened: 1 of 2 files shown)


test/integration/controller/jobset_controller_test.go
```diff
@@ -740,6 +782,36 @@ func checkExpectedServices(js *jobset.JobSet) {
 	}).Should(gomega.Equal(numExpectedServices(js)))
 }
 
+// Check one headless service per job was created successfully.
```
ahg-g, line 785: wrong comment

danielvegamyhre: fixed

PR #140, opened 2023-05-11
Fix suspend tests

kannon92:
Fixes #116 

I notice that Suspend/Resume seem to have cache errors but in a flaky way.  I also found cases where we were modifying the js but not getting the latest when doing test updates.

PR #172, opened 2023-06-01
Reset StartTime before updating the jobs NodeSelector

ahg-g:
Otherwise api-server will reject the nodeSelector update if the job was previously unsuspended then suspended. This unfortunately can't be tested in an integration test because this bug only manifests if we have a job controller

Also fixed the json tag to one of the parameters.

### DIFF
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -43,24 +43,6 @@ const (
 )
 
 var _ = ginkgo.Describe("JobSet validation", func() {
-	// Each test runs in a separate namespace.
-	var ns *corev1.Namespace
-
-	ginkgo.BeforeEach(func() {
-		// Create test namespace before each test.
-		ns = &corev1.Namespace{
-			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: "test-ns-",
-			},
-		}
-		gomega.Expect(k8sClient.Create(ctx, ns)).To(gomega.Succeed())
-
-	})
-
-	ginkgo.AfterEach(func() {
-		gomega.Expect(testutil.DeleteNamespace(ctx, k8sClient, ns)).To(gomega.Succeed())
-	})
-
 	// jobSetUpdate contains the mutations to perform on the jobset and the
 	// checks to perform afterwards.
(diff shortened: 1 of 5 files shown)


ahg-g:
I incorporated some of the changes in https://github.com/kubernetes-sigs/jobset/pull/140 which are necessary to handle conflicts when we update job or jobset inside the tests.

test/integration/controller/jobset_controller_test.go
```diff
@@ -86,23 +60,32 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 		func(tc *testCase) {
 			ctx := context.Background()
 
+			// Create test namespace for each entry.
+			ns := &corev1.Namespace{
```
danielvegamyhre, line 64: What effect does moving the test namespace creation/deletion logic here (instead of in BeforeEach/AfterEach) have? Looking at the integration test logs I still see various reconciler errors about trying to update an out-of-date object, although all the tests pass. Was there some other issue that this resolved?

test/util/util.go
```diff
@@ -132,7 +132,7 @@ func DeleteNamespace(ctx context.Context, c client.Client, ns *corev1.Namespace)
 	if err != nil && !apierrors.IsNotFound(err) {
 		return err
 	}
-	if err := c.Delete(ctx, ns); err != nil && !apierrors.IsNotFound(err) {
+	if err := c.Delete(ctx, ns, client.PropagationPolicy(metav1.DeletePropagationForeground)); err != nil && !apierrors.IsNotFound(err) {
```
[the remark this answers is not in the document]
ahg-g: so we are sure that the namespace is deleted before starting the next entry.

test/integration/controller/jobset_controller_test.go
ahg-g: I am concerned about the global `ns` variable, what if we want to run the entries in parallel?

>  I still see various reconciler errors about trying to update an out-of-date object, although all the tests pass. 

Those reconcile errors happen because the tests update the JobSet object while the controller trying to update the status. The most common one is the first update to the ready/success counters, which always happens on the first reconcile event.

Issue #163, opened 2023-05-26
Add python and go client libraries


ahg-g:
**What would you like to be added**:

python and go client libraries

**Why is this needed**:

To be able to use the apis in code.

ahg-g:
@tenzen-y can you advise how to do this?

tenzen-y:
In go libraries, we can automatically generate by code-generator.

But, in Python libraries, I suggest implementing the client based on OpenAPI Schema.

tenzen-y:
For example

training-operator: https://github.com/kubeflow/training-operator/blob/fc8a64464f364565fcd399d9d7724201e48b0dfa/sdk/python/kubeflow/training/api/training_client.py
katib: https://github.com/kubeflow/katib/blob/6e0069bc7eefe8b7b8a41bce7ea91219253da575/sdk/python/v1beta1/kubeflow/katib/api/katib_client.py

ahg-g:
Thanks @tenzen-y 

> But, in Python libraries, I suggest implementing the client based on OpenAPI Schema.

Can you elaborate more? the examples you shared were written manually?!

tenzen-y:
> Can you elaborate more? the examples you shared were written manually?!

Yes, manually...
Because we can not find to generate a client for Python from the API definition.
So, we manually implement based on open API schema generated from Golang Kubernetes API in the following step:

1. generate open API schema using k8s.io/openapi-gen
2. generate swagger file from openapi schema.
3. generate Python `model` from swagger file.
4. manually implement Python client.

tenzen-y:
If there is any solution to auto-generate the Python client, I'd like to know that...

kannon92:
Have we considered generating proto definitions from our CRD?  That would help you generate python bindings.  I know that k8 apis do generate proto from the go structs.

tenzen-y:
We probably need to do step 4 even If we can generate protocol buffers definitions from our CRD.
IIUC, we can not generate a Python client from protocol buffer definition.

tenzen-y:
The key here is there aren't tools to generate a Python client like https://github.com/kubernetes/code-generator.

tenzen-y:
If we skip step 4, we can probably inform users to use [`CustomObjectsAPI`](https://github.com/kubernetes-client/python/blob/9a2d5e4290920225b528c84aa26d9305bdcc7baf/kubernetes/client/api/custom_objects_api.py#L27).

[...]

PR #173, opened 2023-06-01
Add Go client

danielvegamyhre:
Part of #163


tenzen-y:
@danielvegamyhre We must add `doc.go` to generate appropriate codes.

https://github.com/kubernetes-sigs/kueue/blob/a3c76a537dfd68bb49cd661c6c3e7e1044c375f2/apis/kueue/v1beta1/doc.go

danielvegamyhre:
@tenzen-y thanks, I thought that was a generated file when I saw it at first. I've added it and re-ran the generate command but the client-go files are still empty. Do you know if the package structure `apis/<groupName>/<version>/*.go` is required? This old [article](https://cloud.redhat.com/blog/kubernetes-deep-dive-code-generation-customresources) linked in the code-generator README uses that convention and so does Kueue, so I'm wondering that that's the issue

api/v1alpha1/jobset_types.go
```diff
@@ -93,6 +93,7 @@ type ReplicatedJobStatus struct {
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
 // +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type=date,description="Time this JobSet was created"
+// +genclient
```
tenzen-y, line 96:
```suggestion
// +genclient

```

We need a new line.

danielvegamyhre: Yeah just noticed this is mentioned in the article as well - I added an empty line and moved the `// +genclient` to the top above the kubebuilder tags, following the pattern from the Kueue PR but still no luck.

tenzen-y:
> @tenzen-y thanks, I thought that was a generated file when I saw it at first. I've added it and re-ran the generate command but the client-go files are still empty. Do you know if the package structure `apis/<groupName>/<version>/*.go` is required? This old [article](https://cloud.redhat.com/blog/kubernetes-deep-dive-code-generation-customresources) linked in the code-generator README uses that convention and so does Kueue, so I'm wondering that that's the issue

IIUC, we need to follow the dir structure.

https://github.com/kubernetes/code-generator/blob/65c70a51ae8eca63d4b9ab1e80fe636097a6ce63/generate-groups.sh#L31

api/v1alpha1/jobset_types.go
tenzen-y: What a timing! Nice :)

danielvegamyhre:
> > @tenzen-y thanks, I thought that was a generated file when I saw it at first. I've added it and re-ran the generate command but the client-go files are still empty. Do you know if the package structure `apis/<groupName>/<version>/*.go` is required? This old [article](https://cloud.redhat.com/blog/kubernetes-deep-dive-code-generation-customresources) linked in the code-generator README uses that convention and so does Kueue, so I'm wondering that that's the issue
> 
> IIUC, we need to follow the dir structure.
> 
> https://github.com/kubernetes/code-generator/blob/eec869e1b034acd494e70dee51edafac48a055f1/generate-groups.sh#L30-L32

Yep changing the directory structure worked! Thanks

ahg-g:
Thanks @danielvegamyhre; this PR generates the client libraries for go, what is the next step to generate the apis in python?

danielvegamyhre:
@ahg-g I asked in the api machinery slack about generating protos from our CRD (which we could use to generate the Python types) but was told that this is [not supported]https://kubernetes.slack.com/archives/C0EG7JC6T/p1685662364273809?thread_ts=1685647927.474879&cid=C0EG7JC6T), so the only option I can think of would be to write the protos by hand. This is doable but we will need to manually update the protos every time we make a change to the API, which will be more error prone than the the process being automated as CI step. What do you think?

PR #183, opened 2023-06-08
update jobset api to v1alpha2

vsoch:
Problem: we are preparing to restructure the JobSet Networking sections, and it is appropriate to bump the API version up before this change. Solution: bump from v1alpha1 to v1alpha2

Note that I've never done an api version update for an operator - let me know if I did this incorrectly! I saw in the MPI operator there was only one version, so I opted to rewrite everything for that version and rename directories that were v1alpha1 to v1alpha2.

This is a per-requisite to #171


danielvegamyhre:
@ahg-g in order to bump to v1alpha2 would we move to a multi-version API (as described in kubebuilder tutorial [here](https://book.kubebuilder.io/multiversion-tutorial/tutorial.html)) to have both v1alpha1 and v1alpha2, or do a version replacement such as this?

ahg-g:
> @ahg-g in order to bump to v1alpha2 would we move to a multi-version API (as described in kubebuilder tutorial [here](https://book.kubebuilder.io/multiversion-tutorial/tutorial.html)) to have both v1alpha1 and v1alpha2, or do a version replacement such as this?

We should move to multi-version API now, but for v1alpha1->v1alpha2 we don't need to keep both versions; only when moving from beta to ga we will need that.

vsoch:
@ahg-g @danielvegamyhre do you know what this build error is / how I can resolve it?

vsoch:
ohh I see the bug, I missed a capital letter!

client-go/clientset/versioned/clientset.go
```diff
@@ -21,23 +21,23 @@ import (
 	discovery "k8s.io/client-go/discovery"
 	rest "k8s.io/client-go/rest"
 	flowcontrol "k8s.io/client-go/util/flowcontrol"
-	jobsetv1alpha1 "sigs.k8s.io/jobset/client-go/clientset/versioned/typed/jobset/v1alpha1"
+	jobsetv1alpha2 "sigs.k8s.io/jobset/client-go/clientset/versioned/typed/jobset/v1alpha2"
```
ahg-g, line 24: were the client-go files automatically moved and updated?

vsoch: I moved them - was I not supposed to?

ahg-g: @danielvegamyhre those were supposed to be auto-generated, right?

@vsoch I assume they should get generated automatically via `make generate`

danielvegamyhre: files in client-go/ are auto-generated by hack/update-codegen.sh so i think we need to update that script to use the v1alpha2 package instead of v1alpha1.

vsoch: I did see the docs here: https://book.kubebuilder.io/multiversion-tutorial/tutorial.html but I found it confusing at best. It seems to walk through using the tool to generate a second version, and I didn't think we wanted another directory, but rather just a rename (e..g., the MPI operator just has their latest: https://github.com/kubeflow/mpi-operator/tree/master/pkg/apis/kubeflow). Feel free to close this and just have someone else do it if I'm totally missing the boat.

PR #188, opened 2023-06-13
Added active status of jobset for integration with kueue

### LINKED ISSUE #185: Add Active bool type field in status
status.active should be false if no active child jobs.

**Why is this needed**:
For Kueue integration

/kind feature

stuton:
Fixes #185

/kind feature

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -80,6 +80,9 @@ type JobSetStatus struct {
 	// Restarts tracks the number of times the JobSet has restarted (i.e. recreated in case of RecreateAll policy).
 	Restarts int `json:"restarts,omitempty"`
 
+	// Active shows if there is at least one active child task
+	Active bool `json:"active,omitempty"`
+
 	// ReplicatedJobsStatus track the number of JobsReady for each replicatedJob.
 	// +optional
 	// +listType=map
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -80,9 +80,6 @@ type JobSetStatus struct {
 	// Restarts tracks the number of times the JobSet has restarted (i.e. recreated in case of RecreateAll policy).
 	Restarts int `json:"restarts,omitempty"`
 
-	// Active shows if there is at least one active child task
-	Active bool `json:"active,omitempty"`
-
 	// ReplicatedJobsStatus track the number of JobsReady for each replicatedJob.
 	// +optional
 	// +listType=map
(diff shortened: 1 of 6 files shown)


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -74,6 +74,9 @@ type JobSetStatus struct {
 	// Restarts tracks the number of times the JobSet has restarted (i.e. recreated in case of RecreateAll policy).
 	Restarts int `json:"restarts,omitempty"`
 
+	// Active shows if there is at least one active child task
+	Active bool `json:"active,omitempty"`
```
ahg-g, line 78: This is not useful on of its own for JobSet users. But you can add the Active count under `ReplicatedJobStatus` similar to ready/succeeded/failed.

stuton: as I know, @mcariatm is going to use this property in kueue integration here: 
https://github.com/kubernetes-sigs/kueue/pull/762/files#diff-d3853fca35347a8a9499f4941707f191a6674df29274b49a9b4ec568eacb0605R79

pkg/controllers/jobset_controller.go
```diff
@@ -266,6 +267,7 @@ func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, j
 				log.Error(nil, fmt.Sprintf("job %s missing ReplicatedJobName label", job.Name))
 			}
 		}
+		replicatedJobsReady[job.Labels[jobset.ReplicatedJobNameKey]]["active"] += job.Status.Active
```
ahg-g, line 270: we need to track the number of active jobs not pods, so we should do the following outside this loop (like what we are doing for successful and failed below)

```
	for _, job := range jobs.active {
		replicatedJobsReady[job.Labels[jobset.ReplicatedJobNameKey]]["active"]++
	}
```

stuton: I'm not sure that we need to do another loop to iterate over active jobs when we can do this already inside an existing one

pkg/controllers/jobset_controller.go
```diff
@@ -266,6 +268,9 @@ func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, j
 				log.Error(nil, fmt.Sprintf("job %s missing ReplicatedJobName label", job.Name))
 			}
 		}
+		if job.Status.Active > 0 {
+			replicatedJobsReady[job.Labels[jobset.ReplicatedJobNameKey]]["active"]++
+		}
```
ahg-g, line 273: we need to check if the label exists, we do that in 265 for the ready case, can you pls move that `if` out to the beginning of the function:

```
  log.Error(nil, fmt.Sprintf("job %s missing ReplicatedJobName label, can't update status", job.Name))
  return
}
```

Issue #190, opened 2023-06-15
Python SDK: Install Python and Java from source rather than using package manager


danielvegamyhre:
**What would you like to be added**:
We should install Python and Java from source rather than using an environment specific package manager (apt).

- [openjdk 11 installation](https://github.com/kubernetes-sigs/jobset/blob/main/hack/python-sdk/gen-sdk.sh#L47-L48)
- [python3 and pip installation](https://github.com/kubernetes-sigs/jobset/blob/main/hack/python-sdk/test-sdk.sh#L19-L25)

**Why is this needed**:
To improve the developer experience SDK generation (ran by `make generate`) and testing scripts (ran by `make test` ) more platform agnostic.

kannon92:
Rather than building from source, should we just have all this run in a container?  Building from source is going unlease a maintenance headache as you will have to support issues related to builds on all different kinds of Linux OSes.  

I look forward to this issue as my dev environment is quite flaky now as I don't have apt.

kannon92:
Just an update on this issue, I had to change the python sdk to use a container.  @danielvegamyhre and I think the way forward should be using containers to generate these clients rather than relying on package managers or building from source.

so the client go one should ideally also use a container to generate sdk

jedwins1998:
@chansuke are you still working on this. I will assign it to myself if not.

chansuke:
@jedwins1998 I'm not working.You can take it🫡

epicseven-cup:
The end of life for openjdk-11 will be October 2024 and openjdk-17 is superseded on openjdk's website https://jdk.java.net/17/. 

Will grabbing openjdk-17 from other vendors such as Microsoft, Oracle.. be fine?

danielvegamyhre:
> The end of life for openjdk-11 will be October 2024 and openjdk-17 is superseded on openjdk's website https://jdk.java.net/17/.
> 
> Will grabbing openjdk-17 from other vendors such as Microsoft, Oracle.. be fine?

Yes that should be fine.

epicseven-cup:
Also just a quick question since python sdk is being generated from openapi-generator-cli I'm assuming that we need to update that anymore, just want to double check

danielvegamyhre:
> Just an update on this issue, I had to change the python sdk to use a container. @danielvegamyhre and I think the way forward should be using containers to generate these clients rather than relying on package managers or building from source.
> 
> so the client go one should ideally also use a container to generate sdk

[...]

PR #191, opened 2023-06-16
Install Python and Java from source

chansuke:
Fixes https://github.com/kubernetes-sigs/jobset/issues/190

Update the scripts to install Java and Python from source.

### DIFF
--- a/hack/python-sdk/gen-sdk.sh
+++ b/hack/python-sdk/gen-sdk.sh
@@ -21,6 +21,9 @@ set -o nounset
 set -o pipefail
 set -x
 
+# store initial working directory
+initial_pwd=$(pwd)
+
 repo_root="$(dirname "${BASH_SOURCE}")/../.."
 
 SWAGGER_JAR_URL="https://repo1.maven.org/maven2/org/openapitools/openapi-generator-cli/4.3.1/openapi-generator-cli-4.3.1.jar"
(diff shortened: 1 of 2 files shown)


hack/python-sdk/gen-sdk.sh
```diff
@@ -44,8 +44,27 @@ fi
 
 if [ -z `which java` ]; then
   echo "Installing OpenJDK 11"
+
   apt update -y
-  apt-get install openjdk-11-jdk -y
+
+  apt-get install -y wget unzip zip git autoconf build-essential \
```
danielvegamyhre, line 50: The idea of installing from source was to decouple the installation process from environment specific package managers like `apt` - any way we can do this without using apt?

danielvegamyhre (changes requested):
The idea of installing from source was to decouple the installation process from environment specific package managers like apt - any way we can do this without using apt?

hack/python-sdk/gen-sdk.sh
chansuke: @danielvegamyhre 
Thank you for your review. I will remove package-manager related part.

danielvegamyhre:
Thanks for working on this, it may address #193 as well

chansuke:
@danielvegamyhre 
Thank you so much for your comment.

All the tests are passing except for `pull-jobset-test-unit-main`. The error message that I am seeing is:
> make: *** [Makefile:118: test-python-sdk] Error 1

I am not able to replicate this error on my local machine. Do you have any suggestions on how I could debug this issue further?

kannon92:
Maybe we should look into using a container for generation?

Only way I can think of to avoid doing apt on a user machine and having a more stable build.

danielvegamyhre:
@chansuke Try adding `set -x` to the top of `test-sdk.sh` file near the other `set -o ...` lines. This should enable you to see what line is causing the error in the logs.

danielvegamyhre:
@chansuke are you still working on this? I think @kannon92 is right, we should look into using a container for generation if possible, as we may be able to get downloading and building from source to work in CI and some of our dev environments, but inevitably there will be contributors whose environment will be incompatible with the specific Java/Python builds we're doing here.

danielvegamyhre:
Closing this due to inactivity

PR #201, opened 2023-06-21
update gen-sdk.sh to fallback to openjdk-17

vsoch:
Problem: recent updates to the golang 19 and 20 images are now using debian bookworm, which only has openjdk-17 available. The current builds are broken looking for 11.
Solution: update the install script to try installing openjdk-11, and falling back to openjdk-17 for the interim.

I'm not sure why my computer tweaked the line ending of the Python file - I merely opened it! It must be a Linux thing.


vsoch:
This might be another version update issue:

```
 Processing triggers for libc-bin (2.36-9) ...
error: externally-managed-environment
× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    See /usr/share/doc/python3.11/README.venv for more information.
note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification. 
```
Looking at it now!

vsoch:
Updates for testing:

1. The default Python in the image is 3.11, and the install command would never be hit. If it is hit, it would conflict with the other python installs (that target 3.11) so I updated to just install python3-x of everything, using the default in the image. If we want a specific Python version we need to install everything first and use update-alternatives to make it the default.
2. The error above I believe is because the test-requirements.txt were being installed to the system python. It's better practice to create and source a virtual environment. Reference https://peps.python.org/pep-0668/
3. Since we need to apt-get update anyway for python3-venv, I moved it once to the top.

Next step - let's run the tests and see where we are!

vsoch:
I'm glad I could help! And welcome back - I hope you feel rested and ready for more fun! :partying_face:

PR #203, opened 2023-06-24
add example and update setup.py to semi functional

### LINKED ISSUE #189: Add example using the Python SDK + Dynamic Client
An example script using the Python SDK + Dynamic client to interact with the JobSet API.

**Why is this needed**:
For users who use Python clients to interact with k8s to quickly and easily use the JobSet API.

Also to catch any bugs that may be present in the Python SDK that aren't caught by unit tests.

kannon92:
Using setuptools to create a jobset package so you could use like a normal python package.  

Brief example for how to use it.

Note: there is still a bug I am unsure how to solve. 

I can't seem to import the example when I reference it outside of the src directory.  

Fixes #189


sdk/python/examples/jobset.py
```diff
@@ -0,0 +1,43 @@
...
+                ready=56,
+                succeeded=56,
+            )
+        ],
+        conditions=[None],
+        restarts=56,
+    ),
+)
+
+print(jobset_example)
```
danielvegamyhre, line 43: Can we use the [dynamic client](https://github.com/kubernetes-client/python-base/blob/master/dynamic/client.py) to create this JobSet?

hack/python-sdk/test-sdk.sh
```diff
@@ -15,30 +15,10 @@ set -o errexit
...
 cd "${repo_root}/sdk/python"
 
-# We need to create a virtual environment for testing
-apt-get install -y python3-venv
-python3 -m venv env
-source env/bin/activate
-
-# install test requirements
-pip install -r test-requirements.txt
+docker buildx build -f Dockerfile -t python-unit .
```
kannon92, line 22: It looks like the ci doesn’t detect docker..  any ideas on how to proceed?  I could switch to using the same runner we use for our e2e tests.

sdk/python/examples/jobset.py
```diff
@@ -0,0 +1,43 @@
+import jobset
+from jobset.models.jobset_v1alpha2_job_set import JobsetV1alpha2JobSet
+from kubernetes.client.models.v1_job_template_spec import V1JobTemplateSpec
+
+jobset_example = jobset.models.jobset_v1alpha2_job_set.JobsetV1alpha2JobSet(
```
danielvegamyhre, line 5: Can we use defaults for some of these parameters to simplify the spec?  For example We can let success policy be defaulted, and Network can be defaulted, etc.

sdk/python/examples/jobset.py
```diff
@@ -0,0 +1,43 @@
...
+    kind="0",
+    metadata=None,
+    spec=jobset.models.jobset_v1alpha2_job_set_spec.JobsetV1alpha2JobSetSpec(
+        failure_policy=jobset.models.jobset_v1alpha2_failure_policy.JobsetV1alpha2FailurePolicy(
+            max_restarts=56,
+        ),
+        replicated_jobs=[
+            jobset.models.jobset_v1alpha2_replicated_job.JobsetV1alpha2ReplicatedJob(
+                name="0",
+                network=jobset.models.jobset_v1alpha2_network.JobsetV1alpha2Network(
```
danielvegamyhre, line 16: Network struct is now an [attribute of JobSetSpec](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L53), not ReplicatedJob. We have 1 headless service for the whole jobset.

sdk/python/examples/jobset.py
```diff
@@ -0,0 +1,43 @@
...
+                    enable_dns_hostnames=True,
+                ),
+                replicas=56,
+                template=V1JobTemplateSpec(),
+            )
+        ],
+        success_policy=jobset.models.jobset_v1alpha2_success_policy.JobsetV1alpha2SuccessPolicy(
+            operator="0",
```
danielvegamyhre, line 24: Success policy operator should be "Any" or "All" (see [here](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L155-L158)). However, as mentioned in a previous comment, we can just let the success policy defaulted instead.

hack/python-sdk/test-sdk.sh
danielvegamyhre: Can we install docker in this script?

kannon92: A goal of this was to use docker so I could build on my dev environment (where I use yum).  I realize this is kinda of an opinioned PR but im trying to reproduce depencies on any OS.

sdk/python/examples/jobset.py
kannon92: I'm not really sure about dynamic client.  I found that you can do `create_namespaced_custom_object` which will allow creation of a CRD assuming that the CRD is installed in your cluster.  

Is this sufficient?  Dynamic client seemed a bit more complicated than what I think we need.

sdk/python/examples/jobset.py
kannon92: done.

sdk/python/examples/jobset.py
kannon92: done.

sdk/python/examples/jobset.py
danielvegamyhre: I see, can you update the example to use this function and then we can manually verify that the example works? The goal is to have some copy-pastable example to create a simple JobSet on a cluster.

kannon92: > I see, can you update the example to use this function and then we can manually verify that the example work (i.e. I will try running the script and confirming a JobSet is created in my cluster)?

You can run `python3 examples/create_jobset.py` and that will create it.  

> The goal is to have some copy-pastable example to create a simple JobSet on a cluster, and perhaps does simple CRUD Operations like listing jobsets, updating a jobset, etc. For this I think the dynamic client is necessary, is it not?


lol, tbh I only learned about the python kubernetes client over the weekend.  I see that you can use CRUD operations without a dynamic client (https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md).

We have a CI fix to make this workable but not sure how to get docker installed in the runner.  One option is to switch the runner to use the same runner we use for e2e tests.  Or I could go about installing docker on the CI if we want to keep the same image.

danielvegamyhre: I don't really mind which image we use the same image as we use for e2e tests so that we can run unit tests in docker. I think it would be cleaner than installing docker locally in the script, what do you think @ahg-g?

ahg-g: I think for simply creating a jobset using the customobjectapi is fine, so the example in this PR is good. But for cases where the user is building an operator that requires watching on JobSet objects, then the dynamic client is the better option. It would be nice if we can add another example that uses the dynamic client to create and create a watch, but not urgent.

kannon92: https://github.com/kubernetes/test-infra/pull/29947 is a PR to change the image.  Hopefully it works..

danielvegamyhre:
Seems like docker daemon isn't running, that's odd..

sdk/python/examples/jobset.py
kannon92: I tried to ask in sig-testing why is the docker daemon not started on this runner but no response.  I'm not sure of the path forward if I can't assume that docker is capable of running containers in the CI.

ahg-g:
a different error now :)

kannon92:
> a different error now :)

And the unit tests pass!  Github was funky so retrying the e2e tests but I think this is good to go.

ahg-g:
Leaving this to @danielvegamyhre to lgtm and approve

danielvegamyhre:
This is great @kannon92 thanks for your work on this. I just confirmed the changes in the PR work for my local dev environment as well as in CI so we are good to merge.

/lgtm
/approve

PR #209, opened 2023-06-27
Update examples to use JobSet name as subdomain

### LINKED ISSUE #208: Update examples to use new networking configuration
In #171 we updated the jobset controller to create 1 headless service for the entire jobset, rather than 1 headless service per replicatedJob, and to use the jobset name as the headless service name and subdomain name by default. 

However, our examples still use the old paradigm with the subdomain name as the replicated job name, so they will not work with the v0.2.0 release until we update them.

danielvegamyhre:
Fixes #208

### DIFF
--- a/examples/pytorch/mnist.yaml
+++ b/examples/pytorch/mnist.yaml
@@ -19,7 +19,7 @@ spec:
               - containerPort: 3389
               env:
               - name: MASTER_ADDR
-                value: "pytorch-workers-0-0.pytorch-workers"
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
/retest

Weird the e2e logs are just empty, scheduler timed out after failing to schedule pods..? Never seen this

danielvegamyhre:
/retest

seems like CI has been borked for a couple days now, unable to pull golang:120 image from GCR... :(

PR #211, opened 2023-06-28
Add job hash key (sha1 hash of namespaced job name) as label and annotation for setting exclusive affinities

### LINKED ISSUE #210: Add namedspaced job name label

Add a namespaced job name label to pods and use that as the label selector for 1:1 exclusive placement.

**Why is this needed**:

The upstream job controller adds the `job-name` label, but it is not namespaced. This is a problem for the exclusive feature we have since job-name is not unique across namespaces.

@danielvegamyhre

danielvegamyhre:
Fixes #210

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -24,7 +24,7 @@ const (
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	ReplicatedJobNameKey  string = "jobset.sigs.k8s.io/replicatedjob-name"
 	JobIndexKey           string = "jobset.sigs.k8s.io/job-index"
-	NamespacedJobNameKey  string = "jobset.sigs.k8s.io/namespaced-job-name"
+	JobHashKey            string = "jobset.sigs.k8s.io/job-hash"
 	JobNameKey            string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 	ExclusiveKey          string = "alpha.jobset.sigs.k8s.io/exclusive-topology"
 )
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -24,7 +24,7 @@ const (
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	ReplicatedJobNameKey  string = "jobset.sigs.k8s.io/replicatedjob-name"
 	JobIndexKey           string = "jobset.sigs.k8s.io/job-index"
-	JobHashKey            string = "jobset.sigs.k8s.io/job-hash"
+	JobKey                string = "jobset.sigs.k8s.io/job-key"
 	JobNameKey            string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 	ExclusiveKey          string = "alpha.jobset.sigs.k8s.io/exclusive-topology"
 )
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -24,6 +24,7 @@ const (
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	ReplicatedJobNameKey  string = "jobset.sigs.k8s.io/replicatedjob-name"
 	JobIndexKey           string = "jobset.sigs.k8s.io/job-index"
+	NamespacedJobNameKey  string = "jobset.sigs.k8s.io/namespaced-job-name"
 	JobNameKey            string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 	ExclusiveKey          string = "alpha.jobset.sigs.k8s.io/exclusive-topology"
 )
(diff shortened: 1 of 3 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -701,6 +704,10 @@ func GenSubdomain(js *jobset.JobSet) string {
 	return js.Name
 }
 
+func namespacedJobName(ns string, jobName string) string {
```
ahg-g, line 707: pass the jobset itself, the object include the name and namespace

pkg/controllers/jobset_controller.go
```diff
@@ -701,6 +704,10 @@ func GenSubdomain(js *jobset.JobSet) string {
...
+	return fmt.Sprintf("%s-%s", ns, jobName)
```
ahg-g, line 708:
```suggestion
	return fmt.Sprintf("%s/%s", ns, jobName)
```

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -24,6 +24,7 @@ const (
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	ReplicatedJobNameKey  string = "jobset.sigs.k8s.io/replicatedjob-name"
 	JobIndexKey           string = "jobset.sigs.k8s.io/job-index"
+	NamespacedJobNameKey  string = "jobset.sigs.k8s.io/namespaced-job-name"
```
ahg-g, line 27: we need to use this key instead of JobNameKey in [`setExclusiveAffinities`](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L608)

pkg/controllers/jobset_controller.go
danielvegamyhre: I used a `/` originally but there was a bunch of errors indicating that this character is not allowed in labels, so I switched to `-`.

ahg-g: sg

pkg/controllers/jobset_controller.go
danielvegamyhre: I originally did this until I started updating the unit tests and realized the [makeJob](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller_test.go#L1011-L1021) test util function does not accept the JobSet itself as one of the params, but rather the makeJobArgs struct with various parameters to set for the job. So I had to switch to the current function signature to be compatible from calls within the controller and in the unit tests.

pkg/controllers/jobset_controller.go
ahg-g: I am concerned about the length of the key, perhaps we should hash it to limit its length and call it job-key instead of job-name?

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: right, of course - done!

pkg/controllers/jobset_controller.go
danielvegamyhre: done

pkg/controllers/jobset_controller.go
danielvegamyhre: replaced this with jobHashKey function

pkg/controllers/jobset_controller.go
danielvegamyhre: I used sha1 hash since the hex digest is 40 chars, less than our 63 char limit.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -24,6 +24,7 @@ const (
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
 	ReplicatedJobNameKey  string = "jobset.sigs.k8s.io/replicatedjob-name"
 	JobIndexKey           string = "jobset.sigs.k8s.io/job-index"
+	JobHashKey            string = "jobset.sigs.k8s.io/job-hash"
```
ahg-g, line 27: I would call it `jobset.sigs.k8s.io/job-key`, hash is too specific.

pkg/controllers/jobset_controller.go
```diff
@@ -663,18 +665,21 @@ func shouldCreateJob(jobName string, ownedJobs *childJobs) bool {
...
 	labels[jobset.ReplicatedJobReplicas] = strconv.Itoa(rjob.Replicas)
 	labels[jobset.JobIndexKey] = strconv.Itoa(jobIdx)
+	labels[jobset.JobHashKey] = jobHashKey(js.Namespace, jobName)
 
 	annotations := util.CloneMap(obj.GetAnnotations())
 	annotations[jobset.JobSetNameKey] = js.Name
 	annotations[jobset.ReplicatedJobNameKey] = rjob.Name
 	annotations[jobset.ReplicatedJobReplicas] = strconv.Itoa(rjob.Replicas)
 	annotations[jobset.JobIndexKey] = strconv.Itoa(jobIdx)
+	annotations[jobset.JobHashKey] = jobHashKey(js.Namespace, jobName)
```
ahg-g, line 682: no necessary as annotation, no one will use that as an env var

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: done

pkg/controllers/jobset_controller.go
danielvegamyhre: done

Issue #212, opened 2023-06-28
Multiple issues in JobSet CI


danielvegamyhre:
CI system has been borked this week, making it difficult to get code checked in... repeatedly running into the following issues:
- failing to pull golang:1.20 image from GCR (seems like an intermittent prow or gcr issue that self-resolves but has been happening a lot lately. example: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/211/pull-jobset-test-e2e-main-1-26/1674124616177553408)
- webhook connection refused (example: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/209/pull-jobset-test-e2e-main-1-26/1674119214257934336)
- jobset object was modified, please retry with latest object" error (example: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/211/pull-jobset-test-integration-main/1674124616152387584)
- "pod deleted unexpectedly" (never seen this one before - example: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/209/pull-jobset-test-e2e-main-1-24/1674119214148882432)
- both e2e "pod ping tests" timeout

cc @ahg-g @kannon92

tenzen-y:
Probably, https://github.com/kubernetes/test-infra/pull/29961 fixed the first issue.

tenzen-y:
Regarding the second issue, adding logging to e2e like [kueue](https://github.com/kubernetes-sigs/kueue/blob/cc58315b5c572fd872e354bfafaea5582c04a60c/hack/e2e-test.sh) would be good And then we can investigate the root cause. I guess that the jobset controller takes a bit long time to start. So I believe that we need to increase resource requests in the e2e env.

tenzen-y:
Regarding the third issue, it seems that we updated the jobset without getting created jobset after creating the jobset.
The issue is typically controller issue. So I would suggest getting jobset after https://github.com/kubernetes-sigs/jobset/blob/2903be85cb40273bff9bbaa19fb6b102d0db1d17/test/integration/controller/jobset_controller_test.go#L88.

tenzen-y:
[...]

PR #213, opened 2023-06-28
Make kind cluster creation block until control plane ready

danielvegamyhre:
Part of #212 

cc @tenzen-y


tenzen-y (commented):
LGTM
leaving `/lgtm` on @ahg-g

PR #214, opened 2023-06-29
add kind logs

kannon92:
Grab logs of kind cluster and follow similar pattern to Kueue (https://github.com/kubernetes-sigs/kueue/blob/cc58315b5c572fd872e354bfafaea5582c04a60c/hack/e2e-test.sh#L27-L40).

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -4,8 +4,15 @@ export KIND=$PWD/bin/kind
 
 function cleanup {
     if [ $USE_EXISTING_CLUSTER == 'false' ] 
-    then 
-        $KIND delete cluster --name $KIND_CLUSTER_NAME
+    then
+        if [ ! -d "$ARTIFACTS" ]; then
+            mkdir -p "$ARTIFACTS"
+        fi
+        kubectl logs -n kube-system kube-scheduler-kind-control-plane > $ARTIFACTS/kube-scheduler.log || true
+        kubectl logs -n kube-system kube-controller-manager-kind-control-plane > $ARTIFACTS/kube-controller-manager.log || true
+        kubectl logs -n kueue-system deployment/kueue-controller-manager > $ARTIFACTS/kueue-controller-manager.log || true
+        kubectl describe pods -n kueue-system > $ARTIFACTS/kueue-system-pods.log || true
+        $KIND delete cluster --name $KIND_CLUSTER_NAME || { echo "You need to run make kind-image-build before this script"; exit -1; }
     fi
     (cd config/components/manager && $KUSTOMIZE edit set image controller=gcr.io/k8s-staging-jobset/jobset:main)
 }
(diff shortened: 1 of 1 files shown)


ahg-g:
I don't see any meaningful logs from kube-scheduler for example: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/214/pull-jobset-test-e2e-main-1-24/1674410094915227648/artifacts/kube-scheduler.log

hack/e2e-test.sh
```diff
@@ -4,15 +4,24 @@ export KIND=$PWD/bin/kind
...
-    then 
-        $KIND delete cluster --name $KIND_CLUSTER_NAME
+    then
+        if [ ! -d "$ARTIFACTS" ]; then
+            mkdir -p "$ARTIFACTS"
+        fi
+        kubectl logs -n kube-system kube-scheduler-kind-control-plane > $ARTIFACTS/kube-scheduler.log || true
+        kubectl logs -n kube-system kube-controller-manager-kind-control-plane > $ARTIFACTS/kube-controller-manager.log || true
+        kubectl logs -n kueue-system deployment/kueue-controller-manager > $ARTIFACTS/kueue-controller-manager.log || true
+        kubectl describe pods -n kueue-system > $ARTIFACTS/kueue-system-pods.log || true
```
ahg-g, line 14: jobset instead of kueue :)

ahg-g: can we also add the api-server?

ahg-g:
> I don't see any meaningful logs from kube-scheduler for example: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/214/pull-jobset-test-e2e-main-1-24/1674410094915227648/artifacts/kube-scheduler.log

is it possible that we are not setting a high enough log level for kube-scheduler?

kannon92:
> > I don't see any meaningful logs from kube-scheduler for example: https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/214/pull-jobset-test-e2e-main-1-24/1674410094915227648/artifacts/kube-scheduler.log
> 
> is it possible that we are not setting a high enough log level for kube-scheduler?

If we want that I think we can expand our Kind config to increase logs.  But if we need it I think that can be done as a followup  separately?

hack/e2e-test.sh
```diff
@@ -4,15 +4,25 @@ export KIND=$PWD/bin/kind
 
 function cleanup {
     if [ $USE_EXISTING_CLUSTER == 'false' ] 
-    then 
-        $KIND delete cluster --name $KIND_CLUSTER_NAME
+    then
+        if [ ! -d "$ARTIFACTS" ]; then
+            mkdir -p "$ARTIFACTS"
+        fi
+        kubectl logs -n kube-system kube-scheduler-kind-control-plane > $ARTIFACTS/kube-scheduler.log || true
```
danielvegamyhre, line 11: Perhaps we can do this in a separate PR but since ARTIFACTS is set as the `$PROJECT_DIR/bin` folder [here](https://github.com/kubernetes-sigs/jobset/blob/main/Makefile#L40) think we should think about storing logs in a separate directory, rather than the bin folder.

ahg-g: I can access them from the artifacts tab on prow

danielvegamyhre: ah ok i see

PR #215, opened 2023-06-29
eventually updates

kannon92:
https://github.com/kubernetes-sigs/jobset/issues/212

Aiming to fix some of the integration failures described in the above issue.

### DIFF
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -46,7 +46,7 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 	// jobSetUpdate contains the mutations to perform on the jobset and the
 	// checks to perform afterwards.
 	type jobSetUpdate struct {
-		fn            func(*jobset.JobSet)
+		fn func(*jobset.JobSet)
 	}
 
 	type testCase struct {
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -51,9 +51,8 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 	}
 
 	type testCase struct {
-		makeJobSet                  func(*corev1.Namespace) *testing.JobSetWrapper
-		jobSetCreationShouldSucceed bool
-		updates                     []*jobSetUpdate
+		makeJobSet func(*corev1.Namespace) *testing.JobSetWrapper
+		updates    []*jobSetUpdate
 	}
 
 	ginkgo.DescribeTable("JobSet validation during creation and updates",
@@ -77,15 +76,6 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 			ginkgo.By("creating jobset")
 			js := tc.makeJobSet(ns).Obj()
 
-			// If we are expected a validation error creating the jobset, end the test early.
-			if !tc.jobSetCreationShouldSucceed {
-				ginkgo.By("checking that jobset creation fails")
-				gomega.Eventually(func() error {
-					return k8sClient.Create(ctx, js)
-				}, timeout, interval).ShouldNot(gomega.Succeed())
-				return
-			}
-
 			// Verify jobset created successfully.
 			ginkgo.By("checking that jobset creation succeeds")
 			gomega.Eventually(func() error {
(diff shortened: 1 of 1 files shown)


test/integration/controller/jobset_controller_test.go
```diff
@@ -96,11 +101,23 @@ var _ = ginkgo.Describe("JobSet validation", func() {
...
+								return err
+							}
+							return k8sClient.Update(ctx, &jsGet)
+						}, timeout, interval).ShouldNot(gomega.Succeed())
 						continue
 					}
 					// Verify a valid jobset update succeeded.
-					gomega.Expect(k8sClient.Update(ctx, js)).To(gomega.Succeed())
+					gomega.Eventually(func() error {
+						var jsGet jobset.JobSet
```
ahg-g, line 115: we should call `update.fn(js)`, otherwise we are not updating anything.

test/integration/controller/jobset_controller_test.go
```diff
@@ -96,11 +101,23 @@ var _ = ginkgo.Describe("JobSet validation", func() {
...
 					// If we are expecting a validation error, we can skip the rest of the checks.
 					if !update.shouldSucceed {
-						gomega.Expect(k8sClient.Update(ctx, js)).To(gomega.Not(gomega.Succeed()))
+						gomega.Eventually(func() error {
+							var jsGet jobset.JobSet
+							if err := k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jsGet); err != nil {
+								return err
+							}
+							return k8sClient.Update(ctx, &jsGet)
+						}, timeout, interval).ShouldNot(gomega.Succeed())
```
ahg-g, line 110: hmm, we expect this update to not succeed , so lets leave this alone.

test/integration/controller/jobset_controller_test.go
```diff
@@ -79,13 +80,17 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 			// If we are expected a validation error creating the jobset, end the test early.
 			if !tc.jobSetCreationShouldSucceed {
 				ginkgo.By("checking that jobset creation fails")
-				gomega.Expect(k8sClient.Create(ctx, js)).To(gomega.Not(gomega.Succeed()))
+				gomega.Eventually(func() error {
+					return k8sClient.Create(ctx, js)
+				}, timeout, interval).ShouldNot(gomega.Succeed())
```
ahg-g, line 85: we can remove this logic and the field `jobSetCreationShouldSucceed`, there is only a single test case that doesn't use it.

test/integration/controller/jobset_controller_test.go
ahg-g: actually, there is only a single test case that doesn't set this to false, so we can remove this code and `shouldSucceed` as well

test/integration/controller/jobset_controller_test.go
```diff
@@ -175,7 +194,9 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				} else if up.jobUpdateFn != nil {
 					// Fetch updated job objects so we always have the latest resource versions to perform mutations on.
 					var jobList batchv1.JobList
-					gomega.Expect(k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace))).To(gomega.Succeed())
+					gomega.Eventually(func() error {
+						return k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace))
+					}, timeout, interval).Should(gomega.Succeed())
```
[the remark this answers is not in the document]
kannon92: removed.

test/integration/controller/jobset_controller_test.go
kannon92: Nice catch!  Updated.

test/integration/controller/jobset_controller_test.go
kannon92: Removed this function and the flag.

test/integration/controller/jobset_controller_test.go
kannon92: done.

test/integration/controller/jobset_controller_test.go
```diff
@@ -51,9 +51,8 @@ var _ = ginkgo.Describe("JobSet validation", func() {
 	}
```
ahg-g, line 51: can you please remove the `shouldSucceed` parameter too

Issue #219, opened 2023-06-30
Bug: Changing the API and regenerating client-go does not add changes to code repo.


kannon92:
When I modify the api and try to generate client-go, my client-go is generated and placed in my go src directory (ie /home/ec2-user/go/1.20.2/src/sigs.k8s.io/jobset/client-go/).  

I think that it should go to client-go folder in the repo.  Currently if I modify the API I would have to copy /home/ec2-user/go/1.20.2/src/sigs.k8s.io/jobset/client-go/ and rewrite the folder in client-go.  

This is not ideal.  

Any ideas on what is going wrong?

danielvegamyhre:
When originally developing this I found that `generate-groups.sh` script called by `update-codegen.sh` accepts an "output package" parameter, which I specified as `sigs.k8s.io/jobset/client-go`, and treats each thing between the `/` chars as a directory relative to `$GOROOT/src`. So in my case this writes to `~/go/src/sigs.k8s.io/jobset/client-go`, and since my jobset repo is located at `~/go/src/sigs.k8s.io/jobset`, this is the correct location.

However, I can reproduce this issue by cloning jobset in my /tmp directory and running `make`, it still writes the client-go to the same location relative to my Go root, not relative to my current working directory. I don't see an obvious way to change this in the script, as it only seems to accept an "output package" parameter, not a filesystem location of to create output files relative to.

I think maybe at the end of the script we just need to move the output `client-go` directory to the location relative to the current working directory.

kannon92:
I think copying is fine.  Ideally I’d like to get the this running in a container so I think we would copy the build out of co Rainer anyway.

PR #220, opened 2023-06-30
use int32 rather than int

### LINKED ISSUE #218: API integer types should have explicit size (int32, int64)
This should be `int32` https://github.com/kubernetes-sigs/jobset/blob/b69dd6aad5375ef191e780d35b883fbcb9142656/api/jobset/v1alpha2/jobset_types.go#L124

Given the validation that it's already in place, I think it should be safe to change the type.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#primitive-types

kannon92:
Potential PR to fix #218 

Fixes #218 

Note during this I discovered #219 but I manually copied the changes into client-go.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -559,8 +559,7 @@ func updateCondition(js *jobset.JobSet, condition metav1.Condition) bool {
 }
 func constructJobsFromTemplate(js *jobset.JobSet, rjob *jobset.ReplicatedJob, ownedJobs *childJobs) ([]*batchv1.Job, error) {
 	var jobs []*batchv1.Job
-	var jobIdx int32
-	for jobIdx = 0; jobIdx < rjob.Replicas; jobIdx++ {
+	for jobIdx := 0; jobIdx < int(rjob.Replicas); jobIdx++ {
 		jobName := genJobName(js, rjob, jobIdx)
 		if create := shouldCreateJob(jobName, ownedJobs); !create {
 			continue
@@ -574,7 +573,7 @@ func constructJobsFromTemplate(js *jobset.JobSet, rjob *jobset.ReplicatedJob, ow
 	return jobs, nil
 }
 
-func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int32) (*batchv1.Job, error) {
+func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*batchv1.Job, error) {
 	job := &batchv1.Job{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:      util.CloneMap(rjob.Template.Labels),
@@ -665,7 +664,7 @@ func shouldCreateJob(jobName string, ownedJobs *childJobs) bool {
 	return true
 }
 
-func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int32) {
+func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) {
 	jobName := genJobName(js, rjob, jobIdx)
 	labels := util.CloneMap(obj.GetLabels())
 	labels[jobset.JobSetNameKey] = js.Name
@@ -694,7 +693,7 @@ func JobFinished(job *batchv1.Job) (bool, batchv1.JobConditionType) {
 	return false, ""
 }
 
-func genJobName(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIndex int32) string {
+func genJobName(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIndex int) string {
 	return fmt.Sprintf("%s-%s-%d", js.Name, rjob.Name, jobIndex)
 }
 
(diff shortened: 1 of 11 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -559,7 +559,8 @@ func updateCondition(js *jobset.JobSet, condition metav1.Condition) bool {
 }
 func constructJobsFromTemplate(js *jobset.JobSet, rjob *jobset.ReplicatedJob, ownedJobs *childJobs) ([]*batchv1.Job, error) {
 	var jobs []*batchv1.Job
-	for jobIdx := 0; jobIdx < rjob.Replicas; jobIdx++ {
+	var jobIdx int32
+	for jobIdx = 0; jobIdx < rjob.Replicas; jobIdx++ {
```
danielvegamyhre, line 563: I think it would be simpler to just do type conversion in the loop, then we don't need to change all these function signatures that accept a `jobIdx int` parameter. What do you think?

```suggestion
	for jobIdx := 0; int32(jobIdx) < rjob.Replicas; jobIdx++ {
```

kannon92: updated.

sdk/python/requirements.txt
```diff
@@ -1,4 +1,5 @@
 certifi >= 14.05.14
+future; python_version<="2.7"
```
danielvegamyhre, line 2: Just to clarify, is this for backward compatibility with python 2.7 and earlier?

kannon92: I may have removed it in my refactor of python sdk but this regenerated it so it may be adding this in as part of the sdk generation.

kannon92: I did not manually add that in.

pkg/controllers/jobset_controller.go
```diff
@@ -669,16 +669,16 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
...
+	labels[jobset.JobIndexKey] = strconv.Itoa(int(jobIdx))
 	labels[jobset.JobKey] = jobHashKey(js.Namespace, jobName)
 
 	annotations := util.CloneMap(obj.GetAnnotations())
 	annotations[jobset.JobSetNameKey] = js.Name
 	annotations[jobset.ReplicatedJobNameKey] = rjob.Name
-	annotations[jobset.ReplicatedJobReplicas] = strconv.Itoa(rjob.Replicas)
-	annotations[jobset.JobIndexKey] = strconv.Itoa(jobIdx)
+	annotations[jobset.ReplicatedJobReplicas] = strconv.Itoa(int(rjob.Replicas))
+	annotations[jobset.JobIndexKey] = strconv.Itoa(int(jobIdx))
```
danielvegamyhre, line 681: We don't need to do `int(jobIdex)` anymore since it is now an int instead of int32.

test/e2e/e2e_test.go
```diff
@@ -168,10 +169,11 @@ func pingTestJobSet(ns *corev1.Namespace) *testing.JobSetWrapper {
 func pingTestJobSetSubdomain(ns *corev1.Namespace) *testing.JobSetWrapper {
 	jsName := "js"
 	rjobName := "rjob"
-	replicas := 4
+	var replicas int32 = 4
+	var jobIdx int32 = 0
 	subdomain := "network-subdomain"
 	var podHostnames []string
-	for jobIdx := 0; jobIdx < replicas; jobIdx++ {
+	for jobIdx = 0; jobIdx < replicas; jobIdx++ {
```
danielvegamyhre, line 176: Instead of changing these lines, we can just do a one line changes on line 199 below when setting the wrapper replicas (e.g. `Replicas(int32(replicas))`)

test/util/util.go
```diff
@@ -34,20 +34,20 @@ import (
...
 const interval = time.Millisecond * 250
 
-func NumExpectedJobs(js *jobset.JobSet) int {
-	expectedJobs := 0
+func NumExpectedJobs(js *jobset.JobSet) int32 {
+	var expectedJobs int32 = 0
 	for _, rjob := range js.Spec.ReplicatedJobs {
 		expectedJobs += rjob.Replicas
 	}
 	return expectedJobs
```
danielvegamyhre, line 42:
```suggestion
	return int(expectedJobs)
```
Can we return an int here to avoid the various conversions to int32 when checking number of expected jobs (e.g. `if int32(len(jobList.Items)) != testutil.NumExpectedJobs(js) { ... `)?

pkg/controllers/jobset_controller.go
kannon92: nice suggestion! Fixed!

test/e2e/e2e_test.go
kannon92: Updated.

test/util/util.go
kannon92: updated.

kannon92: Doesn't quite work like you said because Replicas is int32.

test/util/util.go
```diff
@@ -47,7 +47,7 @@ func NumJobs(ctx context.Context, k8sClient client.Client, js *jobset.JobSet) (i
 	if err := k8sClient.List(ctx, &jobList, client.InNamespace(js.Namespace)); err != nil {
 		return -1, err
 	}
-	return len(jobList.Items), nil
+	return int(len(jobList.Items)), nil
```
danielvegamyhre, line 50:
```suggestion
	return len(jobList.Items), nil
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -761,7 +761,7 @@ func matchJobsNodeSelectors(js *jobset.JobSet, nodeSelectors map[string]map[stri
 		jobsUpdated++
 	}
 	// Calculate expected number of updated jobs
-	wantJobsUpdated := 0
+	var wantJobsUpdated int32 = 0
 	for _, rjob := range js.Spec.ReplicatedJobs {
 		if _, exists := nodeSelectors[rjob.Name]; exists {
 			wantJobsUpdated += rjob.Replicas
```
danielvegamyhre, line 767:
```suggestion
			wantJobsUpdated += int(rjob.Replicas)
```
I think it would be better to stick to this same pattern used in the other functions `numJobsExpectedToSucceed()` and `NumExpectedJobs()`.

Looks good to me after this!

kannon92: updated.

PR #221, opened 2023-07-05
Automatically move generated client-go files to repo root

### LINKED ISSUE #219: Bug: Changing the API and regenerating client-go does not add changes to code repo.

danielvegamyhre:
Fixes #219

### DIFF
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -21,6 +21,9 @@ set -o pipefail
 cd "$(dirname "${0}")"
 GO_CMD=${1:-go}
 CODEGEN_PKG=${2:-../bin}
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+
+echo "GOPATH=$GOPATH"
 
 bash "${CODEGEN_PKG}/generate-groups.sh" \
   "all" \
@@ -28,3 +31,6 @@ bash "${CODEGEN_PKG}/generate-groups.sh" \
   sigs.k8s.io/jobset/api \
   jobset:v1alpha2 \
   --go-header-file ./boilerplate.go.txt
+
+echo "moving generated files from $GOPATH/src to $REPO_ROOT/client-go"
+mv $GOPATH/src/client-go $REPO_ROOT
--- a/hack/update-codegen.sh
+++ b/hack/update-codegen.sh
@@ -37,7 +37,7 @@ if [ ! -d "$REPO_ROOT/client-go" ]; then
 
   echo "$REPO_ROOT/client-go does not exist."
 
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
@kannon92 Can you checkout this PR locally, try generating the client-go files, and let me know if the fix works?

kannon92:
> @kannon92 Can you checkout this PR locally, try generating the client-go files, and let me know if the fix works? It works for me locally when I clone jobset into /tmp folder, delete `/tmp/jobset/client-go` then run `make generate`, the client-go files appear in `/tmp/jobset/client-go`.

Doing this now!

hack/update-codegen.sh
```diff
@@ -21,10 +21,29 @@ set -o pipefail
...
+echo "GOPATH=$GOPATH"
 
 bash "${CODEGEN_PKG}/generate-groups.sh" \
   "all" \
   sigs.k8s.io/jobset/client-go \
   sigs.k8s.io/jobset/api \
   jobset:v1alpha2 \
   --go-header-file ./boilerplate.go.txt
+
+# if client-go files were generated outside of repo root, attempt to move them to the repo root.
```
kannon92, line 35: Testing this, it works but you will have to delete your `client-go` directory in order to get the copy to work.  its a bit odd.

Can we just overwrite if we assume that these changes will never be modified manually?

I deleted everything under `client-go` directory but I reread your script and it wouldn't copy those changes so I had to delete the `client-go` directory also.  

Maybe we just assume anything in client-go can be purged if you are running this?

danielvegamyhre: I added a force flag to the mv command so that it will overwrite if necessary. I think it's fine to assume the generated files will be overwritten/purged.

kannon92: The issue is this code:

```
if [ ! -d "$REPO_ROOT/client-go" ]; then
```
This is saying that if the folder exists then we won't copy it.  

Maybe we just remove this if check?

danielvegamyhre: Oh I see what you mean now - thanks! I removed this check and also added a check before moving the files to only move it if it isn't already in the repo root (since if that is the case, the mv command will fail even with the force flag).

kannon92:
/lgtm

Thanks for fixing this!

PR #225, opened 2023-07-06
Use get_namespaced_custom_object for getting jobset

### LINKED ISSUE #224: Bug in create_jobset.py
Running the script results in the following error:
```
...
list_namespaced_custom_object_with_http_info
    raise ApiTypeError(
kubernetes.client.exceptions.ApiTypeError: Got an unexpected keyword argument 'name' to method list_namespaced_custom_object
```

I think this is because to get a specific jobset by name, we need to use [get_namespaced_custom_object](https://github.com/kubernetes-client/python/blob/master/kubernetes/docs/CustomObjectsApi.md#get_namespaced_custom_object) but are currently using list_namespaced_custom_objects, which doesn't accept a name parameter.

I tested these changes locally and the fix works 👍 

/kind bug

danielvegamyhre:
Fixes #224 

cc @kannon92


kannon92 (commented):
/lgtm

Issue #222, opened 2023-07-05
Generate client-go/sdk only if api changed.


kannon92:
Running unit tests/integration tests will now require regeneration of client-go/sdk every time.  I think we should try to make this a bit smarter as it does make unit or integration tests take a lot longer.  

I see two options for development:

1) Only build client-go/sdk if the api was changed. 
2) Move building of client-go/sdk into a separate target and call it separately from `make` or `make test` or `make test-integration`.  

Currently it is built alongside `make` so it will run on every build.

danielvegamyhre:
+1 on this

kannon92:
@danielvegamyhre which preference would you have for this? 

I can see benefits of logic to detect changes being useful but I also wonder if making the client generation explicit and just detecting that api is different would be less brittle and having a check to see if api mismatches and just telling the user they need to run generate step..

danielvegamyhre:
Yeah I think option 2 would be the most straight forward. We could make the change detection part of the verify step in the CI.

A-Hilaly:
I'd like to give to this one a shot :)
/assign

PR #228, opened 2023-07-16
Remove generate target from build, run and test targets

a-hilaly:
Addresses https://github.com/kubernetes-sigs/jobset/issues/222

This patch removes `generate` target from `build`, `run` and `test` targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -109,7 +109,7 @@ vet: ## Run go vet against code.
 	$(GO_CMD) vet ./...
 
 .PHONY: test
-test: manifests generate fmt vet envtest gotestsum test-python-sdk
+test: manifests fmt vet envtest gotestsum test-python-sdk
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" $(GOTESTSUM) --junitfile $(ARTIFACTS)/junit.xml -- ./pkg/... ./api/... -coverprofile  $(ARTIFACTS)/cover.out
 
 .PHONY: test-python-sdk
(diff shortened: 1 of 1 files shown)


A-Hilaly:
I found that there is already a `generate` target. I'm not super sure if this is what we need here. Do we want to add a script to only call `make generate` if the APIs changed?

/assign @kannon92 @danielvegamyhre

kannon92:
> I found that there is already a `generate` target. I'm not super sure if this is what we need here. Do we want to add a script to only call `make generate` if the APIs changed?
> 
> /assign @kannon92 @danielvegamyhre

Kinda. We want to verify that API changed and generate needs to be called.  I think we have something in verify that does this for our manifests.

kannon92:
https://github.com/kubernetes-sigs/jobset/blob/a5f3ee3f1f8e7e7ae8bb6a5050df13021ba1cb1b/Makefile#L121 Is an example

a-hilaly:
>@A-Hilaly Thanks for doing this. Can you confirm that if you make a change to the API Go code without running make generate afterwards, that the command make verify fails? If so, I think we should be good to merge this.

I can confirm `make verify` fails if we don't run `make generate` after an api change

a-hilaly:
> Kinda. We want to verify that API changed and generate needs to be called. I think we have something in verify that does this for our manifests.

I can add a new rule `auto-generate` that will ony execute the generation rules if verify fails.. what do you think about this?

Or maybe we can just call `verify` early in `run`, `build` and `test` targets?

kannon92:
> > Kinda. We want to verify that API changed and generate needs to be called. I think we have something in verify that does this for our manifests.
> 
> I can add a new rule `auto-generate` that will ony execute the generation rules if verify fails.. what do you think about this?
> 
> Or maybe we can just call `verify` early in `run`, `build` and `test` targets?

I don't think we need to do this.  I think we are okay with requiring someone to run `make generate` if their `verify` step fails in the CI.

kannon92:
/lgtm 

With `make verify` failing if one changes the API I think this solves @danielvegamyhre concern.

danielvegamyhre:
/approve

Thanks for working on this!

> I don't think we need to do this. I think we are okay with requiring someone to run make generate if their verify step fails

Agreed we don't need this.

Issue #240, opened 2023-07-29
Question:  Do you plan to create a pipeline of Jobs?


Gekko0114:
Thank you for the development of this interesting OSS.
I have a question.

I think it would be very reasonable to develop `JobSet` to meet the needs of AI/ML workloads. On the other hand, in ML pipelines, it is often required to run multiple Jobs sequentially, such as running Job A, then Job B, and then Job C after that.
Can `JobSet` cover this functionality of executing Jobs in sequence, or should features like these be handled by another OSS like `Argo`?

kannon92:
We have this issue/feature. https://github.com/kubernetes-sigs/jobset/issues/104

Gekko0114:
I thought this feature was only about supporting which one to start first, the worker or the driver.
Does it also cover running multiple Jobs sequentially, like Job A -> Job B -> Job C -> ....?

danielvegamyhre:
> I thought this feature was only about supporting which one to start first, the worker or the driver. Does it also cover running multiple Jobs sequentially, like Job A -> Job B -> Job C -> ....?

That's correct, the idea described in #104 would only address things like the leader/work pattern, where we wait to start the worker until the leader (Job A) is running before starting the worker (Job B) - that is, not waiting until Job A is **finished** but waiting until Job A is **ready**. 

However, I think the proposed design could easily be extended to include an extra knob to configure if we want to wait until the previous Job is **ready** or until it is **finished** before starting the next job.

Gekko0114:
> However, I think the proposed design could easily be extended to include an extra knob to configure if we want to wait until the previous Job is ready or until it is finished before starting the next job.

I see. Sounds useful. Thank you for your explanation!

kannon92:
> > I thought this feature was only about supporting which one to start first, the worker or the driver. Does it also cover running multiple Jobs sequentially, like Job A -> Job B -> Job C -> ....?
> 
> That's correct, the idea described in #104 would only address things like the leader/work pattern, where we wait to start the worker until the leader (Job A) is running before starting the worker (Job B) - that is, not waiting until Job A is **finished** but waiting until Job A is **ready** before launching Job B.
> 
> However, I think the proposed design could easily be extended to include an extra knob to configure if we want to wait until the previous Job is **ready** or until it is **finished** before starting the next job.
> 
> Building off of @kannon92's comment, maybe something like this:
> 
> ```
> type StartupPolicy struct {
[...]

Issue #104, opened 2023-05-02
Support for startup sequence


ahg-g:
**What would you like to be added**:
A way to indicate if the list of Jobs should start in a specific order. 

**Why is this needed**:
Some jobs require a specific start sequence of pods; sometimes the driver is expected to start first (like Ray or Spark), in other cases the workers are expected to be ready before starting the driver (like MPI).

ahg-g:
@vsoch can you please share your thoughts on startup sequence?

I get that in some cases we want a specific job in the set to start first; however, are there requirements on how to handle failures? or should startup sequence be enforced only when the JobSet is first created.

ahg-g:
I am currently leaning towards a solution based on initContainers or sidecars that implements startup sequence (and in-place pod restarts to recover from failures)

vsoch:
> @vsoch can you please share your thoughts on startup sequence?

Yes of course! For our setup (which you can generalize to a main leader and workers) we would always want the leader to start first, so when the workers come up they register to it without issue. We are getting this to work now with basically hacks in the main startup script to ensure the worker tries again if it doesn't connect. I'm not sure how I'd accomplish this with initContainers or sidecar because every pod is fully part of the cluster (e.g., not  launcher) and I'd stay away from additional sidecar because I think for the kueue design, they didn't like that. But I think I have something interesting to discuss (next section)!

## Startup Sequence for Networking?

However, I'm running into some really interesting behavior tonight - it might not be related to this issue (but does relate to starting up an indexed job) so I want to share. Basically, last week we re-designed the operator to not have a certificate generator pod, which was one we created before the indexed job just to generate a certificate for the cluster that would be saved to a ConfigMap. We were able to do this by building zeromq directly into the operator. The weird behavior is that after that change, I've seen the cluster startup + running time (meaning the time it takes the broker to connect to all the pods and complete LAMMPS) slow down, sometimes alarmingly (today I saw 140 seconds when it used to be 11-12!) We were able to figure out that zeromq (what connects the workers to the broker) has an exponential backoff, and so rebuilding flux and setting this to a lower value (5 seconds) I was able to bring 140s down to 20s. However, what I'm puzzled by (and the million dollar question!) is why the network is no longer ready to go, and before it was. Removing a one-off pod that ran before the indexed job shouldn't have influence, but it does!  I can look at logs of previous runs (before the change) and see that there is no time needed to wait for the network created by the headless service, and the average before was ~12 seconds (not ~19-20). 

[...]

PR #242, opened 2023-08-02
add kep template

kannon92:
@ahg-g and @danielvegamyhre As I am thinking a bit about #104 or #240, I realize that doing a design in a github issue seems quite difficult.  

I think this feature would warrant a KEP so wanted to bring over Kueue's structure for KEPs if there is no objection.


danielvegamyhre:
Good idea! Thanks for adding this

PR #243, opened 2023-08-02
add an example for success policy

kannon92:
In a private conversation with @ahg-g and @danielvegamyhre, we wanted to see if it was possible to have a case where workers finish and then JobSet should terminate the driver.  

This is a simple example to demonstrate using `SuccessPolicy` to achieve this.  I think its a useful example to include in the repo.

### DIFF
--- a/examples/simple/driver-worker-success-policy.yaml
+++ b/examples/simple/driver-worker-success-policy.yaml
@@ -3,7 +3,7 @@ kind: JobSet
 metadata:
   name: success-policy
 spec:
-# We want to declare our JobSet successful if workers finish
+# We want to declare our JobSet successful if workers finish.
 # If workers finish we should clean up the remaining replicatedJobs.
   successPolicy:
     operator: All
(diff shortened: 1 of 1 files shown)


examples/simple/driver-worker-success-policy.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: success-policy
+spec:
+# We want to declare our JobSet successful if workers finish
+# If workers finish we should clean up the other jobsets.
```
danielvegamyhre, line 7:
```suggestion
# If workers finish we should clean up the other ReplicatedJobs.
```

Or am I misunderstanding?

kannon92: You are correct. Updated.

examples/simple/driver-worker-success-policy.yaml
```diff
@@ -0,0 +1,47 @@
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: success-policy
+spec:
+# We want to declare our JobSet successful if workers finish
```
danielvegamyhre, line 6:
```suggestion
# We want to declare our JobSet successful if workers finish.
```

Nit: punctuation.

PR #244, opened 2023-08-03
A KEP for StartupPolicy

kannon92:
I wanted to draft a potential API for StartupPolicy as we have discussed it.

### DIFF
--- a/keps/104-StartupPolicy/README.md
+++ b/keps/104-StartupPolicy/README.md
@@ -1,0 +1,428 @@
+# KEP-104: StartupPolicy
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories (Optional)](#user-stories-optional)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Proposal](#api-proposal)
+  - [Implementation](#implementation)
+  - [Defaulting/Validation](#defaultingvalidation)
+  - [User Experience](#user-experience)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+  - [Poor Person Workflow Engine](#poor-person-workflow-engine)
+  - [Increase number of patches to JobSet](#increase-number-of-patches-to-jobset)
+- [Alternatives](#alternatives)
+  - [Operate on Conditions](#operate-on-conditions)
+  - [Allow for both ready and succeeded](#allow-for-both-ready-and-succeeded)
+<!-- /toc -->
+
+## Summary
+
+This KEP adds a StartupPolicy for JobSets.  A StartupPolicy allows for some flexibility on when to start ReplicatedJobs in a JobSet.  
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+High Performance Computing / Machine Learning users usually want some control over how jobs are being created.  
+A Startup policy would allow a user to specify which status a replicated job should have before starting other replicated jobs in a JobSet.
+
+### Goals
+
+- Add a startup policy for JobSet
+
+### Non-Goals
+
+- Comprehensive workflow specifications
+  - This means that we will support simple dependecies rather than processing a full DAG.
+  - Projects like argo-workflows are a better use case for this.
+- Allowing for a percentage of jobs in a ReplicatedJob to be ready to consider the whole ReplicatedJob to ready
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a user, I have one ReplicatedJob that functions as the driver and a series of worker ReplicatedJobs.  
+I want the driver to be ready before I start the worker replicated jobs.  
+This is useful for HPC/AI/ML workloads as some frameworks have a driver pod that must be ready before the workers are started.  
+
+The API below would fit my purpose.
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: driver-ready-worker-start
+spec:
+  startupPolicy:
+     startupPolicyOrder: InOrder
+  replicatedJobs:
+  - name: driver
+    replicas: 1
+    template:
+      spec:
+        # Set backoff limit to 0 so job will immediately fail if any pod fails.
+        backoffLimit: 0 
+        completions: 1
+        parallelism: 1
+        template:
+          spec:
+            containers:
+            - name: driver
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10000
+  - name: workers
+    replicas: 1
+    template:
+      spec:
+        backoffLimit: 0 
+        completions: 2
+        parallelism: 2
+        template:
+          spec:
+            containers:
+            - name: worker
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 2
+
+As a user, I want my replicated jobs to start in order.
+
+One area that this could be useful is if someone has a message queue replicatedJob (e.g., Redis), followed by a driver replicatedJob, followed by a worker replicated job.
+
+In this case, I want message queue to start first, then the driver, and then finally the worker.  
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: messagequeue-driver-worker
+spec:
+  startupPolicy:
+     startupPolicyOrder: InOrder
+  replicatedJobs:
+  - name: messagequeue
+    replicas: 1
+    template:
+      spec:
+        # Set backoff limit to 0 so job will immediately fail if any pod fails.
+        backoffLimit: 0 
+        completions: 1
+        parallelism: 1
+        template:
+          spec:
+            containers:
+            - name: messagequeue
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 100
+  - name: driver
+    replicas: 2
+    template:
+      spec:
+        backoffLimit: 0 
+        completions: 2
+        parallelism: 2
+        template:
+          spec:
+            containers:
+            - name: driver
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+  - name: worker
+    replicas: 2
+    template:
+      spec:
+        backoffLimit: 0 
+        completions: 2
+        parallelism: 2
+        template:
+          spec:
+            containers:
+            - name: worker
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+### Risks and Mitigations
+
+To avoid the JobSet become a half-baked workflow manager, we will not allow for starting particular replicated jobs.
+
+## Design Details
+
+### API Proposal
+  
+```golang
+
+type StartupPolicyOptions string
+
+const (
+ // This is the default setting
+ // AnyOrder means that we will start jobs without any specific order.
+ AnyOrder StartupPolicyOptions = "AnyOrder"
+ // InOrder starts the ReplicatedJobs in order that they are listed. Jobs within a ReplicatedJob will 
+ // still start in any order.
+ InOrder StartupPolicyOptions = "InOrder"
+)
+
+type StartupPolicy struct {
+ // StartupPolicyOrder is an enum of different options for StartupPolicy
+ // AnyOrder means to start replicated jobs in any order.
+ // InOrder means to start them as they are listed in the JobSet.
+ // +kubebuilder:validation:Enum=AnyOrder;InOrder
+ StartupPolicyOrder StartupPolicyOptions `json:"startupPolicyOrder"`
+}
+```
+
+### Implementation
+
+We have three cases where startupPolicy applies:
+
+- when the jobset is first created
+- when the jobset is restarted after a failure (with restart or recreate all failure policy)
+- when the jobset is resumed after suspension
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+On initial creation or recreation, we will apply the startup policy before creating the jobs in a replicated job.
+We will retrieve the `ReplicatedJobStatus` before each call to `createJobs` to verify which jobs have been created.
+
+JobSet views the jobs that it owns and updates the `ReplicatedJobStatus` on each reconcile. For suspended jobs, we will view the `ReplicatedJobStatus` and resume the jobs as they are listed in the `startupPolicyOrder`.
+Suspended is also tracked by `ReplicatedJobStatus`.
+
+One area of contention is if Jobs finish fast and are deleted (ie `ttlAfterFinished` is short). 
+We created [issue](https://github.com/kubernetes-sigs/jobset/issues/360) to address this.
+
+### Defaulting/Validation
+
+- StartupPolicy is immutable.
+- StartupPolicyOrder of `AnyOrder` is default setting.
+
+### User Experience
+
+If Jobs are taking a long time to go to ready, we should have a clear way to check that the JobSet is behaving as expected.
+
+We will add a condition for StartupPolicy to achieve this.
+
+```golang
+metav1.Condition{
+   Type:    "JobSetStartupPolicyComplete",
+   Status:  metav1.ConditionStatus(corev1.ConditionFalse),
+   Reason:  "StartupPolicyInOrder",
+   Message: "replicated job %s is starting",
+  })
+```
+
+A condition of false means that startup policy is running.
+
+A condition of true means that all jobs at least ready. StartupPolicy does not care about succeeded or failed.
+
+Reason will display the type of startupPolicy (`StartupPolicyInOrder`) for now.
+Message will display the replicated job that is starting.
+Once the StartupPolicy is complete, this will read `startup policy successful`.
+
+If `StartupPolicyAnyOrder` (same as no startup policy), then this condition will not be set in the JobSet.
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+### Test Plan
+
+<!--
+**Note:** *Not required until targeted at a release.*
+The goal is to ensure that we don't accept enhancements with inadequate testing.
+
+All code is expected to have adequate tests (eventually with coverage
+expectations). Please adhere to the [Kubernetes testing guidelines][testing-guidelines]
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
+
+<!--
+Based on reviewers feedback describe what additional tests need to be added prior
+implementing this enhancement to ensure the enhancements have also solid foundations.
+-->
+
+#### Unit Tests
+
+<!--
+In principle every added code should have complete unit test coverage, so providing
+the exact set of tests will not bring additional value.
+However, if complete unit test coverage is not possible, explain the reason of it
+together with explanation why this is acceptable.
+-->
+
+<!--
+Additionally, try to enumerate the core package you will be touching
+to implement this enhancement and provide the current unit coverage for those
+in the form of:
+- <package>: <date> - <current test coverage>
+
+This can inform certain test coverage improvements that we want to do before
+extending the production code to implement this enhancement.
+-->
+
+- `controller`: `Aug 7th 2023` - `31.6%`
+
+Unit tests are low in this area as we do more of our testing at the integration level.
+
+We will add a new file called startup_policy to apply our actions.
+
+#### Integration tests
+
+- If Suspend is specified, startupPolicy resume is ignored.
+- A JobSet is suspended while a ReplicatedJob is still starting
+- [Story 1](#story-1)
+- [Story 2](#story-2)
+
+<!--
+Describe what tests will be added to ensure proper quality of the enhancement.
+
+After the implementation PR is merged, add the names of the tests here.
+-->
+
+### Graduation Criteria
+
+<!--
+
+Clearly define what it means for the feature to be implemented and
+considered stable.
+
+If the feature you are introducing has high complexity, consider adding graduation
+milestones with these graduation criteria:
+- [Maturity levels (`alpha`, `beta`, `stable`)][maturity-levels]
+- [Feature gate][feature gate] lifecycle
+- [Deprecation policy][deprecation-policy]
+
+[feature gate]: https://git.k8s.io/community/contributors/devel/sig-architecture/feature-gates.md
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+-->
+
+## Implementation History
+
+- Drafted August 2 2023
+- Updated API December 18th 2023
+<!--
+Major milestones in the lifecycle of a KEP should be tracked in this section.
+Major milestones might include:
+- the `Summary` and `Motivation` sections being merged, signaling SIG acceptance
+- the `Proposal` section being merged, signaling agreement on a proposed design
+- the date implementation started
+- the first Kubernetes release where an initial version of the KEP was available
+- the version of Kubernetes where the KEP graduated to general availability
+- when the KEP was retired or superseded
+-->
+
+## Drawbacks
+
+### Poor Person Workflow Engine
+
+This startup policy can function as a poor person workflow syntax.  And there may be requests to make this more than what it is.  
+
+We ideally want to utilize other workflows engines to provide more complicated dags but I could see simple ML workflows being used for JobSet.  
+
+### Increase number of patches to JobSet
+
+Since we will start all jobs in a suspended state, we will have to patch the spec of all the jobs eventually.  
+
+## Alternatives
+
+### Operate on Conditions
+
+We originally wanted to operate based on Conditions for Jobs.  But it turns out that there is no Ready condition.  
+This means that for some cases we would support statuses and others for conditions.  
+We decided to go forward with operating on status fields for both cases.
+
+### Allow for both ready and succeeded
+
+To avoid JobSets becoming a bad workflow engine.  We want to support startup sequence where once Jobs are ready, then go onto the next job.  
+
+We did not add support for succeeded at this time as this is more in align with a pipeline or workflow.
+
+<!--
+What other approaches did you consider, and why did you rule them out? These do
+not need to be as detailed as the proposal, but should include enough
+information to express the idea and why it was not acceptable.
+-->
--- a/keps/104-StartupPolicy/README.md
+++ b/keps/104-StartupPolicy/README.md
@@ -214,7 +214,8 @@ spec:
 
 ### Risks and Mitigations
 
-To avoid the JobSet become a half-baked workflow manager, we will not allow for starting particular replicated jobs.
+To avoid the JobSet become a half-baked workflow manager, the API will not allow describing DAGs. 
+The goal is to focus on supporting startup sequence that HPC and ML training jobs typically require.
 
 ## Design Details
 
@@ -276,7 +277,7 @@ We will add a condition for StartupPolicy to achieve this.
 
 ```golang
 metav1.Condition{
-   Type:    "JobSetStartupPolicyComplete",
+   Type:    "JobSetStartupPolicyCompleted",
    Status:  metav1.ConditionStatus(corev1.ConditionFalse),
    Reason:  "StartupPolicyInOrder",
    Message: "replicated job %s is starting",
--- a/keps/104-StartupPolicy/kep.yaml
+++ b/keps/104-StartupPolicy/kep.yaml
@@ -1,0 +1,28 @@
+title: Startup Policy
+kep-number: 104
+authors:
+  - "@kannon92"
+status: provisional
+creation-date: 2023-08-01
+reviewers:
+  - "@danielvegamyhre"
+  - "@vsoch"
+approvers:
+  - "@ahg"
+
+see-also:
+  - "NA"
+replaces:
+  - "NA"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.3.0"
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.3.0"


keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,437 @@
...
+
+As a parallel framework user, I have one ReplicatedJob that functions as the driver and a series of worker ReplicatedJobs.  I want the driver to be ready before I start the worker replicated jobs.
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: driver-ready-worker-start
+spec:
+  startupPolicy:
```
danielvegamyhre, line 127: Hmm it feels awkward to have a startupPolicy at the jobSet level that defines a target replicated job to wait for a certain status, before starting x/y/z replicated job(s)... the semantics are unclear, does it apply to all replicatedJobs which are not the target of the policy? It also doesn't seem to have a way to have certain replicated jobs wait on a certain replicated job, but not others. This flexibility might be nice for future use cases. 

What do you think?

kannon92: I was going to push thoughts on this.  

If StartupPolicy is specified, I would suspend all jobs.  And then unsuspend the ones that match the target.  And keep doing that as they are listed.

kannon92: Part of the issue with partial application is that I think the API gets pretty complicated as we operator on ReplicatedJobs rather than Jobs.  So we can have partially completiion of the replicatedJobs with this API.  

I think it would be nice to have a partial readiness at the job level but I tried to bring that up to #wg-batch but I think there was some light pushback as we could utilize the `ready` condition at the Job level.  

I notice that `ReplicatedJobStatus` gives you the overal status of the ReplicatedJob though.  I wasn't sure if this is a big enough change to fix that or not.  

I figured we can tackle a lot of the usecases specified in Stories without having to change ReplicatedJobStatus too much.

vsoch:
This would definitely be useful for us - what I've had to do is typically start the workers and handle race conditions with sleeps / waiting, that sort of thing.

kannon92:
> This would definitely be useful for us - what I've had to do is typically start the workers and handle race conditions with sleeps / waiting, that sort of thing.

And an API like:

```
 startupPolicy:
  - name: driver
    status: Ready
```
Would be sufficent?

Gekko0114:
Thanks @kannon92 so much!

IMO, I prefer defining startUpPolicy under each replicatedJob field.
I think this implementation is easier to read.

```
kind: JobSet
metadata:
  name: driver-ready-worker-start
spec:
  replicatedJobs:
  - name: driver
    replicas: 1
    template:
      spec:
        backoffLimit: 0 
        completions: 1
        parallelism: 1
        template:
          spec:
            containers:
            - name: driver
              image: bash:latest
              command:
              - bash
              - -xc
              - |
                sleep 10000
  - name: workers
    replicas: 1
    startupPolicy:
    - name: driver
       status: Ready
    template:
      spec:
        backoffLimit: 0 
        completions: 2
        parallelism: 2
        template:
          spec:
            containers:
            - name: worker
              image: bash:latest
              command:
              - bash
              - -xc
              - |
                sleep 10
```

FYI
I prefer Tekton's design, especially `runAfter`. 
https://tekton.dev/docs/pipelines/pipelines/
```
apiVersion: tekton.dev/v1 # or tekton.dev/v1beta1
kind: Pipeline
metadata:
  name: pipeline
spec:
  workspaces:
    - name: source
  tasks:
    - name: gen-code
      taskRef:
        name: gen-code # gen-code expects a Workspace named "source"
      workspaces:
        - name: source # <- mapping workspace name
    - name: commit
      taskRef:
        name: commit # commit expects a Workspace named "source"
      workspaces:
        - name: source # <- mapping workspace name
      runAfter:
        - gen-code
```

vsoch:
> IMO, I prefer defining startUpPolicy under each replicatedJob field. I think this implementation is easier to read.

I think he's mirroring `SuccessPolicy`-> `TargetReplicatedJobs` - the idea being that if there is more definition / attributes for such a policy, we have a place to put them and the ReplicatedJob itself doesn't become a single long list of loosely associated features.

https://github.com/kubernetes-sigs/jobset/blob/61ad4b08af84f20ce0e7d641d396b460eb22996c/api/jobset/v1alpha2/jobset_types.go#L177

I think my preference is for that modular (and consistent, since we already have SuccessPolicy and FailurePolicy) approach, ensuring that each of those sections can be produced independently and then added to a replicatedJob.

kannon92:
> > IMO, I prefer defining startUpPolicy under each replicatedJob field. I think this implementation is easier to read.
> 
> I think he's mirroring `SuccessPolicy`-> `TargetReplicatedJobs` - the idea being that if there is more definition / attributes for such a policy, we have a place to put them and the ReplicatedJob itself doesn't become a single long list of loosely associated features.
> 
Correct.

> https://github.com/kubernetes-sigs/jobset/blob/61ad4b08af84f20ce0e7d641d396b460eb22996c/api/jobset/v1alpha2/jobset_types.go#L177
> 
> I think my preference is for that modular (and consistent, since we already have SuccessPolicy and FailurePolicy) approach, ensuring that each of those sections can be produced independently and then added to a replicatedJob.
> 
> And if we wanted to be more consistent it might look like:
> 
> ```go
> type StartupPolicy struct {
> 	TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
> 	Status JobReadyStatus `json:"status"`
> }
> ```
> 
> This makes the assumption that a JobSet will have one condition (set of jobs and a status) to indicate ready. I'm not sure we want to allow more complexity than that (the workflow complexity thing is a non goal).

Yea I believe there was an interest in both `ready` and `succeeded`.  Init containers run after one another and the thought was that we could have a similar idea for this.  

I ended up going with the API I have because of the flexibility of assuming ready or succeeded for the replicated jobs.  

I do acknowledge the irony of calling out that we don't want a workflow engine but then we at least have a poor person workflow engine with this.

Gekko0114:
I commented without understanding the context of jobset, thank you for your explanations

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,551 @@
...
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 2
+
+As an user, I have one ReplicatedJob that I want to finish before going on to the next worker.
```
danielvegamyhre, line 150:
```suggestion
As a user, I have one ReplicatedJob that I want to finish before going on to the next worker.
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,551 @@
...
+}
+```
+
+### Implementation
+
+If StartupPolicy is set and suspend field is off, we will start all jobs in a suspended state.  We will not set a condition for JobSet as suspended.
+
+To track which ReplicatedJobs are suspended, we will add a `Suspended` field to the ReplicatedJobStatus.  This field will track the number of replicas that are suspended.
+
+When a JobSet is started with `StartupPolicy` specified, we will create jobs in a suspended state (ie `Job.Spec.Suspend = true`)  This avoids starting the underlying jobs.
```
danielvegamyhre, line 409: Correct me if I am wrong but I think this saying the same thing as the first sentence in the implementation section?

keps/104-StartupPolicy/README.md
kannon92: TIL that this is the one exception to An with vowels!

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,551 @@
...
+
+If StartupPolicy is set, then we will suspend the other jobs until the StartupPolicy is considered succesful.
+We will unsuspend only the jobs that match `TargetReplicatedJobStartup`.  Once they are considered "started" by matching on targeted status, then we resume all the rest of the jobs.
```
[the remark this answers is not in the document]
kannon92: > Or is this saying that we handle only 1 startup policy at a time, in the order in which they are specified in the spec (e.g. we first see a startup policy for Job A so we unsuspend it, wait for Job A to complete, then we check the next startup policy and it's for Job B, so we unsuspend that, etc.)?

This is where I find it difficult balancing design versus implementation details.  

In the implementation, I apply StartupPolicy in sequential order.  

I would verify if JobA is suspended and then unsuspend.  I would return and trigger a reconcile.  

I loop over the startupPolicy rules and check if JobA is successful via the rules (Ready/Succeeded) and if it is, I move on to JobB and resume if its suspended.  

And repeat.  

If all the startup polices items are successful then we unsuspend everything not in the startup list (because the startup ones are running or succeeded).

keps/104-StartupPolicy/README.md
kannon92: you are correct. updated

keps/104-StartupPolicy/README.md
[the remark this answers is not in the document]
kannon92: updated with this info. PTAL

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a user, I have one ReplicatedJob that functions as the driver and a series of worker ReplicatedJobs.  
+I want the driver to be ready before I start the worker replicated jobs.  
+This is useful for HPC/AI/ML workloads as someone frameworks have a driver pod that must be ready before the workers are started.
```
kannon92, line 97:
```suggestion
This is useful for HPC/AI/ML workloads as some frameworks have a driver pod that must be ready before the workers are started.  
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+An API that fits my purpose is below.
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: driver-ready-worker-start
+spec:
+  startupPolicy:
+  - name: first
```
kannon92, line 163:
```suggestion
  - name: driver
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+          spec:
+            containers:
+            - name: driver
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 100
+  - name: second
```
kannon92, line 184:
```suggestion
  - name: worker
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+An API that fits my purpose is below.  
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: driver-ready-worker-start
+spec:
+  startupPolicy:
+  - name: first
```
kannon92, line 216:
```suggestion
  - name: driver
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+          spec:
+            containers:
+            - name: driver
+              image: bash:latest
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 100
+  - name: second
```
kannon92, line 238:
```suggestion
  - name: worker
```

kannon92:
@ahg-g and @danielvegamyhre please take a look when you can.

kannon92:
friendly ping @ahg-g and @danielvegamyhre

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+An API that fits my purpose is below.
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: driver-ready-worker-start
+spec:
+  startupPolicy:
+  - name: driver
```
danielvegamyhre, line 163: I think the name of this should be `first` to match the replicatedJob below

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 3
+
+As a user I only need a percentage of my targeted replicated job to succeed before going on to the next job.
```
danielvegamyhre, line 205: Interesting idea, what is the use case for this though? What applications/frameworks require a percentage threshold like this?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 2
+
+As a user, I have one ReplicatedJob that I want to finish before going on to the next worker.
```
ahg-g, line 150: I think we need to be careful about turning jobset into a workflow manager. Workflow managers are much more than simple sequencing, and I am concerned that we will be offering a half-baked solution with this approach.

keps/104-StartupPolicy/README.md
ahg-g: +1, what is the use case?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 4
+
+As a user I want to run the jobs in listed order.  Job A -> Job B -> Job C
```
ahg-g, line 259: Similar to the point above about workflows

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+type TargetReplicatedJobStartup struct {
+ // Name of ReplicatedJob to apply startup policy
+ Name string `json:"name"`
+
+ // The status you want the ReplicatedJob to be
+ // +kubebuilder:validation:Enum=Ready;Succeeded
+ Status Status `json:"status"`
+
+ // Number of Statuses is how many ReplicatedJob statuses to consider "started"
+ // Empty will mean all status must match the replicas.
```
ahg-g, line 384: The name is confusing, it is not clear to me what "number of statuses" means? statuses isn't something we usually count

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,549 @@
...
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a user, I have one ReplicatedJob that functions as the driver and a series of worker ReplicatedJobs.  
+I want the driver to be ready before I start the worker replicated jobs.  
+This is useful for HPC/AI/ML workloads as some frameworks have a driver pod that must be ready before the workers are started.
```
ahg-g, line 97: what happens to the second job if the first job gets recreated after all of them became ready?

ahg-g (commented):
I left few comments, I am concerned that we are turning jobset into a half-baked workflow manager.

keps/104-StartupPolicy/README.md
kannon92: Agreed.  I already had some comments in this KEP/issue about this.  I was thinking that this could be fine as we only support sequential and no dag syntax.  

Similar to init containers (start in order).  It gets complicated because if you specify 1 or 2 jobs, what do you do with the others?

Not sure hence why I thought a KEP would be useful at first.

keps/104-StartupPolicy/README.md
kannon92: Good point.  I guess with this API I am not sure.  If the driver gets killed or preempted, I think the downstream objects should be stopped.

vsoch:
> I left few comments, I am concerned that we are turning jobset into a half-baked workflow manager.

I share this concern. I think the workflow manager should be a separate component to manage that logic - e.g., most workflow tools create a DAG, and I think jobset would fit into this idea as a specific step/job in the DAG. The main logic we'd need in that case is if the step (jobset) has some kind of service (database, monitoring tool, etc.) that needs to get some amount of setup done first for the application to use. I've hit this several times, examples include:

- the launcher in a launcher / workers design needs to setup / be ready before workers
- a monitoring tool, database, or general service that the main application in the jobset needs also should be ready first.

> I think we need to be careful about turning jobset into a workflow manager. Workflow managers are much more than simple sequencing, and I am concerned that we will be offering a half-baked solution with this approach.

Also I've seen "workflow" and "workload" so I need to clarify the two. The workload tool is akin to a cluster manager like Flux or Slurm - you give it a queue of work, ideally with an algorithm to determine who gets to run, and it assigns work to resources accordingly. The workflow manager is what takes a specific application / spec for a workflow and turns it into a DAG. In practice what we see is that workflow tools (e.g., Snakemake, Nextflow) will start with a specification file, turn that into steps for a DAG, and then submit steps in a logical way to the workload manager. Let me know if that makes sense - we should be careful about using them interchangeably because they mean different things. In this case, in that jobset is trying to take order logic, we are turning it into more of a workflow manager.

To step back - we do need some policy within a replicated job that says "make sure my service has some status before we start the other jobs" (e.g., "make sure my launcher is ready to receive workers before they try to register." and what we need for that is not logic about a return code (because it would still be running), but some kind of signal from the job that it is ready. I'm not sure how that would work.

> what happens to the second job if the first job gets recreated after all of them became ready?

This to me would need to be handled by the application. If the first job is somehow killed (and it should not be) that is a problem with the application. A successPolicy would be useful here so if/when that happens the entire job completes.

keps/104-StartupPolicy/README.md
kannon92: I'm not sure anymore.  

I was thinking of https://github.com/kubernetes/kubernetes/issues/117758 and [Success/Failure Policy](https://github.com/kubernetes/kubernetes/issues/117600) in the job controller.  

In this case, you could have certain jobs be considered successful or failed based on percentage of indexes that passed/failed.  I thought maybe it could be a good idea to expand that to JobSet where you have certain percentage of replicatedJobs match a targerted status.  

I was thinking that if you had a bunch of replicas of the same ReplicatedJob, then wouldn't you also want to make sure that they all succeed.  But then it could be possible for some of them to fail and others to pass.  

I had a little too much on my last days in my last company so I was feeling inspired.  

Maybe this is YAGNI and we move this to a non goal?

keps/104-StartupPolicy/README.md
kannon92: updated.

keps/104-StartupPolicy/README.md
vsoch: +1 YAGNI.

kannon92:
> > I left few comments, I am concerned that we are turning jobset into a half-baked workflow manager.
> 
> I share this concern. I think the workflow manager should be a separate component to manage that logic - e.g., most workflow tools create a DAG, and I think jobset would fit into this idea as a specific step/job in the DAG. The main logic we'd need in that case is if the step (jobset) has some kind of service (database, monitoring tool, etc.) that needs to get some amount of setup done first for the application to use. I've hit this several times, examples include:
> 
> * the launcher in a launcher / workers design needs to setup / be ready before workers
> * a monitoring tool, database, or general service that the main application in the jobset needs also should be ready first.

+1. However, a Kubernetes Job is a perfect example for how much can actually be in a single step.  A Job has pods which can have multiple init containers, sidecars containers and main containers.  There is some order inherent in that but its not a workflow.  I was thinking that this could be a similar idea for a JobSet.  Its a single step of a workflow (where you are running a distributed ML job) on k8s.  
 

> To step back - we do need some policy within a replicated job that says "make sure my service has some status before we start the other jobs" (e.g., "make sure my launcher is ready to receive workers before they try to register." and what we need for that is not logic about a return code (because it would still be running), but some kind of signal from the job that it is ready. I'm not sure how that would work.

I think this API would fit that.  Hence why I allow for a field for status (ready or succeeded).  I tried to bring up the idea of operating on conditions to wg-batch but there wasn't enough push to do this.  So I figured we can read the status field from the job.  The Job controller has a concept of the job being ready but there isn't a percentage of the pods to be ready to consider the entire job ready.  

Generally, I thought of your use cases for MPI (driver ready and then workers are good to go).  But as you are bringing up databases, I think the API can't really be that simple as I can see other cases where other services should be ready before running other stuff.  

What happens if you want a redis pod, followed by a driver and then you can run the workers.  I think you'd end up following the API I have as you may want to specify a series of services that you want to run before you run the other jobs..  

All that being said, I find myself in a pickle.  I don't want to create a dag-like syntax with JobSet so I was thinking simpler syntax would suffice.   
> > what happens to the second job if the first job gets recreated after all of them became ready?
> 
> This to me would need to be handled by the application. If the first job is somehow killed (and it should not be) that is a problem with the application. A successPolicy would be useful here so if/when that happens the entire job completes.

I think this is a big design decision on K8s.  Applications don't really know if the infrastructure has gone away or scaled up.  Say for example, you are running a job with 10 nodes and your cloud provider takes away all but 1 of your nodes. Ideally you would want your k8s environment to downscale and still let you run your workloads.  Your application layer shouldn't worry about the cloud infrastructure.

keps/104-StartupPolicy/README.md
kannon92: Yea I originally thought we could consider the Job to report the ready status (and use a percentage of Jobs) to decide when a job was ready.  That wasn't accepted in upstream so I decided to try out this idea.

vsoch:
> +1. However, a Kubernetes Job is a perfect example for how much can actually be in a single step. A Job has pods which can have multiple init containers, sidecars containers and main containers. There is some order inherent in that but its not a workflow. I was thinking that this could be a similar idea for a JobSet. Its a single step of a workflow (where you are running a distributed ML job) on k8s.

This is actually something I posted in the slack (unrelated here so apologies) but I tested an init container with an emptyDir vs a sidecar, and the initContainer did not work to persist any data to the main application. I've seen this work outside of Job so I'm wondering if it's a bug.

But I totally agree - within a single step we have need for more than 'run this python script on this nvidia cuda base and call it a day."  Services, monitoring, the whole shabang, are becoming common, even in HPC. That's one of the reasons K8s is so attractive is because we can provide that!

> I think this API would fit that. Hence why I allow for a field for status (ready or succeeded). I tried to bring up the idea of operating on conditions to wg-batch but there wasn't enough push to do this. 

I think I can back you that this is needed. Ready implies it's running, and Succeeded implies it has a successful exit code (if I parsed those correctly). What I've been doing in practice for two containers in the same pod is to have a shared volume with very basic files as indicators. E.g., container A comes up and sets up Flux for container B. When it's ready, it touches a file that container B is waiting for. Then container B proceeds. The way I'd accomplish similar with JobSet is via a service on the same network, albeit polling a service endpoint for a status is hacky at best. But I don't see a better way to do it so likely I'll be trying that soon...

> Generally, I thought of your use cases for MPI (driver ready and then workers are good to go). But as you are bringing up databases, I think the API can't really be that simple as I can see other cases where other services should be ready before running other stuff.

For quick other examples, I've implemented slurm / htcondor / hyperqueue using JobSet and they have various daemons, executors, databases, and things that smell like nodes. For the most part, the only power we have to control order of being ready is with sleeps, waits, and then the application having an expectation of needing to retry (again an example of it being under the jurisdiction of the specific app).

> What happens if you want a redis pod, followed by a driver and then you can run the workers. I think you'd end up following the API I have as you may want to specify a series of services that you want to run before you run the other jobs..

Right - if I were doing this I'd have a few lines in a custom entrypoint for the driver to wait for redis, and then for the workers to wait for the driver. It's doable if you are a developer (e.g., for the metrics operator that uses JobSet I finely tune / control all the entrypoint logic) but for a user that expects to throw a command into a yaml file it's probably hard and annoying.

> All that being said, I find myself in a pickle. I don't want to create a dag-like syntax with JobSet so I was thinking simpler syntax would suffice.

I am still on the same page that we need a simple _something_ to represent this flat logic for startup. The hard part isn't the startup order, but deciding when to start each in context of the rest.

> I think this is a big design decision on K8s. Applications don't really know if the infrastructure has gone away or scaled up. Say for example, you are running a job with 10 nodes and your cloud provider takes away all but 1 of your nodes. 

What a jerk! Just kidding :) I think this scenario would hugely depend on the application / DAG in question, and whether it allows for elasticity or dynamism. For any unexpected resource lost, at least for an HPC-type application that needs to know very fine tuned resources (and then not lose them, a good example is an MPI job) if we lose a node it's game over because they all are communicating and the size of the work is static. But if you have a workload manager that is handling submission and it can see that a job failed, it could wait for resources and request another node. Or if that same job allows for elasticity (e.g., scaling in and out and having the workers adjust accordingly) then it might be OK depending on the loss. So I don't think this is a K8s design decision, but rather up to the flexibility that the application allows. The application uses K8s optimally for what it needs, and if the infrastructure goes away it either needs to bail or or find a way to adjust.

Thinking out loud - even if we had an ability to make decisions based on Conditions, that would still mean that a container with an application inside of a pod needs to bubble up some signal to the jobset operator to say "Hello, I'm ready for you to start the next replicated job!" Is that something that people do / is feasible without having to give the pod permission to interact with the API (which I assume is not good practice)? I've done a small experiment where the application directly talks to an autoscaler, but I had to give it a particular role and it wasn't great.

danielvegamyhre:
> I left few comments, I am concerned that we are turning jobset into a half-baked workflow manager.

I agree, I think JobSet should stay focused on the goal of providing a unified API for running a single distributed ML/HPC workload, and striving to make it a best in class solution for this, rather than expanding to add support for additional use cases.

In my opinion we should simplify the startup policy to strictly support sequential **startup** (start replicatedJob B after replicatedJob A is ready), which is needed to better support leader/worker paradigms (for example MPI where workers must be available before the leader starts SSH-ing in and executing things on them). 

I don't think we need sequential **execution** (so replicatedJob B waits for replicatedJob A to completely finish before starting) or anything else that starts straying into workflow execution engine territory.

From an engineering perspective, I appreciate the more flexible nature of the currently proposed API allowing for these other use cases, however, from a product perspective I think we should keep JobSet focused on modeling a single distributed ML/HPC workload.

What do you all think? I'm hoping to include this feature in v0.3.0 release by the way

vsoch:
I like that, and the one thing that isn't clear to me is:

> In my opinion we should simplify the startup policy to strictly support sequential startup (start replicatedJob B after replicatedJob A is ready), which is needed to better support leader/worker paradigms (for example MPI where workers must be available before the leader starts SSH-ing in and executing things on them).

What is the signal that is used between replicated jobs to indicate when a subsequent one should start, given that we aren't doing sequential execution?

danielvegamyhre:
> What is the signal that is used between replicated jobs to indicate when a subsequent one should start, given that we aren't doing sequential execution?

We can use the ReplicatedJobStatus to check that all child jobs are ready, before starting the subsequent ReplicatedJob.

vsoch:
> We can use the ReplicatedJobStatus to check that all child jobs are ready, before starting the subsequent ReplicatedJob.

Gotcha - and to step up one level, what determines the `ReplicatedJobStatus`?

kannon92:
I am happy to go with this.  I will drop the more flexible API from this KEP.  

How do we want to specify multiple replicated jobs should be ready?

The Leader-Worker pattern is pretty simple but @vsoch has mentioned that I think there is still a need for multiple jobs should be ready before others start.

Maybe someone wants a database ready and then a leader and then the workers.   

One way we could do this is to follow the initContainer route and assume that ReplicatedJobs must be ready in order.

```startupPolicy:
      startUpInOrder: true
```

This would start ReplicatedJobs in order as they are listed in the JobSet.

danielvegamyhre:
> I am happy to go with this. I will drop the more flexible API from this KEP.
> 
> How do we want to specify multiple replicated jobs should be ready?
> 
> The Leader-Worker pattern is pretty simple but @vsoch has mentioned that I think there is still a need for multiple jobs should be ready before others start.
> 
> Maybe someone wants a database ready and then a leader and then the workers.
> 
> One way we could do this is to follow the initContainer route and assume that ReplicatedJobs must be ready in order.
> 
> ```
> startupPolicy:
>       startUpInOrder: true
> ```
> 
> This would start ReplicatedJobs in order as they are listed in the JobSet.

We could even just have a simple boolean field in the jobset spec: `sequentialStartup: true`

This does feel a bit brittle/inflexible but I think it would address the use cases we talk about, and remove any temptation for users from using JobSet as a workflow manager.

vsoch:
> This does feel a bit brittle/inflexible but I think it would address the use cases we talk about, and remove any temptation for users from using JobSet as a workflow manager.

The question for me is still what determines readiness. Just having the container started isn't enough I don't think.

kannon92:
> > This does feel a bit brittle/inflexible but I think it would address the use cases we talk about, and remove any temptation for users from using JobSet as a workflow manager.
> 
> The question for me is still what determines readiness. Just having the container started isn't enough I don't think.

So Job has a field called Ready which is actually the pods reporting whether or not they are ready.  

This typically means that someone has specified ReadinessGates and a Pod is considered ready if those gates have passed.  If you don't specify that, then the pod is ready when the container is started.  

Job has a count of the number of pods that are considered ready: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/2879-ready-pods-job-status

So for users that want more control over this, they can specify [ReadinessGates in the pods](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate)

danielvegamyhre:
> > This does feel a bit brittle/inflexible but I think it would address the use cases we talk about, and remove any temptation for users from using JobSet as a workflow manager.
> 
> The question for me is still what determines readiness. Just having the container started isn't enough I don't think.

I think we can just check the all the expected replicas are ready for the target replicated jobs (i.e., check the number of Ready child jobs in `ReplicatedJobStatus.Ready` == `jobSet.spec.replicatedJobs[x].replicas`).

vsoch:
ah I think I see! So we basically can define readiness gates, and the readiness gates are references to different probes. And a probe can be https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes anything from that set. And for the kind of probe we want readiness: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-readiness-probes. Is that correct?

This is great because it would mean that we can say "Tell replicated job B to start when replicated job A, not the pod but the actual _application inside_ has indicated it is ready based on response from this endpoint / socket / etc. In this case, I think we really could just define the order and then ensure that the containers before it are ready first. If there is a case of A needing to be ready before B and C (not a single dependency but a pair of children) we would still do something like A -> B -> C. 

I didn't know about this Ready thing! That's so cool. TIL!

kannon92:
@danielvegamyhre @ahg-g I pushed an update to this.

I went ahead and simplified the API so we don't allow for sequential execution.  Avoiding the downside of workflow support.  

@danielvegamyhre I decided to keep the field in `StartupPolicy` mostly to keep similar behavior as we have for `SuccessPolicy` and `FailurePolicy`.  I figured we can expand these fields as needed so I wanted to define the field now.  Rather than adding it to the spec field.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+## Motivation
+
+High Performance Computing / Machine Learning users usually want some control over how jobs are being created.  A Startup policy would allow a user to specify which status a replicated job should have before starting other replicated jobs in a jobset.
```
vsoch, line 72: Maybe we could provide an example here, like:

> which status (e.g., ....) a replicated job should have

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+### Goals
+
+- Add a startup policy for JobSet
+
+### Non-Goals
+
+- Complicated workflow specifications
+  - This means that we will support simple dependecies rather than processing a full dag.
```
vsoch, line 81:
```suggestion
  - This means that we will support simple dependecies rather than processing a full DAG.
```
And maybe we could better clarify this - one level of dependencies? One level of flat start up order logic? I don't think we want to imply that we are even processing a DAG here to keep the idea of a workflow entirely separate. This is just a startup policy, no DAG or workflow relevancy. It's for a step in a workload. https://kubernetes.io/docs/concepts/workloads/ (I assume JobSet would eventually be added there).

keps/104-StartupPolicy/README.md
vsoch: If the launcher is essential to everything else, I would have it's (single pod) completion result in a failure in this case. So, no restarts and recreates unless the application stores state separately and is flexible to that.

keps/104-StartupPolicy/README.md
vsoch: I do think if it's at this level we would want to be able to target specific jobs. Arguably if a ReplicatedJob in a JobSet isn't linked to the other ones in a meaningful way, why would it need to be in the Jobset? 

It would make things simpler if at least the readiness was at the level of the Job. There are likely other use cases for that beyond Jobset.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+kind: JobSet
+metadata:
+  name: driver-ready-worker-start
+spec:
+  startupPolicy:
+     replicatedJobsStartInOrder: true
+  replicatedJobs:
+  - name: driver
+    replicas: 1
+    template:
```
vsoch, line 109: stupid question - this is just a Job specification right used as the template? https://github.com/kubernetes-sigs/jobset/blob/4222b34572b8cd8b8a1a7863c9fde5098bd370a3/api/jobset/v1alpha2/jobset_types.go#L136 

So why can't we put the readiness logic in there instead?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+### Notes/Constraints/Caveats (Optional)
+
+In the JobSet API, ReplicatedJobs are the individual Job elements that are created.  These are Replicated so you can have the same Job replicated.  This adds some complexity in the concept of Ready as we must consider the overall ReplicatedJob Ready rather than using the individual Jobs.
```
vsoch, line 217: Nit: in this document I see each of Jobset and JobSet. I think the latter is correct (what I use) so maybe we should be consistent.

keps/104-StartupPolicy/kep.yaml
```diff
@@ -0,0 +1,29 @@
+title: Startup Policy
+kep-number: 104
+authors:
+  - "@kannon92"
+status: provisional
+creation-date: 2023-08-01
+reviewers:
+  - "@daniel"
+  - "@ahg"
+approvers:
```
vsoch, line 10: Am I reviewer here or just an accessory reviewer? :laughing:

kannon92: Sorry, this is the first KEP so I did this very quick and a bit sloppy.  I'll add you as a reviewer!

keps/104-StartupPolicy/README.md
kannon92: @ahg-g and @danielvegamyhre called that out originally in the original jobset doc.  I created https://github.com/kubernetes/kubernetes/issues/117758 for that reason.  

But there wasn't a large drive for that.  I think if we want something at the level of the job, we should bring that up as a use case for the job object.  

Right now, I went ahead with status ready field of Jobs.  This counts the number of pods that are ready and bubbles that information up to the job status.

keps/104-StartupPolicy/README.md
kannon92: I don't really understand what you are asking.  Technically the readiness logic would go there but in the pod template of the JobTemplate (turtles all the way down).  

Are you thinking more that there should be a JobReadiness field?

vsoch: I'm generally thinking how we can simplify the design to put some of the logic into Job, and kill two birds with one stone. It's fairly abstract thinking at the moment and I didn't do a good job of articulating myself (and probably need to think more on this before it's ready for that).

keps/104-StartupPolicy/README.md
kannon92: Updated this.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 2
+
+As a user, I want my replicated jobs to start in order.
+
+One area that this could be useful is if someone has a redis replicatedJob, followed by a driver replicatedJob, followed by a worker replicated job.
```
ahg-g, line 148: Do people use Job to deploy redis?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+  // If this field is set to true, we will start all replicatedJobs in order.
+	// Started means that all the Jobs of a ReplicatedJob are ready.
+	// +optional
+	ReplicatedJobsStartInOrder *bool `json:"replicatedJobsStartInOrder,omitempty"`
+}
+```
+
+### Implementation
+
+If StartupPolicy is set and suspend field is off, we will start all jobs in a suspended state.
```
ahg-g, line 274: How does the implementation interact with failure policy?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+We will not set a condition for JobSet as suspended so we can distinguish between a suspended JobSet and a JobSet with startup policy.
+To track which ReplicatedJobs are suspended, we will add a `Suspended` field to the ReplicatedJobStatus.
```
ahg-g, line 277: This will be confusing, can we use a different name? `WaitingToStart` or something along those lines. Also, when are you going to set this value? it needs to exist from creation time, so I think we need to do it in the defaulting webhook, but I am not sure if status can be set there.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+}
+```
+
+### Implementation
+
+If StartupPolicy is set and suspend field is off, we will start all jobs in a suspended state.
+
+We will not set a condition for JobSet as suspended so we can distinguish between a suspended JobSet and a JobSet with startup policy.
+To track which ReplicatedJobs are suspended, we will add a `Suspended` field to the ReplicatedJobStatus.  
+This field will track the number of replicas that are suspended.
```
ahg-g, line 278:
```suggestion
This field will track ReplicatedJobs that are yet to start.
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+If `ReplicatedJobsStartInOrder` is set, then we will start
+We will loop over all `ReplicatedJobStatuses`.
+We will unsuspend if the ReplicatedJob is suspended and we will trigger a reconcile.
```
ahg-g, line 282:
```suggestion
If `ReplicatedJobsStartInOrder` is set, then we will loop over all `ReplicatedJobStatuses`.
We will start the ReplicatedJob if it is the next in order and trigger a reconcile.
```

ahg-g: Wouldn't we get a reconcile anyway triggered by the previous replicatedjob updates?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+Once all the replicatedJob are unsuspended we go onto to next replicated job in the list.
```
ahg-g, line 283: when will we start the next replicatedJob? until the previous replicatedjob is "ready"?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+If policy is unsuccessful we will return and trigger a reconcile.  If the startup policy for that ReplicatedJob passes then we go onto the next one in the list.
```
ahg-g, line 284: what does it mean that the policy is unsuccessful?

keps/104-StartupPolicy/README.md
kannon92: In my implementation PR, I had startup policy to be immutable so yes.

I'm not sure its worth the complication to allow for updates of the field.

keps/104-StartupPolicy/README.md
kannon92: Left over from previous changes. In this case, it would wait for all the ReplicatedJobs to be ready. And if it fails, then it will be a no-op and on another trigger of the reconcile we will check again.

keps/104-StartupPolicy/README.md
kannon92: I was thinking it would be nice to have a status field for suspended jobs anyway. I halt the execution of the jobs using suspend so it made sense to title this suspended. I can mark it as `WaitingToStart`. The main issue is what do you want for jobs that are all suspended due to the `suspend` field being set?

I have a PR up for this https://github.com/kubernetes-sigs/jobset/pull/250. This mainly adds a new field for JobSet to allow for one to know how many ReplicatedJobs are suspended. And I was able to use this for the implementation PR.

keps/104-StartupPolicy/README.md
kannon92: @vsoch has pointed out to me workflows like https://ieeexplore.ieee.org/abstract/document/9561187?casa_token=aoAoa11fcfoAAAAA:jGlCi6-v5KObaSSlOJooCFCtlpSAQJJkEV8zrVOCTDJfBcf2WihKaDLNTgnd_PqwWQPdQA0 that may require a database/service be deployed as part of a batch job. Maybe a MessageQueue is more likely but there are some examples in the wild.

You are correct that most likely someone should use a Deployment for Redis but I have heard of people wanting batch jobs that utilize some kind of database in the lifecycle of their job and have it complete once the other tasks are complete.

vsoch: That is correct. What distinguishes these services is that they are only needed for the scoped run of the job - it's a strategy for worker communication that can be cleaned up with them.

keps/104-StartupPolicy/README.md
kannon92: As it would normally. We would test failure policy first and if the JobSet failed, then the job is marked as failed.

keps/104-StartupPolicy/README.md
kannon92: I have an implementation of this https://github.com/kubernetes-sigs/jobset/pull/246 if you allow for setting suspend when creating the jobs

keps/104-StartupPolicy/README.md
kannon92: Updated.

keps/104-StartupPolicy/README.md
kannon92: Yes. I updated this to make it clearer.

keps/104-StartupPolicy/README.md
kannon92: In the startup policy function, I was thinking that if a ReplicatedJob is unsuspended, we will update but we should also return and trigger the reconcile. We don't want to unsuspend the other jobs in the list.

keps/104-StartupPolicy/README.md
kannon92: Updated.

keps/104-StartupPolicy/README.md
kannon92: I updated this.

I was thinking in every reconcile, we check `successPolicy`, `failurePolicy` and then `StartupPolicy`. If a Job failed, then we will would mark the JobSet as failed before continuing on to the other jobs.

keps/104-StartupPolicy/README.md
kannon92: I updated the user story to say its a message queue rather than redis.

keps/104-StartupPolicy/README.md
ahg-g: lets clearly state that then.

kannon92: I did have a section on validating saying that startupPolicy is immutable.

keps/104-StartupPolicy/README.md
ahg-g: > I was thinking it would be nice to have a status field for suspended jobs anyway.

ok, we can add that, but suspend and startup sequence are two different concepts and I think it is a mistake to conflate them.

It seems that what we need here is a status to track which replicatedJob we started last to kickoff the next one. This should be separate from the concept of "suspend". We could use a single status field indicating the name of the replicatedJob we processed last.

keps/104-StartupPolicy/README.md
ahg-g: It is not about mutability of the field, more about what we do across restarts, and here we are saying nothing.

We also need to clarify what we want to do across suspends, do we enforce the restart policy after resuming a suspended jobset?

keps/104-StartupPolicy/README.md
kannon92: Do you think that this could just be a label? I don't see why it should be a status field if its an internal implementation detail.

ahg-g: An annotation would be appropriate.

kannon92: I really don't know how we could use annotations to achieve this. The reason I went with Suspend is that there is really no other way in kube to halt the execution of a job once it is created. We have to define the jobs and create them at some point. So I think that starting these jobs in a suspended state is probably the only way to do this with what we have currently.

There is the job readiness gates that @alculquicondor is proposing that could be used here if we don't want to use suspend. But the problem is the same.

We want to create all the jobs in a JobSet and then start them in order. We can't really run the jobs in an order due to the async nature of controllers. It also would make applying failure-policy or success-policy difficult..

ahg-g: We have three cases where we startupPolicy applies:

1) when the jobset is first created
2) when the jobset is restarted after a failure (with restart or recreate all failure policy)
3) when the jobset is resumed after suspension


In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence. In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.

We use the annotation to track which replicatedJob is next to create or unsuspend.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,420 @@
...
+
+ +// Suspended means that the job is suspended
+ +Suspended Status = "Suspended"
+)
+
+type StartupPolicy struct {
+  // If this field is set to true, we will start all replicatedJobs in order.
+	// Started means that all the Jobs of a ReplicatedJob are ready.
+	// +optional
+	ReplicatedJobsStartInOrder *bool `json:"replicatedJobsStartInOrder,omitempty"`
```
ahg-g, line 268: I think we better use an enum here instead to allow to expand into future startup policies.

kannon92: What kind of other startup policies would we support with just an enum? Reverse startup or forward startup?.

It seems that if we want to expand we should have a usecase in mind that can define what they want?

ahg-g: It could be: start the leader first and then the rest in any order (rather than all in strict order). Boolean locks us in two choices, enum offers more future possibilities at pretty much no impact to the API.

The two enum values for now would be `Any` and `InOrder`, with `Any` being the default.

kannon92: I updated the KEP to reflect this. 

I used `AnyOrder` and `InOrder` as `Any` was used for SuccessPolicy.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+We have three cases where startupPolicy applies:
+
+- when the jobset is first created
+- when the jobset is restarted after a failure (with restart or recreate all failure policy)
+- when the jobset is resumed after suspension
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+We use the annotation `jobset.sigs.k8s.io/replicated-job-to-start` to track which replicatedJob is next to create or unsuspend.
```
alculquicondor, line 270: Why do you need this?
In which object is it stored?
How/when is it updated?

keps/104-StartupPolicy/README.md
kannon92: Thanks for clearing that up. I updated the design to call this out explictly.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+extending the production code to implement this enhancement.
+-->
+
+- `controller`: `Aug 7th 2023` - `31.6%`
+
+Unit tests are low in this area as we do more of our testing at the integration level.
+
+We will add a new file called startup_policy to apply our actions.
+
+#### Integration tests
```
alculquicondor, line 333: Worth testing suspending a JobSet when it hasn't finished starting up.

keps/104-StartupPolicy/README.md
kannon92: https://github.com/kubernetes-sigs/jobset/pull/244#discussion_r1424422189

alculquicondor: The answers to these questions should be in the KEP.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+We use the annotation `jobset.sigs.k8s.io/replicated-job-to-start` to track which replicatedJob is next to create or unsuspend.
+
+### Defaulting/Validation
+
+- StartupPolicy is immutable.
+- StartupPolicyOrder of `Any` is default setting.
```
ahg-g, line 275:
```suggestion
- StartupPolicyOrder of `AnyOrder` is default setting.
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+### Goals
+
+- Add a startup policy for JobSet
+
+### Non-Goals
+
+- Complicated workflow specifications
+  - This means that we will support simple dependecies rather than processing a full DAG.
+  - Projects like argo-workflows are a better use case for this.
+- Starting Jobs based on Ready will be at the ReplicatedJob level not at the Job level.
```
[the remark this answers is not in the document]
kannon92: I didn't do a good job of explaining that.  

I meant to say that we don't support a percentage of jobs to be ready (ie 80% of the jobs are ready so the overall ReplicatedJob is ready).

keps/104-StartupPolicy/README.md
kannon92: Updated.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+                sleep 10
+```
+
+#### Story 2
+
+As a user, I want my replicated jobs to start in order.
+
+One area that this could be useful is if someone has a messagequeue replicatedJob, followed by a driver replicatedJob, followed by a worker replicated job.
+
+In this case, I want messagequeue to start, then the driver, and then finally the worker.
```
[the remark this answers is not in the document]
kannon92: Fixed.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+              - -xc
+              - |
+                sleep 10
+```
+
+#### Story 2
+
+As a user, I want my replicated jobs to start in order.
+
+One area that this could be useful is if someone has a messagequeue replicatedJob, followed by a driver replicatedJob, followed by a worker replicated job.
```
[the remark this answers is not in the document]
kannon92: Fixed.

keps/104-StartupPolicy/README.md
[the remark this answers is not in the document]
kannon92: I cleared this up to say we want support partial ready in a ReplicatedJob.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+
+High Performance Computing / Machine Learning users usually want some control over how jobs are being created.  A Startup policy would allow a user to specify which status a replicated job should have before starting other replicated jobs in a JobSet.
+
+### Goals
+
+- Add a startup policy for JobSet
+
+### Non-Goals
+
+- Complicated workflow specifications
```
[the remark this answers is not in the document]
kannon92: updated.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,407 @@
...
+We have three cases where startupPolicy applies:
+
+- when the jobset is first created
+- when the jobset is restarted after a failure (with restart or recreate all failure policy)
+- when the jobset is resumed after suspension
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+We use the annotation `jobset.sigs.k8s.io/replicated-job-to-start` to track which replicatedJob is next to create or unsuspend.
```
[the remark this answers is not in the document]
kannon92: updated.

keps/104-StartupPolicy/README.md
kannon92: Good call. Added

keps/104-StartupPolicy/README.md
kannon92: Thanks for pointing that out. I added a little mention of this.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,411 @@
...
+ ReplicatedJobStart string = "jobset.sigs.k8s.io/startup-policy-next-to-start"
+)
+
+type StartupPolicyOrderOptions string
+
+const (
+ // This is the default setting
+ // AnyOrder means that we will start jobs without any specific order.
+ AnyOrder StartupPolicyOrderOptions = "AnyOrder"
+ // InOrder starts the jobs in order that they are listed.
```
ahg-g, line 247:
```suggestion
 // InOrder starts the ReplicatedJobs in order that they are listed. Jobs within a ReplicatedJob will 
 // still start in any order.
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,411 @@
...
+## Design Details
+
+### API Proposal
+  
+This is an annotation added to a JobSet when `InOrder` is used.
+It will track the next ReplicatedJob that is to be started/resumed.
+
+```golang
+const (
+ ReplicatedJobStart string = "jobset.sigs.k8s.io/startup-policy-next-to-start"
```
ahg-g, line 238:
```suggestion
 StartupPolicyNextToStart string = "jobset.sigs.k8s.io/startup-policy-next-to-start"
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,411 @@
...
+)
+
+type StartupPolicyOrderOptions string
```
ahg-g, line 241:
```suggestion
type StartupPolicyOptions string
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,411 @@
...
+- when the jobset is first created
+- when the jobset is restarted after a failure (with restart or recreate all failure policy)
+- when the jobset is resumed after suspension
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+Since each start/resume of a replicated job will trigger a reconcile, we have to track which replicated job is to start/resume.
+The annotation `jobset.sigs.k8s.io/startup-policy-next-to-start` will be updated on each reconcile to reflect what is the next ReplicatedJob to resume.
+It is implied that the previous all of the jobs of the previous replicated job must be ready before this replicated job is started.
```
ahg-g, line 273: not sure what you mean by implied, but this is something the controller need to explicitly check: start the next replicatedJob only when the previous one is ready.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,411 @@
...
+
+- when the jobset is first created
+- when the jobset is restarted after a failure (with restart or recreate all failure policy)
+- when the jobset is resumed after suspension
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+Since each start/resume of a replicated job will trigger a reconcile, we have to track which replicated job is to start/resume.
+The annotation `jobset.sigs.k8s.io/startup-policy-next-to-start` will be updated on each reconcile to reflect what is the next ReplicatedJob to resume.
```
ahg-g, line 272: Can you explain how this field will be set to prevent race conditions if the JobSet was suspended or a restart triggered mid way through a startup?

Thinking more about this, I think this is probably better be a condition on JobSet, which serves as a way to communicate startup progress.

keps/104-StartupPolicy/README.md
kannon92: yes, the controller will check that. I said implied because we don't have any API or condition that says which job is starting (and we are implicitly waiting for all the jobs to be ready). So its implied in the sense that there is no explict API stating that we are waiting on this job.

The controller should look at the previous job and verify that it is all ready before it starts the next one.

keps/104-StartupPolicy/README.md
kannon92: If its a condition, are you thinking that the reason field could reflect which replicated jobset we are waiting on?

keps/104-StartupPolicy/README.md
alculquicondor: Why not just look at the statuses of all the Jobs in each reconcile? Caching the "next step" is probably going to lead to race conditions.

One corner case to consider is if a Job immediate terminates successfully. What would be the semantics in that case?

kannon92: In this case, the main problem is what replicated job do you want to start next? If replicated job a is all ready, then we should go to b. But looking at the status of all jobs means you may see that both b and c are not yet started but you aren't tracking which one is next to be started.

In the case of job terminate successfully wouldn't that be considered successful eventually in the status field of the jobset?

danielvegamyhre: > In this case, the main problem is what replicated job do you want to start next? If replicated job a is all ready, then we should go to b. But looking at the status of all jobs means you may see that both b and c are not yet started but you aren't tracking which one is next to be started.

Shouldn't the order determined by the order of the replicatedJobs in the spec?

kannon92: Yes, maybe we don't need the annotation after all. I think we have ready, suspended, failed and succeeded as status for ReplicatedJobs.

I think the main issue is on creation or restart, I am thinking you want to track which jobs were created/recreated in case that they are still waiting to be ready.

What I have to do is on `createJobs` I want to only create the jobs in order specified in the startup policy. But it make take a few reconciles for those jobs to start but I think we could look at the status to determine that.

@ahg-g wdyt?

alculquicondor: why is restart different? You should just iterate through all the jobs and see if they are running or not.

> In the case of job terminate successfully wouldn't that be considered successful eventually in the status field of the jobset?

That's a question for your design that should ultimately be reflected in the documentation. It's not necessarily obvious what the semantics are.

alculquicondor: Also what if the job has a TTLAfterFinished?

ahg-g: Yes, we don't need to rely on an explicit state to track what to start next. But I think it is good UX to have a condition indicating that the JobSet is starting up `InOrder` and indicate the progress (as you mentioned, the reason perhaps indicates which replicatedJob we started last or will start next).

I am not sure I follow the question on job termination, if the job terminates, then we are done, why is this relevant to startup sequence? In the same vein, why is TTLAfterFinished relevant as well?

kannon92: IMO its not related to startup policy but its a fair point if we use `replicatedJobStatus` to track state. Since that relies on the Jobs being present in the apiserver, what happens if those jobs are removed from etcd?

If `ttlAfterFinished` is set to a low value (or a job finishes and is deleted quickly), then replicatedJobStatus will no longer detect the status of those replicated jobs. I could see a similar issue happening if `successPolicy` or `failurePolicy` is used. 

Now this seems to be an edge case because the answer I think of is don't set your `ttlAfterFinished` to a time that is less than a reconcile period..

ahg-g: We need to have finalizers on the child jobs to make sure that their status is synced on the replicatedJob status before they get deleted. 

So I agree that we need to take a dependency on the replicatedJob status when determining if the jobs under a replicated job are ready or not. We can address the correctness of status in the face of deleted job objects using finalizers as a followup.

keps/104-StartupPolicy/README.md
ahg-g: Yes.

ahg-g:
@kannon92 this is great, can you please resolve the remaining comments and get this merged?

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,430 @@
...
+
+- when the jobset is first created
+- when the jobset is restarted after a failure (with restart or recreate all failure policy)
+- when the jobset is resumed after suspension
+
+In cases 1 and 2, the operator controls the order of creating the jobs, and so it creates the replicatedJob only when its turn comes in the startup sequence.
+
+In case 3, the operator controls the order of unsuspending the jobs, and so it unsuspends the replicatedJob only when its turn comes in the startup sequence.
+
+On recreation or created, we will apply the startup policy before creating the jobs in a replicated job.
```
ahg-g, line 256:
```suggestion
On initial creation or recreation, we will apply the startup policy before creating the jobs in a replicated job.
```

keps/104-StartupPolicy/README.md
kannon92: I think this  belongs outside of StartupPolicy so I created https://github.com/kubernetes-sigs/jobset/issues/360 to document this.

ahg-g: Yes, but there is still <<Unresolved>> tag in the text.

There are a bunch of suggestions as well that were not addressed yet.

kannon92: I address those and updated this.

keps/104-StartupPolicy/README.md
kannon92: addressed.

keps/104-StartupPolicy/README.md
kannon92: updated this.

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,428 @@
...
+              command:
+              - bash
+              - -xc
+              - |
+                sleep 10
+```
+
+### Risks and Mitigations
+
+To avoid the JobSet become a half-baked workflow manager, we will not allow for starting particular replicated jobs.
```
ahg-g, line 217:
```suggestion
To avoid the JobSet become a half-baked workflow manager, the API will not allow describing DAGs. The goal is to focus on supporting startup sequence that HPC and ML training jobs typically require.
```

keps/104-StartupPolicy/README.md
```diff
@@ -0,0 +1,428 @@
...
+
+### User Experience
+
+If Jobs are taking a long time to go to ready, we should have a clear way to check that the JobSet is behaving as expected.
+
+We will add a condition for StartupPolicy to achieve this.
+
+```golang
+metav1.Condition{
+   Type:    "JobSetStartupPolicyComplete",
```
ahg-g, line 279:
```suggestion
   Type:    "JobSetStartupPolicyCompleted",
```

ahg-g:
A couple final nits, otherwise looks good to me.

keps/104-StartupPolicy/README.md
kannon92: updated.

keps/104-StartupPolicy/README.md
kannon92: updated.

danielvegamyhre:
Thanks for driving this @kannon92! 🥳

PR #250, opened 2023-08-07
add suspend to replicated job status

kannon92:
As part of #244, I think adding a status field for suspend would be helpful if we go this route.  

I added a field for suspend and updated a few of our tests to directly compare against status field.


kannon92:
@danielvegamyhre @ahg-g As we discussed in the startup policy KEP, I'm not sure I will go this route but curious what you think of this. 

If you are happy with, can we merge it into main?

ahg-g:
Thanks, this is great!

/lgtm
/approve

PR #251, opened 2023-08-07
Fix generation

kannon92:
I found a few bugs in our generation.  

OpenAPI had silent errors (list type missing and non matching ReplicatedJobStatus).  

We were never updating the openapi generated code in the API so our python code would not update.

I also fixed a bash script that prevented copying code from temp client-go to our cwd.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -224,7 +224,8 @@ code-generator:
 openapi-gen:
 	@GOBIN=$(PROJECT_DIR)/bin GO111MODULE=on $(GO_CMD) install k8s.io/code-generator/cmd/openapi-gen@latest
 	$(PROJECT_DIR)/bin/openapi-gen --go-header-file hack/boilerplate.go.txt -i sigs.k8s.io/jobset/api/jobset/v1alpha2 -p sigs.k8s.io/jobset/api/jobset/v1alpha2 --alsologtostderr
-
+	rm $(PROJECT_DIR)/api/jobset/v1alpha2/openapi_generated.go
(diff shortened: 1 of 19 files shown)


sdk/python/requirements.txt
```diff
@@ -4,4 +4,3 @@ six >= 1.10
 python_dateutil >= 2.5.3
 setuptools >= 21.0.0
 urllib3 >= 1.15.1
-kubernetes
```
danielvegamyhre, line 7: Did you run post_gen.py? it should add `"kubernetes"` to the requirements.txt file.

kannon92: I thought you said it runs when you run gen.py. 

Reading that script I think it adds kubernetes to test-requirements.txt but not to requirements.txt.

danielvegamyhre: Ah right it is only needed for the unit tests... Nevermind :)

danielvegamyhre:
Checked out the PR locally, made some test changes to the API and confirmed the OpenAPI generation + Python SDK generation sequence is working in my dev environment.

/lgtm
/approve

sdk/python/requirements.txt
kannon92: That’s not exactly true.  I had to add this dependency to pyproject.tomo to import it as a library.  The unit tests pass because we include test dependencies and those work.

kannon92: I noticed we import kubernetes here https://github.com/kubernetes-sigs/jobset/blob/main/sdk/python/jobset/models/__init__.py

kannon92:
@danielvegamyhre wdyt of this one for 0.2?

PR #265, opened 2023-08-23
Fail fast if we cannot build indexes

### LINKED ISSUE #263: Bug: unable to setup indexes
Users are reporting an issue when they apply their JobSet, the JobSet object is created but it never starts running. After examining the logs we found the following error:

```
2023-08-22T21:19:59Z    ERROR    Reconciler error    {"controller": "jobset", "controllerGroup": "jobset.x-k8s.io", "controllerKind": "JobSet", ....  "error": "Index with name field:.metadata.controller does not exist"}
```

The index on `.metadata.controller` is created in [SetupIndexes](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L179) which is called from main.go [here](https://github.com/kubernetes-sigs/jobset/blob/main/main.go#L100). If the indexes cannot be built, we would expect to see [this error logged](https://github.com/kubernetes-sigs/jobset/blob/main/main.go#L101). Looking further in the logs, we do indeed see this error:

```
2023-08-21T14:37:20Z    ERROR    setup    unable to setup indexes    {"error": "failed to get API group resources: unable to retrieve the complete list of server APIs: batch/v1: Get \"https://10.72.0.1:443/apis/batch/v1\": dial tcp 10.72.0.1:443: i/o timeout"}
```

/kind bug

danielvegamyhre:
Fixes #263

### DIFF
--- a/main.go
+++ b/main.go
@@ -99,6 +99,7 @@ func main() {
 	ctx := ctrl.SetupSignalHandler()
 	if err := controllers.SetupIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
 		setupLog.Error(err, "unable to setup indexes")
+		os.Exit(1)
 	}
 
 	// Cert won't be ready until manager starts, so start a goroutine here which


danielvegamyhre:
cc @kannon92 @alculquicondor is this an appropriate way of handling the case where indexes cannot be built, or what is the typical approach here? I notice in the Kueue controller when setting up indexes, errors are only logged but program execution continues uninterrupted: https://github.com/kubernetes-sigs/kueue/blob/main/main.go#L178 

However, as you can see in issue #263 in the case where indexes fail to build, JobSet controller is hitting a reconciliation error [here](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L200) and never starting the jobs due to these indexes not existing, so we need to handle this somehow. Perhaps if we get an error listing child jobs here, we should simply retry the call but without specifying an index (so it will be slower but at least work)?

kannon92 (commented):
/lgtm

PR #269, opened 2023-08-24
Add handling for podFailurePolicy

### LINKED ISSUE #262: Feature: Configurable Failure Policy API
Right now,  the JobSet [FailurePolicy](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L162C13-L162C13) only allows specifying the max number of JobSet restarts to attempt before marking it as failed. 

Users have requested that we should also allow them to configure the failure conditions under which the JobSet will be restarted, or when the JobSet should be failed immediately without going through the allowed number of restarts (see "Example user stories" section below for why this is needed).

As an initial thought, we could define a field in the JobSet FailurePolicy which allows the user to specify the JobSet should respect the Job's [podFailurePolicy](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-failure-policy) (i.e., if a Job failed due to a podFailurePolicy configuring the job to fail immediately without restart under certain conditions, then JobSet should also respect this and not restart the JobSet).

For example, in this list of [common exit codes used by containers](https://komodor.com/learn/exit-codes-in-containers-and-kubernetes-the-complete-guide/), we can see container exit code 143 is used for containers killed by graceful termination (SIGTERM), which is the signal used by maintenance events via graceful node shutdown, as well as by workload pre-emption. 

A user could configure a `podFailurePolicy` on their job to fail the job immediately if there is any exit code **except** 143. If the JobSet respected this, it **would** restart if the child job was killed by a SIGTERM (maintenance event), but **would not** restart if there an application code error.

**Why is this needed**:

Example use cases / user story:

As a user, when I enqueue and run many JobSet workloads using Kueue, I'd like to be able to specify the JobSet failure policy to restart if the containers exited with a SIGTERM exit code (e.g. maintenance event or pre-emption), but NOT restart if the container exited with an application code error, in order to not consume valuable resources repeatedly attempting to run pods that are doomed to fail, while other workloads sit idle in Kueue awaiting free resources to run. 

**Example API**:

```go
type FailurePolicy struct {
	// TargetPodFailurePolicies, if set, will cause the JobSet to fail immediately 
        // without restart if any of its failed child jobs failed due to matching a podFailurePolicy.
	FollowPodFailurePolicy *bool`json:"followPodFailurePolicy,omitempty"`

	// MaxRestarts defines the limit on the number of JobSet restarts.
}
```
Example pod failure policy which the user could use in combination with `.spec.failurePolicy.followPodFailurePolicy = true` to allow jobsets to be restarted if the jobs were killed due to maintenance events, but not if the jobs failed due to application code errors or other bugs:

```yaml
  podFailurePolicy:
    rules:
    - action: FailJob
      onExitCodes:
        containerName: user-container
        operator: NotIn
        values: [143] # SIGTERM
```

danielvegamyhre:
Fixes #262 

cc @kannon92 @alculquicondor


pkg/controllers/jobset_controller.go
```diff
@@ -459,12 +460,30 @@ func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.
...
 
+// If a child job has failed due to triggering a PodFailurePolicy,
+// we should fail the JobSet immediately rather than restarting.
+// This allows the user to configure a PodFailurePolicy such that
+// job failures under certain conditions do not cause the JobSet to
+// restart, while others do.
+func (r *JobSetReconciler) triggeredPodFailurePolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) bool {
+	log := ctrl.LoggerFrom(ctx)
+	for _, failedJob := range ownedJobs.failed {
+		for _, c := range failedJob.Status.Conditions {
```
kannon92, line 477: You could use IsConditionTrue from kubernetes here.

pkg/controllers/jobset_controller.go
```diff
@@ -45,8 +45,9 @@ const (
 )
 
 var (
-	jobOwnerKey = ".metadata.controller"
-	apiGVStr    = jobset.GroupVersion.String()
+	jobOwnerKey                        = ".metadata.controller"
+	apiGVStr                           = jobset.GroupVersion.String()
+	JobConditionReasonPodFailurePolicy = "PodFailurePolicy"
```
kannon92, line 50: https://github.com/kubernetes/kubernetes/blob/714e77595c8b19b693925bda2a96ab80c307d38f/pkg/controller/job/job_controller.go#L59

I know that this change comes from this location and if we go this route I think we’d want to maybe version the constant as a API for job.

I think moving the PodFailurePolicy reason into staging for job would make it so that it would be more difficult to change the constant. Maybe I’m being paranoid but if someone changes this constant name, it could break you pretty easily.  It’d be nice to reference this constant value as part of the API rather than a hard coded constant.


@alculquicondor wdyt?

danielvegamyhre: Yeah it would be ideal if we could simply reference this constant as part of the Job API but since we currently can't, in the interest of velocity I just hard coded a constant here.

pkg/controllers/jobset_controller.go
kannon92: Looking at this more, I don't think you would be able to use this.  That allows you to see if you have a FailedJob conditon but it doesn't match on the reason.  I guess you could match on FailedJob condition and then check reason.

pkg/controllers/jobset_controller.go
kannon92, line 50: Looking at this again I think you want this to be a constant and not a var.  

I would also like to see a comment about why this is defined as is.  If someone changes the job api this would break and it would be confusing to find why this broke.  So I say comment to say a headache in the future.

pkg/controllers/jobset_controller.go
```diff
@@ -459,12 +460,30 @@ func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.
 }
 
 func (r *JobSetReconciler) executeRestartPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) error {
-	if js.Spec.FailurePolicy.MaxRestarts == 0 {
+	if js.Spec.FailurePolicy.MaxRestarts == 0 || r.triggeredPodFailurePolicy(ctx, js, ownedJobs) {
```
kannon92, line 463: So just checking, we only want to obey `PodFailurePolicy` if `RestartPolicy` is specified.  Default behavior would be to fail on any failure?

kannon92 (commented):
Implementation looks good to me!  Its very clean.

Just a few comments.  

Not sure if you want @alculquicondor or @ahg-g thoughts on it.
/hold
/lgtm

pkg/controllers/jobset_controller.go
danielvegamyhre: If no jobset failure policy is specified, the jobset will fail immediately without restarts anyway. So the only place we need to do this podFailurePolicy check is if restarting is an option.

pkg/controllers/jobset_controller.go
danielvegamyhre: Makes sense, updated.

pkg/controllers/jobset_controller.go
alculquicondor: Not sure if this behavior would be immediately intuitive to users:
If one Job fails due to PodFailure policy, the entire job set fails.

Either make sure this is properly documented or make it a JobSet's FailurePolicy whether to respect the Job's.

pkg/controllers/jobset_controller.go
```diff
@@ -459,12 +464,30 @@ func (r *JobSetReconciler) executeFailurePolicy(ctx context.Context, js *jobset.
...
+// If a child job has failed due to triggering a PodFailurePolicy,
+// we should fail the JobSet immediately rather than restarting.
+// This allows the user to configure a PodFailurePolicy such that
+// job failures under certain conditions do not cause the JobSet to
+// restart, while others do.
+func (r *JobSetReconciler) triggeredPodFailurePolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs) bool {
+	log := ctrl.LoggerFrom(ctx)
+	for _, failedJob := range ownedJobs.failed {
+		for _, c := range failedJob.Status.Conditions {
+			if c.Reason == JobConditionReasonPodFailurePolicy && c.Status == corev1.ConditionTrue {
```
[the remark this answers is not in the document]
danielvegamyhre: ah of course, thanks! done.

danielvegamyhre:
>Not sure if this behavior would be immediately intuitive to users:
> If one Job fails due to PodFailure policy, the entire job set fails.
> Either make sure this is properly documented or make it a JobSet's FailurePolicy whether to respect the Job's.

Yeah in my original proposal I wanted to have it configurable as part of the JobSet's FailurePolicy, but I realized modifying the API would probably require a new minor version, and v0.3.0 isn't scheduled until early October, and in the meantime we have users who want this functionality sooner rather than later. 

Personally I would strongly prefer it be a configurable option in the JobSet Failure Policy though so we may have no choice but to wait. I would be curious to get others thoughts on this though.

alculquicondor:
> PodFailurePolicy is one case but what about other FailedJob cases. `ActiveDeadlineExceeded`, `BackoffLimitExceeded`.

I wonder the same. These failures are also indicative of a problem in the user workload.

Let's flip the question: in which scenarios (or use cases), it makes sense to retry the entire job?

kannon92:
> > PodFailurePolicy is one case but what about other FailedJob cases. `ActiveDeadlineExceeded`, `BackoffLimitExceeded`.
> 
> I wonder the same. These failures are also indicative of a problem in the user workload.
> 
> Let's flip the question: in which scenarios (or use cases), it makes sense to retry the entire job?

@danielvegamyhre wdyt?


I can see a more general implementation to obey all failures of a job.

danielvegamyhre:
@kannon92 as long as there is any easy way to check if a Job failed due to `ActiveDeadlineExceeded` or `BackoffLimitExceeded` then we can include those, and perhaps change the name of the configuration to `RespectNonRetriableErrors` or something? We can probably think of a better name, but the general idea seems fine to me.

kannon92:
> @kannon92 as long as there is any easy way to check if a Job failed due to `ActiveDeadlineExceeded` or `BackoffLimitExceeded` then we can include those, and perhaps change the name of the configuration to `RespectNonRetriableErrors` or something? We can probably think of a better name, but the general idea seems fine to me.

We pushed a change in 1.29 to start exposing these as published apis in the reason field.  Both of these work like PodFailurePolicy.

danielvegamyhre:
Closing in favor of revised API discussed in #262

PR #278, opened 2023-08-28
Add validation enforcing that network subdomain adheres to RFC 1035

### LINKED ISSUE #277: Handle service name exceeds max length error gracefully
**Current behavior:**

When the jobset controller generates service names, it's possible they will exceed the 63 character length limit, resulting in an error when attempting to create the service. The Reconcile() call encounters this error and returns early before doing anything useful, so at this point the controller is effectively blocked from reconciling all jobsets it owned (i.e. no jobs being created, no pods being scheduled, etc.)

**Desired behavior**:
Can we find a way to log the error and not reconcile the misconfigured jobset, but continue reconciling any other jobsets?

danielvegamyhre:
Fixes #277 

cc @kannon92 what do you think of this? It seems a bit odd that we are validating against 2 different RFC standards for the subdomain name, since it's used as both as the service name and subdomain name, but this may be simpler than finding a new way to derive a service name and validating that?

### DIFF
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -91,6 +91,11 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
 		for _, errMessage := range validation.IsDNS1123Subdomain(js.Spec.Network.Subdomain) {
 			allErrs = append(allErrs, fmt.Errorf(errMessage))
 		}
+
+		// Since subdomain name is also used as service name, it must adhere to RFC 1035 as well.
+		for _, errMessage := range validation.IsDNS1035Label(js.Spec.Network.Subdomain) {
+			allErrs = append(allErrs, fmt.Errorf(errMessage))
+		}
 	}
 
 	for _, rjob := range js.Spec.ReplicatedJobs {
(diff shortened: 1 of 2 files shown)


kannon92 (commented):
This would be considered a breaking change.

danielvegamyhre:
/hold

target for v0.3.0 release?

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -91,6 +91,11 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
 		for _, errMessage := range validation.IsDNS1123Subdomain(js.Spec.Network.Subdomain) {
```
kannon92, line 91: Is RFC 1035 more emcompassing than DNS 1123?  ie does validating DNS1035 mean that we also satisfy DNS 1123?

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -91,6 +91,11 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
...
 			allErrs = append(allErrs, fmt.Errorf(errMessage))
 		}
+
+		// Since subdomain name is also used as service name, it must adhere to RFC 1035 as well.
+		for _, errMessage := range validation.IsDNS1035Label(js.Spec.Network.Subdomain) {
```
ahg-g, line 96: the subdomain field in the podspec is validated at: https://github.com/kubernetes/kubernetes/blob/6c39a3724d5ff33564d31d5097704f15852c036f/pkg/apis/core/validation/validation.go#L3990

we should do the same here

ahg-g: ah, you are trying to validate for service too, sgtm.

/lgtm

PR #282, opened 2023-08-30
update jobset status without spec

### LINKED ISSUE #141: Unknown fields in logs
As I am looking at the integration logs, I see the following warning:

```
2023-05-12T19:07:27Z  INFO    KubeAPIWarningLogger    unknown field "spec.replicatedJobs[0].template.metadata.creationTimestamp"
  2023-05-12T19:07:27Z  INFO    KubeAPIWarningLogger    unknown field "spec.replicatedJobs[0].template.spec.template.metadata.creationTimestamp"
  2023-05-12T19:07:27Z  INFO    KubeAPIWarningLogger    unknown field "spec.replicatedJobs[1].template.metadata.creationTimestamp"
  2023-05-12T19:07:27Z  INFO    KubeAPIWarningLogger    unknown field "spec.replicatedJobs[1].template.spec.template.metadata.creationTimestamp"
  S
  ```
  
  I don't think its causing any issue but wanted to bring awareness to this.  Not sure of the solution or if it is necessary to solve it.  It happens also in the logs of the integration tests in the CI.

homily707:
to fix #141

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -238,7 +238,7 @@ func (r *JobSetReconciler) calculateAndUpdateReplicatedJobsStatuses(ctx context.
 		return nil
 	}
 	js.Status.ReplicatedJobsStatus = status
-	return r.Status().Update(ctx, js)
+	return r.updateStatusWithoutSpec(ctx, js)
 }
 
 func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, js *jobset.JobSet, jobs *childJobs) []jobset.ReplicatedJobStatus {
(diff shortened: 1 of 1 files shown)


kannon92:
/ok-to-test

can you explain why this fixes the issue you linked?

homily707:
> /ok-to-test
> 
> can you explain why this fixes the issue you linked?

I have the same problem in #141 in my own project, and I searched this solution in [minio-operator](https://github.com/minio/operator/pull/1506).
The reason of the problem is that when you updateStatus, the  "CreationTime" is marshaled as  "null" instead of not rendering at all.
If you don't set spec at all, the problem solved, it's a very elegant solution,  it works in my own project , I think it will also work here. So I open this PR. 

I have build image and it run normally, but it failed the test,  I'm confused about the job failed details. I'll check it out later.

danielvegamyhre:
Interesting, thanks for working on this fix. Let's make sure there are no unintended side effects before merging this. I see the  tests pass, I think it would be good to also manually deploy some of the example workloads in the examples/ directory and verify A) that they work, and B) the "Unknown field" log message doesn't appear.

PR #284, opened 2023-08-31
Validate the generated job names will be dns 1035 compliant

### LINKED ISSUE #283: Add validation to JobSet name to ensure child job names derived from it have no invalid chars
Users have reported issues where the JobSet is successfully created, but special chars that are accepted as part of a JobSet name are not accepted as part of a Job name, and since we derive child job names from the JobSet name, this not only prevents jobs from being created, but seemingly also prevents the jobset controller from reconciling any jobs while this invalid jobset still exists. Deleting the invalid jobset allows the controller to resume reconciling the other valid jobsets.

danielvegamyhre:
Fixes #283

### DIFF
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -27,6 +27,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
 	"sigs.k8s.io/jobset/pkg/util/collections"
 	"sigs.k8s.io/jobset/pkg/util/names"
 
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -107,8 +107,8 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
 			allErrs = append(allErrs, fmt.Errorf("the product of replicas and parallelism must not exceed %d for replicatedJob '%s'", math.MaxInt32, rjob.Name))
 		}
 		// Check that the generated job names for this replicated job will be DNS 1035 compliant.
-		// Use job index 0 as the test example.
-		testJobName := fmt.Sprintf("%s-%s-%d", js.Name, rjob.Name, 0)
+		// Use the largest job index as it will have the longest name.
+		testJobName := fmt.Sprintf("%s-%s-%d", js.Name, rjob.Name, rjob.Replicas-1)
 		for _, errMessage := range validation.IsDNS1035Label(testJobName) {
 			allErrs = append(allErrs, fmt.Errorf(errMessage))
 		}
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -27,8 +27,8 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
-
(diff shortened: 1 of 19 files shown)


danielvegamyhre:
cc @kannon92 am I correct this is considered a breaking change?

kannon92:
> cc @kannon92 am I correct this is considered a breaking change?

Yea I think making validation more strict would be considered a breaking change. OTOH I think having validation to catch errors is a much better UX than failing in reconcile and having someone edit the yaml.

So I don’t think this is being a breaking change should hold it. I was bringing up that in k/k this would be a difficult change to push through.  Usually you would want to guarantee that this wouldn’t break any existing users who may have relied on this (prob unlikely).

I’d like to hear what @ahg-g thinks of changes like this.

api/jobset/v1alpha2/jobset_webhook_test.go
```diff
@@ -409,6 +409,9 @@ func TestValidateCreate(t *testing.T) {
 		{
 			name: "number of pods exceeds the limit",
 			js: &JobSet{
+				ObjectMeta: metav1.ObjectMeta{
```
kannon92, line 412: a nit.  Maybe consider declaring this as a variable (ie `validObjectMeta`) and reuse that in all these places.

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -101,6 +101,12 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
 		if int64(parallelism)*int64(rjob.Replicas) > math.MaxInt32 {
 			allErrs = append(allErrs, fmt.Errorf("the product of replicas and parallelism must not exceed %d for replicatedJob '%s'", math.MaxInt32, rjob.Name))
 		}
+		// Check that the generated job names for this replicated job will be DNS 1035 compliant.
+		// Use job index 0 as the test example.
+		testJobName := fmt.Sprintf("%s-%s-%d", js.Name, rjob.Name, 0)
```
ahg-g, line 106: can we have a shared func between the webhook and the controller for name generation just so when changes are made, they get reflected in both places.

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -101,6 +101,12 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
 		if int64(parallelism)*int64(rjob.Replicas) > math.MaxInt32 {
 			allErrs = append(allErrs, fmt.Errorf("the product of replicas and parallelism must not exceed %d for replicatedJob '%s'", math.MaxInt32, rjob.Name))
 		}
+		// Check that the generated job names for this replicated job will be DNS 1035 compliant.
+		// Use job index 0 as the test example.
```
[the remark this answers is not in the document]
danielvegamyhre: Good point! The length of the job name with the largest job index will be >= the longest job name in the replicatedJob. I updated the test to use that instead of job index 0, since there are certain cases where job index 0 length is less than 63 chars but job index `replicas-1` is not.

kannon92:
I think this looks good to me but I'll leave lgtm for @ahg-g

api/jobset/v1alpha2/jobset_webhook.go
danielvegamyhre: Ok in order to avoid an import/dependency cycle, I had to create a shared util function. However, it didn't seem to fit into any of our existing util pacakges (util/test, util/cert, util/collections) so I made a new one: `pkg/util/names`. I tried to add `GenSubdomain()` into it as well, but it results in an import cycle as well, and unlike `GenJobName` there's no easy fix, so I left it out.

Let me know if you have any thoughts on this.

kannon92 (commented):
One comment https://github.com/kubernetes-sigs/jobset/pull/284/files#r1313274666

/lgtm
/hold

if you want to ignore the comment

api/jobset/v1alpha2/jobset_webhook_test.go
danielvegamyhre: Done

PR #293, opened 2023-09-12
Update docs to use latest release

danielvegamyhre:
We just released v0.2.3, now we need to update the docs.


kannon92 (commented):
/lgtm

It may be worth finding a way to automate this.

PR #303, opened 2023-09-20
Add troubleshooting docs

### LINKED ISSUE #302: Add FAQ for troubleshooting common issues
Common issues and solutions:

1. Jobset webhook not available -> Solution: check if jobset-controller-manager deployment pods are running, usually this means they are unschedulable for some reason (e.g. insufficient CPU -> solution is to scale up CPU node pools or turn on autoscaling)
2. `index .metadata.controller does not exist` -> Bug in older releases, it has been fixed, upgrade to at least JobSet v0.2.3
3. JobSet created but child jobs and/or pods not created  -> Check jobset controller logs, usually means the generated child job names or headless services names (which are derived from the JobSet name) are not valid. Validation has been added to fail the JobSet creation if the generated job/service names will be invalid, but the fix is not included in a release yet. To resolve, simply delete/recreate the jobset with a shorter name with no special chars, ensuring `<jobset-name>-<replicatedJobName>-<jobIndex>-<podIndex>.<subdomain>` will be DNS 1035 compliant.
4. Using JobSet + Kueue, preempted workloads never resume -> Bug in older releases, it has been fixed, upgrade to at least JobSet v0.2.3 + Kueue v0.4.1

Feel free to add any others in the comments and I'll include them in the FAQ

danielvegamyhre:
Fixes #302

### DIFF
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -4,7 +4,7 @@
 
 Example error message:
 
-**Cause**: Usually this means the JobSet controller manager Deployment pods hey are unschedulable for some reason.
+**Cause**: Usually this means the JobSet controller manager Deployment pods are unschedulable for some reason.
 
 **Solution**: Check if jobset-controller-manager deployment pods are running (`kubectl get pods -n jobset-system`).
 If they are in a `Pending` state, describe the pod to see why (`kubectl describe pod <pod> -n jobset-system`), you
--- a/docs/faq/README.md
+++ b/docs/faq/README.md
@@ -1,0 +1,44 @@
+# Troubleshooting Common Issues
+
+### "Webhook not available" error when attempting to create a JobSet
+
+Example error message:
+
+**Cause**: Usually this means the JobSet controller manager Deployment pods hey are unschedulable for some reason.
+
+**Solution**: Check if jobset-controller-manager deployment pods are running (`kubectl get pods -n jobset-system`).
+If they are in a `Pending` state, describ the pod to see why (`kubectl describe pod <pod> -n jobset-system`), you
+should see a message in the pod Events indicating why they are unschedulable. The solution will depend on why the pods
+are unschedulable. For example, if they unschedulable to due insufficient CPU/memory, the solution is to scale up your CPU node pools or turn on autoscaling.
+
+### JobSet is created but child jobs and/or pods are not being created 
+
+Check the jobset controller logs to see why the jobs are not being created:
+
+- `kubectl get pods -n jobset-system`
+- `kubectl logs <pod> -n jobset-system`
+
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
cc @kannon92

docs/faq/README.md
```diff
@@ -0,0 +1,44 @@
+# Troubleshooting Common Issues
+
+### "Webhook not available" error when attempting to create a JobSet
+
+Example error message:
```
kannon92, line 5: Is this missing an error message?

I would wrap this in a code block maybe?

docs/faq/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+**Cause**: Usually this means the JobSet controller manager Deployment pods hey are unschedulable for some reason.
+
+**Solution**: Check if jobset-controller-manager deployment pods are running (`kubectl get pods -n jobset-system`).
+If they are in a `Pending` state, describ the pod to see why (`kubectl describe pod <pod> -n jobset-system`), you
```
kannon92, line 10:
```suggestion
If they are in a `Pending` state, describe the pod to see why (`kubectl describe pod <pod> -n jobset-system`), you
```

docs/faq/README.md
```diff
@@ -0,0 +1,44 @@
...
+should see a message in the pod Events indicating why they are unschedulable. The solution will depend on why the pods
+are unschedulable. For example, if they unschedulable to due insufficient CPU/memory, the solution is to scale up your CPU node pools or turn on autoscaling.
```
kannon92, line 12:
```suggestion
are unschedulable. For example, if they unschedulable due to insufficient CPU/memory, the solution is to scale up your CPU node pools or turn on autoscaling.
```

One could also just change the default resource limits for jobset manager?

kannon92 (commented):
Just a few nits.  

I think we should maybe put FAQ by release or something.  I worry that some of these would look dated over time.  But this is mainly a suggestion.

docs/faq/README.md
```diff
@@ -0,0 +1,44 @@
+# Troubleshooting Common Issues
+
+### "Webhook not available" error when attempting to create a JobSet
+
+Example error message:
+
+**Cause**: Usually this means the JobSet controller manager Deployment pods hey are unschedulable for some reason.
```
ahg-g, line 7:
```suggestion
**Cause**: Usually this means the JobSet controller manager Deployment pods are unschedulable for some reason.
```

docs/faq/README.md
danielvegamyhre: Thanks, added the example error message.

kannon92:
LGTM to me.  I'll leave it for @ahg-g

PR #306, opened 2023-09-21
upgrade to k8s 1.28

### LINKED ISSUE #304: Bump k8s dependencies to 1.28
Bump k8s dependencies to 1.28

kannon92:
Fixes #304


main.go
```diff
@@ -67,9 +68,10 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
-		Scheme:                 scheme,
-		MetricsBindAddress:     metricsAddr,
-		Port:                   9443,
```
kannon92, line 72: @danielvegamyhre this is one issue I am unsure on.  

I am not sure what 9443 corresponds too and what that option is now.

/hold

Makefile
```diff
@@ -231,6 +231,8 @@ CODEGEN_ROOT = $(shell $(GO_CMD) env GOMODCACHE)/k8s.io/code-generator@$(CODEGEN
 code-generator:
 	@GOBIN=$(PROJECT_DIR)/bin GO111MODULE=on $(GO_CMD) install k8s.io/code-generator/cmd/client-gen@$(CODEGEN_VERSION)
 	cp -f $(CODEGEN_ROOT)/generate-groups.sh $(PROJECT_DIR)/bin/
+	cp -f $(CODEGEN_ROOT)/generate-internal-groups.sh $(PROJECT_DIR)/bin/
```
danielvegamyhre, line 234: why did running `generate-internal-groups.sh` become necessary for 1.28?

kannon92: idk but it fails without it.  

See https://github.com/kubernetes-sigs/kueue/pull/1133

kannon92: It is deprecated actually so maybe it won't follow normal behavior. :(

main.go
troy0820: The port was for the webhook server.  

The metricsAddr that you have `:8080` from the flag is being set on the now new `server.Options{BindAddress: ":8080"}`

The webhook has it's own options now set using WebhookServer

troy0820 (commented):
If you want to set the webhook server at :9443, you can do that with the WebhookServer and it's options.

main.go
kannon92: Thanks for helping on that one.  Added this.

test/integration/webhook/suite_test.go
```diff
@@ -91,12 +93,17 @@ var _ = BeforeSuite(func() {
...
-		Scheme:             scheme.Scheme,
-		Host:               webhookInstallOptions.LocalServingHost,
-		Port:               webhookInstallOptions.LocalServingPort,
-		CertDir:            webhookInstallOptions.LocalServingCertDir,
-		LeaderElection:     false,
-		MetricsBindAddress: "0",
+		Scheme:         scheme.Scheme,
+		LeaderElection: false,
+		Metrics: server.Options{
+			BindAddress: "0",
```
danielvegamyhre, line 99: I see setting this to `0` will disable the metrics server. This seems like it would prevent a prometheus operator from scraping metrics as described [here](https://github.com/kubernetes-sigs/jobset/blob/main/docs/setup/install.md#optional-add-metrics-scraping-for-prometheus-operator) (please correct me if I'm wrong)?

kannon92: This is testing for the webhooks. And we had it set to 0 before.

danielvegamyhre: Ah I see, ok thanks

PR #309, opened 2023-09-26
Add pod webhook + reconciler for new "leader pod" exclusive job placement strategy

### LINKED ISSUE #310: Improve scheduling throughput for exclusive job placement per topology
Using pod affinities/anti-affinities for scheduling all pods such that there is 1:1 job per topology domain assignment has resulted in excessively high scheduling latency when running at scale.

We are migrating to a different exclusive placement strategy which scales better.

**Option 1**

For every pod with completion index 0 created ("leader pod"), use a mutating webhook to inject the same exclusive affinities in it, ensuring each leader pod lands in a different topology domain (e.g. node pool). For all other pods, inject a scheduling gate to prevent the scheduler from considering them.

A pod reconciler will do the following: upon pod update events, if the pod has completion index of 0, and the pod has been successfully scheduled to a node, iterate through all pods in the job and set the nodeSelector to match the same node pool label as the leader pod, and then remove the scheduling gate. This will cause all the non-leader pods in the job to follow the leader pod into the same node pool.

**Option 2**: 
Add a pod webhook which does the following:
- if pod being created is a leader pod (index 0), inject pod affinity/anti-affinity (same scheduling constraints used for current exclusive placement implementation). 
- if pod is a follower pod, check if the leader pod for the indexed job is scheduled. if not, reject the pod creation request. otherwise, inject node affinity constraints ensuring it lands in the same topology domain as the leader node.

danielvegamyhre:
Fixes #310 

This PR implements the 2nd option described in the linked issue. Specifically, it adds a pod mutating and admission webhook which does the following:

1) For leader pods, inject exclusive affinities and allow them to be created.
2) For follower pods, reject admission attempts until the leader pod has been scheduled, then inject a nodeSelector for the same topology as the leader is scheduled on.

The pod reconciler is minimal in this PR, and only has the bare minimum logic needed to setup the indexes used by the webhook to look up pods by name and look up all pods for a job. 

Once this PR is submitted, I'll create a follow-up PR with the pod reconciler changes necessary to handle race conditions that can arise in this webhook based exclusive placement strategy.

### DIFF
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -29,7 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"sigs.k8s.io/jobset/pkg/util/collections"
-	"sigs.k8s.io/jobset/pkg/util/names"
+	shared "sigs.k8s.io/jobset/pkg/util/shared"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -109,7 +109,7 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
 		}
 		// Check that the generated job names for this replicated job will be DNS 1035 compliant.
 		// Use the largest job index as it will have the longest name.
-		testJobName := names.GenJobName(js.Name, rjob.Name, int(rjob.Replicas-1))
+		testJobName := shared.GenJobName(js.Name, rjob.Name, int(rjob.Replicas-1))
 		for _, errMessage := range validation.IsDNS1035Label(testJobName) {
 			allErrs = append(allErrs, fmt.Errorf(errMessage))
 		}
--- a/config/components/webhook/manifests.yaml
+++ b/config/components/webhook/manifests.yaml
@@ -24,6 +24,25 @@ webhooks:
     resources:
     - jobsets
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate--v1-pod
+  failurePolicy: Fail
+  name: mpod.kb.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - pods
+  sideEffects: None
 ---
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
@@ -49,4 +68,23 @@ webhooks:
     - UPDATE
     resources:
     - jobsets
+  sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate--v1-pod
+  failurePolicy: Fail
+  name: vpod.kb.io
+  rules:
+  - apiGroups:
+    - ""
+    apiVersions:
+    - v1
+    operations:
+    - CREATE
+    resources:
+    - pods
   sideEffects: None
--- a/main.go
+++ b/main.go
@@ -70,6 +70,7 @@ func main() {
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
 	kubeConfig := ctrl.GetConfigOrDie()
+	// TODO(#338): set QPS and Burst via config flags
 	kubeConfig.QPS = 500
 	kubeConfig.Burst = 500
 
@@ -113,7 +114,7 @@ func main() {
 		setupLog.Error(err, "unable to setup jobset reconciler indexes")
 		os.Exit(1)
 	}
-	if err := controllers.SetupPodReconcilerIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
+	if err := controllers.SetupPodIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
 		setupLog.Error(err, "unable to setup pod reconciler indexes")
 		os.Exit(1)
 	}
@@ -155,14 +156,14 @@ func setupControllers(mgr ctrl.Manager, certsReady chan struct{}) {
 
 	// Set up JobSet validating/defaulting webhook.
 	if err := (&jobset.JobSet{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create validation/defaulting webhook", "webhook", "JobSet")
+		setupLog.Error(err, "unable to create webhook", "webhook", "JobSet")
 		os.Exit(1)
 	}
 
 	// Set up pod mutating and admission webhook.
 	podWebhook := webhooks.NewPodWebhook(mgr)
 	if err := podWebhook.SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create mutating webhook", "webhook", "Pod")
+		setupLog.Error(err, "unable to create webhook", "webhook", "Pod")
 		os.Exit(1)
 	}
 	//+kubebuilder:scaffold:builder
--- a/main.go
+++ b/main.go
@@ -36,6 +36,7 @@ import (
 	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
 	"sigs.k8s.io/jobset/pkg/controllers"
 	"sigs.k8s.io/jobset/pkg/util/cert"
+	"sigs.k8s.io/jobset/pkg/webhooks"
 	//+kubebuilder:scaffold:imports
 )
 
@@ -68,7 +69,11 @@ func main() {
 
 	ctrl.SetLogger(zap.New(zap.UseFlagOptions(&opts)))
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	kubeConfig := ctrl.GetConfigOrDie()
+	kubeConfig.QPS = 500
+	kubeConfig.Burst = 500
+
+	mgr, err := ctrl.NewManager(kubeConfig, ctrl.Options{
 		Scheme: scheme,
 		Metrics: server.Options{
 			BindAddress: metricsAddr,
@@ -104,8 +109,12 @@ func main() {
 	}
 
 	ctx := ctrl.SetupSignalHandler()
-	if err := controllers.SetupIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
-		setupLog.Error(err, "unable to setup indexes")
+	if err := controllers.SetupJobSetIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
+		setupLog.Error(err, "unable to setup jobset reconciler indexes")
+		os.Exit(1)
+	}
+	if err := controllers.SetupPodReconcilerIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
+		setupLog.Error(err, "unable to setup pod reconciler indexes")
 		os.Exit(1)
 	}
 
@@ -130,13 +139,30 @@ func setupControllers(mgr ctrl.Manager, certsReady chan struct{}) {
 	<-certsReady
 	setupLog.Info("certs ready")
 
+	// Set up JobSet controller.
 	jobSetController := controllers.NewJobSetReconciler(mgr.GetClient(), mgr.GetScheme(), mgr.GetEventRecorderFor("jobset"))
 	if err := jobSetController.SetupWithManager(mgr); err != nil {
 		setupLog.Error(err, "unable to create controller", "controller", "JobSet")
 		os.Exit(1)
 	}
+
+	// Set up pod reconciler.
+	podController := controllers.NewPodReconciler(mgr.GetClient(), mgr.GetScheme(), mgr.GetEventRecorderFor("pod"))
+	if err := podController.SetupWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create controller", "controller", "Pod")
+		os.Exit(1)
+	}
+
+	// Set up JobSet validating/defaulting webhook.
 	if err := (&jobset.JobSet{}).SetupWebhookWithManager(mgr); err != nil {
-		setupLog.Error(err, "unable to create webhook", "webhook", "JobSet")
+		setupLog.Error(err, "unable to create validation/defaulting webhook", "webhook", "JobSet")
+		os.Exit(1)
+	}
+
+	// Set up pod mutating and admission webhook.
+	podWebhook := webhooks.NewPodWebhook(mgr)
+	if err := podWebhook.SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create mutating webhook", "webhook", "Pod")
 		os.Exit(1)
 	}
 	//+kubebuilder:scaffold:builder
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -37,7 +37,7 @@ import (
 
 	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
 	"sigs.k8s.io/jobset/pkg/util/collections"
-	"sigs.k8s.io/jobset/pkg/util/names"
+	shared "sigs.k8s.io/jobset/pkg/util/shared"
 )
 
 const (
@@ -176,7 +176,7 @@ func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Complete(r)
 }
 
-func SetupIndexes(ctx context.Context, indexer client.FieldIndexer) error {
+func SetupJobSetIndexes(ctx context.Context, indexer client.FieldIndexer) error {
 	return indexer.IndexField(ctx, &batchv1.Job{}, jobOwnerKey, func(obj client.Object) []string {
 		o := obj.(*batchv1.Job)
 		owner := metav1.GetControllerOf(o)
@@ -232,6 +232,7 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet) 
 	}
 	return &ownedJobs, nil
 }
+
 func (r *JobSetReconciler) calculateAndUpdateReplicatedJobsStatuses(ctx context.Context, js *jobset.JobSet, jobs *childJobs) error {
 	status := r.calculateReplicatedJobStatuses(ctx, js, jobs)
 	// Check if status ReplicatedJobsStatus has changed
@@ -572,7 +573,7 @@ func updateCondition(js *jobset.JobSet, condition metav1.Condition) bool {
 func constructJobsFromTemplate(js *jobset.JobSet, rjob *jobset.ReplicatedJob, ownedJobs *childJobs) ([]*batchv1.Job, error) {
 	var jobs []*batchv1.Job
 	for jobIdx := 0; jobIdx < int(rjob.Replicas); jobIdx++ {
-		jobName := names.GenJobName(js.Name, rjob.Name, jobIdx)
+		jobName := shared.GenJobName(js.Name, rjob.Name, jobIdx)
 		if create := shouldCreateJob(jobName, ownedJobs); !create {
 			continue
 		}
@@ -590,7 +591,7 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:      collections.CloneMap(rjob.Template.Labels),
 			Annotations: collections.CloneMap(rjob.Template.Annotations),
-			Name:        names.GenJobName(js.Name, rjob.Name, jobIdx),
+			Name:        shared.GenJobName(js.Name, rjob.Name, jobIdx),
 			Namespace:   js.Namespace,
 		},
 		Spec: *rjob.Template.Spec.DeepCopy(),
@@ -606,17 +607,17 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 	}
 
 	// If this job should be exclusive per topology, configure the scheduling constraints accordingly.
-	if topologyDomain, ok := js.Annotations[jobset.ExclusiveKey]; ok {
+	if _, ok := js.Annotations[jobset.ExclusiveKey]; ok {
 		// If user has set the nodeSelectorStrategy annotation flag, add the job name label as a
 		// nodeSelector, and add a toleration for the no schedule taint.
 		// The node label and node taint must be added separately by a user/script.
 		if _, exists := js.Annotations[jobset.NodeSelectorStrategyKey]; exists {
-			addNodeSelector(job)
+			addNamespacedJobNodeSelector(job)
 			addTaintToleration(job)
-		} else {
-			// Otherwise, default to using exclusive pod affinities/anti-affinities strategy.
-			setExclusiveAffinities(job, topologyDomain)
 		}
+		// Otherwise, we fall back to the default strategy of the pod webhook assigning exclusive affinities
+		// to the leader pod, preventing follower pod creation until leader pod is scheduled, then
+		// assigning the follower pods to the same node as the leader pods.
 	}
 
 	// if Suspend is set, then we assume all jobs will be suspended also.
@@ -624,60 +625,6 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 	job.Spec.Suspend = ptr.To(jobsetSuspended)
 
 	return job, nil
-}
-
-// Appends pod affinity/anti-affinity terms to the job pod template spec,
-// ensuring that exclusively one job runs per topology domain and that all pods
-// from each job land on the same topology domain.
-func setExclusiveAffinities(job *batchv1.Job, topologyKey string) {
-	if job.Spec.Template.Spec.Affinity == nil {
-		job.Spec.Template.Spec.Affinity = &corev1.Affinity{}
-	}
-	if job.Spec.Template.Spec.Affinity.PodAffinity == nil {
-		job.Spec.Template.Spec.Affinity.PodAffinity = &corev1.PodAffinity{}
-	}
-	if job.Spec.Template.Spec.Affinity.PodAntiAffinity == nil {
-		job.Spec.Template.Spec.Affinity.PodAntiAffinity = &corev1.PodAntiAffinity{}
-	}
-
-	// Pod affinity ensures the pods of this job land on the same topology domain.
-	job.Spec.Template.Spec.Affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(job.Spec.Template.Spec.Affinity.PodAffinity.RequiredDuringSchedulingIgnoredDuringExecution,
-		corev1.PodAffinityTerm{
-			LabelSelector: &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
-				{
-					Key:      jobset.JobKey,
-					Operator: metav1.LabelSelectorOpIn,
-					Values:   []string{job.Labels[jobset.JobKey]},
-				},
-			}},
-			TopologyKey:       topologyKey,
-			NamespaceSelector: &metav1.LabelSelector{},
-		})
-
-	// Pod anti-affinity ensures exclusively this job lands on the topology, preventing multiple jobs per topology domain.
-	job.Spec.Template.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution = append(job.Spec.Template.Spec.Affinity.PodAntiAffinity.RequiredDuringSchedulingIgnoredDuringExecution,
-		corev1.PodAffinityTerm{
-			LabelSelector: &metav1.LabelSelector{MatchExpressions: []metav1.LabelSelectorRequirement{
-				{
-					Key:      jobset.JobKey,
-					Operator: metav1.LabelSelectorOpExists,
-				},
-				{
-					Key:      jobset.JobKey,
-					Operator: metav1.LabelSelectorOpNotIn,
-					Values:   []string{job.Labels[jobset.JobKey]},
-				},
-			}},
-			TopologyKey:       topologyKey,
-			NamespaceSelector: &metav1.LabelSelector{},
-		})
-}
-
-func addNodeSelector(job *batchv1.Job) {
-	if job.Spec.Template.Spec.NodeSelector == nil {
-		job.Spec.Template.Spec.NodeSelector = make(map[string]string)
-	}
-	job.Spec.Template.Spec.NodeSelector[jobset.NamespacedJobKey] = namespacedJobName(job.Namespace, job.Name)
 }
 
 func addTaintToleration(job *batchv1.Job) {
@@ -704,7 +651,7 @@ func shouldCreateJob(jobName string, ownedJobs *childJobs) bool {
 }
 
 func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) {
-	jobName := names.GenJobName(js.Name, rjob.Name, jobIdx)
+	jobName := shared.GenJobName(js.Name, rjob.Name, jobIdx)
 	labels := collections.CloneMap(obj.GetLabels())
 	labels[jobset.JobSetNameKey] = js.Name
 	labels[jobset.ReplicatedJobNameKey] = rjob.Name
@@ -716,8 +663,16 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
 	annotations := collections.CloneMap(obj.GetAnnotations())
 	annotations[jobset.JobSetNameKey] = js.Name
 	annotations[jobset.ReplicatedJobNameKey] = rjob.Name
+	annotations[RestartsKey] = strconv.Itoa(int(js.Status.Restarts))
 	annotations[jobset.ReplicatedJobReplicas] = strconv.Itoa(int(rjob.Replicas))
 	annotations[jobset.JobIndexKey] = strconv.Itoa(jobIdx)
+	annotations[jobset.JobKey] = jobHashKey(js.Namespace, jobName)
+
+	// Only set exclusive key label/annotation on jobs/pods if the parent JobSet
+	// is using exclusive placement.
+	if _, exists := js.Annotations[jobset.ExclusiveKey]; exists {
+		annotations[jobset.ExclusiveKey] = js.Annotations[jobset.ExclusiveKey]
+	}
 
 	obj.SetLabels(labels)
 	obj.SetAnnotations(annotations)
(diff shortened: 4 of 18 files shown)


main.go
```diff
@@ -139,6 +139,8 @@ func setupControllers(mgr ctrl.Manager, certsReady chan struct{}) {
 		setupLog.Error(err, "unable to create webhook", "webhook", "JobSet")
 		os.Exit(1)
 	}
+
+	mgr.GetWebhookServer().Register("/mutate-v1-pod", &webhook.Admission{Handler: &jobset.PodAnnotator{Client: mgr.GetClient()}})
```
ahg-g, line 143: is "/mutate-v1-pod" just the name? where do we say that this is for pods?

danielvegamyhre: ah I was following https://book.kubebuilder.io/reference/webhook-for-core-types but went ahead and changed this to match how we register the validating/defaulting webhook, and specifically made it `For(&corev1.Pod{})`

pkg/util/names/names.go
```diff
@@ -1,9 +1,23 @@
...
+	"encoding/hex"
 	"fmt"
 )
 
 func GenJobName(jsName, rjobName string, jobIndex int) string {
 	return fmt.Sprintf("%s-%s-%d", jsName, rjobName, jobIndex)
 }
+
+// JobHashKey returns the SHA1 hash of the namespaced job name (i.e. <namespace>/<jobName>).
+func JobHashKey(ns string, jobName string) string {
```
ahg-g, line 14: why did you need to move this here?

danielvegamyhre: To avoid an import cycle between v1alpha2 package and controllers package

api/jobset/v1alpha2/jobset_webhook_test.go
```diff
@@ -562,3 +564,119 @@ func TestValidateCreate(t *testing.T) {
...
 }
+
+func TestPatchPod(t *testing.T) {
+	var (
+		topologyKey = "topology-key"
+		jobName     = "job"
+		podName     = "pod"
+		ns          = "default"
+	)
+	tests := []struct {
```
ahg-g, line 575: test that this is added only for pods that have exclusive annotation.

config/components/webhook/manifests.yaml
```diff
@@ -24,6 +24,26 @@ webhooks:
...
     - jobsets
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /mutate-v1-pod
+  failurePolicy: Fail
```
ahg-g, line 34: I am concerned that this will block the whole cluster if jobset-controller is not functioning well. Is there a way to setup the webhook so only specific pods are forwarded? a selector somewhere?

ahg-g: it seems there is: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#matching-requests-objectselector

so we need to set this label selector on pods created (indirectly) via JobSet

pkg/util/names/names.go
ahg-g: but you have the key as a label on the pod, so only the jobset controller should need this function, in all other places we just need to look it up from the label

api/jobset/v1alpha2/jobset_webhook_test.go
```diff
@@ -562,3 +564,119 @@ func TestValidateCreate(t *testing.T) {
...
+			name: "completion index 0",
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: podName,
+					Annotations: map[string]string{
+						batchv1.JobCompletionIndexAnnotation: "0",
+						ExclusiveKey:                         topologyKey,
+					},
+					Labels: map[string]string{
+						JobKey: names.JobHashKey(ns, jobName),
```
ahg-g, line 591: We don't need to use JobHashKey function here (and so we don't need to export it), just set any value manually.

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
+		Complete(r)
+}
+
+func SetupPodReconcilerIndexes(ctx context.Context, indexer client.FieldIndexer) error {
```
danielvegamyhre, line 91: @ahg-g I'm not sure if I'm doing this part correctly, I think the way I'm doing it now will index all pods owned by any job, even that job is not part of a jobset. Seems like we need to do another check to ensure the owner job is part of a jobset? So perhaps make an RPC to get the job and check if it has a JobSetName label/annotation added to it?

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -140,6 +145,101 @@ func (js *JobSet) ValidateDelete() (admission.Warnings, error) {
...
 
+// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io,sideEffects=None,admissionReviewVersions=v1
+
+// podAnnotator for mutating webhook.
+type podAnnotator struct {
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+func NewPodAnnotator(mgr ctrl.Manager) *podAnnotator {
```
ahg-g, line 156: I would name it PodMutatingWebhook to make it clear what this component is about. Also, I would place it in a separate file (pod_webhook.go)

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+	Record record.EventRecorder
+}
+
+func NewPodReconciler(client client.Client, scheme *runtime.Scheme, record record.EventRecorder) *PodReconciler {
+	return &PodReconciler{Client: client, Scheme: scheme, Record: record}
+}
+
+// +kubebuilder:rbac:groups="",resources=events,verbs=create;watch;update;patch
+// +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
+// +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch;create;update;patch;delete
```
ahg-g, line 55: we don't need to create/update/patch/delete nodes, only get/list/watch

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	// TODO: delete logging once we confirmed it works as it will be too noisy.
+	log := ctrl.LoggerFrom(ctx).WithValues("pod", klog.KObj(&pod))
+	ctx = ctrl.LoggerInto(ctx, log)
+	log.V(2).Info("Reconciling Pod")
+
+	// Check if this is the "leader" pod (completion index 0) and if it has been scheduled.
+	if pod.Annotations[batchv1.JobCompletionIndexAnnotation] == "0" && pod.Spec.NodeName != "" {
```
ahg-g, line 72: I think you can set that in the watch filter too

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+
+func SetupPodReconcilerIndexes(ctx context.Context, indexer client.FieldIndexer) error {
+	return indexer.IndexField(ctx, &corev1.Pod{}, podOwnerKey, func(obj client.Object) []string {
+		o := obj.(*corev1.Pod)
+		owner := metav1.GetControllerOf(o)
+		if owner == nil {
+			return nil
+		}
+		// ...make sure it's a Pod owned by a Job...
+		if owner.Kind != "Job" || owner.APIVersion != batchv1.SchemeGroupVersion.String() {
```
ahg-g, line 99: and have the exclusivekey annotation set.

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+			return nil
+		}
+		return []string{owner.Name}
+	})
+}
+
+// nodePoolFromNode gets the node pool name that the given node is part of.
+func (r *PodReconciler) nodePoolFromNode(ctx context.Context, nodeName, ns string) (string, error) {
```
ahg-g, line 107: yeah, we should have no reference to nodepools, lookup the topologyKey label value on the node.

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"sigs.k8s.io/jobset/pkg/util/collections"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+)
+
+const (
+	podOwnerKey = ".metadata.controller"
+	nodePoolKey = "cloud.google.com/gke-nodepool"
```
ahg-g, line 37: nope :)

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+		return "", client.IgnoreNotFound(err)
+	}
+	nodePool, exists := node.Labels[nodePoolKey]
+	if !exists {
+		log.V(2).Error(nil, fmt.Sprintf("missing node pool label: %s", nodePoolKey))
+	}
+	return nodePool, nil
+}
+
+func (r *PodReconciler) assignGatedPodsToNodePool(ctx context.Context, leaderPod *corev1.Pod, nodePool string) error {
```
ahg-g, line 123:
```suggestion
func (r *PodReconciler) assignGatedPodsToTopology(ctx context.Context, leaderPod *corev1.Pod, nodePool string) error {
```

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+	if err := r.List(ctx, &podList, client.InNamespace(leaderPod.Namespace), client.MatchingFields{podOwnerKey: jobName}); err != nil {
+		return err
+	}
+
+	// For every pod besides pod index 0, add a node affinity for this node pool and remove the scheduling gate.
+	for _, pod := range podList.Items {
+		if pod.Annotations[batchv1.JobCompletionIndexAnnotation] == "0" {
+			continue
+		}
+		addNodePoolNodeSelector(&pod, nodePool)
```
ahg-g, line 141: we should do this only if the pod is still gated

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,182 @@
...
+		removeSchedulingGate(&pod)
```
ahg-g, line 142: we need to actually call update on the pod too

config/components/webhook/manifests.yaml
danielvegamyhre: I added an event filter for the pods the controller watches but don't see a similar way to do this for the webhook:

image


I looked at the doc you linked but am still figuring out how to add the objectSelector to Go code

pkg/controllers/pod_controller.go
danielvegamyhre: Good idea! Done

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,227 @@
...
+
+	// Get name of job that owns the leader pod.
+	jobName, err := jobNameFromPod(leaderPod)
+	if err != nil {
+		return err
+	}
+
+	// Get all pods owned by this job.
+	var podList corev1.PodList
+	if err := r.List(ctx, &podList, client.InNamespace(leaderPod.Namespace), client.MatchingFields{podOwnerKey: jobName}); err != nil {
```
ahg-g, line 157: I don't quite recall, but is the jobName really the ownerKey? did you verify this?

config/components/webhook/manifests.yaml
ahg-g: we can change this manually here

pkg/controllers/pod_controller.go
danielvegamyhre: I haven't verified, I assumed based on the Job owner key being JobSet name in the jobset controller. I will verify now.

danielvegamyhre: Verified the owner name is the job name:

```json
        "ownerReferences": [
            {
                "apiVersion": "batch/v1",
                "blockOwnerDeletion": true,
                "controller": true,
                "kind": "Job",
                "name": "pi",
                "uid": "5a3a1823-c35c-4c30-9002-7bf027638864"
            }
        ],
```

Obtained via running this [example Job](https://kubernetes.io/docs/concepts/workloads/controllers/job/#running-an-example-job) and inspecting the ownerReferences.

api/jobset/v1alpha2/pod_webhook.go
```diff
@@ -0,0 +1,127 @@
...
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/mutate-v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create;update,versions=v1,name=mpod.kb.io,sideEffects=None,admissionReviewVersions=v1
```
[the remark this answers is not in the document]
danielvegamyhre: It can be added via the kubebuilder marker? I found a reference to `objectSelector` here https://book.kubebuilder.io/cronjob-tutorial/running-webhook but it links to the `ObjectSelector` field in the k8s admission `ValidatingWebhook` struct which doesn't seem like what we need, since we are using the `controller-runtime` function `NewWebhookManagedBy(...)` to build our webhooks.

Can you clarify how you suggest adding the objectSelector here?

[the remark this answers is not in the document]
ahg-g: I think we can add it to the webhooks manifests.yaml directly, I left a comment there.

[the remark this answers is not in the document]
ahg-g: the controller-runtime is on the client side of the equation, so it will not be where this will be setup. We need it setup on the api-server side, which is what the manifests.yaml file is about

pkg/controllers/jobset_controller.go
```diff
@@ -594,64 +595,13 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
...
-		setExclusiveAffinities(job, topologyDomain)
-	}
 	// if Suspend is set, then we assume all jobs will be suspended also.
 	jobsetSuspended := js.Spec.Suspend != nil && *js.Spec.Suspend
 	job.Spec.Suspend = ptr.To(jobsetSuspended)
 
 	return job, nil
 }
 
-// Appends pod affinity/anti-affinity terms to the job pod template spec,
```
kannon92, line 608: I remember a discussion with @vsoch around these annotations.  Should we implement the new way of placements in case of users using these annotations rather than deleting this entirely?

vsoch: I'm not familiar with the context of this PR, but I think a feature of jobset definitely should be to make it easy to ask for common configurations for affinity (or other similar needs for placement). In my case, I did try using one of the defaults here, but it was slightly off for what I needed (so I couldn't use this strategy) and had to implement an entire function akin to the below, with tweaks. I would say it would make sense to:

1. Define a list of common use cases for affinity / anti-affinity
2. Associate them with names the user can ask for in a section of the spec
3. Implement so it's ridiculously easy to just request a named set!

And in this case, it would make sense to only deprecate the existing when an alternative is available. It might be unlikely that someone is using them, in which case maybe OK to delete first, but perhaps adding the new sets first makes sense so if anyone _is_ using them they aren't really angry :)

pkg/webhooks/pod_webhook.go
```diff
@@ -0,0 +1,145 @@
...
+		For(&corev1.Pod{}).
+		WithDefaulter(p).
+		Complete()
+}
+
+func (p *podMutatingWebhook) Default(ctx context.Context, obj runtime.Object) error {
+	return nil
+}
+
+func (p *podMutatingWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
```
ahg-g, line 57: this handles admission, not mutations; we need to put this logic in the Default function above

sdk/python/docs/JobsetV1alpha2PodAnnotator.md
```diff
@@ -0,0 +1,12 @@
+# JobsetV1alpha2PodAnnotator
```
kannon92, line 1: Looks like something is up with our sdk generation.  I'd expect to see some python code in this PR.

kannon92: Come to think of it, I don't see any client-go changes either.

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,233 @@
...
+	return ctrl.Result{}, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			pod, ok := object.(*corev1.Pod)
+			return ok && usingExclusivePlacement(pod) && isLeaderPod(pod)
```
ahg-g, line 99: as discussed offline, we need to check on all pods, leader and follower

pkg/controllers/pod_controller.go
```diff
@@ -48,6 +49,38 @@ func NewPodReconciler(client client.Client, scheme *runtime.Scheme, record recor
 	return &PodReconciler{Client: client, Scheme: scheme, Record: record}
 }
 
+// SetupWithManager sets up the controller with the Manager.
+func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			pod, ok := object.(*corev1.Pod)
+			return ok && usingExclusivePlacement(pod) && isLeaderPod(pod)
```
ahg-g, line 58: we need to remove the `isLeaderPod` condition from here

pkg/controllers/pod_controller.go
```diff
@@ -48,6 +49,38 @@ func NewPodReconciler(client client.Client, scheme *runtime.Scheme, record recor
...
+		})).
+		Complete(r)
+}
+
+func SetupPodReconcilerIndexes(ctx context.Context, indexer client.FieldIndexer) error {
+	return indexer.IndexField(ctx, &corev1.Pod{}, podOwnerKey, func(obj client.Object) []string {
+		pod := obj.(*corev1.Pod)
+
+		// Make sure the pod has completion index 0 and is part of a JobSet using exclusive placement.
+		if !isLeaderPod(pod) {
```
ahg-g, line 68: ditto, remove this

pkg/controllers/pod_controller.go
```diff
@@ -231,3 +271,11 @@ func usingExclusivePlacement(pod *corev1.Pod) bool {
 func isLeaderPod(pod *corev1.Pod) bool {
 	return pod.Annotations[batchv1.JobCompletionIndexAnnotation] == "0"
 }
+
+func leaderPodName(pod *corev1.Pod) string {
+	parts := strings.Split(pod.Name, "-")
+	n := len(parts)
+	parts = parts[:n-1] // remove pod index
+	parts = append(parts, "0")
```
ahg-g, line 279: This is not correct because the pod have a random suffix added to it; we need to look up the leader pod using the index label

danielvegamyhre: Since index isn't added as a label in version < 1.28, I opted to get pods with the same job-key and then return the one with completion index annotation of 0.

pkg/controllers/pod_controller.go
```diff
@@ -211,16 +211,24 @@ func (r *PodReconciler) assignGatedPodsToTopology(ctx context.Context, leaderPod
...
+	})
 
-	// Get leader pod if it exists.
-	var leaderPod corev1.Pod
-	leaderName := leaderPodName(pod)
-	log.Info(fmt.Sprintf("leader pod is %q for follower pod %q", leaderName, pod.Name))
-	if err := r.Get(ctx, types.NamespacedName{Name: leaderName, Namespace: pod.Namespace}, &leaderPod); err != nil {
-		log.Info(fmt.Sprintf("leader pod %q does not exist", leaderName))
+	var podList corev1.PodList
+	if err := r.List(ctx, &podList, client.InNamespace(pod.Namespace), &client.ListOptions{LabelSelector: labelSelector}); err != nil {
```
ahg-g, line 219: we should list once outside the function (we do a list in assignGatedPodsToTopology) and use that to iterate over the rest of the follower pods

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,283 @@
...
+		log.Info(fmt.Sprintf("%q is a follower pod", pod.Name))
+
+		// If this is not a leader pod, check if the leader pod has been scheduled, and if so,
+		// assign this pod to the same topology.
+		log.Info("getting leader pod for follower")
+		leaderPod, err := r.leaderPodForFollower(ctx, &pod)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+		log.Info("assigning job pods to topology")
```
ahg-g, line 132: we should return if the leader was not scheduled yet

danielvegamyhre: done

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,290 @@
...
+		// If this is not a leader pod, check if the leader pod has been scheduled, and if so,
+		// assign this pod to the same topology.
+		log.Info("getting leader pod for follower")
+		leaderPod, err := r.leaderPodForFollower(ctx, &pod)
+		if err != nil {
+			return ctrl.Result{}, err
+		}
+
+		// If leader pod is not scheduled yet, do nothing and return.
+		if leaderPod.Spec.NodeName == "" {
```
ahg-g, line 134: leaderPod could be nil, in which case we also need to return. I think this line is what caused the crash

pkg/controllers/pod_controller.go
danielvegamyhre: now that we got the functionality working, I did a refactor to remove unnecessary RPCs, simplify logic, reduce code duplication. let me know if you have any additional thoughts

config/components/webhook/podwebhook.yaml
```diff
@@ -0,0 +1,49 @@
+apiVersion: admissionregistration.k8s.io/v1
```
ahg-g, line 1: is this still valid? I guess not, right?

main.go
```diff
@@ -105,7 +106,11 @@ func main() {
 
 	ctx := ctrl.SetupSignalHandler()
 	if err := controllers.SetupIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
-		setupLog.Error(err, "unable to setup indexes")
+		setupLog.Error(err, "unable to setup jobset reconciler indexes")
```
ahg-g, line 109: we need to increase the qps and burst

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,284 @@
...
+		}
+
+		log.Info(fmt.Sprintf("adding node selector to pod: %s", pod.Name))
+		addNodeSelector(&pod, topologyKey, topologyValue)
+
+		log.Info(fmt.Sprintf("removing scheduling gate from pod: %s", pod.Name))
+		removeSchedulingGate(&pod, gateIndex)
+
+		log.Info(fmt.Sprintf("updating pod: %s", pod.Name))
+		if err := r.Update(ctx, &pod); err != nil {
```
ahg-g, line 214: This is a blocking update and so a huge bottleneck, we need to do this in a go routine in parallel; if the update fails, we need to requeue and force another reconile. We need to rate limit re-queueing to avoid hammering the api-server if there is something really wrong.

pkg/controllers/pod_controller.go
```diff
@@ -211,10 +216,17 @@ func (r *PodReconciler) assignGatedPodsToTopology(ctx context.Context, podList *
...
 
 		log.Info(fmt.Sprintf("updating pod: %s", pod.Name))
-		if err := r.Update(ctx, &pod); err != nil {
-			log.Error(err, fmt.Sprintf("updating pod: %s", pod.Name))
-			finalErr = err
-		}
+
+		wg.Add(1)
+		go func(p *corev1.Pod) {
+			err := r.Update(ctx, p)
```
ahg-g, line 248: should we create a context based on the parent one and use here instead?

pkg/controllers/pod_controller.go
```diff
@@ -224,22 +224,23 @@ func (r *PodReconciler) assignGatedPodsToTopology(ctx context.Context, podList *
...
 		// has already processed it and assigned it to the same topology as the leader pod.
-		gateIndex := schedulingGateIndex(&pod)
+		gateIndex := schedulingGateIndex(pod)
 		if gateIndex == gateDoesNotExist {
 			continue
 		}
 
 		log.Info(fmt.Sprintf("adding node selector to pod: %s", pod.Name))
-		addNodeSelector(&pod, topologyKey, topologyValue)
+		addNodeSelector(&p, topologyKey, topologyValue)
```
ahg-g, line 240: this is pod

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,322 @@
...
+	leader := isLeaderPod(&pod)
+	if leader {
+		log.Info(fmt.Sprintf("%q is a leader pod", pod.Name))
+
+		if pod.Spec.NodeName == "" {
+			log.Info("leader pod not scheduled")
+			return ctrl.Result{}, nil
+		}
+	}
+
```
ahg-g, line 122: we can return quick here if not leader and already scheduled (i.e., nodeName is set)

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,322 @@
...
+	return &PodReconciler{Client: client, Scheme: scheme, Record: record}
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
+			pod, ok := object.(*corev1.Pod)
+			return ok && usingExclusivePlacement(pod)
```
ahg-g, line 61: if not leader and nodeName is set, we should filter out

config/components/webhook/podwebhook.yaml
danielvegamyhre: deleted

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,326 @@
...
+// jobKey (SHA1 hash of the namespaced job name) label selector.
+func (r *PodReconciler) listPodsForJob(ctx context.Context, ns, jobKey string) (*corev1.PodList, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	labelSelector := labels.SelectorFromSet(map[string]string{
+		jobset.JobKey: jobKey,
+	})
+
+	var podList corev1.PodList
+	if err := r.List(ctx, &podList, client.InNamespace(ns), &client.ListOptions{LabelSelector: labelSelector}); err != nil {
```
ahg-g, line 287: we should use the index:

`r.List(ctx, &podList, client.InNamespace(ns), client.MatchingFields{podOwnerKey: jobName})`

note we need to pass the jobName that pod belongs to (which is what we used to build the index)

pkg/controllers/pod_controller.go
```diff
@@ -0,0 +1,326 @@
...
+	// If user is not using exclusive placement, we don't need to do anything.
+	topologyKey, exists := pod.Annotations[jobset.ExclusiveKey]
+	if !exists {
+		return ctrl.Result{}, nil
+	}
+
+	log := ctrl.LoggerFrom(ctx).WithValues("pod", klog.KObj(&pod))
+	ctx = ctrl.LoggerInto(ctx, log)
+	log.V(2).Info("Reconciling Pod")
+
```
ahg-g, line 117: can we add and log the time

```
      start := time.Now()
        defer func() {
                end := time.Since(start).Milliseconds()
                log.Info(fmt.Sprintf("time to reconcile pod %q: %d", pod.Name, end))
        }()
```

pkg/webhooks/pod_admission_webhook.go
```diff
@@ -0,0 +1,64 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+	"sigs.k8s.io/jobset/pkg/controllers"
+	"sigs.k8s.io/jobset/pkg/util/shared"
+)
+
+//+kubebuilder:webhook:path=/validate--v1-pod,mutating=false,failurePolicy=fail,sideEffects=None,groups="",resources=pods,verbs=create,versions=v1,name=vpod.kb.io,sideEffects=None,admissionReviewVersions=v1
+
+func (p *podMutatingWebhook) Handle(ctx context.Context, req admission.Request) admission.Response {
```
ahg-g, line 21: Handle is a generic handler for both mutating and admission calls. We already overridden the mutating handler using by implementing the Default interface. We should do the same for Validate. Note that the validating callback is made on create, update and delete, we only need to validate on create.

Here is an example: https://github.com/kubernetes-sigs/controller-runtime/blob/main/examples/builtins/validatingwebhook.go

the ValidateUpdate and ValidateDelete will simply return nil, nil in our case.

pkg/webhooks/pod_mutating_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+
+func NewPodMutatingWebhook(mgr ctrl.Manager) *podMutatingWebhook {
+	return &podMutatingWebhook{client: mgr.GetClient()}
+}
+
+// SetupMutatingWebhook configures the mutating webhook for pods.
+func (p *podMutatingWebhook) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(&corev1.Pod{}).
+		WithDefaulter(p).
```
ahg-g, line 47: .WithValidator(&p{})

pkg/webhooks/pod_mutating_webhook.go
```diff
@@ -0,0 +1,139 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+)
+
+// +kubebuilder:webhook:path=/mutate--v1-pod,mutating=true,failurePolicy=fail,groups="",resources=pods,verbs=create,versions=v1,name=mpod.kb.io,sideEffects=None,admissionReviewVersions=v1
+
+// podMutatingWebhook for mutating webhook.
+type podMutatingWebhook struct {
```
ahg-g, line 34: lets call it `podWebhook` only now that it handles both mutating and validating calls.

pkg/webhooks/pod_admission_webhook.go
danielvegamyhre: I see, thanks! Done

ahg-g:
You can use `make build` on your local environment to check for such compiler errors

pkg/webhooks/pod_admission_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+	"sigs.k8s.io/jobset/pkg/controllers"
+	"sigs.k8s.io/jobset/pkg/util/shared"
+)
+
+//+kubebuilder:webhook:path=/validate--v1-pod,mutating=false,failurePolicy=fail,sideEffects=None,groups="",resources=pods,verbs=create,versions=v1,name=vpod.kb.io,sideEffects=None,admissionReviewVersions=v1
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (v *podWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
ahg-g, line 22: hmm, I thought we must use the same self variable name, shouldn't the `v` be `p`?

pkg/webhooks/pod_mutating_webhook.go
```diff
@@ -0,0 +1,140 @@
...
+			TopologyKey:       pod.Annotations[jobset.ExclusiveKey],
+			NamespaceSelector: &metav1.LabelSelector{},
+		})
+}
+
+func (p *podWebhook) setNodeSelector(ctx context.Context, pod *corev1.Pod) error {
+	log := ctrl.LoggerFrom(ctx)
+	leaderPod, err := p.leaderPodForFollower(ctx, pod)
+	if err != nil {
+		log.Error(err, "finding leader pod for follower")
```
ahg-g, line 134: May be in this case we set the exclusive pod anti-affinity terms?

pkg/webhooks/pod_mutating_webhook.go
```diff
@@ -0,0 +1,140 @@
...
+		return err
+	}
+	pod.Spec.NodeName = leaderPod.Spec.NodeName
```
ahg-g, line 137: we need to set node affinity for the pod to land on the same node pool, not schedule it on the same node :)

pkg/webhooks/pod_mutating_webhook.go
ahg-g: Actually here we just need to return without an error and without changing anything. Mutating webhook is called before the Validating one. So if the leader is not scheduled yet, we return and the validate will reject creating this follower.

pkg/webhooks/pod_admission_webhook.go
```diff
@@ -0,0 +1,112 @@
...
+	_, usingExclusivePlacement := pod.Annotations[jobset.ExclusiveKey]
+
+	// If pod is part of a JobSet that is using the node selector exclusive placement starategy,
+	// we don't need to validate anything.
+	_, usingNodeSelectorStrategy := pod.Annotations[jobset.NodeSelectorStrategyKey]
+
+	// Do not validate anything for leader pods, proceed with creation immediately.
+	// For follower pods, validate leader pod exists and is scheduled.
+	if usingExclusivePlacement && !usingNodeSelectorStrategy && !shared.IsLeaderPod(pod) {
+		leaderScheduled, err := p.leaderPodScheduled(ctx, pod)
```
ahg-g, line 50: we also need to check that the nodeSelector was added the mutating webhook. This is to avoid the race where the leader got scheduled between the two calls.

leader not scheduled
follower mutate webhook called, node affinity not set
leader got scheduled
follower validate called <--- here we should reject to force another call to the mutating webhook to set the node affinity

danielvegamyhre: Ah I thought the validating webhook was called before the mutating webhook. Fixed!

pkg/webhooks/pod_mutating_webhook.go
danielvegamyhre: Thanks, of course, not sure what I was thinking here! Fixed.

pkg/webhooks/pod_admission_webhook.go
danielvegamyhre: fixed

pkg/webhooks/pod_mutating_webhook.go
danielvegamyhre: done

pkg/controllers/pod_controller.go
danielvegamyhre: I think this comment was for debugging the previous design where the logic was all in the reconciler instead of the webhook, so I'm going to close it, but please reopen it if I am mistaken.

pkg/util/shared/shared.go
```diff
@@ -0,0 +1,24 @@
...
+
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+)
+
+func GenJobName(jsName, rjobName string, jobIndex int) string {
+	return fmt.Sprintf("%s-%s-%d", jsName, rjobName, jobIndex)
+}
+
+func IsLeaderPod(pod *corev1.Pod) bool {
```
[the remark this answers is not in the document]
danielvegamyhre: I ended up needing to use this function in the pod reconciler, to check if there is a mismatch between leader and follower pods within a given topology (logic removed from this PR to simplify reviewing), so that's why I placed it in this shared package. 

Yeah naming this package is hard, but I do think we do need a package of shared functions used by both webhooks and controllers, to serve as a single source of truth and avoid code duplication of course.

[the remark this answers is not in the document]
ahg-g: util by definition is a pkg of "shared" functions, may be rename the pkg to `placement` to indicate that those functions are for exclusive placement.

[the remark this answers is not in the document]
ahg-g: you can do that in the followup PR

PR #311, opened 2023-09-28
Add node selector strategy for exclusive job placement per topology

### LINKED ISSUE #312: Node selector + NoSchedule taint exclusive placement strategy
We need to speed up scheduling of pods for very large Jobsets, where the pod affinity/anti-affinity strategy becomes a bottleneck since it scales linearly with the number of pods and nodes.

This solution requires manual actions by the user (running a script) but should dramatically reduce scheduling latency, since use nodeSelectors grows with the number of nodes only, not the number of pods.

**Solution**:

The user compiles a list of target node pool names they want their jobs to run exclusively on, with 1:1 job to node pool placement, then runs a script to add certain labels and taints to the nodes in those nodepools. 

They then pass this list of node pools and their JobSet yaml to a script which generates the list of namespaced job names that will be created by the JobSet, and creating a 1:1 mapping of job to node pool.  

For every given node pool, it looks up the job mapped to that node pool, and applies a namespaced job name node label to every node in that node pool, allowing the job nodeSelectors injected by the JobSet controller to select those nodes to schedule the job pods on. 

It also adds a NoSchedule taint to the nodes (toleration for which is injected by JobSet controller),  preventing any external workloads from running on the nodes.

Together, these allow for exclusive placement of 1 job per GKE node pool.

danielvegamyhre:
Fixes #312

Add node selector strategy for exclusive job placement per topology.

NodeSelectorStrategyKey is an annotation that acts as a flag, the value does not matter. If set, the JobSet controller will automatically inject nodeSelectors for the JobSetNameKey label to ensure exclusive job placement per topology, instead of injecting pod affinity/anti-affinites for this. The user must add the JobSet name node label to the desired topologies separately. 

I manually verified the python script works on a test cluster.

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -27,6 +27,11 @@ const (
 	JobKey                string = "jobset.sigs.k8s.io/job-key"
 	JobNameKey            string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 	ExclusiveKey          string = "alpha.jobset.sigs.k8s.io/exclusive-topology"
+	// NodeSelectorStrategyKey is an annotation that acts as a flag, the value does not matter.
+	// If set, the JobSet controller will automatically inject nodeSelectors for the JobSetNameKey label to
+	// ensure exclusive job placement per topology, instead of injecting pod affinity/anti-affinites for this.
+	// The user must add the JobSet name node label to the desired topologies separately.
+	NodeSelectorStrategyKey string = "alpha.jobset.sigs.k8s.io/node-selector"
 )
 
 type JobSetConditionType string
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -33,6 +33,7 @@ const (
 	// The user must add the JobSet name node label to the desired topologies separately.
 	NodeSelectorStrategyKey string = "alpha.jobset.sigs.k8s.io/node-selector"
 	NamespacedJobKey        string = "alpha.jobset.sigs.k8s.io/namespaced-job"
+	NoScheduleTaintKey      string = "alpha.jobset.sigs.k8s.io/no-schedule"
 )
 
 type JobSetConditionType string
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -32,6 +32,7 @@ const (
 	// ensure exclusive job placement per topology, instead of injecting pod affinity/anti-affinites for this.
 	// The user must add the JobSet name node label to the desired topologies separately.
 	NodeSelectorStrategyKey string = "alpha.jobset.sigs.k8s.io/node-selector"
+	NamespacedJobKey        string = "alpha.jobset.sigs.k8s.io/namespaced-job"
 )
 
 type JobSetConditionType string
--- a/hack/label_nodes/label_nodes.py
+++ b/hack/label_nodes/label_nodes.py
@@ -1,4 +1,28 @@
 #!/usr/bin/python3
+'''
+This script can be used to add certain labels and taints
+to a given list of GKE node pool, in order to dramatically
+speed up scheduling of pods for very large JobSets which are
+using an exclusive placement strategy of one job per node pool.
+
+usage: label_nodes.py [-h] jobset nodepools
+
+positional arguments:
+  jobset      jobset yaml config
+  nodepools   comma separated list of nodepool names
+
+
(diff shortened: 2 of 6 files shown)


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -27,6 +27,11 @@ const (
 	JobKey                string = "jobset.sigs.k8s.io/job-key"
 	JobNameKey            string = "job-name" // TODO(#26): Migrate to the fully qualified label name.
 	ExclusiveKey          string = "alpha.jobset.sigs.k8s.io/exclusive-topology"
+	// NodeSelectorStrategyKey is an annotation that acts as a flag, the value does not matter.
+	// If set, the JobSet controller will automatically inject nodeSelectors for the JobSetNameKey label to
+	// ensure exclusive job placement per topology, instead of injecting pod affinity/anti-affinites for this.
+	// The user must add the JobSet name node label to the desired topologies separately.
+	NodeSelectorStrategyKey string = "alpha.jobset.sigs.k8s.io/node-selector"
```
ahg-g, line 34: lets do that in an annotation, I am not sure if we will carry this long term

pkg/controllers/jobset_controller.go
```diff
@@ -652,6 +660,13 @@ func setExclusiveAffinities(job *batchv1.Job, topologyKey string) {
 		})
 }
 
+func setNodeSelector(js *jobset.JobSet, job *batchv1.Job) {
+	if job.Spec.Template.Spec.NodeSelector == nil {
+		job.Spec.Template.Spec.NodeSelector = make(map[string]string)
+	}
+	job.Spec.Template.Spec.NodeSelector[jobset.JobSetNameKey] = js.Name
```
ahg-g, line 667: Few things:
1) name alone is not unique, it needs to be namespaced
2) we need to inject the job name, not the jobset's name (each job lands on a node pool)
3) Lets also add a noschedule toleration

pkg/controllers/jobset_controller.go
```diff
@@ -594,10 +594,18 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 		job.Spec.Template.Spec.Subdomain = js.Spec.Network.Subdomain
 	}
 
-	// If this job should be exclusive per topology, set the pod affinities/anti-affinities accordingly.
+	// If this job should be exclusive per topology, configure the scheduling constraints accordingly.
 	if topologyDomain, ok := js.Annotations[jobset.ExclusiveKey]; ok {
-		setExclusiveAffinities(job, topologyDomain)
+		// If use has set the nodeSelectorStrategy annotation flag, add the jobset name label as a
```
ahg-g, line 599:
```suggestion
		// If user has set the nodeSelectorStrategy annotation flag, add the job name label as a
```

hack/label_nodes/label_nodes.py
```diff
@@ -0,0 +1,48 @@
...
+    # Create a new Kubernetes client.
+    client = kubernetes.client.CoreV1Api()
+
+    # Get a list of all the nodes in the cluster.
+    nodes = client.list_node().items
+
+    body = {
+        "metadata": {
+            "labels": {
+                JOBSET_NAME_KEY: jobset,
```
ahg-g, line 30: we need to label each node pool using the job name:

np1 gets the label value jobset-workers-slice1
np2 gets the label value jobset-workers-slice2
...

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: This is used as an annotation in the controller - can you clarify what you mean? Do you mean we shouldn't include it in the API types/definitions file since we may not carry it long term?

ahg-g: ah, sorry, tired brain :)

hack/label_nodes/label_nodes.py
danielvegamyhre: done

pkg/controllers/jobset_controller.go
danielvegamyhre: Done

hack/label_nodes/label_nodes.py
```diff
@@ -0,0 +1,104 @@
+#!/usr/bin/python3
```
ahg-g, line 1: some comment on how this script can be used would be good and the args they need to pass

danielvegamyhre: added. let me know if it is too detailed, or if perhaps i should include a concrete example use case?

PR #328, opened 2023-11-07
Parallelize job creation

### LINKED ISSUE #130: Create Jobs in batches
Currently we create one job at at time. We need to create Jobs in parallel (up to some batch size), especially for jobs created for the same ReplicatedJob.

**Why is this needed**:

To scale better, when creating a large number of replicated Jobs.

danielvegamyhre:
Fixes #130 

I parallelized job creation at the replicatedJob level only, since I think parallelizing across all replicatedJobs would complicate the logic for sequential startup.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -41,8 +41,8 @@ import (
 )
 
 const (
-	RestartsKey       string = "jobset.sigs.k8s.io/restart-attempt"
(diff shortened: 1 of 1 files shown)


kannon92 (approved):
/hold
/lgtm

not sure if you want another review so that’s why I put a hold on it.

danielvegamyhre:
It's fine, we can merge this. 

/hold cancel

PR #329, opened 2023-11-09
Migrate to use kube_codegen.sh to generate client

### LINKED ISSUE #305: Migrate to use kube_codegen.sh
```
WARNING: generate-groups.sh is deprecated.
WARNING: Please use k8s.io/code-generator/kube_codegen.sh instead.

WARNING: generate-internal-groups.sh is deprecated.
WARNING: Please use k8s.io/code-generator/kube_codegen.sh instead.
```
When trying to upgrade k8 dependencies, I see the following warnings.  

Kueue has https://github.com/kubernetes-sigs/kueue/issues/1134 tracking this so it would be ideal to have a similar way of generating these dependencies.

ashnehete:
Fixes #305 

Updates hack/update-codegen.sh to use new kube_codegen.sh from k8s.io/code-generator.


hack/update-codegen.sh
```diff
@@ -18,21 +18,34 @@ set -o errexit
...
+kube::codegen::gen_client \
+    --with-watch \
+    --with-applyconfig \
+    --input-pkg-root sigs.k8s.io/jobset/api \
+    --output-base "$REPO_ROOT" \
+    --output-pkg-root sigs.k8s.io/jobset/client-go \
+    --boilerplate "${REPO_ROOT}/hack/boilerplate.go.txt"
 
 # if client-go files were generated outside of repo root, attempt to move them to the repo root.
 CLIENT_GO=$(find $GOPATH -regextype sed -regex ".*jobset.*client-go")
```
ashnehete, line 51: I'm not sure if we need this part anymore. Can anyone please confirm?

danielvegamyhre:
Hi @ashnehete, I will take a look at this tomorrow

hack/update-codegen.sh
danielvegamyhre: Why do you think it is not needed? In the past we've encountered issues where the generated client-go files appear in different locations in the filesystem in different environments, so we have this logic to move the files to the correct location if they end up somewhere in $GOROOT

kannon92: @ashnehete change is trying to fix that issue.  

Both Kueue and Jobset ran into this issue: https://github.com/kubernetes/code-generator/issues/165.  

So I guess the solution is to add a symbolic link for CWD into sigs.k8s.io.  

So what this change does is we should be generating everything into our CWD, but I guess we trick codegen to assume that its actually in sigs.k8s.io.

danielvegamyhre: I see, in that case yes feel free to remove this logic once you've tested manually and confirmed the symbolic link trick works.

ashnehete: The new kube_codegen.sh script seems to be enforcing the output location as specified. I have tested the code manually on MacOS and Ubuntu (Docker).

kannon92: I also pulled this branch and did a quick test with my setup.  I added a dummy api and I was able to confirm that this generates both client-go and python code.  

I don't think I tested any changes where the client-go copying was disabled though.

kannon92:
So I cloned this branch and did some quick testing.  

My work is usually cloned to /home/kehannon/Work/jobset.

I added a comment to a API, ran `make generate manfest`, and I saw both client-go and python-sdk files edited in tree.  

I think this PR is good to go.

/lgtm

@danielvegamyhre any other things you'd want to see from the PR?

danielvegamyhre:
Thanks for testing @kannon92 

/approve

PR #340, opened 2023-11-22
Rename shared pkg to placement

### LINKED ISSUE #339: Rename 'shared' package to 'placement'
Rename 'shared' package to 'placement' to avoid overly generic/redundant "util/shared" naming.

danielvegamyhre:
Fixes #339

### DIFF
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -29,7 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"sigs.k8s.io/jobset/pkg/util/collections"
-	shared "sigs.k8s.io/jobset/pkg/util/placement"
+	"sigs.k8s.io/jobset/pkg/util/placement"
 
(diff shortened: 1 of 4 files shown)


api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -29,7 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"sigs.k8s.io/jobset/pkg/util/collections"
-	shared "sigs.k8s.io/jobset/pkg/util/shared"
+	shared "sigs.k8s.io/jobset/pkg/util/placement"
```
kannon92, line 32:
```suggestion
	placement "sigs.k8s.io/jobset/pkg/util/placement"
```

and maybe update ref just so it’s not confusing

danielvegamyhre: Woops, thanks for catching that, looks like I missed a few references to `shared` - not sure how that happened, I guess the "rename symbol" didn't work as I expected in this case

kannon92 (commented):
/lgtm

I don’t love the name but I can’t think of anything better.

PR #342, opened 2023-11-22
Update pod reconciler to handle race conditions in exclusive placement

### LINKED ISSUE #341: Handle race conditions in exclusive placement
Now that #310 has been addressed via the pod webhook approach, we need to handle certain race conditions which can occur in this strategy.

Consider the following race condition:
 
1) Leader of Job 1 getting killed right after we set the nodeSelector on its follower pods
2) At the same time, leader of Job 2 lands on the same slice before any of Job 1's followers get to schedule on it. In this case Job 1's followers will have a nodeSelector on the node pool, but will never schedule because of the anti-affinity of Job 2's leader.

danielvegamyhre:
Fixes #341 

In the following changes, we aim to enforce that for JobSets using exclusive placement, pods that belong to the same job are scheduled on the same topology domain exclusively. We do this by performing the following steps:

1. Reconcile leader pods which are scheduled and using exclusive placement.
2. For a given leader pod, check all follower pods's nodeSelectors are all configured to select the same topology as the leader pod is currently placed on.
3. If the above condition is false, we delete all the follower pods in this job to allow them to be rescheduled correctly. When deleting pods, we first add a pod condition so that these deletions can be ignored by a podFailurePolicy, to not count against a backoffLimit.

I will add an e2e test forcing the race condition to happen and verifying the reconciler handles it in a follow-up PR.

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -67,8 +67,7 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			// Only reconcile leader pods which have been scheduled which are part of
 			// JobSets using exclusive placement.
 			pod, ok := object.(*corev1.Pod)
-			return ok && usingExclusivePlacement(pod)
-			// return ok && placement.IsLeaderPod(pod) && podScheduled(pod) && usingExclusivePlacement(pod) && !podDeleted(pod)
+			return ok && placement.IsLeaderPod(pod) && podScheduled(pod) && usingExclusivePlacement(pod) && !podDeleted(pod)
 		})).
 		Complete(r)
 }
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -67,7 +67,7 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
 			pod, ok := object.(*corev1.Pod)
 			// Only reconcile leader pods which have been scheduled which are part of
 			// JobSets using exclusive placement.
-			return ok && placement.IsLeaderPod(pod) && isScheduled(pod) && usingExclusivePlacement(pod)
+			return ok && placement.IsLeaderPod(pod) && podScheduled(pod) && usingExclusivePlacement(pod)
 		})).
 		Complete(r)
 }
@@ -170,7 +170,7 @@ func (r *PodReconciler) listPodsForJob(ctx context.Context, ns, jobKey string) (
 func (r *PodReconciler) validatePodPlacements(ctx context.Context, leaderPod *corev1.Pod, podList *corev1.PodList) (bool, error) {
 	// We know exclusive key is set since we have an event filter for this.
 	topologyKey := leaderPod.Annotations[jobset.ExclusiveKey]
-	leaderTopology, err := r.topologyFromPod(ctx, leaderPod, topologyKey)
+	leaderTopology, err := r.leaderPodTopology(ctx, leaderPod, topologyKey)
 	if err != nil {
 		return false, err
 	}
@@ -180,7 +180,7 @@ func (r *PodReconciler) validatePodPlacements(ctx context.Context, leaderPod *co
 		if placement.IsLeaderPod(&pod) {
 			continue
 		}
-		followerTopology, err := r.topologyFromPod(ctx, &pod, topologyKey)
+		followerTopology, err := followerPodTopology(&pod, topologyKey)
 		if err != nil {
 			return false, err
 		}
@@ -189,31 +189,6 @@ func (r *PodReconciler) validatePodPlacements(ctx context.Context, leaderPod *co
 		}
 	}
 	return true, nil
-}
-
-// topologyFromPod returns the topology value (e.g., node pool name, zone, etc.)
-// for a given pod and topology key.
-func (r *PodReconciler) topologyFromPod(ctx context.Context, pod *corev1.Pod, topologyKey string) (string, error) {
-	log := ctrl.LoggerFrom(ctx)
-
-	nodeName := pod.Spec.NodeName
-	ns := pod.Namespace
-
-	// Get node the leader pod is running on.
-	var node corev1.Node
-	if err := r.Get(ctx, types.NamespacedName{Name: nodeName, Namespace: ns}, &node); err != nil {
-		// We'll ignore not-found errors, since there is nothing we can do here.
-		// A node may not exist temporarily due to a maintenance event or other scenarios.
-		log.Error(err, fmt.Sprintf("getting node %s", nodeName))
-		return "", client.IgnoreNotFound(err)
-	}
-
-	// Get topology (e.g. node pool name) from node labels.
-	topology, exists := node.Labels[topologyKey]
-	if !exists {
-		return "", fmt.Errorf("node does not have topology label: %s", topology)
-	}
-	return topology, nil
 }
 
 // deleteFollowerPods deletes follower pods (non-index 0), parallelizing up to `maxParallelism` requests.
@@ -236,6 +211,7 @@ func (r *PodReconciler) deleteFollowerPods(ctx context.Context, pods []corev1.Po
 			Reason:  "ExclusivePlacementViolation",
 			Message: "Pod violated JobSet exclusive placement policy",
 		}
+
 		// If pod status already has this condition, we don't need to send the update RPC again.
 		if updatePodCondition(&pod, condition) {
 			if err := r.Status().Update(ctx, &pod); err != nil {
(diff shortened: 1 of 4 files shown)


danielvegamyhre:
/retest

webhook still isn't ready after 10min somewhat frequently in our CI e2e tests, I think we need to bump the test infra CPU/memory resources.

pkg/controllers/pod_controller.go
```diff
@@ -102,9 +113,187 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
 func (r *PodReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	// Get Pod from apiserver.
```
ahg-g, line 116: add a comment explaining what this reconcile loop is for

pkg/controllers/pod_controller.go
```diff
@@ -102,9 +113,187 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+	var pod corev1.Pod
+	if err := r.Get(ctx, req.NamespacedName, &pod); err != nil {
+		// we'll ignore not-found errors, since there is nothing we can do here.
+		return ctrl.Result{}, client.IgnoreNotFound(err)
+	}
+
+	log := ctrl.LoggerFrom(ctx).WithValues("pod", klog.KObj(&pod))
+	ctx = ctrl.LoggerInto(ctx, log)
+	log.V(2).Info("Reconciling Pod")
+
```
ahg-g, line 126: Explain the "algorithm" in a comment, something like: 

// In the following, we aim to enforce that the pods that belong to the same job are scheduled on the same topology domain exclusively. We do this by ...

pkg/controllers/pod_controller.go
ahg-g, line 116: We should only have reconcile invoked for pods created via jobset and use exclusive placement. We should have a filter for that when setting up the reconciler with the manager, pls check how to do that in kubebuilder.

pkg/controllers/pod_controller.go
```diff
@@ -102,9 +113,187 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
 	return ctrl.Result{}, nil
 }
 
+// listPodsForJobKey returns a list of pods owned by a specific job, using the
+// jobKey (SHA1 hash of the namespaced job name) label selector.
+func (r *PodReconciler) listPodsForJob(ctx context.Context, ns, jobKey string) (*corev1.PodList, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	var podList corev1.PodList
+	if err := r.List(ctx, &podList, client.InNamespace(ns), &client.MatchingFields{podJobKey: jobKey}); err != nil {
```
ahg-g, line 174: do we have an index for this?

pkg/controllers/pod_controller.go
danielvegamyhre: Added

danielvegamyhre: This was already done in the previous PR, so it doesn't show up in the diff here, see the `SetupWithManager` function.

pkg/controllers/pod_controller.go
danielvegamyhre: Yep, see the `SetupPodIndexes` function, it was done in the webhook PR (so the webhook could make use of the indexes), so it doesn't show up in the diff here, but it's there.

danielvegamyhre:
/hold cancel

since #309 and #340 have been merged

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -29,7 +29,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
 
 	"sigs.k8s.io/jobset/pkg/util/collections"
-	shared "sigs.k8s.io/jobset/pkg/util/shared"
+	shared "sigs.k8s.io/jobset/pkg/util/placement"
```
ahg-g, line 32: I thought you had this change merged already, why is it showing up here?

danielvegamyhre: Needed to rebase, sorry about that. PR is updated now.

ahg-g (commented):
l have few more comments, will discuss f2f

pkg/controllers/pod_controller.go
```diff
@@ -99,12 +110,198 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+}
+
+// leaderPodForFollower returns the Pod object for the leader pod (job completion index 0) for
+// a given follower pod.
+func (r *PodReconciler) leaderPodForFollower(ctx context.Context, pod *corev1.Pod) (*corev1.Pod, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	var leaderPod *corev1.Pod
+	if placement.IsLeaderPod(pod) {
+		log.Info(fmt.Sprintf("%q is a leader pod", pod.Name))
```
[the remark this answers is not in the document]
danielvegamyhre: True, alright I reduced the amount of non-critical logs.

pkg/controllers/pod_controller.go
```diff
@@ -99,12 +110,198 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+	if len(podList.Items) != 1 {
+		return nil, fmt.Errorf("expected 1 pod with name %q, got %d", podName, len(podList.Items))
+	}
+
+	return &podList.Items[0], nil
+}
+
+// leaderPodForFollower returns the Pod object for the leader pod (job completion index 0) for
+// a given follower pod.
+func (r *PodReconciler) leaderPodForFollower(ctx context.Context, pod *corev1.Pod) (*corev1.Pod, error) {
```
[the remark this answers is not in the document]
danielvegamyhre: Ended up not needing this function in the pod controller after revising the way we handle race conditions (only reconcile scheduled leader pods using exclusive placement, and only delete follower pods when a placement violation is found).

pkg/controllers/pod_controller.go
```diff
@@ -99,12 +110,198 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+	}
+	return true, nil
+}
+
+// topologyFromPod returns the topology value (e.g., node pool name, zone, etc.)
+// for a given pod and topology key.
+func (r *PodReconciler) topologyFromPod(ctx context.Context, pod *corev1.Pod, topologyKey string) (string, error) {
+	log := ctrl.LoggerFrom(ctx)
+
+	nodeName := pod.Spec.NodeName
```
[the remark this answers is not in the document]
danielvegamyhre: Since we are only reconciling scheduled leader pods using exclusive placement now, there's no need for that check here anymore.

pkg/controllers/pod_controller.go
```diff
@@ -99,12 +110,198 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+}
+
+// deletePods deletes the given pods, parallelizing up to `parallelDeletions` requests.
+func (r *PodReconciler) deletePods(ctx context.Context, pods []corev1.Pod) error {
+	lock := &sync.Mutex{}
+	var finalErrs []error
+
+	workqueue.ParallelizeUntil(ctx, parallelDeletions, len(pods), func(i int) {
+		pod := pods[i]
+		backgroundPolicy := metav1.DeletePropagationBackground
```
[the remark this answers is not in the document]
danielvegamyhre: Done

pkg/controllers/pod_controller.go
```diff
@@ -99,12 +110,198 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+			return ctrl.Result{}, nil
+		}
+	}
+
+	// We need a reference to the scheduled leader pod of this job, to find the topology domain
+	// it is scheduled in.
+	leaderPod, err := r.leaderPodForFollower(ctx, &pod)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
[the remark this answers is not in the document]
danielvegamyhre: Since we are only reconciling scheduled leader pods using exclusive placement now, there's no need for that check here anymore.

pkg/controllers/pod_controller.go
```diff
@@ -56,8 +65,9 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&corev1.Pod{}).
 		WithEventFilter(predicate.NewPredicateFuncs(func(object client.Object) bool {
 			pod, ok := object.(*corev1.Pod)
-			// Only reconcile pods that are part of JobSets using exclusive placement.
-			return ok && usingExclusivePlacement(pod)
+			// Only reconcile leader pods which have been scheduled which are part of
+			// JobSets using exclusive placement.
+			return ok && placement.IsLeaderPod(pod) && isScheduled(pod) && usingExclusivePlacement(pod)
```
ahg-g, line 70: we should skip deleted pods too, we can do that by checking if the `DeletionTimestamp` is set.

pkg/controllers/pod_controller.go
```diff
@@ -99,17 +109,188 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
...
+		// Add condition to pod status so that a podFailurePolicy can be used to ignore
+		// deletions by this controller done to handle race conditions.
+		condition := corev1.PodCondition{
+			Type:    corev1.DisruptionTarget,
+			Status:  v1.ConditionTrue,
+			Reason:  "ExclusivePlacementViolation",
+			Message: "Pod violated JobSet exclusive placement policy",
+		}
+
+		// If pod status already has this condition, we don't need to send the update RPC again.
```
ahg-g, line 215:
```suggestion
		// If pod status already has this condition, we don't need to send the update again.
```

pkg/controllers/pod_controller.go
```diff
@@ -99,17 +109,188 @@ func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
 // +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=core,resources=nodes,verbs=get;list;watch
 
-// Reconcile is part of the main kubernetes reconciliation loop which aims to
-// move the current state of the cluster closer to the desired state.
+// Reconcile attempts to enforce that the pods that belong to the same job are
+// scheduled on the same topology domain exclusively (if the parent JobSet is using
+// exclusive placement).
```
ahg-g, line 114:
```suggestion
// Reconcile attempts to enforce that the pods that belong to the same job are
// scheduled on the same topology domain exclusively if the parent JobSet is using
// exclusive placement.
```

ahg-g (commented):
looks good, just a couple of nits

pkg/controllers/pod_controller.go
danielvegamyhre: done

pkg/controllers/pod_controller.go
danielvegamyhre: done

danielvegamyhre:
@ahg-g I cleaned up the logging in the webhook by removing 1 superfluous log line and marking the others at V(3), which is higher than the controller logs which are V(2), so the user can choose to not view them unless they want to debug something.

PR #358, opened 2023-12-20
Use jobset-system instead of kind-system for jobset

### LINKED ISSUE #356: JobSet logs are not displayed in our e2e tests

We should have the jobset controller pods in our e2e tests. Currently, they are missing from the logs.

See https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/244/pull-jobset-test-e2e-main-1-29/1737240901450731520/artifacts/ and notice that jobset does not have any logs.

**Why is this needed**:
JobSet controller is critical for debugging issues and we should have the logs on the e2e test.

kannon92:
We are not finding the correct namespace so we have no logs.

Fixes #356

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -8,7 +8,7 @@ function cleanup {
         if [ ! -d "$ARTIFACTS" ]; then
             mkdir -p "$ARTIFACTS"
         fi
-        kubectl logs -n kube-system deployment/jobset-controller-manager > $ARTIFACTS/jobset-controller-manager.log || true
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
Thanks for looking into this

/lgtm
/approve

kannon92:
/hold

Want to confirm that it works before merge.

kannon92:
Can confirm that this works: https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/358/pull-jobset-test-e2e-main-1-28/1737556386608320512/artifacts/

You can browse there and see the logs.

/hold cancel

Issue #377, opened 2024-01-17
Job validation bug occurs in JobSet but not standalone Job: podFailurePolicy's "onPodConditions" field not treated as optional


danielvegamyhre:
A while ago we fixed this bug causing validation failure when the podFailurePolicy `onPodConditions` field is unset, even though it is an optional field: https://github.com/kubernetes/kubernetes/pull/120204. This change was cherry picked into 1.27 and my understanding is that using 1.27.7+ should allow me to use this change.

Despite using 1.27.9 for my GKE cluster master version, and v0.28.5 for k8s Go packages in [JobSet](https://github.com/kubernetes-sigs/jobset), I get the following error when onPodConditions is unset:

```
~/go/src/sigs.k8s.io/jobset$ k apply -f examples/simple/configurable-failure-policy.yaml 
The JobSet "configurable-failure-policy" is invalid: 
* spec.replicatedJobs[0].template.spec.podFailurePolicy.rules[0].onPodConditions: Invalid value: "null": spec.replicatedJobs[0].template.spec.podFailurePolicy.rules[0].onPodConditions in body must be of type array: "null"
* spec.replicatedJobs[1].template.spec.podFailurePolicy.rules[0].onPodConditions: Invalid value: "null": spec.replicatedJobs[1].template.spec.podFailurePolicy.rules[0].onPodConditions in body must be of type array: "null"
* <nil>: Invalid value: "null": some validation rules were not checked because the object was invalid; correct the existing errors to complete validation
```

Here is the JobSet yaml:

```
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata:
  name: configurable-failure-policy
  annotations:
    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool # 1:1 job replica to node pool assignment
spec:
  failurePolicy:
    rules:
    - action: FailJobSet
      onJobFailureReasons: 
      - PodFailurePolicy
    maxRestarts: 3
  replicatedJobs:
  - name: buggy-job 
    replicas: 1 # set to number of node pools
    template:
      spec:
        parallelism: 1 # number of nodes per pool
        completions: 1 # number of nodes per pool
        backoffLimit: 0
        # fail job immediately if job was not killed by SIGTERM (graceful node shutdown for maintenance events)
        podFailurePolicy:
          rules:
          - action: FailJob
            onExitCodes:
              containerName: main
              operator: NotIn
[...]

PR #381, opened 2024-01-19
KEP 262: Configurable Failure Policy API

### LINKED ISSUE #262: Feature: Configurable Failure Policy API

danielvegamyhre:
Fixes #262


keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,374 @@
+# KEP-NNNN: Your short, descriptive title
```
kannon92, line 1:
```suggestion
# KEP-262: Configurable Failure Policy
```

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,374 @@
...
+
+**Example Failure Policy Configuration for this use case**:
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: configurable-failure-policy-1
+  annotations:
+    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool # 1:1 job replica to node pool assignment
```
kannon92, line 87: I would suggest leaving out this if its not necessary. Project should be cloud neutral.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,374 @@
...
+              image: docker.io/library/bash:5
+              command: ["bash"]
+              args:
+              - -c
+              - echo "Hello world! I'm going to exit with exit code 1 to simulate a software bug." && sleep 20 && exit 1
+```
+
+### Story 2
+
+As a user, I have a JobSet with 2 replicated jobs: one which runs distributed training processes across a pool of GPU/TPU
```
kannon92, line 128: Again, these examples would be hard for someone to reproduce if they aren't using GKE. Are these details necessary for the KEP?

kannon92 (commented):
Haven't read the API yet but seeing a lot of GKE specifics in the examples.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,377 @@
...
+	FailJob FailurePolicyAction = "FailJob"
+
+	// Restart the failed child job only, not the whole JobSet.
+	RestartJob FailurePolicyAction = "RestartJob"
+)
+
+// FailurePolicyRule defines a FailurePolicyAction to be executed if a child job
+// fails due to a reason listed in OnJobFailureReasons.
+// Only only rule can match a unique (Job failure reason, ReplicatedJob) pair.
+type FailurePolicyRule struct {
```
kannon92, line 236: Check indentation on this API. Its a bit uneven.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,377 @@
...
+
+type FailurePolicy struct {
+	// MaxRestarts defines the limit on the number of JobSet restarts.
+	// A restart is achieved by recreating all active child jobs.
+	MaxRestarts int32 `json:"maxRestarts,omitempty"`
+	// List of failure policy rules for this JobSet.
+  // Only one rule can match a unique (Job failure reason, ReplicatedJob)
+  // pair. If such a rule exists, it is executed. Otherwise, the default
+  // failure policy of attempting to restart the JobSet is applied.
+	Rules []FailurePolicyRule `json:"rules,omitempty"`
```
kannon92, line 259: Should there be an optional here?

I think this requires `MaxRestarts` and `Rules` but we should probably support either or in this case.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,377 @@
...
+}
+
+```
+
+### Constraints
+
+- Each unique (Job failure reason, target Replicated Job) pair can only be associated with one FailurePolicyRule.
```
kannon92, line 266: I think you need to comment a lot more on the job failure reason.

How does the JobSet detect this? Is a condition reason in the Job controller?

What about items like `activeDeadlineSeconds` for Jobs?

What about `BackoffLimit` on a job?

keps/262-ConfigurableFailurePolicy/kep.yaml
```diff
@@ -0,0 +1,28 @@
+title: Configurable Failure Policy
+kep-number: 262
+authors:
+  - "@danielvegamyhre"
+status: provisional
+creation-date: 2023-08-01
+reviewers:
+  - "@kannon92"
```
kannon92, line 8: Maybe @mimowo can weigh in as a reviewer? If possible.

As I see a big overlap between this and `PodFailurePolicy`.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Good point, added a section on the constraints for the OnJobFailureReasons field. Let me know if there's anything else that would be useful here.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Updated

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Updated example to be more cloud neutral.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,385 @@
...
+This gives the user the flexibility to specify different failure and restart semantics for jobs failing for the same
+reason, depending on what Replicated Job they are part of. However, within a Replicated Job, there cannot be more than
+one way of handling a particular Job failure reason.
+
+- OnJobFailureReasons must be a valid Job failure reason as defined [here](https://github.com/kubernetes/kubernetes/blob/2d4100335e4c4ccc28f96fac78153f378212da4c/staging/src/k8s.io/api/batch/v1/types.go#L537-L554)
+in the batchv1 Job API. At the time of writing this KEP, these include:
+  - `PodFailurePolicy`
+  - `BackoffLimitExceeded`
+  - `DeadlineExceeded`
+  - `MaxFailedIndexesExceeded`
```
kannon92, line 277: These are alpha features so maybe we leave those off for now.

We should also think about SuccessPolicy in Job if that feature gets accepted.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Fixed.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,387 @@
...
+  // Restart the failed child job only, not the whole JobSet.
+  RestartJob FailurePolicyAction = "RestartJob"
+)
+
+// FailurePolicyRule defines a FailurePolicyAction to be executed if a child job
+// fails due to a reason listed in OnJobFailureReasons.
+// Only only rule can match a unique (Job failure reason, ReplicatedJob) pair.
+type FailurePolicyRule struct {
+  // The action to take if the rule is matched.
+  Action FailurePolicyAction `json:"action"`
```
kannon92, line 239: Since this this is really an enum. I think we can add KubeBuilder enum validation here.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: > These are alpha features so maybe we leave those off for now.

Looks like they are in beta now, as of 1.29.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Good idea, added. Also just added `// +kubebuilder:validation:UniqueItems` validation to the slices which should have unique elements.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Fixed

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Right, the intent is to support either or. Added +optional kubebuilder marker.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,390 @@
...
+  // +optional
+  // +listType=atomic
+  // +kubebuilder:validation:UniqueItems
+  TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
+}
+
+type FailurePolicy struct {
+  // MaxRestarts defines the limit on the number of JobSet restarts.
+  // A restart is achieved by recreating all active child jobs.
+  // +optional
```
kannon92, line 258: This would actually be a breaking change. MaxRestarts was not marked as optional before. I also think since its a int32 the nil is just 0 which is probably fine.

Maybe just rules should be optional and default means that we restart with ignore?

danielvegamyhre: > This would actually be a breaking change. MaxRestarts was not marked as optional before. I also think since its a int32 the nil is just 0 which is probably fine.

Ah good catch, thanks, fixed.

> Maybe just rules should be optional and default means that we restart with ignore?

In this proposal, if no rules are specified, we fall back to the default behavior of restarting the JobSet and **not** ignoring the restart (i.e., incrementing `jobset.sigs.k8s.io/restart-attempt`) so it counts against MaxRestarts.

If the user doesn't want a restart to count against MaxRestarts, they'll need to define a FailurePolicyRule with action type Ignore, configured for any (or all) failure reasons, and any (or all) target replicated jobs.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,389 @@
...
+    replicas: 1
+    template:
+      spec:
+        parallelism: 1
+        completions: 1
+        backoffLimit: 0
+        # podFailurePolicy which fails job immediately if job was not killed by SIGTERM (i.e., graceful node shutdown for maintenance events)
+        podFailurePolicy:
+          rules:
+          - action: FailJob
```
kannon92, line 109: Did you find that you don't need `onPodConditions` here after all?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,389 @@
...
+4) `Ignore`: To "ignore" the failure (i.e., restart the JobSet without counting it against `MaxRestarts`), the controller
+will delete all child jobs and allow the normal reconciliation process to recreate them. This is in contrast to how
+we restart the JobSet **without** ignoring the failure, which is done by incrementing the counter on the JobSet annotation
+`jobset.sigs.k8s.io/restart-attempt`, which on subsequent reconiliations will trigger job deletion and recreation for any
+child job with a `jobset.sigs.k8s.io/restart-attempt` counter value less than that of the JobSet.
+
+
+### Test Plan
+
+The testing plan will focus on integration tests. Specific test cases and scenarios are defined in the integration test
```
kannon92, line 306: I think you probably should have unit tests tracking this feature also.

You probably want tests for validation and I'm sure there will be some unit tests required for this.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,389 @@
...
+in the form of:
+- <package>: <date> - <current test coverage>
+
+This can inform certain test coverage improvements that we want to do before
+extending the production code to implement this enhancement.
+-->
+
+- `<package>`: `<date>` - `<test coverage>`
+
+#### Integration tests
```
kannon92, line 337: What is our policy for e2e tests?

@ahg-g @danielvegamyhre 

It could be nice to start adding some e2e tests for features like this. But we don't have a lot of coverage of e2e tests at the moment.

keps/262-ConfigurableFailurePolicy/README.md
kannon92: That second part is worth a comment in the API.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: I actually do need to set it to `onPonConditions: []` to pass validation still, but I omitted it because I thought that unsightly bug might confuse readers and distract from the KEP 😆

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: In general, our informal/de fact policy for e2e tests has been only adding e2e tests for things which cannot be tested via unit tests or integration tests. The pod networking test cases are an example of something which can only be verified via an e2e test, since we need the Job controller to be running and to actually create pods. 

In this case, we don't care about pods, just about how the JobSet controller interacts with child jobs based on the failure policy, which is something we can test with integration tests.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+
+Users have requested that we should also allow them to configure the failure conditions under which the JobSet will be restarted,
+or when the JobSet should be failed immediately without going through the allowed number of restarts. 
+
+This will allow users to make more efficient use of their computing resources, by avoiding `maxRestarts` number of unnecessary 
+JobSet restarts in the case of a non-retriable failure (e.g., an application code bug), while still restarting in the event of
+retriable failures (e.g., a host maintenace event).
+
+This is especially important for large scale distributed training workloads on expensive hardware accelerators. Recreating the 
+entire JobSet may take ~1 minute or so (per latest scale tests on several node pools of ~3k nodes each), and if `maxRestarts`
```
ahg-g, line 57:
```suggestion
entire JobSet may take ~1 minute or so (per latest scale tests on ~3k nodes), and if `maxRestarts`
```

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+is set to a very high number to allow for disruptions (which occur often at a large scale like this), the workload could
+waste roughly `maxRestarts` number of minutes repeatedly recreating despite the fact that it is doomed to fail due to a non-retriable
```
ahg-g, line 59: could be more in case the failure doesn't manifest itself immediately

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+metadata:
+  name: configurable-failure-policy-1
+spec:
+  # Failure Policy configured to fail the JobSet immediately if a job failure reason for any child job was NOT due to a SIGTERM (e.g. host maintenance).
+  # Otherwise, restart up to 10 times.
+  failurePolicy:
+    rules:
+    - action: FailJobSet
+      onJobFailureReasons: 
+      - PodFailurePolicy
```
ahg-g, line 99: I think it would be great if we can send an upstream feature request to batchv1.Job to optionally allow setting the reason for the podFailurePolicy as part of the action in that api, or if we don't want to change the api, in the case of onExitCodes to have the code appended to the reason, like "PodFailurePolicy-ExitCode143".

This will give users of jobSet more flexibility since they may want the failure policy to be configured differently depending on the exact reason of pod failure (even if in all cases that caused the job to fail).

ahg-g: @liggitt @alculquicondor what do you think? Which option is more realistic to pursue?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+// FailurePolicyAction defines the action the JobSet controller will take for
+// a given FailurePolicyRule.
+type FailurePolicyAction string
+
+const (
+  // Fail JobSet immediately, regardless of maxRestarts.
+  FailJobSet FailurePolicyAction = "FailJobSet"
+
+  // Don't count the failure against maxRestarts.
+  Ignore FailurePolicyAction = "Ignore"
```
ahg-g, line 227: can we add a story for this case

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+
+  // The failed child job is permanently failed and should not be
+  // restarted, active child jobs will continue to run to completion.
+  FailJob FailurePolicyAction = "FailJob"
```
ahg-g, line 231: ditto, for each action, we should have a story, if we don't have a clear one, we can hold off adding it.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+        template:
+          spec:
+            containers:
+            - name: main
+              image: pytorch:latest
+              command: ["..."]
+            resources:
+              limits:
+                nvidia.com/gpu: 1
+  - name: gpu-worker
```
ahg-g, line 171:
```suggestion
  - name: driver 
```

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+
+  // Don't count the failure against maxRestarts.
+  Ignore FailurePolicyAction = "Ignore"
+
+  // The failed child job is permanently failed and should not be
+  // restarted, active child jobs will continue to run to completion.
+  FailJob FailurePolicyAction = "FailJob"
+
+  // Restart the failed child job only, not the whole JobSet.
+  RestartJob FailurePolicyAction = "RestartJob"
```
ahg-g, line 234: What about `RestartReplicatedJob`? In story 2, if one of the worker jobs fails, perhaps we only need to restart all worker child jobs, but not the driver job too.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+)
+
+// FailurePolicyRule defines a FailurePolicyAction to be executed if a child job
+// fails due to a reason listed in OnJobFailureReasons.
+// Only only rule can match a unique (Job failure reason, ReplicatedJob) pair.
```
ahg-g, line 239: I think this comment is not relevant here, this is related to the Rules list.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+  // +kubebuilder:validation:UniqueItems
+  TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
+}
+
+type FailurePolicy struct {
+  // MaxRestarts defines the limit on the number of JobSet restarts.
+  // A restart is achieved by recreating all active child jobs.
+  MaxRestarts int32 `json:"maxRestarts,omitempty"`
+  // List of failure policy rules for this JobSet.
+  // Only one rule can match a unique (Job failure reason, ReplicatedJob)
```
ahg-g, line 263: Are we actually going to validate that? I think it is easier to say that the rules will be evaluated in-order and the first rule to match will be applied. This semantic makes it possible to add new conditions in the future other than onJobFailureReason (can't think of any now though).

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+### Implmentation
+
+The core part of the implementation will be defining what specific mechanisms the JobSet controller uses to implement
+the behavior defined for each FailurePolicyAction type:
+
+1) `RestartJob`: To restart a single child job without restarting the entire JobSet, the controller will delete that 
+particular child job and allow the normal reconciliation process to recreate it.
+
+2) `FailJob`: To leave a particular child job in a failed state without restarting it or restarting the JobSet, the
+controller will simply do nothing, taking no action on this job.
```
ahg-g, line 296: What about maxRestarts? will it count against it?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+
+3) `FailJobSet`: To fail the JobSet immediately without restarting, the controller updates the JobSet status to failed.
+
+4) `Ignore`: To "ignore" the failure (i.e., restart the JobSet without counting it against `MaxRestarts`), the controller
+will delete all child jobs and allow the normal reconciliation process to recreate them. This is in contrast to how
+we restart the JobSet **without** ignoring the failure, which is done by incrementing the counter on the JobSet annotation
+`jobset.sigs.k8s.io/restart-attempt`, which on subsequent reconiliations will trigger job deletion and recreation for any
+child job with a `jobset.sigs.k8s.io/restart-attempt` counter value less than that of the JobSet.
```
ahg-g, line 304: what if we want to ignore counting it against maxRestarts but restart the child job or the replicatedJob only?

keps/262-ConfigurableFailurePolicy/kep.yaml
ahg-g: Agree, that would be great.

keps/262-ConfigurableFailurePolicy/README.md
kannon92: KEPs are free documentation for the features so I like to make the user stories as real test cases. So users could copy these and try out the feature. But that is just me.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,392 @@
...
+
+### Implmentation
+
+The core part of the implementation will be defining what specific mechanisms the JobSet controller uses to implement
+the behavior defined for each FailurePolicyAction type:
+
+1) `RestartJob`: To restart a single child job without restarting the entire JobSet, the controller will delete that 
+particular child job and allow the normal reconciliation process to recreate it.
+
+2) `FailJob`: To leave a particular child job in a failed state without restarting it or restarting the JobSet, the
```
mimowo, line 295: Question, say we have 2 Jobs failed. One matched the `FailJob` rule, the other does not match any rules, but we have `maxRestarts>0`, and it is not reached yet. Now, two cases, depending on the order of matching the Job failures:
1. [job1 (FailJob), job2(no matching rule)] - so after job1 we do nothing, but when we process job2, do we restart the JobSet, including Job1?
2. [job2 (no matching rule), job1 (FailJob)] - so we restart the JobSet, and then job1 (FailJob) is essentially ignored.

Is my understanding correct, is this expected behavior? I find it a little bit tricky in both cases actually.

examples/simple/configurable-failure-policy.yaml
```diff
@@ -0,0 +1,44 @@
...
+        backoffLimit: 0
+        # fail job immediately if job was not killed by SIGTERM (graceful node shutdown for maintenance events)
+        podFailurePolicy:
+          rules:
+          - action: FailJob
+            onExitCodes:
+              containerName: main
+              operator: NotIn
+              values: [143] # SIGTERM = exit code 143
+            onPodConditions: [] # Note this field must be specified as an empty list, even if unused
```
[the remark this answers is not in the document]
danielvegamyhre: Yes, see #377 which has more detail

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: `RestartJob` will restart the child job without counting it against maxRestarts.

I added `RestartReplicatedJob` which restarts the replicatedJob without counting it against maxRestarts.

examples/simple/configurable-failure-policy.yaml
```diff
@@ -0,0 +1,44 @@
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: configurable-failure-policy
+spec:
+  failurePolicy:
+    rules:
+    # This rule will fail the job set immediately if a job failure not triggered by a pod failure happens.
```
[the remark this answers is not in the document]
danielvegamyhre: Yes that is what I meant, I forgot the word "policy" after "pod failure" which made it confusing. I rephrased this sentence to clarify the meaning.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Removed

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Ok makes sense, I've added 3 user stories, one for each of the following actions:

- `FailJobSet`
- `Ignore`
- `RestartReplicatedJob`

However, I haven't been able to think of a concrete use case for `FailJob` or `RestartJob`. It seems ReplicatedJob level granularity is sufficient for the use cases I have thought of so far. 

I asked @vsoch about potential HPC use cases, to better understand the requirements there, but based on our initial conversation I think the 3 actions above are sufficient.

If anyone has any concrete use cases for restarting an individual child job (or leaving it failed), please feel free to add a coment.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Yeah that makes sense, that is the idea I was trying to get across with User Story 2, although I hadn't defined the API for it yet. Updated.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Added

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Clarified by saying it will '...waste **atleast** maxRestarts number of minutes`

vsoch:
Chatting with @danielvegamyhre in slack - I want to share our discussion here (and perhaps move up a level to a higher level picture):

When thinking about success/failure of JobSet, we have several levels that this might operate on:

- *lowest level only* we have the failure / success policy defined on the level of Job (and indexed job, for example) so then that is honored, and replicatedJob is just a shell to put that in (and honor those policies), and that trickles up to replicatedJob->JobSet.
- "global": if something in the JobSet fails, we have to start everything again (this doesn't seem ideal to me) but certainly was an OK starting approach, before the full set of use cases was known for JobSet.
- "middle level": if something fails in a replicatedJob, take this action.
- "multiple level": any of the above, with needing to resolve conflicting preferences (probably not ideal)

I do agree that just the global level is not ideal (I think this is a main thrust of this KEP) Likely the original implication of JobSet was to make it easy to wrap jobs together, and assume they fail together, but the use case has expanded beyond that to not want that anymore. As an example, I often use JobSet to run an indexedJob (some HPC app running in parallel) and then I have services that run alongside it, nicely on the same headless service provided by JobSet. If the service fails, I would just want the service to restart, and I wouldn't want my main compute/simulation/ML job to be failed. Within that, I might have logic on the indexed job that says something about what happens if a single index fails. 

So the "fail the entire JobSet if one component fails" design is not good for that use case. Then the question becomes where to add the logic - does it belong with Job or replicatedJob? I'm assuming we don't want to manage conflicts (the last case) for now.

For having the logic on replicatedJob, that is theoretically adding additional context to a Job, so I kind of like that, but if we want the failure/success to work for Job outside of JobSet, then the first bullet makes most sense (and then replicated Job inherits) but that might be a harder change if it needs to go into Kubernetes proper. If we aren't able to OR it's much harder to make a change to there, then likely you'd put this logic on replicatedJob for now.

So TLDR: we definitely have use cases for this. The challenging part, for me, is in terms of design - where to allow representing the policies, and how that is managed given conflict.

vsoch:
Also just a high level observation - I understand ML (and ML jobs) are leading the industry space now, but I think it's dangerous to develop Kubernetes components that are optimized primarily for that. I think the vision here needs to be flexible that the future could (likely will be) different.

ahg-g:
Thanks @vsoch, I see in https://github.com/kubernetes-sigs/jobset/pull/381#discussion_r1465571802 that HPC were discussed, do you have feedback on whether or not the current proposal addresses them? Can you help suggest user stories to add to the KEP to evaluate if the proposed approach handles them and if not, discuss how to adjust it to accommodate them?

ahg-g:
> Chatting with @danielvegamyhre in slack - I want to share our discussion here (and perhaps move up a level to a higher level picture):
> 
> When thinking about success/failure of JobSet, we have several levels that this might operate on:
> 
> * _lowest level only_ we have the failure / success policy defined on the level of Job (and indexed job, for example) so then that is honored, and replicatedJob is just a shell to put that in (and honor those policies), and that trickles up to replicatedJob->JobSet.
> * "global": if something in the JobSet fails, we have to start everything again (this doesn't seem ideal to me) but certainly was an OK starting approach, before the full set of use cases was known for JobSet.
> * "middle level": if something fails in a replicatedJob, take this action.
> * "multiple level": any of the above, with needing to resolve conflicting preferences (probably not ideal)
> 
> I do agree that just the global level is not ideal (I think this is a main thrust of this KEP) Likely the original implication of JobSet was to make it easy to wrap jobs together, and assume they fail together, but the use case has expanded beyond that to not want that anymore. As an example, I often use JobSet to run an indexedJob (some HPC app running in parallel) and then I have services that run alongside it, nicely on the same headless service provided by JobSet. If the service fails, I would just want the service to restart, and I wouldn't want my main compute/simulation/ML job to be failed. Within that, I might have logic on the indexed job that says something about what happens if a single index fails.
> 
> So the "fail the entire JobSet if one component fails" design is not good for that use case. Then the question becomes where to add the logic - does it belong with Job or replicatedJob? I'm assuming we don't want to manage conflicts (the last case) for now.
> 
> For having the logic on replicatedJob, that is theoretically adding additional context to a Job, so I kind of like that, but if we want the failure/success to work for Job outside of JobSet, then the first bullet makes most sense (and then replicated Job inherits) but that might be a harder change if it needs to go into Kubernetes proper. If we aren't able to OR it's much harder to make a change to there, then likely you'd put this logic on replicatedJob for now.
> 
> So TLDR: we definitely have use cases for this. The challenging part, for me, is in terms of design - where to allow representing the policies, and how that is managed given conflict.

I think it is better to handle concrete use cases than discussing hypothetical or abstract scenarios, so lets list the user stories that you would like to be taken into account.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: This is a good point - we need to think carefully about how to handle conflicting actions for different failed jobs which fail at the same time. 

I think broader failure policy rules should still apply to failed jobs which match narrower failure policies. 

In your example 1 above, the first job will be left failed, then the next job will trigger a recreate of the whole JobSet. That is the intended behavior.

In example 2 above, the first failed job will trigger a JobSet recreation, and the second failed job will also be restarted, regardless of the fact that it matched a narrower failure policy rule, which is the intended behavior.

The narrower failure policies are there to enable the user to avoid more expensive restart actions (e.g., recreating the whole JobSet) if there is **only** a failed job matching that rule (for example, enable the user not restart the whole JobSet if ONLY some auxiliary metrics collecting jobs failed).  

However, we still want the broader, more expensive restart policies to apply to all applicable jobs, including that failed job which matched some other failure policy. I'm not aware of a use case for a broad restart policy ignoring failed jobs which matched a narrower failure policy.

@ahg-g what do you think?

vsoch:
> I think it is better to handle concrete use cases than discussing hypothetical or abstract scenarios, so lets list the user stories that you would like to be taken into account.

Sure, the concrete use case is an indexed job running a simulation (wrapped in a replicatedJob) an then having several services on the same headless service (other replicatedJob) and then not wanting the entire JobSet to fail when one of the services needs to restart.

danielvegamyhre:
> > I think it is better to handle concrete use cases than discussing hypothetical or abstract scenarios, so lets list the user stories that you would like to be taken into account.
> 
> Sure, the concrete use case is an indexed job running a simulation (wrapped in a replicatedJob) an then having several services on the same headless service (other replicatedJob) and then not wanting the entire JobSet to fail when one of the services needs to restart.

For this use case, I think the `RestartReplicatedJob` action in the current API would work. If a job running one of the services fails, only the ReplicatedJob for the services will be recreated, not the ReplicatedJob for the main workload.

vsoch:
> For this use case, I think the RestartReplicatedJob action in the proposed API would work. If a job running one of the services fails, only the ReplicatedJob for the services will be recreated, not the ReplicatedJob for the main workload.

And I think as long as we had some kind of policy for within the indexed job (e.g., don't restart failed workers) then it would cover the set of (near term) foreseeable cases.

danielvegamyhre:
> > For this use case, I think the RestartReplicatedJob action in the proposed API would work. If a job running one of the services fails, only the ReplicatedJob for the services will be recreated, not the ReplicatedJob for the main workload.
> 
> And I think as long as we had some kind of policy for within the indexed job (e.g., don't restart failed workers) then it would cover the set of (near term) foreseeable cases.

To clarify, do you mean within a single indexed job, don't restart a pod that has failed?

vsoch:
Yes, we have [an example](https://github.com/flux-framework/flux-operator/tree/test-refactor-modular/examples/elasticity/downsize) of downsizing a Flux MiniCluster and this means:

1. A sidecar that is monitoring the main queue notices (via an algorithm) that we have too many pods, we can downsize
2. We have a nice way to kill that worker
3. With [this setting](https://github.com/flux-framework/flux-operator/blob/289128a8ccbda725bccc7f24742e9a8dcf5d4308/examples/elasticity/downsize/minicluster.yaml#L10) the pod is allowed to exit with code 0 instead of trying to reconnect
4. The pod completes and does not restart

To be clear, this isn't a failure policy that is supported natively, we had to come up with that solution via the return code - a 0 allows completion, and anything else will have the worker try to reconnect. Then the "restart" case is having completions == parallelism so we can enable scaling up/down of the actual cluster. It would create a new pod, with (unfortunately) a higher index, but we can keep track of that within the sidecar that is monitoring things.

danielvegamyhre:
> Yes, we have [an example](https://github.com/flux-framework/flux-operator/tree/test-refactor-modular/examples/elasticity/downsize) of downsizing a Flux MiniCluster and this means:
> 
> 1. A sidecar that is monitoring the main queue notices (via an algorithm) that we have too many pods, we can downsize
> 2. We have a nice way to kill that worker
> 3. With [this setting](https://github.com/flux-framework/flux-operator/blob/289128a8ccbda725bccc7f24742e9a8dcf5d4308/examples/elasticity/downsize/minicluster.yaml#L10) the pod is allowed to exit with code 0 instead of trying to reconnect
> 4. The pod completes and does not restart
> 
> To be clear, this isn't a failure policy that is supported natively, we had to come up with that solution via the return code - a 0 allows completion, and anything else will have the worker try to reconnect. Then the "restart" case is having completions == parallelism so we can enable scaling up/down of the actual cluster. It would create a new pod, with (unfortunately) a higher index, but we can keep track of that within the sidecar that is monitoring things.

Ok I see, yeah this proposal wouldn't interfere with the behavior you have implemented there. As long as the Job itself is not marked as failed, the JobSet controller will do nothing here, regardless of the internal state of the indexed job. It only looks at failed jobs.

keps/262-ConfigurableFailurePolicy/README.md
vsoch: For the HPC use case it would be hugely useful to be able to act on cases down to the exit code.

keps/262-ConfigurableFailurePolicy/README.md
vsoch: Why is an empty list required? Does the absence of the field not indicate that?

keps/262-ConfigurableFailurePolicy/README.md
vsoch: `FailJob` or `RestartJob` might be relevant if you are running some kind of simulation (a number of replicas within a ReplicatedJob) that would generate a result that is valid or invalid, and you had a way to measure the distance from that. For example, if each Job within the Replicated Job generates a random parameter matrix for the simulation within that Job (to produce a result that is either valid or invalid) if your resulting value is too far off from some metric of goodness you would want to fail that specific Job to generate a new parameter matrix - the one you have isn't good. If the result is within a reasonable distance of your metric of goodness (close to valid, but still invalid) you might want to `RestartJob` because you have some confidence it could produce a valid result. You'd want to run the simulation again.

vsoch:
> Thanks @vsoch, I see in https://github.com/kubernetes-sigs/jobset/pull/381#discussion_r1465571802 that HPC were discussed, do you have feedback on whether or not the current proposal addresses them? Can you help suggest user stories to add to the KEP to evaluate if the proposed approach handles them and if not, discuss adjustments to the API?

See https://github.com/kubernetes-sigs/jobset/pull/381#discussion_r1465826615

That's a dummy example from a real workflow I'm implementing in Kubernetes now - not using jobset (yet) but could eventually get there.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: No, only restarting the JobSet will count against maxRestarts. However, I still haven't thought of a concrete use case for `FailJob` so I may end up removing it.

vsoch: @danielvegamyhre did you see https://github.com/kubernetes-sigs/jobset/pull/381#discussion_r1465826615 ? I think that's a valid case - a workflow I'm developing for K8s right now has that property.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: > `FailJob` or `RestartJob` might be relevant if you are running some kind of simulation (a number of replicas within a ReplicatedJob) that would generate a result that is valid or invalid, and you had a way to measure the distance from that. For example, if each Job within the Replicated Job generates a random parameter matrix for the simulation within that Job (to produce a result that is either valid or invalid) if your resulting value is too far off from some metric of goodness you would want to fail that specific Job to generate a new parameter matrix - the one you have isn't good. If the result is within a reasonable distance of your metric of goodness (close to valid, but still invalid) you might want to `RestartJob` because you have some confidence it could produce a valid result. You'd want to run the simulation again.

How would this metric be surfaced to the JobSet controller though, in order to determine if it should FailJob or RestartJob? It only can look at failed Jobs and examine the reason it failed.

vsoch: It wouldn't - it would be internal to the application, which would then exit with the appropriate code to determine the desired action.

danielvegamyhre: > It wouldn't - it would be internal to the application, which would then exit with the appropriate code to determine the desired action.

Right, but the podFailurePolicy would be configured to do what based on the container exit code?

The only podFailurePolicy options are to Ignore or FailJob right? If I understand correctly, FailJob is what the podFailurePolicy would be applying here, regardless of whether the application exit code indicates the job should restart or stay failed. 

The "reason" on the failure condition applied to the Job will be `PodFailurePolicy` regardless of which container exit code caused the podFailurePolicy to fail the job, so the JobSet controller won't know whether to RestartJob or FailJob. Am I missing something?

vsoch: The replicatedJob has multiple jobs under it, each generated from a template. If the top level replicatedJob has a global Failure or Restart policy, that means it doesn't vary on the level of the container. But what we need:
 - if JobA under the replicated Job (let's say it's one pod) has exit codes A-C, I want FailJob. I don't want that particular Job (pod) to restart, I know it's a doozie.
 - If JobB under the replicated Job  (also one pod) has exit codes D-F, that's a signal from the application to try again. I restart the Job. 
 
The above is mapping 1 container --> 1 pod --> 1 node, which is what we often do for our workloads because we need the full resources, in case that is relevant. So when I talk about 1 Job I am talking about one Pod and one container.

If I'm still not understanding then I'm having a hard time understanding the many levels of this proposal, and that's probably a bad thing from a user perspective.

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: Yes indeed! From JobSet perspective, I think we have the right API. The change needs to happen in the Job controller since it is the one dealing with the pods.

keps/262-ConfigurableFailurePolicy/README.md
vsoch: Another concern I have (that I've expressed before) is where to draw the line between what a workflow manager is supposed to do (that has a DAG, monitors statuses and makes decisions) vs. what Kubernetes is doing. We keep stepping into representing functionality that (traditionally) a workflow manager would handle and I'm not sure where the line is being drawn. In situations like these there should be enough information provided about any kind of status for the workload manager to make the right decision, and it's up to the abstraction to just provide that (but not to act like a workflow manager).

danielvegamyhre: Yeah, I understand your use case, but the just the limitations of the Job controller and PodFailurePolicies don't allow us to support this in JobSet yet. JobA and JobB would need distinct JobFailureReasons in order for JobSet controller to apply different actions to them. However, the Job controller currently only supports a single Job failure reason for podFaliurePolicies ([link](https://sourcegraph.com/github.com/kubernetes/kubernetes@f7f1ea6fa23ec0fc527ff10174bbe0dc3bec77da/-/blob/staging/src/k8s.io/api/batch/v1/types.go?L542:37-542:55)), which are the only mechanism the Job controller has for distinguishing between different application exit codes. So the underlying infrastructure in the Job API just doesn't exist yet to support this, unfortunately.

vsoch: okay understood - so the use case is valid, but not possible yet. And it's not that you don't see a case for `FailJob` or `RestartJob` but rather that there is no way to actually implement it. I just want to clarify that, because those are two different things (although they result in the same outcome that we can't add the cases here yet).

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: Created https://github.com/kubernetes/kubernetes/issues/122972 to track the issue

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: It's a bug in one of the k8s api packages we are using, haven't tracked it down yet

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: The most important thing is whether the API we are proposing now allows supporting this case in the future when the enhancement to the Job API is implemented: https://github.com/kubernetes/kubernetes/issues/122972

If we agree, then I think we can resolve this, right?

danielvegamyhre: Ok that makes sense, we can add FailJob and RestartJob proactively in anticipation of the Job API enhancements. Are we fairly sure https://github.com/kubernetes/kubernetes/issues/122972 will be approved and implemented? If so, I will work on adding some user stories for FailJob and RestartJob, based on the use case @vsoch described.

vsoch: I don't think we can be sure of anything, but if it's a valid and needed use case, we should continue advocating for it (I certainly will)!

kannon92: Even if 1229721 will be implemented (we would probably need a KEP in the next week) for 1.30. We still would have to support previous versions without this change so I think it would be best to implement without it and then maybe create a follow up issue once 1229721 is implemented and all supported releases of kube have this change.

ahg-g: > If so, I will work on adding some user stories for FailJob and RestartJob, based on the use case @vsoch described.

We can do that citing the open issue as a dependency.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Resolved this in the [comment chain](https://github.com/kubernetes-sigs/jobset/pull/381#discussion_r1465826615)

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Ok I added user story 4 which covers the FailJob and RestartJob use cases, and cited https://github.com/kubernetes/kubernetes/issues/122972 as a dependency.

examples/simple/configurable-failure-policy.yaml
```diff
@@ -0,0 +1,44 @@
...
+    # this failure policy will allow the JobSet to restart upon host maintenance, but fail immediately
+    # for non-retriable errors (we've broadly included all non-SIGTERM exit codes as non-retriable for
+    # this example).
+    - action: FailJobSet
+      onJobFailureReasons: 
+      - PodFailurePolicy
+    maxRestarts: 10
+  replicatedJobs:
+  - name: my-job
+    replicas: 1 # set to number of node pools
```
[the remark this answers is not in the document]
danielvegamyhre: Removed

keps/262-ConfigurableFailurePolicy/README.md
mimowo: I think separate KEP is preferable as we would like to go GA in 1.31 with pod failure policy, which is already incubating for a while, and IIUC the new API field should be protected by Alpha level feature gate first to work properly when the API servers are in skewed versions.

In the meanwhile you could also consider matching the `message` field in the condition added by pod failure policy. It contains details information about the exit code, or pod condition, see: https://github.com/kubernetes/kubernetes/blob/7340ce932bf3071c73ca371e4dfdd5ba00365668/pkg/controller/job/pod_failure_policy_test.go#L86

keps/262-ConfigurableFailurePolicy/README.md
mimowo: Daniel, thanks for explaining. This makes sense. 

I'm still wondering if there would also be useful concept of a FailedJob (failed for good, not retried even on JobSet restarts)? 

Recently in the core k8s Job we added the FailIndex action which marks some indexes as no-longer retriable, even if the BackoffLimitPerIndex would still allow for the restarts. The analogy I see is that with the FailJob action here some users may expect that using "FailJob" action they fail it for good. Still, this could be another action like "NonRestartableFailJob", added later if needed.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+  - `FailedIndexes`
+
+### Implmentation
+
+The core part of the implementation will be defining what specific mechanisms the JobSet controller uses to implement
+the behavior defined for each FailurePolicyAction type:
+
+1) `FailJobSet`: To fail the JobSet immediately without restarting, the controller updates the JobSet status to failed.
+
+2) `Ignore`: To "ignore" the failure (i.e., restart the JobSet without counting it against `MaxRestarts`), the controller
```
ahg-g, line 414: We need to use another annotation to track which jobs we need to restart since now we will not be incrementing it always, while restart-attempt will be a counter against maxRestarts.

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: We need to give users a way to control how many times to restart the replicatedJob or the job in those cases, they can't just restart forever.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+jobs of the target replicated job, **without incrementing the restart attempt annotation**. The jobs will then be 
+recreated via the normal reconciliation process.
+
+4) `RestartJob`: To restart a single child job without restarting the entire JobSet, the controller will delete that 
+particular child job, **without incrementing the restart attempt annotation**, and allow the normal reconciliation
+process to recreate it.
+
+5) `FailJob`: To leave a particular child job in a failed state without restarting it or restarting the JobSet, the
+controller will simply do nothing, taking no action on this job.
+
```
ahg-g, line 430: I would hold off implementing those three actions, we can add them later. For the first two, we need to think of how to define limits on the number of restarts, for the last one I think we need to mark the job as failed forever just like FailIndex in the Job API, and to do that, we need to use finalizers, which will complicate the implementation.

Perhaps move them to a future work section and leave the above points as open questions for discussion.

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: hmm, I was actually assuming that FailJob is similar to FailIndex, failed for good.

keps/262-ConfigurableFailurePolicy/README.md
alculquicondor: Maybe it should be `IgnoreAndRestartJobSet`?

It's not clear what happens with the jobs that didn't directly fail.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+```go
+
+// FailurePolicyAction defines the action the JobSet controller will take for
+// a given FailurePolicyRule.
+type FailurePolicyAction string
+
+const (
+  // Fail JobSet immediately, regardless of maxRestarts.
+  FailJobSet FailurePolicyAction = "FailJobSet"
```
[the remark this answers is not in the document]
alculquicondor: I prefer more explicit enums. `RestartJobSet` is compatible with your recommendations for `RestartJob` and so on.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+  // +kubebuilder:validation:Enum:=FailJobSet;Ignore;FailJob;RestartJob
+  Action FailurePolicyAction `json:"action"`
+  // The requirement on the job failure reasons. The requirement
+  // is satisfied if at least one reason matches the list.
+  // Each unique failure reason can only be associated with one FailurePolicyRule.
+  // An empty list applies the rule to any job failure reason.
+  // +kubebuilder:validation:UniqueItems:true
+  OnJobFailureReasons []string `json:"onJobFailureReasons"`
+  // TargetReplicatedJobs are the names of the replicated jobs the operator will apply to.
+  // A null or empty list will apply to all replicatedJobs.
```
alculquicondor, line 375:
```suggestion
  // An empty list will apply to all replicatedJobs.
```

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+  // +optional
+  // +listType=atomic
+  // +kubebuilder:validation:UniqueItems
+  TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
+}
+
+type FailurePolicy struct {
+  // MaxRestarts defines the limit on the number of JobSet restarts.
+  // A restart is achieved by recreating all active child jobs.
```
alculquicondor, line 384: What if I have a rule for `RestartJob` or `RestartReplicatedJob`? Do those count towards this counter?

Do you already store the number of failures in the JobSet status? If not, that should be part of the KEP too.

keps/262-ConfigurableFailurePolicy/README.md
alculquicondor: But these will be arbitrary strings in the future.

alculquicondor (commented):
Take my review comments as nits :)

keps/262-ConfigurableFailurePolicy/README.md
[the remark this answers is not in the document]
ahg-g: sounds good, lets continue to use explicit naming.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: To restart all child jobs without incrementing restart-attempt, can't we simply delete the child jobs without incrementing the restart-attempt annotation, and allow the reconciliation process to recreate them? 

If we use a separate annotation, the controller will still ultimately react to the change in that annotation by deleting all the child jobs without incrementing the restart-attempt annotation.

The separate annotation could be used to set a limit / ensure we don't restart forever, but then we are just building a parallel version of the existing maxRestarts logic.

What is the benefit of the separate annotation here?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+
+// FailurePolicyRule defines a FailurePolicyAction to be executed if a child job
+// fails due to a reason listed in OnJobFailureReasons.
+type FailurePolicyRule struct {
+  // The action to take if the rule is matched.
+  // +kubebuilder:validation:Enum:=FailJobSet;Ignore;FailJob;RestartJob
+  Action FailurePolicyAction `json:"action"`
+  // The requirement on the job failure reasons. The requirement
+  // is satisfied if at least one reason matches the list.
+  // Each unique failure reason can only be associated with one FailurePolicyRule.
```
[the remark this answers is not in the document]
danielvegamyhre: Updated

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: That makes sense, especially since 2 of the 3 rely on an upstream dependency which in the best case scenario, will only be in alpha in 1.30. 

Moved these to a "Future Work" section at the bottom, explaining the unresolved ambiguity and the upstream dependencies under review.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: done

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: > To restart all child jobs without incrementing restart-attempt, can't we simply delete the child jobs without incrementing the restart-attempt annotation, and allow the reconciliation process to recreate them?

How would you know if a child job was recreated or not? Remember that the jobset controller may restart in the middle of the reconciliation loop.

More concretely, in the following pseudo code which mimics what the jobset controller does:
```
for j in childJobs {
  recreate(j)
}
```
if the controller restarts before going through the list, the next reconcile call should continue from where we left off, not iterate over all the child jobs again.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -411,24 +273,12 @@ the behavior defined for each FailurePolicyAction type:
 
 1) `FailJobSet`: To fail the JobSet immediately without restarting, the controller updates the JobSet status to failed.
 
-2) `Ignore`: To "ignore" the failure (i.e., restart the JobSet without counting it against `MaxRestarts`), the controller
+2) `RestartJobSetAndIgnore`: To "ignore" the failure (i.e., restart the JobSet without counting it against `MaxRestarts`), the controller
```
ahg-g, line 276: `Ignore` is fine, I don't think we should change this one. @alculquicondor did you make this suggestion?

alculquicondor: I did, actually. With just `Ignore`, it's not immediately obvious that one Job failure causes the restart of all the others.

`Ignore` in Job just means restart that pod.

danielvegamyhre: Yeah, I tried to come up with a more explicit enum here because I agreed with Aldo that it was a bit vague. In fact, to me `Ignore` is a bit counter-intuitive, since I think "ignore" implies "do nothing," when it in fact means "restart the entire JobSet without incrementing restart attempts."

ahg-g: ok, I was assuming that we would continue to match rules down the order and only if we go through all the list without matching any, we restart jobset because it is the default behavior. 

But if the semantics is that we don't continue down the rules list, then we I agree with the name change, but perhaps we should say `RestartJobSetAndIgnoreMaxRestarts` :)

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,517 @@
...
+  TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
+}
+
+type FailurePolicy struct {
+  // MaxRestarts defines the limit on the number of JobSet restarts.
+  // A restart is achieved by recreating all active child jobs.
+  MaxRestarts int32 `json:"maxRestarts,omitempty"`
+  // List of failure policy rules for this JobSet.
+  // For a given Job failure, the rules will be evaluated in order,
+  // and only the first matching rule will be executed.
```
ahg-g, line 250: add that if no rule matches, then the default is to RestartJobSet

keps/262-ConfigurableFailurePolicy/README.md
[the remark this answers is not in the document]
ahg-g: @danielvegamyhre what about what about RestartJobSet action?

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: This is not yet addressed.

keps/262-ConfigurableFailurePolicy/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+  - "@vsoch"
+  - "@mimowo"
+approvers:
+  - "@ahg"
+
+see-also:
+  - "NA"
```
ahg-g, line 15: Point to PodFailurePolicy and your KEP to add the reason to it

keps/262-ConfigurableFailurePolicy/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+replaces:
+  - "NA"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
```
ahg-g, line 20: lets remove this and the milestone section at the end, we don't have feature gates, so this will be confusing to the reader.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Yeah sorry I had to shift my focus to something high priority that came up in another workstream. Thanks for the explanation, I will do another pass through the comments here today.

ahg-g: no worries, I thought it fell through the cracks

danielvegamyhre:
One additional idea I have is to add a `DefaultAction` field to the failure policy, in case the user doesn't want RestartJobSet to be the behavior for any failure not matching a failure policy rule. I think it would make the default restart/failure recovery behavior more explicitly clear in the API, rather than the user needing to read the docs/comments to understand what happens if they don't set one. 

If unset, it would default to `RestartJobSet` so our current default behavior is preserved.

```go
type FailurePolicy struct {
  // MaxRestarts defines the limit on the number of JobSet restarts.
  // List of failure policy rules for this JobSet.
  // For a given Job failure, the rules will be evaluated in order,
  // and only the first matching rule will be executed.
  Rules []FailurePolicyRule `json:"rules,omitempty"`
  // The default action that is executed for any Job failure that does
  // not match any of the failure policy rules.
  // If unset, this defaults to "RestartJobSet"
  DefaultAction FailurePolicyAction `json:"defaultAction,omitempty"`
}
```

What do you think?

alculquicondor:
I wouldn't add such thing.
It might make more sense for the user to add a Rule at the end that kind of matches anything

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,517 @@
...
+  // rule is executed.
+  // An empty list applies the rule to any job failure reason.
+  // +kubebuilder:validation:UniqueItems:true
+  OnJobFailureReasons []string `json:"onJobFailureReasons"`
+  // TargetReplicatedJobs are the names of the replicated jobs the operator applies to.
+  // An empty list will apply to all replicatedJobs.
+  // +optional
+  // +listType=atomic
+  // +kubebuilder:validation:UniqueItems
+  TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
```
kannon92, line 241: Could you add a story for this? 

It isn't quite clear to me how one would use TargetReplicatedJobs with `PodFailurePolicy`?

You would short circuit restarts if the replicated job matched a pod failure policy for instance? And other jobs you would continue as usually and restart while ignoring `PodFailurePolicy`?

danielvegamyhre:
> I wouldn't add such thing. It might make more sense for the user to add a Rule at the end that kind of matches anything

Hmm, can you elaborate on this a bit? That seems like a more complicated way of achieving the same thing.

alculquicondor:
Perhaps the last rule has empty OnConditionReason, so it matches any Failed condition.

It's easier to document because rules apply in order.

ahg-g:
Re DefaultAction: I would keep the api simple. The default action is RestartJobSet, and if the user wants to change that, they can add a catch all rule at the end.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Added user story 4 for this.

keps/262-ConfigurableFailurePolicy/kep.yaml
danielvegamyhre: Removed

keps/262-ConfigurableFailurePolicy/kep.yaml
danielvegamyhre: Added

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Updated this

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -213,18 +313,23 @@ Consider including folks who also work outside the SIG or subproject.
 type FailurePolicyAction string
 
 const (
-  // Fail JobSet immediately, regardless of maxRestarts.
+  // Fail the JobSet immediately, regardless of maxRestarts.
   FailJobSet FailurePolicyAction = "FailJobSet"
 
+  // Restart the JobSet if the "restart-attempts" annotation is less than
```
ahg-g, line 319: The annotation is an implementation detail. Just say "Restart the JobSet if the number of restarts didn't reach MaxRestarts" or something to that effect.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,516 @@
...
+      - simulations
+      onJobFailureReasons:
+      - PodFailurePolicy-ExitCode2
+    # If Job fails due to a pod failing with exit code 3, restart that Job.
+    - action: RestartJob
+      targetReplicatedJobs:
+      - simulations
+      onJobFailureReasons:
+      - PodFailurePolicy-ExitCode3
+    maxRestarts: 10
```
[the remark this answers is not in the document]
danielvegamyhre: Good point, updated the example in this story to have 1 podFailurePolicy rule and corresponding JobSet failure policy rule, to `RestartJob` for exit code 3, and a catch all failure policy rule to leave the Job failed otherwise.

keps/262-ConfigurableFailurePolicy/README.md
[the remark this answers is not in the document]
danielvegamyhre: Ah, I had only imagined use cases where restarting the whole JobSet would be left as the implicit default behavior (when no rules are matched), without an Action explicitly defined for it. However, I think you're right, there will definitely be cases where the user wants to explicitly define a failure policy rule to restart the whole JobSet upon X,Y conditions.

Added.

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Updated

keps/262-ConfigurableFailurePolicy/README.md
danielvegamyhre: Updated to `RestartJobSetAndIgnoreMaxRestarts`

ahg-g:
/lgtm
/hold

holding just in case, feel free to remove the hold

Thanks Daniel!

danielvegamyhre:
/hold cancel

Thanks for the feedback everyone!

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -0,0 +1,608 @@
...
+
+  // Don't count the failure against maxRestarts.
+  RestartJobSetAndIgnoreMaxRestarts FailurePolicyAction = "RestartJobSetAndIgnoreMaxRestarts"
+)
+
+// FailurePolicyRule defines a FailurePolicyAction to be executed if a child job
+// fails due to a reason listed in OnJobFailureReasons.
+type FailurePolicyRule struct {
+  // The action to take if the rule is matched.
+  // +kubebuilder:validation:Enum:=FailJobSet;RestartJobSetAndIgnoreMaxRestarts;FailJob;RestartJob
```
danielvegamyhre, line 331: note for implementation: remove FailJob;RestartJob from kubebuilder marker. We'll add them back after upstream changes are in place to support these use cases.

PR #389, opened 2024-01-27
Move exclusive placement annotation to ReplicatedJob template

### LINKED ISSUE #344: Update exclusive placement to be configurable per ReplicatedJob instead for the entire JobSet
Update exclusive placement to be configurable per ReplicatedJob instead for the entire JobSet

**Why is this needed**:
Users have reported use cases where they only want 1 replicated job to use exclusive placement, and don't want the others to use exclusive placement.

danielvegamyhre:
Fixes #344

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -612,11 +612,11 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 	}
 
 	// If this job should be exclusive per topology, configure the scheduling constraints accordingly.
-	if _, ok := js.Annotations[jobset.ExclusiveKey]; ok {
+	if _, ok := rjob.Template.Annotations[jobset.ExclusiveKey]; ok {
 		// If user has set the nodeSelectorStrategy annotation flag, add the job name label as a
 		// nodeSelector, and add a toleration for the no schedule taint.
 		// The node label and node taint must be added separately by a user/script.
-		if _, exists := js.Annotations[jobset.NodeSelectorStrategyKey]; exists {
+		if _, exists := rjob.Template.Annotations[jobset.NodeSelectorStrategyKey]; exists {
 			addNamespacedJobNodeSelector(job)
 			addTaintToleration(job)
 		}
@@ -675,8 +675,12 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
 
 	// Only set exclusive key label/annotation on jobs/pods if the parent JobSet
 	// is using exclusive placement.
-	if _, exists := js.Annotations[jobset.ExclusiveKey]; exists {
-		annotations[jobset.ExclusiveKey] = js.Annotations[jobset.ExclusiveKey]
+	if topologyDomain, exists := rjob.Template.Annotations[jobset.ExclusiveKey]; exists {
+		annotations[jobset.ExclusiveKey] = topologyDomain
+		// Add nodeSelectorStrategy annotation if it is being used.
+		if value, ok := rjob.Template.Annotations[jobset.NodeSelectorStrategyKey]; ok {
+			annotations[jobset.NodeSelectorStrategyKey] = value
+		}
 	}
 
 	obj.SetLabels(labels)
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -612,11 +612,15 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 	}
 
 	// If this job should be exclusive per topology, configure the scheduling constraints accordingly.
-	if _, ok := rjob.Template.Annotations[jobset.ExclusiveKey]; ok {
+	_, jsExclusivePlacement := js.Annotations[jobset.ExclusiveKey]
+	_, rjobExclusivePlacement := rjob.Template.Annotations[jobset.ExclusiveKey]
+	if jsExclusivePlacement || rjobExclusivePlacement {
 		// If user has set the nodeSelectorStrategy annotation flag, add the job name label as a
 		// nodeSelector, and add a toleration for the no schedule taint.
 		// The node label and node taint must be added separately by a user/script.
-		if _, exists := rjob.Template.Annotations[jobset.NodeSelectorStrategyKey]; exists {
+		_, jobsetUsingNodeSelectorStrategy := js.Annotations[jobset.NodeSelectorStrategyKey]
+		_, rjobUsingNodeSelectorStrategy := rjob.Template.Annotations[jobset.NodeSelectorStrategyKey]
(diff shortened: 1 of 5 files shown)


ahg-g:
For backwards compatibility, we shouldn't move, but rather we should support both. If set at the jobset level, then apply to all replicatedJobs regardless whether or not it is set at the replicatedJob.

pkg/controllers/jobset_controller.go
```diff
@@ -612,11 +612,11 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 	}
 
 	// If this job should be exclusive per topology, configure the scheduling constraints accordingly.
-	if _, ok := js.Annotations[jobset.ExclusiveKey]; ok {
+	if _, ok := rjob.Template.Annotations[jobset.ExclusiveKey]; ok {
```
ahg-g, line 615: To support setting it at both levels, here we should check both `rjob.Template.Annotations[jobset.ExclusiveKey]` and `js.Annotations[jobset.ExclusiveKey]`

pkg/controllers/jobset_controller.go
```diff
@@ -675,8 +675,12 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
...
 	// Only set exclusive key label/annotation on jobs/pods if the parent JobSet
 	// is using exclusive placement.
-	if _, exists := js.Annotations[jobset.ExclusiveKey]; exists {
-		annotations[jobset.ExclusiveKey] = js.Annotations[jobset.ExclusiveKey]
+	if topologyDomain, exists := rjob.Template.Annotations[jobset.ExclusiveKey]; exists {
+		annotations[jobset.ExclusiveKey] = topologyDomain
+		// Add nodeSelectorStrategy annotation if it is being used.
+		if value, ok := rjob.Template.Annotations[jobset.NodeSelectorStrategyKey]; ok {
+			annotations[jobset.NodeSelectorStrategyKey] = value
+		}
```
ahg-g, line 683: is this a bug?

ahg-g: if so, this is a reason to patch back.

pkg/controllers/jobset_controller.go
```diff
@@ -675,8 +675,12 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
 
 	// Only set exclusive key label/annotation on jobs/pods if the parent JobSet
 	// is using exclusive placement.
-	if _, exists := js.Annotations[jobset.ExclusiveKey]; exists {
-		annotations[jobset.ExclusiveKey] = js.Annotations[jobset.ExclusiveKey]
+	if topologyDomain, exists := rjob.Template.Annotations[jobset.ExclusiveKey]; exists {
```
ahg-g, line 678: same thing here, check both `js.Annotations[jobset.ExclusiveKey]` and `rjob.Template.Annotations[jobset.ExclusiveKey]` and

danielvegamyhre:
> For backwards compatibility, we shouldn't move, but rather we should support both. If set at the jobset level, then apply to all replicatedJobs regardless whether or not it is set at the replicatedJob.

I see, yeah I was planning on messaging you on Monday to discuss the best way to support backward compatibilty for users still using the JobSet level annotation. I think the most intuitive way to do so would be to document that if the JobSet level annotation is set, the ReplicatedJob level annotations will not matter, regardless if they are set or not. I will update the PR to support both.

pkg/controllers/jobset_controller.go
danielvegamyhre: No, previously we just were only setting the nodeSelectorStrategy annotation at the JobSet level. Now that we are supporting exclusive placement strategy at the ReplicatedJob level, the nodeSelectorStrategy annotation also needs to be able to be set on the ReplicatedJob Job template, since the user may be using the nodeSelectorStrategy implementation of exclusive placement for this particular ReplicatedJob, not the entire JobSet.

pkg/controllers/jobset_controller.go
```diff
@@ -612,11 +612,15 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) (*b
 	}
 
 	// If this job should be exclusive per topology, configure the scheduling constraints accordingly.
-	if _, ok := js.Annotations[jobset.ExclusiveKey]; ok {
+	_, jsExclusivePlacement := js.Annotations[jobset.ExclusiveKey]
+	_, rjobExclusivePlacement := rjob.Template.Annotations[jobset.ExclusiveKey]
+	if jsExclusivePlacement || rjobExclusivePlacement {
```
ahg-g, line 617: actually here, we just need to check `job.Annotations[jobset.ExclusiveKey]`, and below the same, use `job.Annotations[jobset.NodeSelectorStrategyKey]` since the job object was annotated above, right?

danielvegamyhre: Ah good point, that's simpler - done.

pkg/controllers/jobset_controller_test.go
```diff
@@ -899,14 +1068,15 @@ func TestCalculateReplicatedJobStatuses(t *testing.T) {
...
 type makeJobArgs struct {
-	jobSetName        string
-	replicatedJobName string
-	jobName           string
-	ns                string
-	replicas          int
-	jobIdx            int
-	restarts          int
-	topology          string
+	jobSetName           string
```
ahg-g, line 1071: We can do this as a followup, but we are using the wrapper pattern, we shouldn't need a makeJobArgs struct, we should be able to do something like

```
makeJob(name, namespace)
  .ReplicatedJobName(replicatedJobName)
  .Replicas(n)
```

pkg/controllers/jobset_controller.go
```diff
@@ -673,10 +669,21 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
...
+		// Check if we are using nodeSelectorStrategy implementation of exclusive placement at the JobSet level.
+		if value, ok := js.Annotations[jobset.NodeSelectorStrategyKey]; ok {
+			annotations[jobset.NodeSelectorStrategyKey] = value
+		}
+	}
+	// Check for ReplicatedJob level exclusive placement.
+	if topologyDomain, exists := rjob.Template.Annotations[jobset.ExclusiveKey]; exists {
+		annotations[jobset.ExclusiveKey] = topologyDomain
+		// Check if we are using nodeSelectorStrategy implementation of exclusive placement at the ReplicatedJob level.
+		if value, ok := rjob.Template.Annotations[jobset.NodeSelectorStrategyKey]; ok {
```
ahg-g, line 684: This labelAndAnnotate function is invoked for both the job template and the pod template. The job template inherits gets this already because the job is created out of `rjob.Template` anyways, and we don't need to set this annotation on the pod.

ahg-g: anyways, we can clean this up when we add the canonical placement policy API.

PR #393, opened 2024-02-01
Migrate from background to foreground cascading deletion policy

### LINKED ISSUE #392: Use foreground deletion when deleting child jobs

Use Foreground deletion when deleting child jobs at https://github.com/kubernetes-sigs/jobset/blob/42fd64765ddc4c308fb4742f384bce4c32d66e8b/pkg/controllers/jobset_controller.go#L517

**Why is this needed**:
The current background policy allows deleting the job object without waiting for its pods to be cleaned up first. This means JobSet will directly recreate a replacement job in the case of recreate failure policy, and so pods from the failed and new job may exist at the same time. In most cases, the new pods will likely target the same nodes of the failed job, and so the scheduling of those pods will block. 

The above behavior is especially problematic when exclusive placement policy is used. The major issue is that the followers pods will be rejected until the leader (index 0) is scheduled first, and the job controller will handle those repeated follower pods creation failures by placing their recreation retries in exponential backoff (starting from 1 sec to the max of 1min). 

By doing foreground deletion, we block the creation of the new job, and so the follower pods, until the pod they replace are cleaned up, and so the leader of the new job will likely schedule faster, and so avoiding too many creation failures of the follower pods that may lead to long backoffs.

danielvegamyhre:
Fixes #392 

This change involves:
- Using foreground deletion policy for deleting jobs
- Skipping deletion attempts for jobs with deletion timestamp already set
- Retrying job creations (with exponential backoff) which fail
- Refactoring jobset controller integration tests to remove foreground deletion finalizers for any tests which involve jobs being deleted

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strconv"
 	"sync"
+	"time"
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -28,6 +29,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -411,6 +402,28 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
...
+		return err
+	}
+
+	// Create the job with an exponential back off policy if it already exists.
+	return wait.ExponentialBackoff(backoff, func() (bool, error) {
+		if err := r.Create(ctx, job); err != nil {
+			return false, err
+		}
+		return true, nil
+	})
```
ahg-g, line 424: We shouldn't do the exponential backoff explicitly, we should just try to force another reconcile based on exponential backoff.

I actually found that controller-manager does that by default when we return an error: https://github.com/kubernetes-sigs/controller-runtime/blob/4000e996a202917ad7d40f02ed8a2079a9ce25e9/pkg/reconcile/reconcile.go#L95

so we shouldn't change anything here

danielvegamyhre: Ah interesting, I didn't know that. Done.

ahg-g:
/kind bug
/lgtm

marking as a bug since it caused a marked performance degradation

PR #400, opened 2024-02-09
add suspend field to printcolumn

### LINKED ISSUE #399: JobSet should display whether it is suspended or not.
Related: https://github.com/kubernetes/kubernetes/issues/123221

It would be nice to have a boolean value for suspend when you do `kubectl get jobset`.

kannon92:
Fixes #399

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -116,6 +116,7 @@ type ReplicatedJobStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
+// +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="JobSet suspended"
 // +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type=date,description="Time this JobSet was created"
 
 // JobSet is the Schema for the jobsets API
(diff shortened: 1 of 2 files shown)


kannon92:
Maybe https://github.com/kubernetes/test-infra/issues/31889 is related to these failures...

danielvegamyhre:
I think this would be useful to have. Looks like the current implementation results in a nil dereference error, though.

kannon92:
Are you bringing up the seg fault?

I am getting the same failure in the startup policy. I tested this and it worked with specifiy suspend: true, suspend:false and not specifying it.

kannon92:
I had a feeling something is up with the CI because I am getting the same error in https://github.com/kubernetes-sigs/jobset/pull/246.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -116,6 +116,7 @@ type ReplicatedJobStatus struct {
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
+// +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="Is this JobSet suspended"
```
ahg-g, line 119:
```suggestion
// +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="JobSet suspended"
```

ahg-g (commented):
Thanks Kevin, one suggestion.

api/jobset/v1alpha2/jobset_types.go
kannon92: updated.

PR #409, opened 2024-02-13
Validate longest pod name for jobset will not exceed 63 chars

### LINKED ISSUE #408: Validate longest generated pod name will not exceed 63 characters
**Problem**

In the JobSet webhook, we currently only [validate](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_webhook.go#L106-L109) the longest child job that will be created for this JobSet is DNS 1035 compliant, part of which is enforcing a [max length of 63 characters](https://github.com/kubernetes/kubernetes/blob/73f19e4c0162a2ee711c254b2af226b4e135d0e3/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L228). This constraint is enforced by the Job controller, so we reject the JobSet creation request if it would result in attempting to create child jobs with names longer than 63 characters.

However, we currently have no similar validation logic for pod names, which also have a max size of 63 characters. We previously thought there was validation in the Job API that would throw an error if the pod name it is attempting to create for a Job is longer than 63 characters, but it turns out that it actually only [validates the pod hostname](https://github.com/kubernetes/kubernetes/blob/3bf9477b521f21d374914dc7d0046e58259d2bf4/pkg/apis/batch/validation/validation.go#L147-L152) has a max length of 63 characters (which does NOT include the 5 character random suffix); for generating the pod name itself, it actually [truncates](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/indexed_job_utils.go#L501-L508) the job name to make room for the pod index and 5 character random suffix.

This led to an issue where a long JobSet name results in pod names longer than 63 characters, so the Job controller truncates them before creating the pods, which breaks JobSets using JobSet’s exclusive Job per topology feature. 

This is because the JobSet pod admission webhook blocks creation of all follower pods (pods with completion index != 0) until the leader pod (completion index 0) for that Job is scheduled, then it admits the pod creation requests and injects nodeSelectors into their specs, which enforce scheduling constraints ensuring the follower pods land in the same topology as their leader pod.

The way it does this depends on the leader pod name following the expected format: [jobset name]-[job index]-[pod index-[random 5 character suffix], and maintains an index mapping [pod name without random suffix] to the k8s corev1.Pod object. The pod webhook can then query the leader pod name without the random suffix to get the leader Pod object and check if it is scheduled, and have conditional logic based on this.

Therefore, since the leader pod name was truncated before creation and no longer follows the expected format, the pod webhook cannot look up the leader pod, and blocks follower pods from creation indefinitely, preventing the JobSet from running properly.

**Solution**:
- Short-term: Add JobSet validation which checks the longest pod name that will be generated will be less than or equal to 63 characters. If not, reject the JobSet creation and return a clear, user-facing error message.
Long-term:
- Long-term: Explore ways to look up the leader pod without depending on the pod name, since this is somewhat brittle and there are no upstream guarantees about the pod suffix length, among other things (see [here](https://github.com/kubernetes/kubernetes/pull/116430#issuecomment-1894547268)).

danielvegamyhre:
Fixes #408

### DIFF
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -17,6 +17,7 @@ import (
 	"errors"
 	"fmt"
 	"math"
+	"strconv"
 
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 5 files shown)


api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -105,10 +106,21 @@ func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
...
 		}
+		// Check that the generated pod names for the replicated job is DNS 1035 compliant.
+		isIndexedJob := rjob.Template.Spec.CompletionMode != nil && *rjob.Template.Spec.CompletionMode == batchv1.IndexedCompletion
+		if isIndexedJob && rjob.Template.Spec.Completions != nil {
+			maxJobIndex := strconv.Itoa(int(rjob.Replicas - 1))
+			maxPodIndex := strconv.Itoa(int(*rjob.Template.Spec.Completions - 1))
+			// Add 5 char suffix to the deterministic part of the pod name to validate the full pod name is compliant.
+			longestPodName := placement.GenPodName(js.Name, rjob.Name, maxJobIndex, maxPodIndex) + "-abcde"
+			for _, errMessage := range validation.IsDNS1035Label(longestPodName) {
+				allErrs = append(allErrs, fmt.Errorf(errMessage))
```
ahg-g, line 121: Will the error message contain the `-abcde`? If so, I am afraid that will be confusing and perhaps better to have our own error message.

danielvegamyhre: I'll test and confirm really quick

danielvegamyhre: Error message looks good to me:

```
$ k apply -f examples/simple/exclusive-placement.yaml 
Error from server (Forbidden): error when creating "examples/simple/exclusive-placement.yaml": admission webhook "vjobset.kb.io" denied the request: must be no more than 63 characters
```

PR #421, opened 2024-02-13
Improve error messages and logging in webhooks

danielvegamyhre:


api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -36,6 +37,24 @@ import (
...
+	// is longer than 63 characters.
+	dns1035MaxLengthExceededErrorMsg = "must be no more than 63 characters"
+
+	// Error message returned by JobSet validation if the generated child jobs
+	// will be longer than 63 characters.
+	jobNameTooLongErrorMsg = "JobSet name is too long, job names generated for this JobSet will exceed 63 characters"
+
+	// Error message returned by JobSet validation if the generated pod names
+	// will be longer than 63 characters.
+	podNameTooLongErrorMsg = "JobSet name is too long, job names generated for this JobSet will exceed 63 characters"
```
ahg-g, line 51:
```suggestion
	podNameTooLongErrorMsg = "JobSet name is too long, pod names generated for this JobSet will exceed 63 characters"
```

pkg/webhooks/pod_admission_webhook.go
```diff
@@ -87,7 +87,7 @@ func (p *podWebhook) leaderPodForFollower(ctx context.Context, pod *corev1.Pod)
 	log := ctrl.LoggerFrom(ctx)
 	leaderPodName, err := genLeaderPodName(pod)
 	if err != nil {
-		log.Error(err, "getting leader pod name for follower pod")
+		log.V(3).Info("leader pod (%s) for follower pod does not exist or has not been scheduled yet", leaderPodName)
```
ahg-g, line 90: this is not where we lookup the leader, this is just generating its name. The error for looking up the leader happens below at line 102

api/jobset/v1alpha2/jobset_webhook.go
danielvegamyhre: done

pkg/webhooks/pod_admission_webhook.go
```diff
@@ -99,7 +99,7 @@ func (p *podWebhook) leaderPodForFollower(ctx context.Context, pod *corev1.Pod)
 
 	// Validate there is only 1 leader pod for this job.
 	if len(podList.Items) != 1 {
-		return nil, fmt.Errorf("incorrect number of leader pods for this job (expected 1, got %d)", len(podList.Items))
+		return nil, fmt.Errorf("leader pod (%s) does not exist or has not been scheduled yet (expected 1 pod, got %d)", leaderPodName, len(podList.Items))
```
ahg-g, line 102: We wouldn't know here that it has not been scheduled, it will be above (the call to `leaderPodScheduled`); also leader pod doesn't exist is applicable only if the len is 0

pkg/webhooks/pod_admission_webhook.go
danielvegamyhre: Updated. Let me know if you have any thoughts on the new error message. I thought about changing the upstream caller which logs the full error to log at INFO level instead, since multiple users have reached out thinking this was the problem when it was actually a problem with their Job template or something similiar, simply because the message is logged at ERROR level.

Example:

`ERROR	admission	finding leader pod for follower	{"webhookGroup": "", "webhookKind": "Pod", "Pod": {"name":"","namespace":"default"}, "namespace": "default", "name": "", .....  "error": "incorrect number of leader pods for this job (expected 1, got 0)"}`



For now though, I just changed the error message to mention it is an expected, transient error and include the leader pod name.

pkg/webhooks/pod_admission_webhook.go
danielvegamyhre: sorry, it's late and my brain is getting tired 😅 

updated this error message, and added a log line in the leaderPodScheduled function.

PR #433, opened 2024-02-22
Validate follower pod owned by same Job as leader pod

danielvegamyhre:
This change validates the leader pod has same owner UID as the follower, to ensure they are part of the same Job.
This is necessary to handle a potential race condition between index updates and pod rescheduling during JobSet restarts.

1. A job failure occurs and the JobSet is restarted (deleting and recreating all jobs)
2. Leader pods may land on different node pools than they were originally scheduled on.
3. When the follower pods are recreated, and we look up the leader pod using the index which maps
[pod name without random suffix] -> corev1.Pod object, if this occurs before the index updates for the leader pod have been pushed to the controller, we may get a stale index entry and inject the the wrong nodeSelector, using the topology the leader pod was originally scheduled on before the
restart.

### DIFF
--- a/pkg/webhooks/pod_admission_webhook.go
+++ b/pkg/webhooks/pod_admission_webhook.go
@@ -83,7 +83,7 @@ func (p *podWebhook) leaderPodScheduled(ctx context.Context, pod *corev1.Pod) (b
 	}
 	scheduled := leaderPod.Spec.NodeName != ""
 	if !scheduled {
-		log.V(0).Info("leader pod %s is not yet scheduled", leaderPod.Name)
+		log.V(2).Info("leader pod %s is not yet scheduled", leaderPod.Name)
 	}
 	return scheduled, nil
 }
--- a/pkg/webhooks/pod_admission_webhook.go
+++ b/pkg/webhooks/pod_admission_webhook.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 
 	ctrl "sigs.k8s.io/controller-runtime"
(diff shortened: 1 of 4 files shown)


pkg/webhooks/pod_admission_webhook.go
```diff
@@ -82,7 +83,7 @@ func (p *podWebhook) leaderPodScheduled(ctx context.Context, pod *corev1.Pod) (b
 	}
 	scheduled := leaderPod.Spec.NodeName != ""
 	if !scheduled {
-		log.V(3).Info("leader pod %s is not yet scheduled", leaderPod.Name)
+		log.V(0).Info("leader pod %s is not yet scheduled", leaderPod.Name)
```
kannon92, line 86:
```suggestion
		log.V(2).Info("leader pod %s is not yet scheduled", leaderPod.Name)
```

Either we should use level 2 or just remove the V(0) and use log.Info(...)

What log level do we deploy jobset with? I think 2 is usually default logging anyway.

pkg/webhooks/pod_admission_webhook.go
```diff
@@ -131,3 +142,20 @@ func genLeaderPodName(pod *corev1.Pod) (string, error) {
 	leaderPodName := placement.GenPodName(jobSet, replicatedJob, jobIndex, "0")
 	return leaderPodName, nil
 }
+
+// validatePodsOnSameRestartAttempt returns an error if the leader pod and
+// follower pod are not owned by the same Job UID. Otherwise, it returns nil.
+func podsOwnedBySameJob(leaderPod, followerPod *corev1.Pod) error {
```
kannon92, line 148: Could you add some tests for this?

kannon92:
I am going to leave LGTM for @ahg-g on this one. I don't really have much context into this problem.

pkg/webhooks/pod_admission_webhook.go
ahg-g: yeah, I think we generally have been using 2 I think

pkg/webhooks/pod_admission_webhook.go
ahg-g: +1

danielvegamyhre: Added unit tests

pkg/webhooks/pod_admission_webhook.go
danielvegamyhre: Done

Issue #279, opened 2023-08-29
JobSet TTL to clean up completed workloads


danielvegamyhre:
We should look into having JobSet respect TTL so completed JobSets can be deleted from etcd and not take up space

https://kubernetes.io/docs/concepts/workloads/controllers/ttlafterfinished/

kannon92:
So I am a bit confused by this.

Are you saying we should have a field in JobSet that will purge JobSet from etcd giving a TTL?

Or are you wanting someone to test if one sets TTL in a Job, will the Job finish and then get deleted.  And even if the JobSet was marked as success (when Jobs all finish), the JobSet will recreate the finished job because it was deleted?

kannon92:
So individual jobs can use ttl without any issue.

```
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata:
  name: simple-with-ttl
spec:
  replicatedJobs:
  - name: leader
    replicas: 1
    template:
      spec:
        # Set backoff limit to 0 so job will immediately fail if any pod fails.
        backoffLimit: 0
        ttlSecondsAfterFinished: 60
        completions: 1
        parallelism: 1
        template:
          spec:
            containers:
            - name: leader
              image: bash:latest
              command:
              - bash
              - -xc
              - |
                sleep 100
  - name: workers
    replicas: 1
    template:
      spec:
        backoffLimit: 0
        ttlSecondsAfterFinished: 60
        completions: 2
        parallelism: 2
        template:
          spec:
            containers:
            - name: worker
              image: bash:latest
              command:
              - bash
              - -xc
              - |
                sleep 100
~                            
```

Jobs are cleaned up and deleted from etcd but JobSet will still mark this was successful.  

I think we could add a ttl for JobSet as the JobSet will not be deleted.  IMO this would be a nice feature for any CRD though..

kannon92:
Yea I didn’t see jobs recreated in my case.

danielvegamyhre:
Sounds good, thanks for taking a look at this.

danielvegamyhre:
I think we actually need a TTL in JobSet to improve the user experience with Kueue integration. Right now, users have to manually delete all Workloads since JobSet has no TTL, so they either need to do it manually or have some process doing it periodically. If JobSet had a TTL, the JobSet + Workload object would be cleaned up automatically.

kannon92:
Can you clean up the description? I think this could be a good first issue if we clear the objectives.

[...]

PR #374, opened 2023-12-29
feat: add support for ttl after finished controller for cleaning up finished jobsets

dejanzele:
**CHANGELOG**
* Added `TTLSecondsAfterFinished *int32` in `JobSetSpec` with validation for minimal value 0 and default value 0
* Added TTLAfterFinished controller which watches for finished JobSets with configured TTL, similarly how it is implemented in Kubernetes for Job controller
* Added unit and integration tests
* Edited Dockerfile to include client-go folder in build container

This PR closes https://github.com/kubernetes-sigs/jobset/issues/279

PR #443, opened 2024-03-10
feat: add support for ttl cleanup for finished jobsets

dejanzele:
CHANGELOG

* Added `TTLSecondsAfterFinished *int32` in `JobSetSpec` with validation for minimal value 0
* Run `make manifests` and `make generate` to update autogenerated code
* Updated JobSet Controller to handle cleanup after TTL expires
* Added unit and integration tests

This PR closes https://github.com/kubernetes-sigs/jobset/issues/279


dejanzele:
Reference to the closed PR with original comments - https://github.com/kubernetes-sigs/jobset/pull/374

Should I add the metrics here also (I added the `jobset_deletion_duration_seconds` in the closed PR)?

pkg/controllers/jobset_controller.go
```diff
@@ -114,12 +119,30 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	}
 
-	// If JobSet is already completed or failed, clean up active child jobs.
+	// If JobSet is already completed or failed, clean up active child jobs and requeue if TTLSecondsAfterFinished is set.
 	if jobSetFinished(&js) {
 		if err := r.deleteJobs(ctx, ownedJobs.active); err != nil {
 			log.Error(err, "deleting jobs")
 			return ctrl.Result{}, err
 		}
+		if js.Spec.TTLSecondsAfterFinished != nil {
```
danielvegamyhre, line 128: If TTLSecondsAfterFinished is set, we want to wait to delete child Jobs until after the TTL has expired. This gives the user time to inspect/debug the pods before everything is deleted, without having to formulate queries against cloud (or on-prem) log storage.

pkg/controllers/jobset_controller.go
```diff
@@ -913,3 +936,88 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
 	}
 	return jobset.ReplicatedJobStatus{}
 }
+
+func (r *JobSetReconciler) deleteJobSet(ctx context.Context, js *jobset.JobSet) error {
+	log := ctrl.LoggerFrom(ctx)
+
+	policy := metav1.DeletePropagationForeground
+	options := []client.DeleteOption{client.PropagationPolicy(policy), client.Preconditions{UID: &js.UID}}
```
danielvegamyhre, line 944: Do we need the jobset UID as a precondition here, since the UID is already part of the jobset object being passed to `Delete()`?

Please correct me if I'm wrong but I thought we would only need to pass the UID as a precondition if we were deleting by name or by namespaced-name or something that may not be unique over time (e.g., same JobSet rapidly created/deleted could lead to race condition).

pkg/controllers/jobset_controller.go
```diff
@@ -913,3 +936,88 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+	log.V(2).Info("Cleaning up JobSet", "jobset", klog.KObj(js))
+
+	return r.Client.Delete(ctx, js, options...)
```
danielvegamyhre, line 947:
```suggestion
	return r.Delete(ctx, js, options...)
```

pkg/controllers/jobset_controller.go
```diff
@@ -913,3 +936,88 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+}
+
+// checkIfTTLExpired checks whether a given JobSet's TTL has expired.
+func (r *JobSetReconciler) checkIfTTLExpired(ctx context.Context, jobSet *jobset.JobSet) (expiredAt *time.Time, err error) {
```
danielvegamyhre, line 951: I don't think named return parameters are necessary here, we can just use unnamed return params

pkg/controllers/jobset_controller.go
```diff
@@ -913,3 +936,88 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+	}
+
+	now := r.clock.Now()
+	remaining, expiredAt, err := timeLeft(log, jobSet, &now)
+	if err != nil {
+		return nil, err
+	}
+
+	// TTL has expired
+	if *remaining <= 0 {
```
danielvegamyhre, line 966: Can we add a nil check before dereferencing here? It may technically be safe now but I think it is good practice just in case upstream code changes in the future, for example what if the `timeLeft()` function changes and now we encounter a nil value here?

pkg/controllers/jobset_controller.go
```diff
@@ -913,3 +936,88 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+}
+
+// timeLeft returns the time left until the JobSet's TTL expires and the time when it will expire.
+func timeLeft(log logr.Logger, js *jobset.JobSet, since *time.Time) (*time.Duration, *time.Time, error) {
+	finishAt, expireAt, err := getJobSetFinishAndExpireTime(js)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	if finishAt.After(*since) {
```
danielvegamyhre, line 980: We should nil check the pointer before proceeding with the rest of the function which depends on dereferencing it. 

I think each function as a standalone unit of code should be safe from nil dereference errors, and not depend on safety checks elsewhere in the code base, since if we ever change those safety checks it may be easy to miss one place where we depended on them and the controller crashes.

pkg/controllers/jobset_controller.go
```diff
@@ -913,3 +936,88 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+
+// jobSetFinishTime takes an already finished JobSet and returns the time it finishes.
+func jobSetFinishTime(finishedJobSet *jobset.JobSet) (metav1.Time, error) {
+	for _, c := range finishedJobSet.Status.Conditions {
+		if (c.Type == string(jobset.JobSetCompleted) || c.Type == string(jobset.JobSetFailed)) && c.Status == metav1.ConditionTrue {
+			finishAt := c.LastTransitionTime
+			if finishAt.IsZero() {
+				return metav1.Time{}, fmt.Errorf("unable to find the time when the JobSet %s/%s finished", finishedJobSet.Namespace, finishedJobSet.Name)
+			}
+			return c.LastTransitionTime, nil
```
danielvegamyhre, line 1008:
```suggestion
			return finishAt, nil
```

danielvegamyhre (commented):
Thanks for working on this @dejanzele! I did a first pass reviewing the PR, once those comments are addressed I'll take another look.

danielvegamyhre:
@dejanzele are you still working on this?

dejanzele:
> @dejanzele are you still working on this?

yes, I was off due to KubeCon, now I can resolve your comments

pkg/controllers/jobset_controller.go
dejanzele: I guess not, I will remove it

pkg/controllers/jobset_controller.go
dejanzele: agree

pkg/controllers/jobset_controller.go
dejanzele: agree

pkg/controllers/jobset_controller.go
dejanzele: added safety checks above

pkg/controllers/jobset_controller.go
dejanzele: makes sense, I reordered the job deletion to be below

pkg/controllers/jobset_controller.go
```diff
@@ -916,3 +939,98 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+		return nil, nil, err
+	}
+	// The following 2 checks do sanity checking for nil pointers in case of changes to the above function.
+	// This logic should never be executed.
+	if since == nil {
+		log.V(2).Info("Warning: Received invalid since time. JobSet cleanup will be deferred.")
+		return nil, nil, nil
+	}
+	if finishAt == nil || expireAt == nil {
+		log.V(2).Info("Warning: Received invalid finish or expire time. JobSet cleanup will be deferred.")
```
danielvegamyhre, line 989: I think these 3 nil checks be combined into 1 right? Any reason you kept them separate? If not let's combine them.

pkg/controllers/jobset_controller.go
```diff
@@ -916,3 +939,98 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+	// TTL has expired
+	if remaining != nil && *remaining <= 0 {
+		return expiredAt, nil
+	}
+
+	return nil, nil
+}
+
+// timeLeft returns the time left until the JobSet's TTL expires and the time when it will expire.
+func timeLeft(log logr.Logger, js *jobset.JobSet, since *time.Time) (*time.Duration, *time.Time, error) {
```
danielvegamyhre, line 977: For consistency with the rest of the code base, instead of passing `log` as a param, let's pass in `ctx` and do `log := ctrl.LoggerFrom(ctx)`

pkg/controllers/jobset_controller.go
```diff
@@ -916,3 +939,98 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
...
+	}
+
+	now := r.clock.Now()
+	remaining, expiredAt, err := timeLeft(log, jobSet, &now)
+	if err != nil {
+		return nil, err
+	}
+
+	// TTL has expired
+	if remaining != nil && *remaining <= 0 {
```
danielvegamyhre, line 969: Let's nil check `remaining` before dereferencing. I see it won't be nil from the function below, but let's not rely on that logic always being the same.

pkg/controllers/jobset_controller.go
danielvegamyhre, line 977: Can we rename `since` to something like `now` or `currTime` since it holds the current time? The variable name `since` makes the code a bit more confusing for me to read

pkg/controllers/jobset_controller.go
```diff
@@ -114,8 +119,26 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
-	// If JobSet is already completed or failed, clean up active child jobs.
+	// If JobSet is already completed or failed, clean up active child jobs and requeue if TTLSecondsAfterFinished is set.
 	if jobSetFinished(&js) {
+		if js.Spec.TTLSecondsAfterFinished != nil {
+			expiredAt, err := r.checkIfTTLExpired(ctx, &js)
```
danielvegamyhre, line 125: Any reason why this function returns a *time.Time instead of a bool? Seems like we just use it as a boolean flag basically

pkg/controllers/jobset_controller.go
dejanzele: We can, I was wondering should have different logs for each case, but that might be overkill for a scenario which we won't ever encounter most probably

pkg/controllers/jobset_controller.go
dejanzele: makes sense

pkg/controllers/jobset_controller.go
dejanzele: It is being used to add a field to the log below when did the TTL expire, if that info isn't specially important to be logged, we can easily switch to bool

danielvegamyhre: Ah ok I missed that, sounds good, the current implementation is fine.

danielvegamyhre:
@dejanzele can you add an example JobSet yaml to the examples/simple/ folder which utilizes this feature?

ahg-g:
/hold

I will take a quick look

pkg/controllers/jobset_controller.go
```diff
@@ -110,8 +114,26 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+			// otherwise, we requeue it for the remaining TTL duration.
+			if expired {
+				log.V(5).Info("JobSet TTL expired")
+				if err := r.deleteJobSet(ctx, &js); err != nil {
+					log.Error(err, "deleting jobset")
+					return ctrl.Result{}, err
+				}
+			} else {
+				return ctrl.Result{RequeueAfter: requeueJobSetAfter(&js)}, nil
+			}
```
ahg-g, line 135: can we please more this logic to a function named `executeTTLAfterFinishedPolicy` similar to the pattern we have for failure and success policies below?

pkg/controllers/jobset_controller_test.go
```diff
@@ -1219,3 +1308,55 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
 		PodAnnotations(annotations)
 	return jobWrapper
 }
+
+func newJobSet(completionTime, failedTime metav1.Time, ttl *int32) *jobset.JobSet {
```
ahg-g, line 1312: please avoid adding such functions and use the wrappers functions from https://github.com/kubernetes-sigs/jobset/blob/main/pkg/util/testing/wrappers.go (please add missing wrappers if any)

pkg/controllers/jobset_controller.go
```diff
@@ -946,3 +968,91 @@ func managedByExternalController(js jobset.JobSet) *string {
 	}
 	return nil
 }
+
+func (r *JobSetReconciler) deleteJobSet(ctx context.Context, js *jobset.JobSet) error {
```
ahg-g, line 972: can we please move all those functions to a `ttl_after_finished.go` file with its own unit tests file as well? like `startup_policy.go`?

this and the function below can receive the client and the clock as parameters instead of having the functions as members of jobsetReconciler.

test/integration/controller/jobset_controller_test.go
```diff
@@ -1499,6 +1575,26 @@ func checkNoActiveJobs(js *jobset.JobSet, numFinishedJobs int) {
 	}, timeout, interval).Should(gomega.Equal(true))
 }
 
+// Check that the jobs' deletion timestamp is set or not set for the provided number of jobs.
+func checkJobsDeletionTimestamp(js *jobset.JobSet, set bool, numJobs int) {
```
ahg-g, line 1579: can we please move this to https://github.com/kubernetes-sigs/jobset/blob/main/test/util/util.go

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,82 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			gomega.Expect(k8sClient.List(ctx, &jobList, client.InNamespace(js2.Namespace))).Should(gomega.Succeed())
+			gomega.Expect(len(jobList.Items)).To(gomega.Equal(testutil.NumExpectedJobs(js2)))
+			failJob(&jobList.Items[0])
+
+			// Verify jobset is marked as completed.
+			testutil.JobSetCompleted(ctx, k8sClient, js1, timeout)
+			testutil.JobSetFailed(ctx, k8sClient, js2, timeout)
+
+			// Verify active jobs have not been deleted if ttl has not passed.
+			checkJobsDeletionTimestamp(js2, false, testutil.NumExpectedJobs(js2))
```
ahg-g, line 1278: how do we know that at this point TTL didn't pass yet? I don't think we can...

test/integration/controller/jobset_controller_test.go
ahg-g: I also suggest to use the format: `expectJobsDeletionTimestampSet` and `expectJobsDeletionTimestampNotSet` instead of passing a flag.

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,82 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+
+			// Verify jobset has not been deleted if ttl has not passed.
+			ginkgo.By("checking that jobset has not been deleted before configured seconds pass")
+			var fresh1, fresh2 jobset.JobSet
+			gomega.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(js1), &fresh1)).To(gomega.Succeed())
+			gomega.Expect(fresh1.DeletionTimestamp).To(gomega.BeNil())
+			gomega.Expect(k8sClient.Get(ctx, client.ObjectKeyFromObject(js2), &fresh2)).To(gomega.Succeed())
+			gomega.Expect(fresh2.DeletionTimestamp).To(gomega.BeNil())
```
ahg-g, line 1286: ditto, how can we make sure that at this point the time has not passed yet?

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,82 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+
+			// Verify active jobs have been deleted after ttl has passed.
+			checkJobsDeletionTimestamp(js2, true, testutil.NumExpectedJobs(js2)-1)
```
ahg-g, line 1289: this one and the check below make sense since eventually the TTL will pass.

pkg/controllers/jobset_controller.go
dejanzele: Should `deleteJobSet` function stay in the `jobset_controller.go`? It is a generic CRUD function, all of the other CRUD functions are in the same file

test/integration/controller/jobset_controller_test.go
dejanzele: Yeah, I don't think there is a reliable way to check, unless we increase the TTL, which then causes different issues. Should I just remove the check?

pkg/controllers/jobset_controller.go
ahg-g: It is generic, but only ttl after finished uses it, and likely will continue to be the only one

test/integration/controller/jobset_controller_test.go
ahg-g: yes, otherwise the test will flake

pkg/controllers/jobset_controller.go
danielvegamyhre: +1

dejanzele:
@danielvegamyhre yes, I plan on submiting a new commit today

test/integration/controller/jobset_controller_test.go
dejanzele: agreed, it can flake

pkg/controllers/ttl_after_finished_test.go
```diff
@@ -0,0 +1,189 @@
...
+		},
+		{
+			name:   "jobset completed now, nil since",
+			jobset: testutils.MakeJobSet(jobSetName, ns).TTLSecondsAfterFinished(0).CompletedCondition(now).Obj(),
+			since:  nil,
+		},
+		{
+			name:             "jobset completed now, 0s TTL",
+			jobset:           testutils.MakeJobSet(jobSetName, ns).TTLSecondsAfterFinished(0).CompletedCondition(now).Obj(),
+			since:            &now.Time,
```
danielvegamyhre, line 134: In the `timeLeft()` function being tested here, we changed the `since` parameter to be called `now` (since the purpose of the variable is to hold the current time). We should make the corresponding change here, and call this `now` or `currTime`.

danielvegamyhre (commented):
LGTM, one last minor comment.

danielvegamyhre:
/lgtm

I'll leave approval to @ahg-g so he can confirm his comments have been resolved.

pkg/controllers/ttl_after_finished.go
```diff
@@ -0,0 +1,130 @@
...
+
+	if js.Spec.TTLSecondsAfterFinished != nil {
+		expired, err := checkIfTTLExpired(ctx, r.clock, js)
+		if err != nil {
+			return 0, fmt.Errorf("error checking if ttl expired: %w", err)
+		}
+		// if expired is true, that means the TTL has expired, and we should delete the JobSet
+		// otherwise, we requeue it for the remaining TTL duration.
+		if expired {
+			log.V(5).Info("JobSet TTL expired")
```
ahg-g, line 33:
```suggestion
			log.V(5).Info("JobSet TTL expired, deleting")
```

ahg-g: also please increase the log level to 2, I think this is useful info

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				gomega.Expect(testutil.DeleteNamespace(ctx, k8sClient, ns1)).To(gomega.Succeed())
+				gomega.Expect(testutil.DeleteNamespace(ctx, k8sClient, ns2)).To(gomega.Succeed())
+			}()
+			// Create JobSet.
+			js1 := testJobSet(ns1).TTLSecondsAfterFinished(2).Obj()
+			js2 := testJobSet(ns2).TTLSecondsAfterFinished(2).Obj()
+
+			// Verify jobsets created successfully.
+			ginkgo.By(fmt.Sprintf("creating jobSet %s/%s", js1.Name, js1.Namespace))
+			gomega.Eventually(k8sClient.Create(ctx, js1), timeout, interval).Should(gomega.Succeed())
```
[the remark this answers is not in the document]
dejanzele: I see it being wrapped in a couple of places so I just followed the pattern:
* https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L84
* https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L160
* https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L1153

Should I revert to `Expect()`?

pkg/controllers/ttl_after_finished.go
```diff
@@ -0,0 +1,130 @@
...
+	log := ctrl.LoggerFrom(ctx)
+
+	finishAt, expireAt, err := getJobSetFinishAndExpireTime(js)
+	if err != nil {
+		return nil, err
+	}
+	// The following 2 checks do sanity checking for nil pointers in case of changes to the above function.
+	// This logic should never be executed.
+	if now == nil || finishAt == nil || expireAt == nil {
+		log.V(2).Info("Warning: Calculated invalid expiration time. JobSet cleanup will be deferred.")
```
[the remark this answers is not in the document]
dejanzele: Should I log it as error or return an error?

I don't know should we fail Reconciliation if it goes into this loop, in theory we should never execute this logic.

pkg/controllers/ttl_after_finished.go
```diff
@@ -0,0 +1,130 @@
...
+
+	// This should never happen if the JobSets have finished
+	return metav1.Time{}, fmt.Errorf("unable to find the status of the finished JobSet %s/%s", finishedJobSet.Namespace, finishedJobSet.Name)
+}
+
+// requeueJobSetAfter returns the duration after which the JobSet should be requeued if TTLSecondsAfterFinished is set, otherwise returns 0.
+func requeueJobSetAfter(js *jobset.JobSet) time.Duration {
+	var requeueAfter time.Duration = 0
+	if js.Spec.TTLSecondsAfterFinished != nil {
+		requeueAfter = time.Duration(*js.Spec.TTLSecondsAfterFinished) * time.Second
```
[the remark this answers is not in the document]
dejanzele: makes sense

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			gomega.Expect(k8sClient.Create(ctx, ns1)).To(gomega.Succeed())
+			gomega.Expect(k8sClient.Create(ctx, ns2)).To(gomega.Succeed())
+
+			defer func() {
+				gomega.Expect(testutil.DeleteNamespace(ctx, k8sClient, ns1)).To(gomega.Succeed())
+				gomega.Expect(testutil.DeleteNamespace(ctx, k8sClient, ns2)).To(gomega.Succeed())
+			}()
+			// Create JobSet.
+			js1 := testJobSet(ns1).TTLSecondsAfterFinished(2).Obj()
+			js2 := testJobSet(ns2).TTLSecondsAfterFinished(2).Obj()
```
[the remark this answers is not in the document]
dejanzele: agreed

pkg/controllers/ttl_after_finished.go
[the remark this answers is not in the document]
ahg-g: I would actually return an error here

test/integration/controller/jobset_controller_test.go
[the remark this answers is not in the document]
ahg-g: those are not necessary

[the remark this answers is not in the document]
ahg-g: yes, pls revert them to expect

pkg/controllers/ttl_after_finished.go
```diff
@@ -0,0 +1,130 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+)
+
+// executeTTLAfterFinishedPolicy checks if the JobSet has a TTLSecondsAfterFinished set.
+// If the JobSet has expired, it deletes the JobSet.
+// If the JobSet has not expired, it returns the time after which the JobSet should be requeued.
+// If the JobSet does not have a TTLSecondsAfterFinished set, it returns 0.
+func (r *JobSetReconciler) executeTTLAfterFinishedPolicy(ctx context.Context, js *jobset.JobSet) (time.Duration, error) {
```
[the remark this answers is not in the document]
dejanzele: I also added unit tests using the fake client

dejanzele:
@danielvegamyhre @ahg-g is there a chance this feature makes it into the v0.5.0 release?

kannon92:
> @danielvegamyhre @ahg-g is there a chance this feature makes it into the v0.5.0 release?

Lightly following this, we called this out in the release tracking issue. And it seems that the PR is almost there. So I think its possible. its the last item for the release.

danielvegamyhre:
> @danielvegamyhre @ahg-g is there a chance this feature makes it into the v0.5.0 release?

I think so, we plan to cut the release tomorrow but we can wait an extra day or two if necessary.

dejanzele:
Cool, I think all comments are addressed

ahg-g:

/lgtm
/approve

Thanks!

dejanzele:
@ahg-g I think you need to unhold also

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			gomega.Expect(k8sClient.List(ctx, &jobList, client.InNamespace(js2.Namespace))).Should(gomega.Succeed())
+			gomega.Expect(len(jobList.Items)).To(gomega.Equal(testutil.NumExpectedJobs(js2)))
+			completeAllJobs(&jobList)
+
+			// Verify jobset is marked as completed.
+			testutil.JobSetFailed(ctx, k8sClient, js1, timeout)
+			testutil.JobSetCompleted(ctx, k8sClient, js2, timeout)
+
+			// Verify active jobs have been deleted after ttl has passed.
+			testutil.ExpectJobsDeletionTimestamp(ctx, k8sClient, js1, testutil.NumExpectedJobs(js1)-1, timeout)
```
ahg-g, line 1278: Why are we just checking for the timestamp instead of checking that the jobs are actually deleted?

test/integration/controller/jobset_controller_test.go
```diff
@@ -1225,6 +1225,72 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			var fresh1, fresh2 jobset.JobSet
+			ginkgo.By("checking that ttl after finished controller deletes only the jobset with ttl set after configured seconds pass")
+			gomega.Eventually(func() bool {
+				if err := k8sClient.Get(ctx, client.ObjectKeyFromObject(js1), &fresh1); err != nil {
+					return false
+				}
+				if err := k8sClient.Get(ctx, client.ObjectKeyFromObject(js2), &fresh2); err != nil {
+					return false
+				}
+				return !fresh1.DeletionTimestamp.IsZero() && fresh2.DeletionTimestamp.IsZero()
```
ahg-g, line 1290: same here, why are we only checking that the deletion timestamp is set instead of checking that the jobset is notFound?

PR #448, opened 2024-03-13
Emit Job creation failed event

### LINKED ISSUE #447: Emit event when Job creation fails
Emit event when Job creation fails

**Why is this needed**:
To propagate the Job creation failure up to be more visible to the user, so they don't need to dig through the logs to know that their Job creation failed.

danielvegamyhre:
Fixes #447 

I opted to only emit 1 event if any Job creation fails, rather than emit a separate event for each Job failure (since there could be many).

While unconventional, I included the error message in the event message since the intent is to allow the user to quickly see why Jobs aren't being created without digging through logs.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -475,6 +475,9 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 	}
 	allErrs := errors.Join(finalErrs...)
 	if allErrs != nil {
+		// Emit event to propagate the Job creation failures up to be more visible to the user.
+		// TODO(#422): Investigate ways to validate Job templates at JobSet validation time.
+		r.Record.Eventf(js, corev1.EventTypeWarning, "JobCreationFailed", "Job creation(s) failed with error: %s", allErrs)
 		return allErrs
 	}
 	// Skip emitting a condition for StartupPolicy if JobSet is suspended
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -475,6 +475,9 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 	}
 	allErrs := errors.Join(finalErrs...)
 	if allErrs != nil {
+		// Emit event to propagate the Job creation failures up to be more visible to the user.
+		// TODO(#422): Investigate ways to validate Job templates at JobSet validation time.
+		r.Record.Eventf(js, corev1.EventTypeWarning, "JobCreationFailed", "Job creation(s) failed with error: %s", allErrs)
```
kannon92, line 480: Could you add a bit more information to this event? If we had multiple jobs, which job creation failed? I guess validation would be in logs or kubetctl somewhere so they could dig into this.

danielvegamyhre: I think the `allErrs` variable should have these details

kannon92: Are we sure that events actually shows the errors in all its entirity?

danielvegamyhre: To be safe, I wrapped the `err` returned by `r.Create(...)` in a custom error which includes the Job name at the beginning of the error message, let me know what you think of this

kannon92 (commented):
/lgtm
/hold

maybe @ahg-g wants to weigh in.

PR #452, opened 2024-03-13
Add features overview to README

danielvegamyhre:
Add features overview to README


README.md
```diff
@@ -13,19 +13,21 @@ Take a look at the [concepts](/docs/concepts/README.md) page for a brief descrip
...
 
-Read the [installation guide](/docs/setup/install.md) to learn more.
+- **Exclusive Placement Per Topology Domain**: JobSet includes an [annotation](https://github.com/kubernetes-sigs/jobset/blob/1ae6c0c039c21d29083de38ae70d13c2c8ec613f/examples/simple/exclusive-placement.yaml#L6) which can be set by the user, specifying that there should be a 1:1 mapping between child job and a particular topology domain, such as a datacenter rack or zone. This means that all the pods belonging to a child job will be colocated in the same topology domain, while pods from other jobs will not be allowed to run within this domain. This gives the child job exclusive access to computer resources in this domain.
```
danielvegamyhre, line 26: Question for reviewers: I think this feature will make little sense to users without a concrete use case, but the only one I can think of is TPU Multislice training, and since TPUs are specific to Google I didn't include it here. If anyone has any a suggestion for a concrete use case here I would appreciate it. I am happy to include TPU multislice training as well, based on feedback.

README.md
```diff
@@ -13,19 +13,21 @@ Take a look at the [concepts](/docs/concepts/README.md) page for a brief descrip
...
 
-```shell
-kubectl apply --server-side -f https://github.com/kubernetes-sigs/jobset/releases/download/v0.4.0/manifests.yaml
-```
+- **Configurable success policies**: JobSet has [configurable success policies](https://github.com/kubernetes-sigs/jobset/blob/1ae6c0c039c21d29083de38ae70d13c2c8ec613f/examples/simple/success-policy.yaml) which target specific ReplicatedJobs, with operators to target `Any` or `All` of their child jobs. For example, you can configure the JobSet to be marked complete if and only if all pods that are part of the “worker” ReplicatedJob are completed. This enables users to use their compute resources more efficiently, allowing a workload to be declared successful and release the resources for the next workload more quickly.
```
kannon92, line 22: Should we use a perm link? 

A user could see a different example than what is in the repo.

README.md
kannon92: maybe @vsoch has some ideas of a general example?

README.md
danielvegamyhre: The configurable success example link is a permalink to a specific commit.

danielvegamyhre: I see, so you think for these examples we should link to main?

danielvegamyhre: Updated links to use v0.4.0 for example yamls for now.

We can think about having examples directly in the docsite separately, for now I just want the feature overview list on the Github README.

README.md
danielvegamyhre: If we come up with a better concrete example we can add it in a follow up PR. For now I think we should get the feature overview list into the README so it's clear to potential users glancing at the Github landing page what JobSet offers.

vsoch: Sorry missed this comment! Mapping to the level of a rack isn't particularly useful, at least it doesn't belong at this level - when we deploy to Google Cloud we usually ask for COMPACT mode when we want some guarantee of rack closeness. For mapping topology that is interesting, a better example is 1 pod per node. I think that can typically be achieved with resource requests / limits that are slightly below the node max capacity, and (maybe) a suggestion to the scheduler with affinity rules (but in practice I have found this is not enough). The topology that we are really interested in is more fine grained than that, and probably would need to be under the jurisdiction of the kubelet.

vsoch: Also I'm designing a new project idea that (I think) will use JobSet again, will ping you / keep you in the loop if/when it manifests. No pun intended! :P

README.md
```diff
@@ -44,6 +48,20 @@ Read the [installation guide](/docs/setup/install.md) to learn more.
 - ✔️ Security: RBAC based accessibility.
 - ✔️ Stable release cycle(2-3 months) for new features, bugfixes, cleanups.
 
+## Installation
+
+**Requires Kubernetes 1.26 or newer**.
```
kannon92, line 53: Can we say that we follow Kubernetes release process?

In 1-2 months I think we would want to bump this to Kubernetes 1.27..

danielvegamyhre: Sure, so something like:

**Maintains support for latest 3 Kubernetes minor versions. Current: 1.27, 1.28, 1.29**

(I know we currently run e2e-tests with 1.26 as well, but we should remove this and just focus on support for latest 3 minors, to align with upstream k8s).

What are your thoughts on this?

kannon92: My goal would be to avoid having to PR to keep these versions up to date..

Maintains support for latest 3 Kubernetes minor versions.

danielvegamyhre: Ok I added a line to "production readiness" bullets about this, and then here (installation instructiosn) I mentioned one of the last 3 minor versions is required.

kannon92 (commented):
/lgtm

PR #460, opened 2024-03-20
Move JobSet webhook into same webhooks package as pod webhook

### LINKED ISSUE #422: Investigate ways to validate Job templates at JobSet validation time
Currently we don't have a good way of validating the Job templates at JobSet validation time, so JobSets containing some error in the Job template somewhere are created successfully anyway, giving the user the false impression their workload spec is valid, only for them to be confused when no pods are created (either because the Job spec was invalid, or the pod spec was invalid, etc).

Debugging what the issue is requires checking the JobSet controller logs, which is not ideal.

We should see if we can validate at least some key parts of the Job template at JobSet creation time, in order to improve the user experience by rejecting the JobSet creation and giving them a clear error message.

danielvegamyhre:
Included in this PR:

- Refactoring the JobSet webhook (which was previously in the api pkg `v1alpha2`) into `pkg/webhooks` with the pod webhook. This decoupling of the JobSet struct itself from the webhook used for JobSet defaulting/validation seemed necessary for a couple reasons:
    - Consistency across the codebase (all webhooks implemented the same way, exist in same location)
    - We may eeded to add a `Validator` field (from kubectl-validate) to the webhook for #422 so we only create the OpenAPI client for kubectl-validate once, and I don't think the JobSet struct is the proper place for that openapi client based validator.


api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -107,6 +118,15 @@ var _ webhook.Validator = &JobSet{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (js *JobSet) ValidateCreate() (admission.Warnings, error) {
+	client := openapiclient.NewComposite(
```
danielvegamyhre, line 121: Obviously this would need to be refactored to not create a new client every time on every validation attempt, but this is just a very rough proof-of-concept to determine if we want to actually go this direction or not. 

If so, then I will refactor all of this

Makefile
```diff
@@ -84,11 +84,11 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
 		crd:generateEmbeddedObjectMeta=true output:crd:artifacts:config=config/components/crd/bases\
 		webhook output:webhook:artifacts:config=config/components/webhook\
-		paths="./..."
+		paths="{./api/..., ./pkg/...}"
```
danielvegamyhre, line 87: Note: I had to change some Makefile rules to target specific paths rather than everything (`./...`) since it was picking up some javascript dependencies in the site/ directory, now that it is populated.

danielvegamyhre:
@ahg-g @kannon92 this is ready for review now when you have some time. Changes are outlined in the PR description.

go.mod
```diff
@@ -21,27 +21,40 @@ require (
 )
 
 require (
+	github.com/NYTimes/gziphandler v1.1.1 // indirect
```
kannon92, line 24: Do you know how much larger this makes the binary?

It seems to be a lot of dependencies we brought in..

pkg/webhooks/jobset_webhook.go
```diff
@@ -11,30 +11,38 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package v1alpha2
+package webhooks
```
kannon92, line 14: Could we do this PR in stages?

There is a lot of changes here but its mixed with some cleanup/imports.

Maybe let's move the pkg and change the imports first.

And then we add the validation?

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -13,6 +14,10 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/ptr"
+
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
```
kannon92, line 20: This change seems to make the diffs really large..

As I stated above, could we see about a PR with just this change? And then the validation?

pkg/webhooks/jobset_webhook.go
danielvegamyhre: Sounds good, what do you think of the concept/approach though? And about adding https://github.com/kubernetes-sigs/kubectl-validate as a dependency?

Makefile
```diff
@@ -79,16 +79,20 @@ help: ## Display this help.
...
+manifests: controller-gen crd ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) \
 		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
 		crd:generateEmbeddedObjectMeta=true output:crd:artifacts:config=config/components/crd/bases\
 		webhook output:webhook:artifacts:config=config/components/webhook\
-		paths="./..."
+		paths="{./api/..., ./pkg/...}"
+
+.PHONY: crd
+crd: kustomize # Write CRD to independent file used for validation.
```
kannon92, line 90: Do we have any protections to make sure that someone commits an update to the crd when bumping the api?

kannon92: I think we would actually. https://github.com/kubernetes-sigs/jobset/blob/9102d3f75b08a40457504350700c761c4702ef36/Makefile#L137

danielvegamyhre: Not yet, one thing I was thinking is we could add the path to the crd directory into those we perform git diffs on in the `verify` make rule

kannon92: I think its covered here actually.

`make verify` calls manifests generates and then checks the diff.

It looks like you added `make crd` to the manifest or generate step so I think this should be fine.

pkg/webhooks/jobset_webhook.go
kannon92: I think its an elegant solution. I think the dependency increase is probably going from including apiserver.

Makefile
danielvegamyhre: Ah right, perfect

pkg/webhooks/jobset_webhook.go
```diff
@@ -53,26 +61,66 @@ const (
 	// Error message returned by JobSet validation if the network subdomain
 	// will be longer than 63 characters.
 	subdomainTooLongErrMsg = ".spec.network.subdomain is too long, must be less than 63 characters"
+
+	// k8sMinorForHardcodedBuitins is used for creating the OpenAPI client.
+	k8sMinorForHardcodedBuitins = "1.27"
```
kannon92, line 66: We are still supporting 1.26 until 1.30 is released right?

kannon92: I'm wrong. https://kubernetes.io/releases/#release-v1-26

1.26 was dropped as of last month. 

Not sure what your LTS support will be for GKE on this. but I guess its fine from upstream perspective.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -727,23 +801,100 @@ func TestValidateCreate(t *testing.T) {
...
+						Operator: jobset.OperatorAll,
+					},
 				},
 			},
 			want: errors.Join(
 				fmt.Errorf(podNameTooLongErrorMsg),
 			),
 		},
+		{
+			name: "missing containers",
```
kannon92, line 819: At a glance I think this is the main change in this file.

We should maybe consider more test cases here..

Does kubectl-validate buy us anything more in jobset validation? Maybe we can mix in a few validations on the spec?

kannon92: Ah I see. You have successPolicy but we are missing a few others.

go.mod
danielvegamyhre: I'm not sure, I'll have to look into it.

pkg/webhooks/jobset_webhook_test.go
danielvegamyhre: Yeah, containers and successPolicy are the new things being validated here. 

The kind of test cases we can add here are limited since many types of errors (typos in field names) are not possible to represent with Go code since those would be caught at compile time.

I was thinking perhaps in a separate PR we could add some E2E tests which create JobSets from yaml directly.

pkg/webhooks/jobset_webhook.go
kannon92: @danielvegamyhre @ahg-g WDYT of this?

danielvegamyhre: I think we should drop support for 1.26 (remove 1.26 e2e test) and keep 1.27 for the hardcoded builtins here

danielvegamyhre: Resolving since we dropped the 1.26 test

danielvegamyhre:
/hold

until I can talk to @alexzielenski when he is back from kubecon to learn more about kubectl-validate and it's roadmap for improved support for k8s native types

pkg/webhooks/jobset_webhook_test.go
danielvegamyhre: I am splitting this into separate PRs. This PR is now purely focused on moving the jobset webhook into the pkg/webhooks pkg the pod webhook lives in.

pkg/webhooks/jobset_webhook.go
danielvegamyhre: Updated this PR to only focus on migrating jobset webhook to webhooks pkg and very minor cleanup

kannon92:
Please rebase.

Otherwise, lgtm.

danielvegamyhre:
> Please rebase.
> 
> Otherwise, lgtm.

Finished rebasing

kannon92 (commented):
/lgtm

PR #465, opened 2024-03-21
Add ElasticJob support

### LINKED ISSUE #463: Support Elastic JobSets

**Why is this needed**:

**Implementation**:

kannon92:
Fixes #463 

Elastic JobSet provides a way to downscale/upscale replicated jobs. The goal of this PR is to support ElasticJobSet sets.

In order to support this, we need to relax our validation of ReplicatedJobs.

### DIFF
--- a/api/jobset/v1alpha2/jobset_webhook.go
+++ b/api/jobset/v1alpha2/jobset_webhook.go
@@ -178,8 +178,7 @@ func (js *JobSet) ValidateUpdate(old runtime.Object) (admission.Warnings, error)
 		}
 	}
 	// Note that SucccessPolicy and failurePolicy are made immutable via CEL.
-	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
-	errs = append(errs, apivalidation.ValidateImmutableField(js.Labels[LabelManagedBy], oldJS.Labels[LabelManagedBy], field.NewPath("metadata").Child("labels").Key(LabelManagedBy))...)
+	errs := apivalidation.ValidateImmutableField(js.Labels[LabelManagedBy], oldJS.Labels[LabelManagedBy], field.NewPath("metadata").Child("labels").Key(LabelManagedBy))
 	return nil, errs.ToAggregate()
 }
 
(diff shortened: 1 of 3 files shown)


kannon92:
/hold

tests are failing and we should discuss validation for ReplicatedJobs in more detail.

api/jobset/v1alpha2/jobset_webhook.go
```diff
@@ -178,8 +178,7 @@ func (js *JobSet) ValidateUpdate(old runtime.Object) (admission.Warnings, error)
 		}
 	}
 	// Note that SucccessPolicy and failurePolicy are made immutable via CEL.
-	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
```
kannon92, line 181: @ahg-g this is probably not what we want.

If we drop all validation of a replicated job, updates could add/remove entire replicated jobs.

I think we wouldn't want that.

api/jobset/v1alpha2/jobset_webhook_test.go
```diff
@@ -747,3 +747,192 @@ func TestValidateCreate(t *testing.T) {
...
+			oldJs: &JobSet{
+				ObjectMeta: validObjectMeta,
+				Spec: JobSetSpec{
+					Suspend:        ptr.To(true),
+					ReplicatedJobs: validReplicatedJobs,
+				},
+			},
+		},
+		{
+			name: "dropping a replicated job is valid",
```
kannon92, line 901: @ahg-g @danielvegamyhre added this here so we can discuss what kind of validation do we want on updates for replicated jobs.

I think this is probably not what we want..  OTOH if a jobset is suspended maybe someone can drop/add replicated jobs..

test/integration/controller/jobset_controller_test.go
```diff
@@ -1125,6 +1125,25 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("elastic jobset; scale up replicated jobs", &testCase{
```
kannon92, line 1128: note: this isn't working yet. Once I figure out validation, I'll come back to this.

PR #468, opened 2024-03-22
Add jobset roadmap to README

danielvegamyhre:
We should include a high level roadmap for the known features we want to work on over the next year or so.


kannon92 (commented):
/lgtm
/hold

for @ahg-g thoughts

PR #477, opened 2024-03-26
Include first failed job name in event emitted when JobSet fails, as well as the JobSet failure condition

### LINKED ISSUE #466: Emit event containing name of first failed Job which ultimately triggered JobSet failure
Emit event containing name of first (oldest) failed Job which ultimately triggered JobSet failure

**Why is this needed**:
Users would like a quicker way of immediately knowing the original failed Job that triggered a JobSet failure, so they can quickly look up the relevant logs.

danielvegamyhre:
Fixes #466 

- Include first failed job name in event emitted when JobSet fails, as well as the JobSet failure condition.
- Move constants into a separate constants package
- Move all success policy related helper functions into a separate file


pkg/controllers/jobset_controller.go
```diff
@@ -916,3 +914,42 @@ func findReplicatedStatus(replicatedJobStatus []jobset.ReplicatedJobStatus, repl
 	}
 	return jobset.ReplicatedJobStatus{}
 }
+
+// messageWithFirstFailedJob appends the first failed job to the original event message in human readable way.
```
kannon92, line 918: Could we move these out of this file?

They don't require the controller so it may be useful to move them to a separate file.

pkg/controllers/jobset_controller.go
```diff
@@ -43,6 +43,18 @@ import (
 const (
 	RestartsKey    string = "jobset.sigs.k8s.io/restart-attempt"
 	maxParallelism int    = 50
+
+	// Event reasons and messages
```
kannon92, line 47: If these constants keep growing, maybe we should consider moving them out of this file.

Its mostly a nit.

pkg/controllers/jobset_controller.go
danielvegamyhre: Yeah jobset_controller.go is overdue for refactoring. To start I think we can refactor some of the many helper functions into separate files based on the feature (similar to what you did with startup policy). 

I did some refactoring in this PR (e.g. moving some functions into success_policy.go, adding a constants pkg, etc.)

However, for these particular functions, I'm not sure of the best place to put them yet. They are mostly about finding the first failed job for a Jobset, which doesn't lend itself to a cleanly separated feature file.

pkg/controllers/jobset_controller.go
danielvegamyhre: I agree, moved to a constants file

pkg/controllers/constants/constants.go
```diff
@@ -0,0 +1,56 @@
+/*
```
kannon92, line 1: can we move this to pkg/constants/constants.go?

pkg/controllers/success_policy.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	batchv1 "k8s.io/api/batch/v1"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+
+	"sigs.k8s.io/jobset/pkg/util/collections"
+)
+
+// TODO: add unit tests for the functions in this file.
```
kannon92, line 24: +1

kannon92:
/hold
/lgtm

I have a minor nit around constants naming but otherwise this looks good to me.

pkg/controllers/constants/constants.go
danielvegamyhre: Yeah that makes sense, done

pkg/controllers/success_policy.go
kannon92: Do you want this done as part of this PR? or a TODO for others?

danielvegamyhre: I'll create a ticket to do it in follow up PR

kannon92:
> All tests are passing locally but the build fails in e2e CI? odd...

Not sure why we did this but it bit us again...

https://github.com/kubernetes-sigs/jobset/blob/main/Dockerfile

Docker file is copying elements one by one in pkg.

danielvegamyhre:
> > All tests are passing locally but the build fails in e2e CI? odd...
> 
> Not sure why we did this but it bit us again...
> 
> https://github.com/kubernetes-sigs/jobset/blob/main/Dockerfile
> 
> Docker file is copying elements one by one in pkg.

Ahhhh not again... this dockerfile will be the death of me. Fixing

danielvegamyhre:
@kannon92 I fixed the dockerfile but lgtm was removed because of this, can you take another look please?

Issue #476, opened 2024-03-26
Increase unit test coverage


googs1025:
**What would you like to be added**:
Try to increase the scope of unit testing, the current coverage is relatively low.
**Why is this needed**:
Improve unit test coverage, improve code robustness, and use this issue to track progress.

reference：https://github.com/kubernetes-sigs/jobset/pull/473

googs1025:
@danielvegamyhre
I tried to follow your suggestion and rewrote the demo. Please check if it meets your expectations. This is just a demo.

```go
func TestCreateJobs(t *testing.T) {
	var (
		jobSetName = "test-jobset"
		ns         = "default"
	)
	tests := []struct {
		name                string
		js                  *jobset.JobSet
		jobs                childJobs
		expected            error
		expectedJobsCreated []*batchv1.Job
	}{
		{
			name: "create jobs",
			js: testutils.MakeJobSet(jobSetName, ns).
				ReplicatedJob(testutils.MakeReplicatedJob("replicated-job-1").
					Job(testutils.MakeJobTemplate("test-job", ns).Obj()).
					Replicas(1).
					Obj()).Obj(),
			jobs: childJobs{
				active: []*batchv1.Job{
					makeJob(&makeJobArgs{
						jobSetName:        jobSetName,
						replicatedJobName: "replicated-job-1",
						jobName:           "test-jobset-replicated-job-1-test-job-0",
						ns:                ns,
						replicas:          1,
						jobIdx:            0,
					}).Parallelism(1).
						Completions(2).
						Ready(1).
						Active(1).Obj(),
				},
			},
			expectedJobsCreated: []*batchv1.Job{
				makeJob(&makeJobArgs{
					jobSetName:        jobSetName,
					replicatedJobName: "replicated-job-1",
					jobName:           "test-jobset-replicated-job-1-test-job-0",
					ns:                ns,
					replicas:          1,
					jobIdx:            0,
				}).Parallelism(1).
					Completions(2).
					Ready(1).
					Active(1).Obj(),
			},
		},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			var jobsCreated []*batchv1.Job
			scheme := runtime.NewScheme()
			utilruntime.Must(jobset.AddToScheme(scheme))
			utilruntime.Must(corev1.AddToScheme(scheme))
			utilruntime.Must(batchv1.AddToScheme(scheme))
			fakeClient := fake.NewClientBuilder().
[...]

PR #479, opened 2024-03-27
test: add unit test for createJobs deleteJobs

googs1025:
issue from #476

### DIFF
--- a/pkg/controllers/jobset_controller_test.go
+++ b/pkg/controllers/jobset_controller_test.go
@@ -17,15 +17,18 @@ import (
 	"context"
 	"strconv"
 	"testing"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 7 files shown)


pkg/controllers/jobset_controller_test.go
```diff
@@ -1150,3 +1161,1070 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
 		PodAnnotations(annotations)
 	return jobWrapper
 }
+
+type fakeClient struct {
```
danielvegamyhre, line 1165: Can we move the helper functions and option structs (including `makeJob` and `makeJobArgs`) to the bottom of the file, and have the actual tests at the top?

googs1025: ya, that makes sense

PR #480, opened 2024-03-27
Lower default resource request for controller manager but keep high limit

### LINKED ISSUE #462: default resource request should be lower so it fits on e2-medium
default resource request should be lower so it fits on e2-medium

**Why is this needed**:

Too many people struggle to figure out why their jobset controller can't schedule on a default GKE cluster that uses e2-medium.

danielvegamyhre:
Fixes #462


danielvegamyhre:
Had to lower from 2 CPU to 500mi  to get it of it on e2-medium by default

danielvegamyhre:
cc @kannon92 this is ready now

PR #483, opened 2024-03-28
Move headless service creation outside of createJobs

danielvegamyhre:
The fact that headless service creation is inside the createJobs function is a relic of the first JobSet release where we created 1 headless service per replicatedJob.

When we changed this to 1 headless service per JobSet, we should have moved this outside of the createJobs function, to better follow the single responsibility principle. The headless service creation now has no dependency on jobs or replicated jobs so we should move it out to the main reconcile loops before job creation begins.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -145,6 +145,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 	// If pod DNS hostnames are enabled, create a headless service for the JobSet
 	if err := r.createHeadlessSvcIfNecessary(ctx, &js); err != nil {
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -143,6 +143,11 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		}
 	}
 
+	// If pod DNS hostnames are enabled, create a headless service for the JobSet
+	if err := r.createHeadlessSvcIfNecessary(ctx, &js); err != nil {
```
kannon92, line 147: Maybe we could have a log here also?

danielvegamyhre: Added logging of the error here

PR #484, opened 2024-03-28
Code cleanup and refactoring

danielvegamyhre:
Code cleanup and refactoring
- Move more event reasons and messages into constants package
- Refactor `suspendJobSet` to `suspendJobs` to follow same pattern as createJobs and deleteJobs, as well as improve readability (since before the logic read "if jobset is suspended, suspend jobset"
- Refactor `resumeJobsIfNecessary` similarly to above
- Split "`generateStartupPolicyCondition(metav1.True or metav1.False)`" into two functions for readability, which better indicate the possible conditions and their semantic meaning:
  - inOrderStartupPolicyExecutingCondition()
  - inOrderStartupPolicyCompletedCondition
- Add some comments throughout the code base where I think things could some explanation
- Rename `rjobStarted` to `allReplicasStarted` to better indicate what exactly is required for a replicatedJob to be considered started.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -153,12 +153,12 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	// Handle suspending a jobset or resuming a suspended jobset.
 	jobsetSuspended := jobSetSuspended(&js)
 	if jobsetSuspended {
-		if err := r.suspendJobSet(ctx, &js, ownedJobs); err != nil {
+		if err := r.suspendJobs(ctx, &js, ownedJobs.active); err != nil {
 			log.Error(err, "suspending jobset")
 			return ctrl.Result{}, err
 		}
 	} else {
-		if err := r.resumeJobSetIfNecessary(ctx, &js, ownedJobs, status); err != nil {
+		if err := r.resumeJobsIfNecessary(ctx, &js, ownedJobs.active, status); err != nil {
 			log.Error(err, "resuming jobset")
 			return ctrl.Result{}, err
 		}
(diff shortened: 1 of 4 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -355,22 +355,32 @@ func (r *JobSetReconciler) resumeJobSetIfNecessary(ctx context.Context, js *jobs
...
+				condition:        inOrderStartupPolicyExecutingCondition(),
 			})
 		}
 	}
+	// At this point all replicated jobs have had their jobs resumed.
+	// If using an in order startup policy. add a condition to the JobSet indicating the
+	// in order startup policy has completed.
 	if inOrderStartupPolicy(startupPolicy) {
-		return r.ensureCondition(ctx, ensureConditionOpts{
+		if err := r.ensureCondition(ctx, ensureConditionOpts{
```
kannon92, line 373: This is changing the code.

We want to return here if startup policy is done.

danielvegamyhre: Yeah sorry I forgot to put a note here explaining this change. My thought was this was a bug that we are returning early here, because otherwise we never set the condition below that indicates the JobSet has been resumed; we only add the condition that startup policy is done. Wdyt?

kannon92: I did that by design.

I return the jobset to trigger a reconcile. I only want to resume the jobset once all the jobs have resumed.

kannon92: Why I return is that I don't want to go to the next job in the loop until the previous job is started.

I think this could potentially start the next job once the previously is resumed.

danielvegamyhre: >Why I return is that I don't want to go to the next job in the loop until the previous job is started.

This is outside the loop though. Inside the loop we still do the early return on line 361, to wait for the replicatedJob we just resumed to be ready before moving on to the next.

kannon92: Ha. Yea I guess that is so..

kannon92: imo I would really try to keep bug fixes separate from refactor PRs. It makes cherry-picks and such much more difficult..

/hold

kannon92: Could you move this fix into a separate PR?

danielvegamyhre:
Thanks for the quick review @kannon92 ! We can go ahead and merge since this is just refactoring and a minor bug fix. I will keep bug fixes separate from refactors going forward as you mentioned

/hold cancel

Issue #440, opened 2024-03-04
Introduce `managedBy` field


tenzen-y:
**What would you like to be added**:
Remove the `alpha.jobset.sigs.k8s.io/managed-by` label and introduce `.spec.managedBy` field.

@mimowo @ahg-g @alculquicondor WDYT?

/kind feature

**Why is this needed**:
Since https://github.com/kubernetes-sigs/jobset/pull/407, we introduced the new `managed-by` label based on the batch/job KEP: https://github.com/kubernetes/enhancements/pull/4370.

But, we decided to introduce  a new `managedBy` field to batch/job here: https://github.com/kubernetes/enhancements/pull/4530
So, I think that we should use the `managedBy` field instead of the dedicated label as well as batch/job.

ahg-g:
We added in the previous release, see https://github.com/kubernetes-sigs/jobset/issues/379

alculquicondor:
+1 for replacing with a field.

I suppose the alpha annotation already made it to the release. I would suggest keeping both mechanism for at least one release, to facilitate upgrades to Kueue+Jobset.

tenzen-y:
> +1 for replacing with a field.
> 
> I suppose the alpha annotation already made it to the release. I would suggest keeping both mechanism for at least one release, to facilitate upgrades to Kueue+Jobset.

That makes sense.

kannon92:
I tried a quick revert option on the PR but that failed so reverting may be a bit of a manual process..

mimowo:
+1 for using field, for ecosystem consistency, but also so that we fail fast when someone tries to use it on an old version (same argument as for the k8s job).

danielvegamyhre:
@trasc have you already started this? I was hoping to have a team member of mine work on this as a good first issue to ramp up on JobSet, we just hadn't assigned it to him yet, so i'd like to assign it to them if that's okay

trasc:
@danielvegamyhre Not yet.

/unassign

danielvegamyhre:
Thanks!

/assign @jedwins1998

kannon92:
What is our deprecation policy for the managed-by label?

danielvegamyhre:
Update: we can safely drop the label entirely per conversation here https://github.com/kubernetes-sigs/kueue/pull/1870#discussion_r1536007381

danielvegamyhre:
Some additional notes for implementation: 

[...]

PR #487, opened 2024-03-28
Introduce `managedBy` field and Remove `managed-by` label

jedwins1998:
This pull request is to introduce `managedBy` field and Remove `managed-by` label.

Original Issue: https://github.com/kubernetes-sigs/jobset/issues/440


pkg/controllers/jobset_controller.go
```diff
@@ -86,7 +86,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(&js))
 	ctx = ctrl.LoggerInto(ctx, log)
 
-	if manager, found := js.Labels[jobset.LabelManagedBy]; found && manager != jobset.JobSetManager {
+	if manager := ptr.Deref(js.Spec.ManagedBy, ""); js.Spec.ManagedBy != nil && manager != jobset.JobSetManager {
```
kannon92, line 89:
```suggestion
	if !ptr.Equal(js.Spec.ManagedBy, jobset.JobSetManager) {
```

jedwins1998: Will `ptr.Equal` work if `jobset.JobSetManager` is of type `string` and not `*string`?

kannon92: Not without a cast I think.

jedwins1998: In that case I propose leaving the line as is.

danielvegamyhre: Either way is fine but let's extract this check into a helper function, so the main entry point into the JobSet controller consists of high level functions that read like English, and put the low level pointer manipulation, string comparison logic into helper functions with clear names. 

(I realize we have been doing the raw checking logic here so far, but this is a good opportunity to improve the readability, which is becoming increasingly important with time as we add more features and the code base grows).

Let's also add a comment explaining why this check is even necessary, and what's its purpose is, since to an unfamiliar reader it probably makes no sense for the JobSet controller to not be managing the JobSet (see MultiKueue KEP https://github.com/kubernetes-sigs/kueue/tree/main/keps/693-multikueue, let's include a link to this in the comment as well).

danielvegamyhre (commented):
LGTM overall, left one minor comment

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -94,6 +91,9 @@ type JobSetSpec struct {
 
 	// Suspend suspends all running child Jobs when set to true.
 	Suspend *bool `json:"suspend,omitempty"`
+
+	// ManagedBy is used to indicate the controller or entity that manages a JobSet
+	ManagedBy *string `json:"managedBy,omitempty"`
```
mimowo, line 96: I would suggest that the `managedBy` value follows the format in the core k8s: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/4368-support-managed-by-for-batch-jobs#api, with the reserved value for k8s being: `kubernetes.io/job-controller`.

The upstream validation: https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/pkg/apis/batch/validation/validation.go#L220-L223. I see JobSet already depends on `k8s.io/apimachinery/pkg/api/validation` so I assume `IsDomainPrefixedPath` can be used here also. 

For jobset the reserved value could be: `jobset.sigs.k8s.io/jobset-controller`, or `jobset.x-k8s.io/jobset-controller` (or something along the lines).

pkg/controllers/jobset_controller.go
jedwins1998: This is done. I chose the helper function named based on what the Job controller uses. See [here](https://github.com/kubernetes/kubernetes/blob/3dedb8eb8c122d0a3221a5842c1d6697d8958151/pkg/controller/job/job_controller.go#L1988).

api/jobset/v1alpha2/jobset_types.go
jedwins1998: This is done. I went with `jobset.sigs.k8s.io/jobset-controller` for the reserved value. I also updated the variable name in `jobset_types.go` to `JobSetControllerName` to match  the name used in the Job API.

mimowo: Sg

danielvegamyhre:
Let's also validate the ManagedBy field like is done upstream here https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/pkg/apis/batch/validation/validation.go#L220-L223

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -40,12 +40,9 @@ const (
...
 
-	// LabelManagedBy is used to indicate the controller or entity that manages an JobSet
-	LabelManagedBy = "alpha.jobset.sigs.k8s.io/managed-by"
-
-	// JobSetManager is used as the value for LabelManagedBy to identify the jobset controller manager
-	// as the manager of a specific JobSet.
-	JobSetManager = "jobset"
+	// JobSetControllerName is the reserved value for the managedBy field for the built-in
+	// JobSet controller.
+	JobSetControllerName = "jobset.sigs.k8s.io/jobset-controller"
```
mimowo, line 45: Should we to introduce validation in the webhook with `IsDomainPrefixedPath`, analogously as for Job: https://github.com/kubernetes/kubernetes/blob/f4e246bc93ffb68b33ed67c7896c379efa4207e7/pkg/apis/batch/validation/validation.go#L220-L223. Potentially also length.

mimowo (commented):
Question regarding validation, other than that lgtm,

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: Yes I think so, we have been bitten before by not having strict enough validation, so I'd like to enforce the field is restricted to expected format and length

jedwins1998: I updated the code. Please resolve if nothing else is needed.

danielvegamyhre: Can you add a unit tests ensuring the validation works as expected

jedwins1998: Done.

Issue #478, opened 2024-03-26
Add pod controller unit tests


danielvegamyhre:
**What would you like to be added**:
Unit tests for https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/pod_controller.go in a pod_controller_test.go file.

Should use a fake client with interceptor functions as described [here](https://github.com/kubernetes-sigs/jobset/pull/473#discussion_r1537055795) to verify we delete the follower pods when the leader pod's topology does not match.

**Why is this needed**:
For improved test coverage.

googs1025:
Do we need to close this issue? @danielvegamyhre

PR #490, opened 2024-03-30
test: add pod controller unit test

googs1025:
Added unit test for podcontroller
issue from https://github.com/kubernetes-sigs/jobset/issues/478

### DIFF
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -56,6 +56,7 @@ func TestValidatePodPlacements(t *testing.T) {
 			podName:           "test-jobset-replicated-job-1-test-job-0-1",
 			ns:                ns,
 			jobIdx:            0})
+		nodeSelector = map[string]string{"test-node-topologyKey": "topologyDomain"}
 	)
 	tests := []struct {
 		name           string
@@ -73,7 +74,7 @@ func TestValidatePodPlacements(t *testing.T) {
 			podList: &corev1.PodList{
 				Items: []corev1.Pod{
 					leaderPodWrapper.AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
-					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
+					podWrapper.NodeSelector(nodeSelector).Obj(),
 				},
 			},
 			node: &corev1.Node{
@@ -81,8 +82,7 @@ func TestValidatePodPlacements(t *testing.T) {
 					Name: "test-node",
 				},
 			},
-			wantMatched: false,
-			wantErr:     fmt.Errorf("node does not have topology label: %s", "test-node-topologyKey"),
+			wantErr: fmt.Errorf("node does not have topology label: %s", "test-node-topologyKey"),
 		},
 		{
 			name: "valid pod placements with matched",
@@ -90,22 +90,21 @@ func TestValidatePodPlacements(t *testing.T) {
 				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 			podList: &corev1.PodList{
 				Items: []corev1.Pod{
-					leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
+					leaderPodWrapper.NodeSelector(nodeSelector).
 						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
-					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
+					podWrapper.NodeSelector(nodeSelector).Obj(),
 				},
 			},
 			node: &corev1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   "test-node",
-					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+					Labels: nodeSelector,
 				},
 			},
 			wantMatched: true,
-			wantErr:     nil,
 		},
 		{
-			name: "valid pod placements with pod nodeSelector is nil",
+			name: "follower pod nodeSelector is nil",
 			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
 				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 			podList: &corev1.PodList{
@@ -119,14 +118,14 @@ func TestValidatePodPlacements(t *testing.T) {
 			node: &corev1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   "test-node",
-					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+					Labels: nodeSelector,
 				},
 			},
 			wantMatched: false,
 			wantErr:     fmt.Errorf("pod %s nodeSelector is nil", "test-jobset-replicated-job-1-test-job-0-1"),
 		},
 		{
-			name: "valid pod placements with pod nodeSelector is missing key",
+			name: "valid pod placements with pod nodeSelector is empty",
 			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
 				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 			podList: &corev1.PodList{
@@ -138,10 +137,9 @@ func TestValidatePodPlacements(t *testing.T) {
 			node: &corev1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   "test-node",
-					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+					Labels: nodeSelector,
 				},
 			},
-			wantMatched: false,
 			wantErr: fmt.Errorf("pod %s nodeSelector is missing key: %s",
 				"test-jobset-replicated-job-1-test-job-0-1", "test-node-topologyKey"),
 		},
@@ -151,7 +149,7 @@ func TestValidatePodPlacements(t *testing.T) {
 				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 			podList: &corev1.PodList{
 				Items: []corev1.Pod{
-					leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
+					leaderPodWrapper.NodeSelector(nodeSelector).
 						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain1"}).Obj(),
 				},
@@ -159,10 +157,9 @@ func TestValidatePodPlacements(t *testing.T) {
 			node: &corev1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   "test-node",
-					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+					Labels: nodeSelector,
 				},
 			},
-			wantMatched: false,
 			wantErr: fmt.Errorf("follower topology %q != leader topology %q",
 				"topologyDomain1", "topologyDomain"),
 		},
@@ -172,7 +169,7 @@ func TestValidatePodPlacements(t *testing.T) {
 				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 			podList: &corev1.PodList{
 				Items: []corev1.Pod{
-					leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
+					leaderPodWrapper.NodeSelector(nodeSelector).
 						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
 					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain1"}).Obj(),
 				},
@@ -180,10 +177,9 @@ func TestValidatePodPlacements(t *testing.T) {
 			node: &corev1.Node{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:   "test-node",
-					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+					Labels: nodeSelector,
 				},
 			},
-			wantMatched:    false,
 			forceClientErr: true,
 			wantErr:        errors.Join(errors.New("example error")),
 		},
@@ -298,7 +294,7 @@ func TestDeleteFollowerPods(t *testing.T) {
 					Type:               corev1.DisruptionTarget,
 					Status:             corev1.ConditionFalse,
 					Reason:             constants.ExclusivePlacementViolationReason,
-					LastTransitionTime: metav1.Now().Rfc3339Copy(),
+					LastTransitionTime: metav1.Now(),
 					Message:            constants.ExclusivePlacementViolationMessage,
 				},
 				}).Obj(),
@@ -336,7 +332,8 @@ func TestDeleteFollowerPods(t *testing.T) {
 					if tc.forceClientErr || pod == nil {
 						return errors.Join(errors.New("example error"))
 					}
-					// This is to solve the problem that the timestamps do not match. There will be a slight gap.
+					// This is to solve the problem that the timestamps do not match.
+					// There will be a slight gap.
 					pod.Status.Conditions[0].LastTransitionTime =
 						tc.wantPodsDeleted[0].Status.Conditions[0].LastTransitionTime
 					gotPodsDeleted = append(gotPodsDeleted, *pod)
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -162,12 +162,6 @@ func TestValidatePodPlacements(t *testing.T) {
 					followerPodWrapper.NodeSelector(map[string]string{testTopologyKey: "topologyDomain"}).Obj(),
 				},
 			},
-			node: &corev1.Node{
-				ObjectMeta: metav1.ObjectMeta{
-					Name:   "test-node",
-					Labels: nodeSelector,
-				},
-			},
 			forceClientErr: true,
 			wantErr:        errors.New("example error"),
 		},
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -172,7 +172,7 @@ func TestValidatePodPlacements(t *testing.T) {
 				},
 			},
 			forceClientErr: true,
-			wantErr:        errors.Join(errors.New("example error")),
+			wantErr:        errors.New("example error"),
 		},
 	}
 	for _, tc := range tests {
@@ -185,7 +185,7 @@ func TestValidatePodPlacements(t *testing.T) {
 						return nil
 					}
 					if tc.forceClientErr || node == nil {
-						return errors.Join(errors.New("example error"))
+						return errors.New("example error")
 					}
 					// Set returned node value to be the node defined in the test case.
 					*node = *tc.node
@@ -197,7 +197,9 @@ func TestValidatePodPlacements(t *testing.T) {
 				Record: fc.record,
 			}
 			gotMatched, gotErr := r.validatePodPlacements(context.Background(), &tc.leaderPod, tc.podList)
-			assert.Equal(t, tc.wantErr, gotErr)
+			if tc.wantErr != nil && gotErr != nil {
+				assert.Equal(t, tc.wantErr.Error(), gotErr.Error())
+			}
 			assert.Equal(t, tc.wantMatched, gotMatched)
 		})
 	}
@@ -293,7 +295,7 @@ func TestDeleteFollowerPods(t *testing.T) {
 				}).Obj(),
 			},
 			forceClientErr: true,
-			wantErr:        errors.Join(errors.Join(errors.New("example error"))),
+			wantErr:        errors.New("example error"),
 		},
 		{
 			name: "delete follower pods with delete error",
@@ -307,7 +309,7 @@ func TestDeleteFollowerPods(t *testing.T) {
 				}).Obj(),
 			},
 			forceClientErr: true,
-			wantErr:        errors.Join(errors.Join(errors.New("example error"))),
+			wantErr:        errors.New("example error"),
 		},
 	}
 	for _, tc := range tests {
@@ -321,7 +323,7 @@ func TestDeleteFollowerPods(t *testing.T) {
 						return nil
 					}
 					if tc.forceClientErr || pod == nil {
-						return errors.Join(errors.New("example error"))
+						return errors.New("example error")
 					}
 					// This is to solve the problem that the timestamps do not match.
 					// There will be a slight gap.
@@ -337,7 +339,7 @@ func TestDeleteFollowerPods(t *testing.T) {
 						return nil
 					}
 					if tc.forceClientErr || pod == nil {
-						return errors.Join(errors.New("example error"))
+						return errors.New("example error")
 					}
 					return nil
 				},
@@ -349,7 +351,9 @@ func TestDeleteFollowerPods(t *testing.T) {
 			}
 
 			gotErr := r.deleteFollowerPods(context.Background(), tc.pods)
-			assert.Equal(t, gotErr, tc.wantErr)
+			if tc.wantErr != nil && gotErr != nil {
+				assert.Equal(t, tc.wantErr.Error(), gotErr.Error())
+			}
 			sort.Slice(gotPodsDeleted, func(i, j int) bool {
 				return gotPodsDeleted[i].Name < gotPodsDeleted[j].Name
 			})
@@ -357,7 +361,7 @@ func TestDeleteFollowerPods(t *testing.T) {
 				return tc.wantPodsDeleted[i].Name < tc.wantPodsDeleted[j].Name
 			})
 
-			if !assert.Equal(t, gotPodsDeleted, tc.wantPodsDeleted) {
+			if !assert.Equal(t, tc.wantPodsDeleted, gotPodsDeleted) {
 				t.Errorf("deleteFollowerPods() did not make the expected pod deletion calls")
 			}
 		})
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -55,7 +55,7 @@ func TestValidatePodPlacements(t *testing.T) {
 			jobName:           "test-jobset-replicated-job-1-test-job-0",
 			podName:           "test-jobset-replicated-job-1-test-job-0-1",
 			ns:                ns,
-			jobIdx:            0})
+			jobIdx:            0}).AddAnnotation(batchv1.JobCompletionIndexAnnotation, "1")
 		testTopologyKey = "test-node-topologyKey"
 		nodeSelector    = map[string]string{testTopologyKey: "topologyDomain"}
 	)
@@ -222,14 +222,14 @@ func TestDeleteFollowerPods(t *testing.T) {
 			jobName:           "test-jobset-replicated-job-1-test-job-0",
 			podName:           "test-jobset-replicated-job-1-test-job-0-1",
 			ns:                ns,
-			jobIdx:            0})
+			jobIdx:            0}).AddAnnotation(batchv1.JobCompletionIndexAnnotation, "1")
 		wantPodWrapper = makePod(&makePodArgs{
 			jobSetName:        jobSetName,
 			replicatedJobName: "replicated-job-1",
 			jobName:           "test-jobset-replicated-job-1-test-job-0",
 			podName:           "test-jobset-replicated-job-1-test-job-0-1",
 			ns:                ns,
-			jobIdx:            0})
+			jobIdx:            0}).AddAnnotation(batchv1.JobCompletionIndexAnnotation, "1")
 	)
 	tests := []struct {
 		name            string
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -1,0 +1,430 @@
+/*
+Copyright 2023 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    htcp://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"sort"
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+	"sigs.k8s.io/jobset/pkg/constants"
+	testutils "sigs.k8s.io/jobset/pkg/util/testing"
+)
+
+func TestValidatePodPlacements(t *testing.T) {
+	var (
+		jobSetName       = "test-jobset"
+		ns               = "default"
+		leaderPodWrapper = makePod(&makePodArgs{
+			jobSetName:        jobSetName,
+			replicatedJobName: "replicated-job-1",
+			jobName:           "test-jobset-replicated-job-1-test-job-0",
+			podName:           "test-jobset-replicated-job-1-test-job-0-0",
+			ns:                ns,
+			nodeName:          "test-node",
+			jobIdx:            0})
+		podWrapper = makePod(&makePodArgs{
+			jobSetName:        jobSetName,
+			replicatedJobName: "replicated-job-1",
+			jobName:           "test-jobset-replicated-job-1-test-job-0",
+			podName:           "test-jobset-replicated-job-1-test-job-0-1",
+			ns:                ns,
+			jobIdx:            0})
+	)
+	tests := []struct {
+		name           string
+		leaderPod      corev1.Pod
+		podList        *corev1.PodList
+		node           *corev1.Node
+		wantErr        error
+		forceClientErr bool
+		wantMatched    bool
+	}{
+		{
+			name: "topology node label not found",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
+				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+			podList: &corev1.PodList{
+				Items: []corev1.Pod{
+					leaderPodWrapper.AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
+				},
+			},
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-node",
+				},
+			},
+			wantMatched: false,
+			wantErr:     fmt.Errorf("node does not have topology label: %s", "test-node-topologyKey"),
+		},
+		{
+			name: "valid pod placements with matched",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
+				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+			podList: &corev1.PodList{
+				Items: []corev1.Pod{
+					leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
+						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
+				},
+			},
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "test-node",
+					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+				},
+			},
+			wantMatched: true,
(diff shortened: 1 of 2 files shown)


googs1025:
@danielvegamyhre /PTAL thank a lot!

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,757 @@
...
+	jobSetName        string
+	replicatedJobName string
+	jobName           string
+	podName           string
+	nodeName          string
+	ns                string
+	replicas          int
+	jobIdx            int
+	restarts          int
+}
```
danielvegamyhre, line 707: It seems like `makePodArgs` has several parameters that aren't necessary for to test the functions in the pod controller, and as a result the unit test cases are bloated with extra lines we don't need. Can you reduce this to only needing the parameters which are strictly necessary?

googs1025: Oh, this is my question. It seems reasonable. I didn’t think it through at the time. I’ll change it.

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+					podList.Items = append(podList.Items, *tc.pod)
+					return nil
+				},
+			}).Build()
+			r := &PodReconciler{
+				Client: fc,
+				Scheme: scheme,
+				Record: recorder,
+			}
+			actualPodList, actual := r.listPodsForJob(context.TODO(), tc.pod.Namespace, tc.pod.GetLabels()[jobset.JobKey])
```
dejanzele, line 131:
```suggestion
			actualPodList, actualErr := r.listPodsForJob(context.TODO(), tc.pod.Namespace, tc.pod.GetLabels()[jobset.JobKey])
```

Also I think it is fine to use the `context.Background`

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+			assert.Equal(t, actual, tc.expectedErr)
```
dejanzele, line 132:
```suggestion
			assert.Equal(t, actualErr, tc.expectedErr)
```

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+					*node = *tc.node
+					return nil
+
+				}})
+			r := &PodReconciler{
+				Client: fc.client,
+				Scheme: fc.scheme,
+				Record: fc.record,
+			}
+			isMatched, actualErr := r.validatePodPlacements(context.TODO(), tc.leaderPod, tc.podList)
```
dejanzele, line 428: I think it is fine to use the `context.Background`

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+				},
+			}).Build()
+			r := &PodReconciler{
+				Client: fc,
+				Scheme: scheme,
+				Record: recorder,
+			}
+			actualPodList, actual := r.listPodsForJob(context.TODO(), tc.pod.Namespace, tc.pod.GetLabels()[jobset.JobKey])
+			assert.Equal(t, actual, tc.expectedErr)
+			if !reflect.DeepEqual(actualPodList, tc.expectPodList) {
```
dejanzele, line 133: Can you use `assert` package here also as it is already being used in other tests?

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+			pod: makePod(&makePodArgs{
+				jobSetName:        jobSetName,
+				replicatedJobName: "replicated-job-1",
+				jobName:           "test-jobset-replicated-job-1-test-job-0",
+				podName:           "test-jobset-replicated-job-1-test-job-0-0",
+				ns:                ns,
+				jobIdx:            0}).
+				Obj(),
+			expectPodList: nil,
+			errorFlag:     true,
```
dejanzele, line 96: Do you need `errorFlag`? Will it be enough just to check `expectedErr != nil`?

pkg/controllers/pod_controller_test.go
googs1025: @dejanzele Thank you for cr. After reviewing the source code of assert.Equal and reflect.DeepEqual, I found that assert.Equal internally calls the reflect.DeepEqual method as well. So, do I need to replace all the code with assert.Equal?

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+		jobSetName = "test-jobset"
+		ns         = "default"
+	)
+	tests := []struct {
+		name          string
+		job           *batchv1.Job
+		pod           *corev1.Pod
+		expectPodList *corev1.PodList
+		expectedErr   error
+		errorFlag     bool
```
dejanzele, line 39: Do you need errorFlag? Will it be enough just to check expectedErr != nil?

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,688 @@
...
+		ns         = "default"
+	)
+	tests := []struct {
+		name            string
+		leaderPod       *corev1.Pod
+		podList         *corev1.PodList
+		node            *corev1.Node
+		expectPodList   *corev1.PodList
+		expectedErr     error
+		errorFlag       bool
```
dejanzele, line 152: Same question for `errorFlag` as above

pkg/controllers/pod_controller_test.go
googs1025: I need the errorFlag variable to control whether the fakeclient has an error. Setting it to true represents an error, while setting it to false represents no error.

pkg/controllers/pod_controller_test.go
dejanzele: It does the same call, but I think the `assert` package also does nice prints of diffs in case of assertfion failure

pkg/controllers/pod_controller_test.go
dejanzele: Doesn't `errorFlag` always equal to the value of the following assertion `expectedErr != nil`?

googs1025: oh I understand what you mean, it makes sense

googs1025: @dejanzele After reviewing the code, I found that the errorFlag is used to indicate errors specifically related to the fakeClient operations, rather than representing all errors. This variable needs to be kept. So, would it be more appropriate for me to change it to "fakeClientOperationError"?

googs1025:
@danielvegamyhre /PTAL I have made the necessary modifications. Please take the time to review them. thanks a lot.

danielvegamyhre:
Been busy lately but will review this as soon as I can

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+	"sigs.k8s.io/jobset/pkg/constants"
+	testutils "sigs.k8s.io/jobset/pkg/util/testing"
+)
+
+func TestListPodsForJob(t *testing.T) {
```
danielvegamyhre, line 27: We can remove this test.

The only thing the method [listPodsForJob](https://github.com/kubernetes-sigs/jobset/blob/8860f832dee2c35d9cb2951260208fcef87cd2db/pkg/controllers/pod_controller.go#L161) does is make a List() API call for pod objects with a specific annotation. 

So we can't really inject an fake List() function via interceptorFunc here for testing, because there's no logic to test, so we would just be testing the test code.

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+	var (
+		jobSetName = "test-jobset"
+		ns         = "default"
+	)
+	tests := []struct {
+		name                       string
+		leaderPod                  *corev1.Pod
+		podList                    *corev1.PodList
+		node                       *corev1.Node
+		expectPodList              *corev1.PodList
```
danielvegamyhre:
```suggestion
		wantPodList              *corev1.PodList
```

Let's use more concise variable names following the typical "want/got" pattern

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+		expectedErr                error
```
danielvegamyhre:
```suggestion
		wantErr                error
```

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+		fakeClientOperationErrFlag bool
```
danielvegamyhre, line 151:
```suggestion
		forceClientErr bool
```

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+		name                       string
+		leaderPod                  *corev1.Pod
+		podList                    *corev1.PodList
+		node                       *corev1.Node
+		wantErr                    error
+		fakeClientOperationErrFlag bool
+		wantMatched                bool
+	}{
+		{
+			name: "valid pod placements with not found node label",
```
danielvegamyhre, line 155:
```suggestion
			name: "topology node label not found",
```

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+				replicatedJobName: "replicated-job-1",
+				jobName:           "test-jobset-replicated-job-1-test-job-0",
+				podName:           "test-jobset-replicated-job-1-test-job-0-0",
+				ns:                ns,
+				nodeName:          "test-node",
+				jobIdx:            0}).AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
+				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+			podList: &corev1.PodList{
+				Items: []corev1.Pod{
+					*makePod(&makePodArgs{
```
danielvegamyhre, line 167: It seems like this podList with these 2 makePod() calls is repeated many times, can we just store them in a variable and re-use it in all the tests?

Or atleast store the pods created by `makePod()` and then in each test case you can use `pod.NodeSelectors(...)` to test different cases (selector matches leader topology, selector doesn't match, etc.)

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+		leaderPod                  *corev1.Pod
+		podList                    *corev1.PodList
+		node                       *corev1.Node
+		wantErr                    error
+		fakeClientOperationErrFlag bool
+		wantMatched                bool
+	}{
+		{
+			name: "valid pod placements with not found node label",
+			leaderPod: makePod(&makePodArgs{
```
danielvegamyhre, line 156: Seems like this leader pod is created via the same makePod() call many times, let's store it in a variable and re-use it. That will reduce length of these test cases and improve readability

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+		name                       string
+		pods                       []corev1.Pod
+		wantPodsDeleted            []*corev1.Pod
+		wantErr                    error
+		fakeClientOperationErrFlag bool
+	}{
+		{
+			name: "delete follower pods",
+			pods: []corev1.Pod{
+				*makePod(&makePodArgs{
```
danielvegamyhre, line 450: Same in this test, instead of calling makePod() with the same arguments repeatedly let's store 1-2 pods in variables and reuse them in test cases.

pkg/util/testing/wrappers.go
```diff
@@ -326,3 +326,68 @@ func (j *JobWrapper) NodeSelector(nodeSelector map[string]string) *JobWrapper {
...
 	return &j.Job
 }
+
+// PodWrapper wraps a Pod.
+type PodWrapper struct {
+	corev1.Pod
+}
+
+// MakePod creates a wrapper for a Pod.
+func MakePod(podName, ns string) *PodWrapper {
```
danielvegamyhre, line 336: Is this used anywhere? I'm not seeing it used

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,687 @@
...
+		jobset.JobIndexKey:          strconv.Itoa(args.jobIdx),
+		jobset.JobKey:               jobHashKey(args.ns, args.jobName),
+	}
+	annotations := map[string]string{
+		jobset.JobSetNameKey: args.jobSetName,
+		jobset.JobIndexKey:   strconv.Itoa(args.jobIdx),
+		jobset.JobKey:        jobHashKey(args.ns, args.jobName),
+	}
+
+	podWrapper := testutils.MakePod(args.podName, args.ns).Labels(labels).Annotations(annotations)
```
googs1025, line 658: @danielvegamyhre here testutils.MakePod is called

pkg/controllers/pod_controller_test.go
googs1025: This is a bool value. Should I change it to forceClientErrFlag instead of forceClientErr?

pkg/controllers/pod_controller_test.go
danielvegamyhre: Ah ok, I will resolve that comment, thanks.

pkg/util/testing/wrappers.go
danielvegamyhre: Discussed below, resolving.

pkg/controllers/pod_controller_test.go
danielvegamyhre: I think the name is clear for a booelan value. You can add the `Flag` suffix if you want, not a big deal either way.

googs1025:
I will revise and modify the code in the next two days.

pkg/controllers/pod_controller_test.go
googs1025: So, does that mean I should remove the injected interceptorFunc method?

googs1025: If the interceptorFunc method is not injected, it will not be possible to test the scenario where the listing fails.

googs1025: @danielvegamyhre

pkg/controllers/pod_controller_test.go
googs1025: I tried to reuse the podWrapper method for each test case, but it caused some errors because these variables are shared across all the test cases, and certain fields may be modified. While we can list leaderPodWrapper for reuse, it might be better to create a separate podWrapper instance for each variable in every test case.

Here's an example:
```go
func TestValidatePodPlacements(t *testing.T) {
	var (
		jobSetName       = "test-jobset"
		ns               = "default"
		leaderPodWrapper = makePod(&makePodArgs{
			jobSetName:        jobSetName,
			replicatedJobName: "replicated-job-1",
			jobName:           "test-jobset-replicated-job-1-test-job-0",
			podName:           "test-jobset-replicated-job-1-test-job-0-0",
			ns:                ns,
			nodeName:          "test-node",
			jobIdx:            0})
		podWrapper = makePod(&makePodArgs{
			jobSetName:        jobSetName,
			replicatedJobName: "replicated-job-1",
			jobName:           "test-jobset-replicated-job-1-test-job-0",
			podName:           "test-jobset-replicated-job-1-test-job-0-1",
			ns:                ns,
			jobIdx:            0})
	)
	tests := []struct {
		name           string
		leaderPod      *corev1.Pod
		podList        *corev1.PodList
		node           *corev1.Node
		wantErr        error
		forceClientErr bool
		wantMatched    bool
	}{
		{
			name: "topology node label not found",
			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
			podList: &corev1.PodList{
				Items: []corev1.Pod{
					*leaderPodWrapper.AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
					*podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
				},
			},
			node: &corev1.Node{
					Name: "test-node",
				},
			},
			wantMatched: false,
			wantErr:     fmt.Errorf("node does not have topology label: %s", "test-node-topologyKey"),
		},
		{
			name: "valid pod placements with matched",
			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
			podList: &corev1.PodList{
				Items: []corev1.Pod{
					*leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
					*podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
				},
			},
			node: &corev1.Node{
					Name:   "test-node",
					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
				},
			},
			wantMatched: true,
			wantErr:     nil,
		},
		{
			name: "valid pod placements with pod nodeSelector is nil",
			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
			podList: &corev1.PodList{
				Items: []corev1.Pod{
					*leaderPodWrapper.NodeSelector(map[string]string{
						"test-node-topologyKey": "topologyDomain",
					}).AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
					*podWrapper.NodeSelector(nil).Obj(),
				},
			},
			node: &corev1.Node{
					Name:   "test-node",
					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
				},
			},
			wantMatched: false,
			wantErr:     fmt.Errorf("pod %s nodeSelector is nil", "test-jobset-replicated-job-1-test-job-0-1"),
		},
		{
			name: "valid pod placements with pod nodeSelector is missing key",
			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
			podList: &corev1.PodList{
				Items: []corev1.Pod{
					*leaderPodWrapper.AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
					*podWrapper.NodeSelector(map[string]string{}).Obj(),
				},
			},
			node: &corev1.Node{
					Name:   "test-node",
					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
				},
			},
			wantMatched: false,
			wantErr: fmt.Errorf("pod %s nodeSelector is missing key: %s",
				"test-jobset-replicated-job-1-test-job-0-1", "test-node-topologyKey"),
		},
		{
			name: "valid pod placements with followerTopology != leaderTopology",
			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
			podList: &corev1.PodList{
				Items: []corev1.Pod{
					*leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
					*podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain1"}).Obj(),
				},
			},
			node: &corev1.Node{
					Name:   "test-node",
					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
				},
			},
			wantMatched: false,
			wantErr: fmt.Errorf("follower topology %q != leader topology %q",
				"topologyDomain1", "topologyDomain"),
		},
		{
			name: "valid pod placements with get node error",
			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
			podList: &corev1.PodList{
				Items: []corev1.Pod{
					*leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
						AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
					*podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain1"}).Obj(),
				},
			},
			node: &corev1.Node{
					Name:   "test-node",
					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
				},
			},
			wantMatched:    false,
			forceClientErr: true,
			wantErr:        errors.Join(errors.New("example error")),
		},
	}
	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			fc := makeFakeClient(interceptor.Funcs{
				Get: func(ctx context.Context, client client.WithWatch,
					key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
					node, ok := obj.(*corev1.Node)
					if !ok {
						return nil
					}
					if tc.forceClientErr || node == nil {
						return errors.Join(errors.New("example error"))
					}
					*node = *tc.node
					return nil
				}})
			r := &PodReconciler{
				Client: fc.client,
				Scheme: fc.scheme,
				Record: fc.record,
			}

			gotMatched, gotErr := r.validatePodPlacements(context.Background(), tc.leaderPod, tc.podList)
			assert.Equal(t, tc.wantErr, gotErr)
			assert.Equal(t, tc.wantMatched, gotMatched)
		})
	}
}
```

googs1025: done

pkg/controllers/pod_controller_test.go
googs1025: done

pkg/controllers/pod_controller_test.go
danielvegamyhre: We should remove this test entirely.

googs1025: done

googs1025:
By now I think all comments have been addressed. @danielvegamyhre  /PTAL  thank a lot!
Test coverage increased from 0% to 52%.
image

googs1025:
@kannon92 @danielvegamyhre Are there any other comments/messages? If not, I think we can proceed with merging so that I can contribute to other issues. If there are, I'll continue with the modifications!

googs1025:
@danielvegamyhre /PTAL

danielvegamyhre:
Thanks @googs1025, I will review this again soon

danielvegamyhre:
Thanks for working on this @googs1025, in the future when addressing comments could you add a new commit instead of rebasing? This way reviewers can see the diff of what changes you made to address comments. Thanks!

googs1025:
@danielvegamyhre I apologize for the inconvenience. To avoid excessive commits, I have developed a habit of using rebase to reduce the number of commits. I will make sure to pay attention to this issue in the future. Thank you for pointing it out!

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,430 @@
...
+					leaderPodWrapper.AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+					podWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).Obj(),
+				},
+			},
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-node",
+				},
+			},
+			wantMatched: false,
```
danielvegamyhre, line 84: Omit this line and let it default to false. Same everywhere in the file

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,430 @@
...
+					if tc.forceClientErr || pod == nil {
+						return errors.Join(errors.New("example error"))
+					}
+					// This is to solve the problem that the timestamps do not match. There will be a slight gap.
+					pod.Status.Conditions[0].LastTransitionTime =
+						tc.wantPodsDeleted[0].Status.Conditions[0].LastTransitionTime
+					gotPodsDeleted = append(gotPodsDeleted, *pod)
+					return nil
+				},
+				SubResourceUpdate: func(ctx context.Context, client client.Client,
```
danielvegamyhre, line 345: Can you clarify why this was added?

If tc.forceClientError == true, the Delete call above will return an error already. So it's unclear to me why this SubresourceUpdate interceptor func is needed, since it doesn't return anything or seem to do anything else.

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,430 @@
...
+					Status:             corev1.ConditionFalse,
+					Reason:             constants.ExclusivePlacementViolationReason,
+					LastTransitionTime: metav1.Now().Rfc3339Copy(),
+					Message:            constants.ExclusivePlacementViolationMessage,
+				},
+				}).Obj(),
+			},
+			wantPodsDeleted: nil,
+			forceClientErr:  true,
+			wantErr:         errors.Join(errors.Join(errors.New("example error"))),
```
danielvegamyhre, line 308:
```suggestion
			wantErr:         errors.New("example error"),
```

Same for other places with this

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,430 @@
...
+		},
+		{
+			name: "delete follower pods with update pod status error",
+			pods: []corev1.Pod{
+				leaderPodWrapper.AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+				followerPodWrapper.SetConditions([]corev1.PodCondition{{
+					Type:               corev1.DisruptionTarget,
+					Status:             corev1.ConditionFalse,
+					Reason:             constants.ExclusivePlacementViolationReason,
+					LastTransitionTime: metav1.Now().Rfc3339Copy(),
```
danielvegamyhre, line 301: Why is Rfc3339Copy() necessary?

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,430 @@
...
+			wantMatched: false,
+			wantErr:     fmt.Errorf("node does not have topology label: %s", "test-node-topologyKey"),
+		},
+		{
+			name: "valid pod placements with matched",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
+				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
+			podList: &corev1.PodList{
+				Items: []corev1.Pod{
+					leaderPodWrapper.NodeSelector(map[string]string{"test-node-topologyKey": "topologyDomain"}).
```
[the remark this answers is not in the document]
googs1025: done

pkg/controllers/pod_controller_test.go
googs1025: I previously used `errors.New` for handling errors, but I encountered the following error that indicated a type mismatch.
image

pkg/controllers/pod_controller_test.go
googs1025: yes, we don't need it. I will remove it.

pkg/controllers/pod_controller_test.go
googs1025: [Here](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/pod_controller.go#L219), we have the relevant operation to update the pod status. Therefore, we need the SubresourceUpdate interceptor function. Additionally, our test cases also cover scenarios where updating the status results in an error.

pkg/controllers/pod_controller_test.go
danielvegamyhre: I see, what about `fmt.Errorf`, will that work or does it still require a join?

googs1025: use `fmt.Errorf`the same error
image

pkg/controllers/pod_controller_test.go
googs1025: May I ask if you have any other suggestions, please? @danielvegamyhre

pkg/controllers/pod_controller_test.go
danielvegamyhre: ok sounds good

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+		ns               = "default"
+		leaderPodWrapper = makePod(&makePodArgs{
+			jobSetName:        jobSetName,
+			replicatedJobName: "replicated-job-1",
+			jobName:           "test-jobset-replicated-job-1-test-job-0",
+			podName:           "test-jobset-replicated-job-1-test-job-0-0",
+			ns:                ns,
+			nodeName:          "test-node",
+			jobIdx:            0})
+		podWrapper = makePod(&makePodArgs{
```
danielvegamyhre, line 52: Let's call this `followerPodWrapper`

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+		podList        *corev1.PodList
+		node           *corev1.Node
+		wantErr        error
+		forceClientErr bool
+		wantMatched    bool
+	}{
+		{
+			name: "topology node label not found",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
+				AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0").Obj(),
```
danielvegamyhre, line 73: Call `AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0")` just once at the top when you define the variable rather than calling it in every test case:

`leaderPod = makePod(...).AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0")`

Values which are common to the variable for every test case (like completion index) should be set once in the variable definition.

Only values which change depending on the test case should be set in the test case itself.

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+		leaderPod      corev1.Pod
+		podList        *corev1.PodList
+		node           *corev1.Node
+		wantErr        error
+		forceClientErr bool
+		wantMatched    bool
+	}{
+		{
+			name: "topology node label not found",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, "test-node-topologyKey").
```
danielvegamyhre, line 72: Store `"test-node-topologyKey"` in a variable since it is re-used many times in this test. This makes it easier to see what values are supposed to correspond to one another in the tests.

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+					podWrapper.NodeSelector(nil).Obj(),
+				},
+			},
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "test-node",
+					Labels: nodeSelector,
+				},
+			},
+			wantMatched: false,
```
danielvegamyhre, line 124: don't set this explicitly, let it default to `false`

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,430 @@
...
+					Name:   "test-node",
+					Labels: map[string]string{"test-node-topologyKey": "topologyDomain"},
+				},
+			},
+			wantMatched: false,
+			wantErr: fmt.Errorf("pod %s nodeSelector is missing key: %s",
+				"test-jobset-replicated-job-1-test-job-0-1", "test-node-topologyKey"),
+		},
+		{
+			name: "valid pod placements with followerTopology != leaderTopology",
```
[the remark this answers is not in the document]
danielvegamyhre: This has not been addressed

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+			},
+			node: &corev1.Node{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-node",
+				},
+			},
+			wantErr: fmt.Errorf("node does not have topology label: %s", "test-node-topologyKey"),
+		},
+		{
+			name: "valid pod placements with matched",
```
danielvegamyhre, line 88:
```suggestion
			name: "valid pod placements",
```

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "test-node",
+					Labels: nodeSelector,
+				},
+			},
+			wantMatched: false,
+			wantErr:     fmt.Errorf("pod %s nodeSelector is nil", "test-jobset-replicated-job-1-test-job-0-1"),
+		},
+		{
+			name: "valid pod placements with pod nodeSelector is empty",
```
danielvegamyhre, line 128:
```suggestion
			name: "follower pod nodeSelector is empty",
```

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "test-node",
+					Labels: nodeSelector,
+				},
+			},
+			wantErr: fmt.Errorf("follower topology %q != leader topology %q",
+				"topologyDomain1", "topologyDomain"),
+		},
+		{
+			name: "valid pod placements with get node error",
```
danielvegamyhre, line 167: Please don't prefix every test case with "valid pod placements."

"valid pod placements" implies the pod placements are valid, but many of the tests are for invalid pod placements.

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,427 @@
...
+				Get: func(ctx context.Context, client client.WithWatch,
+					key client.ObjectKey, obj client.Object, opts ...client.GetOption) error {
+					node, ok := obj.(*corev1.Node)
+					if !ok {
+						return nil
+					}
+					if tc.forceClientErr || node == nil {
+						return errors.Join(errors.New("example error"))
+					}
+					*node = *tc.node
```
danielvegamyhre, line 199: can you add a comment here saying `"Set returned node value to be the node defined in the test case."`

Same for other interceptor funcs we follower a similar pattern, please add a comment explaining it

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+			ns:                ns,
+			nodeName:          "test-node",
+			jobIdx:            0}).AddAnnotation(batchv1.JobCompletionIndexAnnotation, "0")
+		followerPodWrapper = makePod(&makePodArgs{
+			jobSetName:        jobSetName,
+			replicatedJobName: "replicated-job-1",
+			jobName:           "test-jobset-replicated-job-1-test-job-0",
+			podName:           "test-jobset-replicated-job-1-test-job-0-1",
+			ns:                ns,
+			jobIdx:            0})
```
danielvegamyhre, line 58: Add job completion index of a non-zero number to the follower pod.

`AddAnnotation(batchv1.JobCompletionIndexAnnotation, "1")`

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+		forceClientErr  bool
+	}{
+		{
+			name: "delete follower pods",
+			pods: []corev1.Pod{
+				leaderPodWrapper.Obj(),
+				followerPodWrapper.Obj(),
+			},
+			wantPodsDeleted: []corev1.Pod{
+				wantPodWrapper.ResourceVersion("999").SetConditions([]corev1.PodCondition{{
```
danielvegamyhre, line 248: Can you remind me why we need to explicitly set the resourceVersion here?

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,418 @@
...
+					Type:               corev1.DisruptionTarget,
+					Status:             corev1.ConditionFalse,
+					Reason:             constants.ExclusivePlacementViolationReason,
+					LastTransitionTime: metav1.Now(),
+					Message:            constants.ExclusivePlacementViolationMessage,
+				},
+				}).Obj(),
+			},
+			forceClientErr: true,
+			wantErr:        errors.Join(errors.Join(errors.New("example error"))),
```
danielvegamyhre, line 296: I think there must be a better way to do this than chained joined errors. I'll try to check out this PR locally and figure it out

pkg/controllers/pod_controller_test.go
googs1025: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/fake/client.go#L285
I guess the reason for using a magic number here

pkg/controllers/pod_controller_test.go
googs1025: There is a point of view, maybe we should not care about the error type, because it is an error itself, we only need to know `err != nil` in the code. We can not use `assert.Equal` to check whether the type is completely matched, because it is not important

danielvegamyhre: Yeah that would be fine

googs1025: I changed the error type to`errors.New("example error")`, and change code to the following
```go
if tc.wantErr != nil {
   assert.Equal(t, tc.wantErr.Error(), gotErr.Error())
}
```
We should focus on the error message rather than whether the types are consistent

pkg/controllers/pod_controller_test.go
danielvegamyhre: I see, honestly I think we should just count the number of pod deletion calls made via the Delete() interceptor func. Needing to dive into the weeds of another code base to understand the unit tests isn't worth it for the marginal improvement of the precision of our tests.

Sorry this came up rather late in the review process, I hope you don't mind making the change.

googs1025: No problem, it's also a learning experience for me, haha. I'll make the change later.

googs1025: fix. @danielvegamyhre

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,385 @@
...
+				},
+			},
+			wantErr: fmt.Errorf("node does not have topology label: %s", testTopologyKey),
+		},
+		{
+			name:      "valid pod placements",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, testTopologyKey).Obj(),
+			podList: &corev1.PodList{
+				Items: []corev1.Pod{
+					leaderPodWrapper.NodeSelector(nodeSelector).Obj(),
```
danielvegamyhre, line 91: Leader pod has no nodeSelector, only follower pods do.

validatePodPlacements() looks up the node the leader pod is scheduled on, gets the topology node label from that node, and compares it against the follower pod nodeSelectors.

googs1025: my bad,done

pkg/controllers/pod_controller_test.go
```diff
@@ -0,0 +1,383 @@
...
+		{
+			name:      "get node error",
+			leaderPod: leaderPodWrapper.AddAnnotation(jobset.ExclusiveKey, testTopologyKey).Obj(),
+			podList: &corev1.PodList{
+				Items: []corev1.Pod{
+					leaderPodWrapper.Obj(),
+					followerPodWrapper.NodeSelector(map[string]string{testTopologyKey: "topologyDomain"}).Obj(),
+				},
+			},
+			node: &corev1.Node{
```
danielvegamyhre, line 165: Since we're forcing the client to return an error in this test case, we don't need this node, let's remove it to simplify the test case.

googs1025: sure! done

danielvegamyhre:
/lgtm
/approve

Thanks @googs1025 for improving our test coverage! This was a much needed improvement.

googs1025:
Finally merged! Haha, thank you for the review. It did take some time, but I also learned a lot from it. @danielvegamyhre

PR #494, opened 2024-04-04
Call Status.Update once in each reconcile attempt

### LINKED ISSUE #405: Call Status.Update once in reconcile

`Status().Update` is invoked multiple times in Reconcile, we should refactor the code so that it is called once.

**Why is this needed**:

Reduces calls to the api sever and avoids conflicts in the case status is actually called more that once.

danielvegamyhre:
Fixes #405 

This approach removes the JobSet status update API calls from the controller, and simply modifies the status in memory, and does a single status update call at the end of each reconciliation attempt.

These changes also include:
- Refactoring to improve readability and organization
- Fixing bug causing unnecessary status update calls ([here](https://github.com/kubernetes-sigs/jobset/blob/547c8eb8e3448e7a15223afe047b99caa8390cfa/pkg/controllers/jobset_controller.go#L258) the replicated job statuses must be sorted before doing the semantic equality test)
- Refactor "in order startup policy" condition into 2 mutually exclusive conditions: one indicating the startup policy condition is in progress, and one indicating it has completed.
- Remove `forceFalseUpdate` flag from `conditionOpts` since we are no longer using a condition status of false/true on the in order startup policy condition to indicate if the startup policy is in-progress/completed (see prior bullet point).

**Note**: the operation conflict errors are still in the integration tests. One idea I had is that we are still doing Job status updates in the middle of reconcile logic. These will trigger Jobset reconciles which then update the jobSet.status.replicatedJobStatuses as it looks at those child jobs. The controller runtime work queue should be handling these sequentially though since MaxConcurrency defaults to 1 and we don't override that, so I'm not sure of the root cause yet. Kueue has these errors in the integration test logs as well though so they may be an expected side effect of the integration test framework or someting.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -419,18 +419,20 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 			return nil
 		}
 	}
+
+	// If no jobs were resumed / no action was taken, there's nothing more to do here.
+	if numJobsResumed == 0 {
+		return nil
+	}
 	// At this point all replicated jobs have had their jobs resumed.
 	// If using an in order startup policy. add a condition to the JobSet indicating the
 	// in order startup policy has completed.
 	if inOrderStartupPolicy(startupPolicy) {
 		setInOrderStartupPolicyCompleted(js, updateStatusOpts)
 	}
-
 	// Finally, set the suspended condition on the JobSet to false to indicate
 	// the JobSet is no longer suspended.
-	if numJobsResumed > 0 {
-		setJobSetResumed(js, updateStatusOpts)
-	}
+	setJobSetResumed(js, updateStatusOpts)
 	return nil
 }
 
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -198,10 +198,12 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 			return ctrl.Result{}, err
 		}
 	} else {
-		if err := r.resumeJobsIfNecessary(ctx, js, ownedJobs.active, rjobStatuses, updateStatusOpts); err != nil {
+		requeue, err := r.resumeJobsIfNecessary(ctx, js, ownedJobs.active, rjobStatuses, updateStatusOpts)
+		if err != nil {
 			log.Error(err, "resuming jobset")
 			return ctrl.Result{}, err
 		}
+		return ctrl.Result{Requeue: requeue}, nil
 	}
 	return ctrl.Result{}, nil
 }
@@ -379,7 +381,11 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
 	return nil
 }
 
-func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job, replicatedJobStatuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) error {
+// resumeJobsIfNecessary iterates through each replicatedJob, resuming any suspended jobs if the JobSet
+// is not suspended. Returns a boolean value indicating if the JobSet should be requeued for reconciliation.
+// This is so in-order startup policy can be respected, where after resuming one replicatedJob, we must
+// wait for it to become ready before resuming the next.
+func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job, replicatedJobStatuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) (bool, error) {
 	// Store node selector for each replicatedJob template.
 	nodeAffinities := map[string]map[string]string{}
 	for _, replicatedJob := range js.Spec.ReplicatedJobs {
@@ -394,7 +400,6 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 	}
 
 	startupPolicy := js.Spec.StartupPolicy
-	numJobsResumed := 0
 	// If JobSpec is unsuspended, ensure all active child Jobs are also
 	// unsuspended and update the suspend condition to true.
 	for _, replicatedJob := range js.Spec.ReplicatedJobs {
@@ -409,31 +414,21 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 				continue
 			}
 			if err := r.resumeJob(ctx, job, nodeAffinities); err != nil {
-				return err
+				return false, err
 			}
-			numJobsResumed += 1
 		}
-		// If using in order startup policy, return early and wait for the replicated job to be ready.
-		if numJobsResumed > 0 && inOrderStartupPolicy(startupPolicy) {
+		// If in order startup policy, we need to return early and allow for
+		// this replicatedJob to become ready before resuming the next.
+		if inOrderStartupPolicy(startupPolicy) {
 			setInOrderStartupPolicyInProgressCondition(js, updateStatusOpts)
-			return nil
+			return true, nil
 		}
 	}
 
-	// If no jobs were resumed / no action was taken, there's nothing more to do here.
-	if numJobsResumed == 0 {
-		return nil
-	}
-	// At this point all replicated jobs have had their jobs resumed.
-	// If using an in order startup policy. add a condition to the JobSet indicating the
-	// in order startup policy has completed.
-	if inOrderStartupPolicy(startupPolicy) {
-		setInOrderStartupPolicyCompletedCondition(js, updateStatusOpts)
-	}
 	// Finally, set the suspended condition on the JobSet to false to indicate
 	// the JobSet is no longer suspended.
 	setJobSetResumedCondition(js, updateStatusOpts)
-	return nil
+	return false, nil
 }
 
 func (r *JobSetReconciler) resumeJob(ctx context.Context, job *batchv1.Job, nodeAffinities map[string]map[string]string) error {
@@ -471,7 +466,7 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 
 		status := findReplicatedJobStatus(replicatedJobStatus, replicatedJob.Name)
 
-		// For startup policy, if the job is started we can skip this loop.
+		// For startup policy, if the replicatedJob is started we can skip this loop.
 		// Jobs have been created.
 		if !jobSetSuspended(js) && inOrderStartupPolicy(startupPolicy) && allReplicasStarted(replicatedJob.Replicas, status) {
 			continue
@@ -501,7 +496,7 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
 
 		// If we are using inOrder StartupPolicy, then we return to wait for jobs to be ready.
 		// This updates the StartupPolicy condition and notifies that we are waiting
-		// for this replicated job to finish.
+		// for this replicated job to start up before moving onto the next one.
 		if !jobSetSuspended(js) && inOrderStartupPolicy(startupPolicy) {
 			setInOrderStartupPolicyInProgressCondition(js, updateStatusOpts)
 			return nil
@@ -905,7 +900,7 @@ type conditionOpts struct {
 // setCondition will add a new condition to the JobSet status (or update an existing one),
 // and enqueue an event for emission if the status update succeeds at the end of the reconcile.
 func setCondition(js *jobset.JobSet, condOpts *conditionOpts, updateStatusOpts *statusUpdateOpts) {
-	// Return early if this condition is already set.
+	// Return early if no status update is required for this condition.
 	if !updateCondition(js, condOpts) {
 		return
 	}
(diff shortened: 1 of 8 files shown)


kannon92:
There are two TODOs in this PR. Do you want to address them in the PR? or as part of the review?

pkg/controllers/jobset_controller.go
```diff
@@ -652,16 +664,14 @@ type ensureConditionOpts struct {
...
+	// Return early if this condition is already set.
 	if !updateCondition(opts.jobset, opts.condition, opts.forceFalseUpdate) {
-		return nil
-	}
-	if err := r.Status().Update(ctx, opts.jobset); err != nil {
-		return err
+		return
 	}
-
+	// TODO: should we remove this event emission, since now it is occuring before the actual
```
kannon92, line 672: I'm torn.

Having the events could be helpful but you are correct that they don't occur at the same time..

@ahg-g wdyt?

ahg-g: I think we should create a struct (like `updateStatusOpts`) tracking if status changed and needs to be updated and the event that we want to emit alongside the status update. We return this struct and pass it to the status update function which does the update and emit an event.

pkg/controllers/jobset_controller.go
```diff
@@ -83,7 +84,23 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
-	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(&js))
+	// Store the original status so at the end of the reconciliation attempt,
+	// we can check if it has changed and do all the status updates in a single API call.
+	originalStatus := js.Status.DeepCopy()
```
ahg-g, line 89: I am not sure we need to this, see the other comment.

pkg/controllers/jobset_controller.go
danielvegamyhre: This is ready for another look, I liked the updateStatusOpts idea @ahg-g and so I went ahead and implemented it. I All the status updates that need to be performed are tracked in this struct, as well as events which should be conditionally emitted if the status update API call succeeds.

I tried to document the new types and functions introduced clearly and concisely, please let me know if you have any questions or comments.

pkg/controllers/jobset_controller.go
```diff
@@ -950,9 +884,159 @@ func findJobFailureTime(job *batchv1.Job) *metav1.Time {
...
+			LastTransitionTime: metav1.Now(),
+			Reason:             constants.JobSetResumedReason,
+			Message:            constants.JobSetResumedMessage,
+		},
+	}, updateStatusOpts)
+}
+
+// function parameters for setCondition
+type conditionOpts struct {
+	jobset *jobset.JobSet
```
ahg-g, line 983: we don't need to carry a reference to jobset, we will not be updating the status of more than one jobset in each reconcile and the reference should be available at the time we call whatever function that consumes conditionOpts.

pkg/controllers/jobset_controller.go
```diff
@@ -950,9 +884,159 @@ func findJobFailureTime(job *batchv1.Job) *metav1.Time {
...
+//     no change was made).
+//  2. If a condition of the same type exists but with a different status, update
+//     the condition in place and return true (indicating a condition change was made).
+//  3. If no condition of the given type exists yet, add the new condition if the status
+//     is true or if forceFalseUpdate flag is set to true.
+//
+// The `forceFalseUpdate` flag is needed for the in-order startup policy implementation.
+// We set the startup policy condition status to false to indicate the startup policy is
+// in progress, then set it to true once it has completed.
+func updateCondition(js *jobset.JobSet, condition metav1.Condition, forceFalseUpdate bool) bool {
```
ahg-g, line 1018: pass `conditionOpts` instead of the last two parameters

pkg/controllers/jobset_controller.go
```diff
@@ -264,14 +297,21 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+// replicatedJobs has changed.
+func updateReplicatedJobsStatuses(ctx context.Context, js *jobset.JobSet, statuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) {
 	// Check if status ReplicatedJobsStatus has changed
 	if apiequality.Semantic.DeepEqual(js.Status.ReplicatedJobsStatus, statuses) {
 		return
 	}
-	js.Status.ReplicatedJobsStatus = statuses
+	// Add a new status update to perform at the end of the reconciliation attempt.
+	enqueueStatusUpdate(updateStatusOpts, func(jsStatus *jobset.JobSetStatus) {
+		jsStatus.ReplicatedJobsStatus = statuses
```
ahg-g, line 309: Why not just update js here instead of enqueuing an update? we are doing that for the conditions case, so better follow one pattern.

I suggest we drop the `enqueueStatusUpdate` concept.

danielvegamyhre: Oh I forgot to update `updateCondition` to follow this pattern, but to answer your question - I added it to address t [this](https://github.com/kubernetes-sigs/jobset/pull/494#discussion_r1554189775) which suggested we shouldn't compare the oldStatus and new status to determine if an update should be made, but simply track any status updates that should be made. 

The 2 ways I thought of to avoid a deepCopy and deepEqual were either hold a boolean flag in `updateStatusOpts` indicating if status update call should be made, or hold a list of status update functions and only perform the status update call if we have statusUpdateFuncs to run.

In general I tend to prefer other solutions over ones depending on boolean flags but in retrospect I think in this case this alternative is introducing too much extra complexity which is not justified for the amount of value added by it, so I'll refactor to just use a flag.

pkg/controllers/jobset_controller.go
danielvegamyhre: Moved jobset reference out of conditionOpts into its own function parameter in `setCondition()`, to match the pattern in `updateCondition()`.

pkg/controllers/jobset_controller.go
```diff
@@ -484,30 +497,20 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, ow
...
-				forceFalseUpdate: true,
-				condition:        inOrderStartupPolicyExecutingCondition(),
-			})
+			setInOrderStartupPolicyInProgress(js, updateStatusOpts)
+			return nil
 		}
 	}
 	allErrs := errors.Join(finalErrs...)
 	if allErrs != nil {
-		// Emit event to propagate the Job creation failures up to be more visible to the user.
```
danielvegamyhre, line 497: Note for reveiwers: I removed this event since the one we added in #466 (name of first failed job that led to jobset failure) is more useful than this (and this event potentially got very large as well since it contained all errors from parallel job creations).

danielvegamyhre:
/hold

I'll continue work on this after the v0.5.0 release

pkg/controllers/jobset_controller.go
```diff
@@ -86,7 +102,22 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+	updateStatusOpts := statusUpdateOpts{}
+
+	// Reconcile the JobSet.
+	result, err := r.reconcile(ctx, &js, &updateStatusOpts)
+	if err != nil {
+		return result, err
+	}
+
+	// At the end of this Reconcile attempt, do one API call to persist all the JobSet status changes.
+	return ctrl.Result{}, r.updateJobSetStatus(ctx, &js, &updateStatusOpts)
```
dejanzele, line 115:
```suggestion
	return ctrl.Result{RequeueAfter: result.RequeueAfter}, r.updateJobSetStatus(ctx, &js, &updateStatusOpts)
```

This should fix the failing test, as it currently isn't setting the `RequeueAfter` and the JobSet doesn't get requeued after TTL expiration and we the TTL logic is never executed.

danielvegamyhre: Thanks @dejanzele! I will test this out locally and update the PR If so

danielvegamyhre: It worked, thanks!

pkg/controllers/jobset_controller.go
```diff
@@ -371,54 +405,39 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 		}
 		jobsFromRJob := replicatedJobToActiveJobs[replicatedJob.Name]
 		for _, job := range jobsFromRJob {
+			if !jobSuspended(job) {
+				return nil
+			}
```
ahg-g, line 410: By returning here, aren't we missing to resume the rest of the jobs? did you mean to use `continue`?

pkg/controllers/jobset_controller.go
```diff
@@ -371,54 +405,39 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
...
+		if numJobsResumed > 0 && inOrderStartupPolicy(startupPolicy) {
+			setInOrderStartupPolicyInProgress(js, updateStatusOpts)
+			return nil
 		}
 	}
+
+	// If no jobs were resumed / no action was taken, there's nothing more to do here.
+	if numJobsResumed == 0 {
+		return nil
+	}
```
ahg-g, line 426: Why did we need this? should the set conditions below ensure that the condition is updated only if it really needed to anyway?

ahg-g: Also, for conditions that sets the status to false, `updateCondition` should handle this by simply equating the nonexistence of the condition with one that requests to set it to false.

pkg/controllers/jobset_controller.go
```diff
@@ -951,9 +881,159 @@ func findJobFailureTime(job *batchv1.Job) *metav1.Time {
...
+		}
+	}
+
+	// Condition doesn't exist, add it.
+	js.Status.Conditions = append(js.Status.Conditions, condition)
+	return true
+}
+
+// setJobSetCompleted sets a condition on the JobSet status indicating it has completed.
+func setJobSetCompleted(js *jobset.JobSet, updateStatusOpts *statusUpdateOpts) {
```
ahg-g, line 956: I would suffix this and the functions below with `Condition`

pkg/controllers/jobset_controller.go
```diff
@@ -100,22 +131,22 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
-	// Calculate JobsReady and update statuses for each ReplicatedJob
-	status := r.calculateReplicatedJobStatuses(ctx, &js, ownedJobs)
-	if err := r.updateReplicatedJobsStatuses(ctx, &js, ownedJobs, status); err != nil {
-		log.Error(err, "updating replicated jobs statuses")
-		return ctrl.Result{}, err
+	// Calculate JobsReady and update statuses for each ReplicatedJob.
+	// If status ReplicatedJobsStatus has changed, update the JobSet status.
+	rjobStatuses := r.calculateReplicatedJobStatuses(ctx, js, ownedJobs)
+	if !replicatedJobStatusesEqual(js.Status.ReplicatedJobsStatus, rjobStatuses) {
```
ahg-g, line 143: This check probably fits better inside `updateReplicatedJobsStatuses`

pkg/controllers/jobset_controller.go
```diff
@@ -138,52 +169,44 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 
 	// If any jobs have succeeded, execute the JobSet success policy.
 	if len(ownedJobs.successful) > 0 {
-		completed, err := r.executeSuccessPolicy(ctx, &js, ownedJobs)
-		if err != nil {
-			log.Error(err, "executing success policy")
-			return ctrl.Result{}, err
-		}
+		completed := executeSuccessPolicy(ctx, js, ownedJobs, updateStatusOpts)
 		if completed {
```
ahg-g, line 179:
```suggestion
		if completed := executeSuccessPolicy(ctx, js, ownedJobs, updateStatusOpts); completed {
```

pkg/controllers/jobset_controller.go
danielvegamyhre: >Why did we need this? should the set conditions below ensure that the condition is updated only if it really needed to anyway?

The main reconcile loop doesn't distinguish between a jobset that has never been suspended and a jobset that was previously suspended and now needs to be resumed, so it always enters the `resumeJobSetIfNecessary()` function. This resulted in the "JobSet resumed" condition being set on jobsets that were never suspended, so I checked here if we actually had suspended jobs that were resumed before adding that condition.

pkg/controllers/jobset_controller.go
danielvegamyhre: Woops yes I meant to continue, thanks

pkg/controllers/jobset_controller.go
danielvegamyhre: Done

pkg/controllers/jobset_controller.go
danielvegamyhre: Done

pkg/controllers/jobset_controller.go
danielvegamyhre: Done

pkg/controllers/jobset_controller.go
ahg-g: but then we can fix that via the below suggestion, which should fix it for all cases, not just here.

> Also, for conditions that sets the status to false, `updateCondition` should handle this by simply equating the nonexistence of the condition with one that requests to set it to false.

danielvegamyhre: > Also, for conditions that sets the status to false, updateCondition should handle this by simply equating the nonexistence of the condition with one that requests to set it to false.

If I understand your suggestion correctly, we could do this but we would have to change how we set the "in-order startup policy in progress" condition.  

We currently add a new condition with status of false to track that an in-order startup policy is currently in progress but has not completed yet. Once it completed, we update that false status to true. So if we want to maintain this current logic, we can't treat the condition not existing and a false condition as equivalent, otherwise the initial "in progres" (status = false) condition wouldn't be set.

I recently suggested to @kannon92 we migrate to using 2 separate conditions for tracking "in progress" vs "completed" in-order startup policy, since I personally find the false->true thing a bit unintuitive. What do you think?

ahg-g: Yes, I think we should have two: InProgress and Complete and they should be mutually exclusive. See how we did something similar in LWS: https://github.com/kubernetes-sigs/lws/blob/45854111b905cc2acc474cda34dd9e9fc6a293de/pkg/controllers/leaderworkerset_controller.go#L557

danielvegamyhre: Sounds good, I refactored to split in order startup policy condition into 2 separate conditions, "in progress" and "copmleted." This is ready for another look.

pkg/controllers/jobset_controller.go
```diff
@@ -375,11 +377,15 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
...
-	setJobSetSuspended(js, updateStatusOpts)
+	setJobSetSuspendedCondition(js, updateStatusOpts)
 	return nil
 }
 
-func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job, replicatedJobStatuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) error {
+// resumeJobsIfNecessary iterates through each replicatedJob, resuming any suspended jobs if the JobSet
+// is not suspended. Returns a boolean value indicating if the JobSet should be requeued for reconciliation.
+// This is so in-order startup policy can be respected, where after resuming one replicatedJob, we must
+// wait for it to become ready before resuming the next.
```
ahg-g, line 387: Do we really need to requeue? wouldn't the status update of the child job trigger a requeue anyway?

danielvegamyhre: If there's multiple replicated jobs, and we are resuming one at a time (returning early after each replicatedJob is resumed so we can wait for it to be ready before resuming the next), then only the first status update to "in order startup policy in progress" will actually result in a status update API call, since the next replicated job we resume would attempt to set this same "in progress" status again, which would have no affect, since for duplicate conditions we make no update.

ahg-g: The update to the status of the child jobs when they become ready will trigger a reconcile, i was not referring to the update to jobset status.

danielvegamyhre: Ah I see what you mea now, yes you're right. Updated the code to remove the requeue.

ahg-g:
/lgtm
/approve
/hold cancel

Thanks!

Issue #429, opened 2024-02-16
Comprehensive example task running training workload on GPUs using JobSet


danielvegamyhre:
**What would you like to be added**:
A comprehensive example showing how to run a training workload on GPUs using JobSet. We could have one example per major cloud provider.

**Why is this needed**:
We need more concrete examples to reduce friction of user onboarding. Right now we mostly have toy examples with sleep containers to demonstrate functionality of different features.

danielvegamyhre:
@uroy-personal are you still working on these? If not I am going to unassign them so someone else can work on them.

uroy-personal:
Hi @danielvegamyhre,
Yes I am on it. I need to add the example here right?  
https://github.com/kubernetes-sigs/jobset/blob/main/docs/concepts/README.md 

Also please help me on what content ( example yaml ) to put there. I hope to finish all the open task by this week-end.

danielvegamyhre:
> Hi @danielvegamyhre, Yes I am on it. I need to add the example here right? https://github.com/kubernetes-sigs/jobset/blob/main/docs/concepts/README.md
> 
> Also please help me on what content ( example yaml ) to put there. I hope to finish all the open tasks ( assigned to me ) by this week-end.

Yes, you can reference some examples in the `examples/` directory to help you get started.

danielvegamyhre:
Also note it would be nice in the provisioning step to show example commands for all 3 major cloud providers (AWS, GCP, Azure)

uroy-personal:
> Also note it would be nice in the provisioning step to show example commands for all 3 major cloud providers (AWS, GCP, Azure)

Thanks. I am working on it. Hope to raise the PR in the next few days.

uroy-personal:
Yes @danielvegamyhre , I am on it. I made the changes but found that the above README page removed. Will complete it within this week for sure! Thanks

uroy-personal:
Good Morning @danielvegamyhre ,
Started the ball rolling here. So far I have added the examples present in examples/ into the site concepts page. Where to get the example commands for the cloud providers ( GCP, AWS & Azure ) ? Please help. I will modify the PR again.

uroy-personal:
It seems this issue needs GPU access.  Is there a way to get GPU access @danielvegamyhre ?

danielvegamyhre:
[...]

PR #500, opened 2024-04-08
Comprehensive example task running training workload on GPUs using Jobset

uroy-personal:
Comprehensive example task running training workload on GPUs using Jobset

Addresses issue#: https://github.com/kubernetes-sigs/jobset/issues/429


kannon92:
@uroy-personal The goal of that issue was to have real examples with GPUs. This seems to be restating our examples.

kannon92:
If you don't have access to GPUs, it may be worth leaving this issue open for someone else. I also do not have access to GPUs or the means to acquire them so I am out also.

Our hope is to have a real example with training on a GPU so people can see how this is used in practice.

uroy-personal:
Thanks @kannon92 ,
Let me unassign the issue.

PR #503, opened 2024-04-11
fix: add IsNotFoundErr when get headlessSvc

googs1025:
get client error, there are two types of errors: "NotFound" error and others. If it is a "NotFound" error, it should be created. If it is not, the original error should be returned.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -27,6 +27,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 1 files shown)


kannon92:
Code looks good to me but not sure if this is the logic we want.
@ahg-g @danielvegamyhre WDYT?

pkg/controllers/jobset_controller.go
```diff
@@ -552,29 +553,33 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js
...
-		headlessSvc := corev1.Service{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      subdomain,
-				Namespace: js.Namespace,
-			},
-			Spec: corev1.ServiceSpec{
-				ClusterIP: "None",
-				Selector: map[string]string{
-					jobset.JobSetNameKey: js.Name,
+		if k8serrors.IsNotFound(err) {
```
ahg-g, line 556: can you flip this just so we reduce indentation:

```
if !k8serrors.IsNotFound(err) {
  return err
}
// proceed to create headless service
```

googs1025: done

PR #505, opened 2024-04-12
add publish not ready headless service to jobset

### LINKED ISSUE #504: Allow setting publishNotReadyAddresses=True when creating headless service
**What would you like to be added:**

An option to set publishNotReadyAddresses=True when creating the headless service of pod hostnames.

**Why is this needed:**

The headless service is used to simplify discovery of pod network IDs within a group for distributed computing, in most cases this is not related to any particular readiness status.

kannon92:
Fixes #504 

Adds PublishNotReadyHeadlessService to Jobset.

Defaulting logic sets to true. 

Did one optimization in network defaulting where I moved it out of a loop.


kannon92:
Man this is a rough one..

I am not able to generate CRDs anymore so the integration tests are failing due to the fact that crd are not generated.

Looking into this now.

/hold

Makefile
```diff
@@ -83,6 +83,11 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 	$(CONTROLLER_GEN) \
 		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
 		paths="./api/..."
+
+	$(CONTROLLER_GEN) \
```
kannon92, line 87: I noticed that I wasn’t generating crds anymore. Added this allows me to generate the CRD.

not sure when this broke.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -35,6 +35,8 @@ var TestPodTemplate = corev1.PodTemplateSpec{
 	},
 }
 
+var defaultNetwork = &jobset.Network{EnableDNSHostnames: ptr.To(true), PublishNotReadyAddresses: ptr.To(true)}
```
danielvegamyhre, line 38: Is this only used in `TestJobSetDefaulting`? If so let's move it into TestJobSetDefaulting next to the defaultSuccessPolicy and defaultStartupPolicy variables.

kannon92: done.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -198,6 +198,12 @@ type Network struct {
 	// Defaults to <jobSet.name> if not set.
 	// +optional
 	Subdomain string `json:"subdomain,omitempty"`
+
+	// PublishNotReadyAddresses indicates that any agent
+	// which deals with endpoints for this Service should disregard any indications of ready/not-ready.
+	// Defaults to true if not set
```
ahg-g, line 204: "Indicates if DNS records of pods should be published before the pods are ready. Defaults to True."

kannon92: updated.

ahg-g:
/lgtm

leaving approve to Daniel

danielvegamyhre:
This change makes sense to me, thanks for working on it. 

/approve

Issue #506, opened 2024-04-12
Add some EventRecorder events from jobset


googs1025:
**What would you like to be added**:
I noticed that the events generated by EventRecorder are not comprehensive enough. I plan to add more events in the controller, such as events for handling deletion errors or generating some common errors
**Why is this needed**:
The events generated by EventRecorder can be viewed for more critical events when executing kubectl describe jobset.

googs1025:
@danielvegamyhre @kannon92 Can I attempt to enhance this feature?

googs1025:
I'm not sure if the pod controller also needs the related events generated by EventRecorder. I don't think we would specifically look at the pods generated by the jobset.

danielvegamyhre:
> **What would you like to be added**: I noticed that the events generated by EventRecorder are not comprehensive enough. I plan to add more events in the controller, such as events for handling deletion errors or generating some common errors **Why is this needed**: The events generated by EventRecorder can be viewed for more critical events when executing kubectl describe jobset.

I don't think an event for failing to delete a job is needed. Failing to create jobs is what we care about more; specifically, the first failed job which led to the jobset failing. We have an event emitted for this already.

googs1025:
@danielvegamyhre  I have also added events for the failure to delete a job and the failure to create a headless service. So, the meaning is to keep the latter and remove the former, right?

googs1025:
Done！thanks @danielvegamyhre @kannon92

PR #507, opened 2024-04-12
feat: add event recorder event

googs1025:
refer to: https://github.com/kubernetes-sigs/jobset/issues/506

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -45,11 +45,11 @@ const (
 	AllJobsCompletedMessage = "jobset completed successfully"
 
 	// Event reason used when a Job creation fails.
-	// The event uses the error(s) as the reason.
+	// The event uses the error(s) as the message.
 	JobCreationFailedReason = "JobCreationFailed"
 
 	// Event reason used when a Headless Service creation fails.
-	// The event uses the error(s) as the reason.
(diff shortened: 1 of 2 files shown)


danielvegamyhre:
/ok-to-test

Not sure if we necessarily need to emit an event for this. What are your thoughts @kannon92?

danielvegamyhre:
> > /ok-to-test
> > Not sure if we necessarily need to emit an event for this. What are your thoughts @kannon92?
> 
> What is the behavior if the service creation fails? I think it would be an error in the jobset controller logs which is not the best UX. And someone would have to know that a service was created.
> 
> Events are a better user experience than logs of the controller.

True. Alright let's include this change

pkg/constants/constants.go
```diff
@@ -48,6 +48,10 @@ const (
 	// The event uses the error(s) as the reason.
 	JobCreationFailedReason = "JobCreationFailed"
 
+	// Event reason used when a Headless Service creation fails.
+	// The event uses the error(s) as the reason.
```
danielvegamyhre, line 52:
```suggestion
	// The event uses the error(s) as the message.
```

Can you make the same change for `JobCreationFailedReason` above as well? This is a typo.

danielvegamyhre:
/approve

Thanks!

PR #510, opened 2024-04-13
Update envtest and add back crd generation when updating the api

kannon92:
As I was working on https://github.com/kubernetes-sigs/jobset/pull/505#discussion_r1563622468, I found an interesting bug where my CRDs were no longer being updated.

It was a pain in the butt to detect as the code detects the API but e2e/integration report that the field didn't exist.

We need this back so people could generate CRDs if they updated the API.

And I also update envtest to 1.29.0 as 1.26 is out of support.


danielvegamyhre:
/lgtm
/approve

Thanks for fixing this!

PR #511, opened 2024-04-13
feat: add e2e test for ttl seconds after finished in jobset

dejanzele:
CHANGELOG
* add e2e test for ttl seconds after finished in jobset

This PR is an improvement for https://github.com/kubernetes-sigs/jobset/issues/279

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -105,6 +105,31 @@ var _ = ginkgo.Describe("JobSet", func() {
 			util.JobSetCompleted(ctx, k8sClient, js, timeout)
 		})
 	})
+	ginkgo.When("ttl seconds after finished is set", func() {
+		ginkgo.It("should clean up the completed jobset after configured ttl seconds expire", func() {
+			ctx := context.Background()
+
+			// Create JobSet.
+			testFinalizer := "fake.example.com/blockDeletion"
+			ginkgo.By("creating jobset with ttl seconds after finished")
+			js := sleepTestJobSet(ns).Finalizers([]string{testFinalizer}).TTLSecondsAfterFinished(5).Obj()
+
+			// Verify jobset created successfully.
+			ginkgo.By("checking that jobset creation succeeds")
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
+
+			// Check jobset status if specified.
+			ginkgo.By("checking jobset condition")
+			util.JobSetCompleted(ctx, k8sClient, js, timeout)
+
+			// We remove the jobset finalizer, so it can get deleted when ttl expires.
+			util.RemoveJobSetFinalizer(ctx, k8sClient, js, testFinalizer, timeout)
+
+			// Check jobset is cleaned up after ttl seconds.
+			ginkgo.By("checking jobset is cleaned up after ttl seconds")
+			util.JobSetDeleted(ctx, k8sClient, js, timeout)
+		})
+	})
 
 }) // end of Describe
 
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -105,6 +105,33 @@ var _ = ginkgo.Describe("JobSet", func() {
 			util.JobSetCompleted(ctx, k8sClient, js, timeout)
 		})
 	})
+	ginkgo.When("ttl seconds after finished is set", func() {
+		ginkgo.It("should clean up the completed jobset after configured ttl seconds expire", func() {
+			ctx := context.Background()
+
+			// Create JobSet.
+			ginkgo.By("creating jobset with ttl seconds after finished")
+			js := pingTestJobSetSubdomain(ns).TTLSecondsAfterFinished(5).Obj()
```
danielvegamyhre, line 114: Let's not have the TTL test also be doing an unrelated networking test, this can cause issues/flakiness unrelated to the TTL feature being tested here. We should isolate the feature being tested to reduce flakiness and make debugging test failures easier.

danielvegamyhre (changes requested):
Thanks for working on this, left one comment.

test/e2e/e2e_test.go
```diff
@@ -105,6 +105,33 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+
+			// Verify jobset created successfully.
+			ginkgo.By("checking that jobset creation succeeds")
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
+
+			// We'll need to retry getting this newly created jobset, given that creation may not immediately happen.
+			gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).Should(gomega.Succeed())
```
ahg-g, line 121: if create succeeded above, we shouldn't need to use eventually here

ahg-g: in fact, we don't need to do a get at all

test/e2e/e2e_test.go
```diff
@@ -105,6 +105,33 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+
+			ginkgo.By("checking all jobs were created successfully")
+			gomega.Eventually(util.NumJobs, timeout, interval).WithArguments(ctx, k8sClient, js).Should(gomega.Equal(util.NumExpectedJobs(js)))
+
+			// Check jobset status if specified.
+			ginkgo.By("checking jobset condition")
+			util.JobSetCompleted(ctx, k8sClient, js, timeout)
```
ahg-g, line 128: It is possible that we face a race here where the jobset is deleted before we get a chance to check that it is completed here, I think we should remove this check and the one above, and just do the one below checking that it is deleted.

test/e2e/e2e_test.go
```diff
@@ -105,6 +105,33 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+
+			// Verify jobset created successfully.
+			ginkgo.By("checking that jobset creation succeeds")
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
+
+			// We'll need to retry getting this newly created jobset, given that creation may not immediately happen.
+			gomega.Eventually(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &jobset.JobSet{}), timeout, interval).Should(gomega.Succeed())
+
+			ginkgo.By("checking all jobs were created successfully")
+			gomega.Eventually(util.NumJobs, timeout, interval).WithArguments(ctx, k8sClient, js).Should(gomega.Equal(util.NumExpectedJobs(js)))
```
ahg-g, line 124: pls remove this check, we may face a race where the jobset is completed and marked as deleted before we get here.

test/e2e/e2e_test.go
ahg-g: To avoid flakiness, we need to create the jobset with a finalizer to ensure that the jobset doesn't get deleted before we check that it exists and is complete; and then before we do the `JobSetDeleted` check, we remove the finalizer.

dejanzele:
@danielvegamyhre @ahg-g can you take another look please?

test/e2e/e2e_test.go
```diff
@@ -105,6 +105,35 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+			ginkgo.By("checking that jobset creation succeeds")
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
+
+			// Check jobset status if specified.
+			ginkgo.By("checking jobset condition")
+			util.JobSetCompleted(ctx, k8sClient, js, timeout)
+
+			// We get the latest version of the jobset before removing the finalizer
+			var fresh jobset.JobSet
+			gomega.Expect(k8sClient.Get(ctx, types.NamespacedName{Name: js.Name, Namespace: js.Namespace}, &fresh)).Should(gomega.Succeed())
```
ahg-g, line 127: move this get to inside removeJobSetFinalizer within Eventually (see comment there)

dejanzele:
@ahg-g thanks for the review, I have addressed your comments

ahg-g:
Thanks! one thing I usually do is to force a bug in the test to check if it fails, just to verify that it is working as expected. I wonder if you did that?

dejanzele:
@ahg-g what kind of bug? i.e. the test will fail if you don't get latest version of the jobset before updating as it will fail due to conflict, or it will fail if you delete the jobset manually in the middle of the test run

PR #522, opened 2024-04-16
use kube-openapi rather than code generator openapi-gen

### LINKED ISSUE #514: Move from using openapi-gen
Upstream kube has dropped openapi-gen from 1.30.

We should look into using something else. It should no longer be needed.

kannon92:
In 1.30, k8s.io/code-generator/cmd/openapi-gen was removed from the repo.

It looks like k8s.io/kube-openapi is the new way of doing this (https://github.com/kubernetes/kube-openapi).

https://github.com/kubernetes/kubernetes/commit/1a4450f6ffdad44bfc12fae5911c93e3bd238f0f drops it from kubernetes in 1.30.

So we should move to this new approach.

Fixes #514


danielvegamyhre:
/lgtm
/approve

Thanks!

kannon92:
@danielvegamyhre could you test this and make sure the changes don't break you?

I ran it on my machine but I don't clone code into GOPATH.

danielvegamyhre:
Sure, just checked it out locally and it worked 👍 

/hold cancel

PR #525, opened 2024-04-16
Allow passing args to ginkgo for integration tests

danielvegamyhre:
@jedwins1998 pointed out there wasn't an easy way to run specific integration tests.

This change allows for passing arguments to Ginkgo for the integration tests to help with this issue.

For example, to run only tests with "TTL" in the full concatenated test description, one can use:

`GINKGO_ARGS="--focus TTL"  make test-integration`

Docs: https://onsi.github.io/ginkgo/#description-based-filtering

I tested and confirmed this works, and does not break the make rule when ran without GINKGO_ARGS.


kannon92 (commented):
/lgtm

PR #526, opened 2024-04-16
add unit tests for createHeadlessSvcIfNecessary

dejanzele:
Add unit tests for createHeadlessSvcIfNecessary

### DIFF
--- a/pkg/controllers/jobset_controller_test.go
+++ b/pkg/controllers/jobset_controller_test.go
@@ -15,9 +15,21 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
+
+	"k8s.io/client-go/tools/record"
+
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/interceptor"
+
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/klog/v2/ktesting"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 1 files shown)


dejanzele:
cc @danielvegamyhre

pkg/controllers/jobset_controller_test.go
```diff
@@ -1205,3 +1212,119 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
...
+				Client: fakeClient,
+				Scheme: scheme,
+			}
+
+			// Execute the function under test
+			gotErr := r.createHeadlessSvcIfNecessary(ctx, tc.jobSet)
+			if gotErr != nil {
+				t.Errorf("createHeadlessSvcIfNecessary() error = %v", gotErr)
+			}
+			if tc.expectServiceCreate {
```
kannon92, line 1295: Could we flip this so that we assume this is true and continue?

```suggestion
			if !tc.expectServiceCreate {
			  continue
			}
```

kannon92: Since only 1 test case has this set to true, I don't see why we need a large chunk of this code in this if statement.

pkg/controllers/jobset_controller_test.go
```diff
@@ -1205,3 +1212,119 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
...
+		},
+		{
+			name:                           "service does not exist and should be created, subdomain set",
+			jobSet:                         testutils.MakeJobSet(jobSetName, ns).EnableDNSHostnames(true).NetworkSubdomain("test-subdomain").Obj(),
+			expectServiceCreate:            true,
+			expectServiceName:              "test-subdomain",
+			expectPublishNotReadyAddresses: true,
+		},
+		{
+			name:                           "service does exist and should be created, publishNotReadyAddresses is false",
```
kannon92, line 1251:
```suggestion
			name:                           "service does not exist and should be created, publishNotReadyAddresses is false",
```

pkg/controllers/jobset_controller_test.go
```diff
@@ -1205,3 +1212,121 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
...
+			if tc.existingService {
+				svc := &corev1.Service{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "test-jobset",
+						Namespace: ns,
+					},
+				}
+				fakeClientBuilder.WithObjects(svc)
+			}
+			fakeClient := fakeClientBuilder.Build()
```
danielvegamyhre, line 1282: I think we should use an interceptorFunc in the fake client (as described [here](https://github.com/kubernetes-sigs/jobset/pull/473#discussion_r1537055795)) in order to validate the actual `r.Create(svc)` calls done in the function.

dejanzele:
thanks for the review @danielvegamyhre, I have addressed your comments

pkg/controllers/jobset_controller_test.go
```diff
@@ -1205,3 +1217,152 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
...
+			scheme := runtime.NewScheme()
+			utilruntime.Must(jobset.AddToScheme(scheme))
+			utilruntime.Must(corev1.AddToScheme(scheme))
+			fakeClientBuilder := fake.NewClientBuilder().WithScheme(scheme).WithInterceptorFuncs(interceptor.Funcs{
+				Create: func(ctx context.Context, client client.WithWatch, obj client.Object, opts ...client.CreateOption) error {
+					if tc.expectErr {
+						return errors.New("unexpected error")
+					}
+					if !tc.expectServiceCreate {
+						t.Fatal("unexpected service creation")
```
danielvegamyhre, line 1292: Why not t.Error() here? Don't we want to continue running the rest of the tests?

pkg/controllers/jobset_controller_test.go
```diff
@@ -1205,3 +1217,152 @@ func makeJob(args *makeJobArgs) *testutils.JobWrapper {
...
+						Controller:         ptr.To(true),
+						BlockOwnerDeletion: ptr.To(true),
+					}
+					if diff := cmp.Diff(expectedOwnerRef, svc.OwnerReferences[0]); diff != "" {
+						t.Errorf("unexpected service owner reference value (+got/-want): %s", diff)
+					}
+					if svc.Spec.ClusterIP != corev1.ClusterIPNone {
+						t.Errorf("expected service to have ClusterIP None, got %s", svc.Spec.ClusterIP)
+					}
+					selector, ok := svc.Spec.Selector[jobset.JobSetNameKey]
```
danielvegamyhre, line 1314: The variable name `selector` implies it contains the full key-value pair of a label selector. 

Let's use `selectorValue` or something similar so it's clear it only contains the node label value.

danielvegamyhre (commented):
LGTM after 2 minor comments.

pkg/controllers/jobset_controller_test.go
dejanzele: I am not sure, it is an error if an unexpected service creation happens, but does it make sense to do further assertion on that service, if we didn't expect creaton in the first place?
At the end we also have assertions how many services we expected to be created (1 if we expect, 0 if we don't) which would also get skipped.

I don't have a strong opinion, you can make the decision.

PR #527, opened 2024-04-18
Fix path for the error when attempting to mutate managedBy

mimowo:
Fix for the path in error which I discovered wrong during manual e2e testing.

### DIFF
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -228,7 +228,7 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 	}
 	// Note that SucccessPolicy and failurePolicy are made immutable via CEL.
 	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
-	errs = append(errs, apivalidation.ValidateImmutableField(mungedSpec.ManagedBy, oldJS.Spec.ManagedBy, field.NewPath("spec").Child("labels").Key("managedBy"))...)
+	errs = append(errs, apivalidation.ValidateImmutableField(mungedSpec.ManagedBy, oldJS.Spec.ManagedBy, field.NewPath("spec").Child("managedBy"))...)
 	return nil, errs.ToAggregate()
 }
 
--- a/pkg/webhooks/jobset_webhook_test.go
+++ b/pkg/webhooks/jobset_webhook_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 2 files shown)


kannon92:
/lgtm

We should backport this to 0.5.0.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1097,10 +1120,8 @@ func TestValidateUpdate(t *testing.T) {
 			newObj := tc.js.DeepCopyObject()
 			oldObj := tc.oldJs.DeepCopyObject()
 			_, err = webhook.ValidateUpdate(context.TODO(), oldObj, newObj)
-			if tc.want != nil {
-				assert.True(t, strings.Contains(err.Error(), tc.want.Error()))
-			} else {
-				assert.Nil(t, err)
+			if diff := cmp.Diff(tc.want, err, cmpopts.IgnoreFields(field.Error{}, "BadValue")); diff != "" {
```
danielvegamyhre, line 1123: Why do we need to ignore the BadValue field here

danielvegamyhre (commented):
Thanks for catching this @mimowo!

pkg/webhooks/jobset_webhook_test.go
mimowo: We don't need to ignore, tentatively I ignored it because I got a bigger diff for the test case: "replicated_jobs_are_immutable", and it wasn't asserted before:
```diff
            - 		BadValue: string(""),
            + 		BadValue: []v1alpha2.ReplicatedJob{
            + 			{
            + 				Name:     "test-jobset-replicated-job-0",
            + 				Template: v1.JobTemplateSpec{Spec: v1.JobSpec{...}},
            + 				Replicas: 2,
            + 			},
            + 			{
            + 				Name:     "test-jobset-replicated-job-1",
            + 				Template: v1.JobTemplateSpec{Spec: v1.JobSpec{...}},
            + 				Replicas: 1,
            + 			},
            + 		},
```
I'm happy either way, WDYT?

mimowo: The last commit pushed contains validation of the BadValue. PTAL.

pkg/webhooks/jobset_webhook.go
```diff
@@ -228,7 +228,7 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 	}
 	// Note that SucccessPolicy and failurePolicy are made immutable via CEL.
 	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
-	errs = append(errs, apivalidation.ValidateImmutableField(mungedSpec.ManagedBy, oldJS.Spec.ManagedBy, field.NewPath("spec").Child("labels").Key("managedBy"))...)
+	errs = append(errs, apivalidation.ValidateImmutableField(mungedSpec.ManagedBy, oldJS.Spec.ManagedBy, field.NewPath("spec").Child("managedBy"))...)
```
kannon92, line 231: We can try cherry-picking this but this PR seems to be two in one.

We have a bug fix (which is useful!) and some test cleanup. I'll see if I can get both into but I would usually suggest leaving cleanups out of a bug fix.

kannon92:
/lgtm

Cherry-pick seems fine but just a suggestion for next time.

pkg/webhooks/jobset_webhook.go
mimowo: The test changes are in order to assert on the error path. Previously path it was ignored in the test. I think this is a good practice to accompany each change with a test in the same PR.

One thing potentially optional is asserting on the "BadValue", but it raised a question, so I added: https://github.com/kubernetes-sigs/jobset/pull/527#discussion_r1571040757

pkg/webhooks/jobset_webhook_test.go
danielvegamyhre: Yeah I think the bad value can be helpful to see what value was mutated in the immutable construct, so it's easier to find where the mutation is coming from in the code and fix it.

danielvegamyhre:
/approve

Thanks!

PR #528, opened 2024-04-18
Do not default the managedBy field

mimowo:
1. To be consistent with the batch Job which does not default
2. To let MultiKueue default the field to `kueue.x-k8s.io/multikueue` on the management cluster. Otherwise we would need to mutate it.

Not defaulting is fine, because both nil and the reserved value are handled by the JobSet controller the same way.

### DIFF
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -121,9 +121,6 @@ func (j *jobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
 		js.Spec.Network.PublishNotReadyAddresses = ptr.To(true)
 	}
 
-	if js.Spec.ManagedBy == nil {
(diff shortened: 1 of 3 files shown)


danielvegamyhre:
/lgtm
/approve

Thanks @mimowo!

mimowo:
@danielvegamyhre can we please get this cherry-picked into 0.5.1?

kannon92:
> @danielvegamyhre can we please get this cherry-picked into 0.5.1?

No strong objection but is this really a bug? We should probably wait for the webhook fix before we release a new patch?

mimowo:
> No strong objection but is this really a bug? 

I'm not sure, but it is a deviation from what the batch Job is doing, and we would like to keep consistency in the ecosystem, as much as possible.

>  We should probably wait for the webhook fix before we release a new patch?

Sure.

kannon92:
@mimowo we added this in support of Kueue so I opened a cherry-pick into release 0.5 for that reason.

PR #529, opened 2024-04-18
support elastic jobset

### LINKED ISSUE #463: Support Elastic JobSets

kannon92:
Fixes #463 

Add elastic jobset support (downscale/upscaling replicas of a Replicated Job).

Before this change, we treated ReplicateJobs (in JobSet.Spec) as immutable. This relaxes that restriction.

I choose to validate that names cannot be changed and the job templates must not change during updates. I don't see a reason for names to change and we also don't allow order change of the replicated job list so we can detect differences easily.

### DIFF
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -24,6 +24,7 @@ import (
 
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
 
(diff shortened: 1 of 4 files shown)


danielvegamyhre:
So the use case for this auto-provisioning adding new node pools, and relaxing this validation allows us to potentially autoscale the number of jobs in the replicatedJob to utilize the newly provisioned infrastructure?

danielvegamyhre:
/lgtm

Will leave approval for @ahg-g

kannon92:
@danielvegamyhre @googs1025 im on leave so im not going to be addressing these rebase for a little while. If you want this in, I would suggest maybe taking it up.

PR #531, opened 2024-04-18
Fix bug when checking if a JobSet is active during tests.

### LINKED ISSUE #530: Bug in testing utility function JobSetActive
@jedwins1998 encountered a bug when working on tests for configurable failure policy changes, tests would pass when using `JobSetFailed()` to check the JobSet has a failed condition type with a status of true, but also would pass when `JobSetFailed()` was changed to `JobSetActive()`. 

I think this is because in [JobSetActive](https://github.com/kubernetes-sigs/jobset/blob/9dc4358670f119164517f90d6005bd4742a4ff5d/test/util/util.go#L120) we are simply passing in an empty list of conditions to checkJobSetStatus, which verifies all given conditions are set on the jobset status. 

However, even for a failed JobSet, the number of conditions matching an empty list of conditions will be 0, and the check will return ok, which it should not.

Instead, we should change JobSetActive to verify there are no mutually exclusive conditions set on the JobSet status, like failed or completed.

jedwins1998:
Change `checkJobSetActive` function to check for JobSet conditions `JobSetFailed`, `JobSetComplete`, and `JobSetSuspended` as indicators that the JobSet is not active.

This fixes #530

### DIFF
--- a/test/util/util.go
+++ b/test/util/util.go
@@ -16,6 +16,7 @@ package util
 import (
 	"context"
 	"fmt"
+	"slices"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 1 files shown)


jedwins1998:
cc @danielvegamyhre

jedwins1998:
Update as I realized the code I pushed did not match the definition I left on [this comment](https://github.com/kubernetes-sigs/jobset/issues/530#issuecomment-2065388816) in issue #531.

test/util/util.go
```diff
@@ -119,7 +120,27 @@ func JobSetStartupPolicyNotFinished(ctx context.Context, k8sClient client.Client
...
+	var fetchedJS jobset.JobSet
+	if err := k8sClient.Get(ctx, types.NamespacedName{Namespace: js.Namespace, Name: js.Name}, &fetchedJS); err != nil {
+		return false, err
+	}
+
+	forbiddenTypes := []string{string(jobset.JobSetFailed), string(jobset.JobSetCompleted), string(jobset.JobSetSuspended)}
+	forbiddenStatuses := []metav1.ConditionStatus{metav1.ConditionTrue}
+
+	for _, c := range fetchedJS.Status.Conditions {
+		if slices.Contains(forbiddenTypes, c.Type) && slices.Contains(forbiddenStatuses, c.Status) {
```
danielvegamyhre, line 139: All of these conditions must have a status of true in order to be failed/completed/suspended, so let's just directly check `slices.Contains(forbiddenTypes, c.Type) && c.Status == metav1.ConditionTrue`

jedwins1998: Updated the code.

danielvegamyhre:
/approve

Thanks!

PR #533, opened 2024-04-19
[Release-0.5] Do not default the managedBy field

kannon92:
Cherry pick this into release-0.5.

This is to keep the same behavior as upstream Job and to support Kueue use case. This was a major point of release 0.5 so I think a cherry-pick is useful.

We also need #527 in the patch so we should not release a patch until that PR is also in the release.


mimowo:
LGTM

kannon92:

This and the other managedBy change should probably be released for 0.5.1.

PR #544, opened 2024-04-22
fix: fix ci error caused by typo

googs1025:
- fix ci error caused by typo


googs1025:
We should deal with this PR first. The ci error seems to be wrong because of spelling. @danielvegamyhre @kannon92

danielvegamyhre:
/ok-to-test

@googs1025 Can you run `./hack/update-toc.sh` please? This will update the table of contents based on the changes to the section headers

PR #547, opened 2024-04-25
cleanup duplicate func

googs1025:
/kind cleanup 
- clean duplicate func


danielvegamyhre:
My thinking with having this function called inside the conditions instead of outside, is that it can be expensive to compute for jobsets with a large number of jobs, so maybe we don't want to unconditionally call it even if we won't end up using the result.

Anyone else have thoughts on this? @kannon92

kannon92:
I don't think this is really needed. I like the policy to report performance numbers to really get a fix for it. This could be a premature optimization. If it comes up, its an easy fix but no reason to fix it at the moment.

googs1025:
Understood, I was just a bit confused while reading the code, but now I understand the reason. I'll close it out. Thank you all.

PR #562, opened 2024-05-09
Don't reconcile JobSets with deletion timestamp set

### LINKED ISSUE #561: JobSet controller should not reconcile JobSets with deletion timestamp set (bug when deleting JobSets using foreground cascading deletion policy)
JobSets deleted with foreground cascading deletion policy set the deletion timestamp, and a finalizer is added to the jobset which is only removed once all it's owned resources are deleted.

Right now, the jobset controller isn't currently filtering out jobsets with deletion timestamps set, so the controller is continuing to reconcile the jobset and recreate it's resources that are being deleted.

Unfortunately this went unnoticed until now because users have been using the default deletion policy (background cascading deletion) where the jobset object is deleted immediately and owned resources are cleaned up asynchronously, during which time no jobset object exists to reconcile anymore.

However, a user using foreground deletion to delete jobSets just reported this issue, so we need to add an event filter to prevent the controller from reconciling jobsets with deletion timestamp set.

danielvegamyhre:
Fixes #561 

Confirmed the fix works via manual testing.

Without the changes in this PR, I deployed a JobSet with 100 Jobs and attempted to delete with foreground cascading deletion policy via  `k delete jobset paralleljobs --cascade=foreground`, and observed the behavior described in #561 where the child jobs and services are continually being recreated after being deleted.

With the changes in this PR, I deployed the same JobSet of 100 Jobs and deleted it with foreground cascading deletion policy, and confirmed the foreground deletion worked as expected.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -103,6 +103,11 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, client.IgnoreNotFound(err)
 	}
 
+	// Don't reconcile JobSets marked for deletion.
+	if jobSetMarkedForDeletion(&js) {
+		return ctrl.Result{}, nil
+	}
+
 	// Track JobSet status updates that should be performed at the end of the reconciliation attempt.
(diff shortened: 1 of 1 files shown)


danielvegamyhre:
cc @ahg-g @kannon92

PR #565, opened 2024-05-09
Improve the API generated docs for managedBy

mimowo:
/kind documentation

- mention this is immutable (motivated by the [thread](https://github.com/kubernetes-sigs/kueue/pull/2048#discussion_r1595354688))
- describe validation rules as in k/k for Job [link](https://github.com/kubernetes/kubernetes/blob/57b406a18afc54c84725488e0ca3d4b4cabd61db/pkg/apis/batch/types.go#L472-L484)

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -94,7 +94,18 @@ type JobSetSpec struct {
 	// Suspend suspends all running child Jobs when set to true.
 	Suspend *bool `json:"suspend,omitempty"`
 
-	// ManagedBy is used to indicate the controller or entity that manages a JobSet
+	// ManagedBy is used to indicate the controller or entity that manages a JobSet.
+	// The built-in JobSet controller reconciles JobSets which don't have this
+	// field at all or the field value is the reserved string
+	// `jobset.sigs.k8s.io/jobset-controller`, but skips reconciling JobSets
+	// with a custom value for this field.
+	//
+	// The value must be a valid domain-prefixed path (e.g. acme.io/foo) -
+	// all characters before the first "/" must be a valid subdomain as defined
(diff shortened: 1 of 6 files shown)


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -94,7 +94,18 @@ type JobSetSpec struct {
...
+	// `jobset.sigs.k8s.io/jobset-controller`, but skips reconciling JobSets
+	// with a custom value for this field.
+	//
+	// The value must be a valid domain-prefixed path (e.g. acme.io/foo) -
+	// all characters before the first "/" must be a valid subdomain as defined
+	// by RFC 1123. All characters trailing the first "/" must be valid HTTP Path
+	// characters as defined by RFC 3986. The value cannot exceed 63 characters.
+	// The field is immutable.
+	// +optional
 	ManagedBy *string `json:"managedBy,omitempty"`
```
tenzen-y, line 109: This is just curious. Why we don't use CEL validation instead of webhook validation here?
https://github.com/kubernetes-sigs/jobset/blob/7752a365fa5c9ab497c5b61d4dcd7c9e3609e9cb/pkg/webhooks/jobset_webhook.go#L231

tenzen-y (commented):
LGTM
Thank you!

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: No specific reason for using webhook validation here. We can look into migrating to CEL as a follow up.

danielvegamyhre:
/lgtm
/approve

Thanks for improving the documentation here!

api/jobset/v1alpha2/jobset_types.go
tenzen-y: I see. Thank you.

PR #567, opened 2024-05-11
chore: Upgrade e2e local image

googs1025:
Currently we support k8s versions: v1.27, v1.28, v1.29, so our e2e test image should be upgraded to v1.29.1


danielvegamyhre:
/lgtm
/approve

Thanks!

PR #590, opened 2024-06-01
Propagate Job pod template updates to suspended jobs when resuming

### LINKED ISSUE #586: Propagate JobTemplate updates made to suspended JobSets to child Jobs
Propagate JobTemplate updates made to suspended JobSets to child Jobs

**Why is this needed**:
Kueue is now updating JobSet JobTemplates while the JobSet is suspended. A validation relaxation was implemented in #579 to support this. However, since a suspended JobSet just suspends child Jobs then unsuspends them when the JobSet is resumed, the JobTemplate updates have no affect on the child Jobs. 

To support this, we need to either delete child jobs when a JobSet is suspended and recreate them upon unsuspension, or propagate the JobTemplate update to the existing child Job objects.

cc @alculquicondor

danielvegamyhre:
Fixes #586 

This PR refactors the JobSet controller logic (which previously only propagated Job pod template spec updates to nodeSelectors to child Jobs when resuming them) to also propagate updates to labels, annotations, and tolerations.

It also updates the relevant integration test to verify updates to all of these values are propagated correctly.


danielvegamyhre:
@alculquicondor @ahg-g for review, please take a look when you can :)

alculquicondor:
Thanks! Please cherry-pick to 0.5

Issue #545, opened 2024-04-22
Simpler top level status fields for total, succeeded and failed jobs


shrinandj:
**What would you like to be added**:
Currently, the `status` of a jobset object consists only of two arrays: `conditions` and `replicatedJobsStatus`. This is great for providing detailed status of individual conditions and replicated jobs.

However, it makes referencing the overall status of the jobset a little harder.

Along with the current `conditions` and `replicatedJobsStatus` fields, it will be useful to have at least 3 additional simple fields in the status indicating the overall status of the jobset, similar to K8s jobs:

```
status:
  totalJobs: 2
  succeededJobs: 2
  failedJobs: 0
  conditions:
  - lastTransitionTime: "2024-04-22T18:10:09Z"
....
  replicatedJobsStatus:
  - active: 0
...
```

**Why is this needed**:
We were trying to run Jobsets via Argo-workflows as part of a distributed training setup. Argo-workflows can create [Kubernetes resources](https://argo-workflows.readthedocs.io/en/latest/walk-through/kubernetes-resources/). In order to do this correctly, Argo allows users to specify a `successCondition` and `failureCondition`. These conditions need to be simple `key=value` conditions. Referencing array elements, filtering for specific conditions, etc. is not possible.

If the new fields are available, adding them to Argo or even using any other K8s tools would become that much easier:

```
      successCondition: status.succeededJobs = status.totalJobs
      failureCondition: status.failedJobs > 0
```

kannon92:
Is it not possible to use Conditions for this purpose? We emit a conditon with the job is successful or failed and you could check that.

kannon92:
The reason we have an array of this is because replicated jobs can all finish at different times so we can't really flatten the array to report jobs without that information.

shrinandj:
In some cases, it is not possible to use array notations and filters for specific conditions.

In the case of Argo workflows, they use a label selector to parse the success and failure conditions (which I see only works with `key` = `value` pattern).

Is it possible to have `totalJobs` in status populated when the jobset is created and then populate `succeededJobs` and `failedJobs` when the `AllJobsCompleted` condition is being populated?

kannon92:
@ahg-g @danielvegamyhre WDYT?

shrinandj:
The goal wouldn't be to remove existing `conditions` or `replicatedJobStatuses`. They have to be there to present the complete picture. The top level numerical metrics provide (a) a summary of overall status and (b) usable for cases like the one mentioned above.

[...]

PR #594, opened 2024-06-05
feat: add terminalState to jobset status

googs1025:
Fix: https://github.com/kubernetes-sigs/jobset/issues/545

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -50,8 +50,6 @@ type JobSetConditionType string
 
 // These are built-in conditions of a JobSet.
 const (
-	// JobSetRunning means the job is running.
-	JobSetRunning JobSetConditionType = "Running"
 	// JobSetCompleted means the job has completed its execution.
 	JobSetCompleted JobSetConditionType = "Completed"
 	// JobSetFailed means the job has failed its execution.
@@ -136,9 +134,9 @@ type JobSetStatus struct {
 	// RestartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
-	// Phase of the JobSet.
-	// +kubebuilder:default="Running"
-	Phase string `json:"phase,omitempty"`
+	// TerminalState the state of the JobSet when it finishes execution.
+	// It can be either Complete or Failed. Otherwise, it is empty by default.
+	TerminalState string `json:"terminalState,omitempty"`
 
 	// ReplicatedJobsStatus track the number of JobsReady for each replicatedJob.
 	// +optional
@@ -175,7 +173,7 @@ type ReplicatedJobStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
-// +kubebuilder:printcolumn:name="Phase",JSONPath=".status.phase",type=string,description="Phase of the JobSet"
+// +kubebuilder:printcolumn:name="TerminalState",JSONPath=".status.terminalState",type=string,description="Final state of JobSet"
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
 // +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="JobSet suspended"
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -50,6 +50,8 @@ type JobSetConditionType string
 
 // These are built-in conditions of a JobSet.
 const (
+	// JobSetRunning means the job is running.
+	JobSetRunning JobSetConditionType = "Running"
 	// JobSetCompleted means the job has completed its execution.
 	JobSetCompleted JobSetConditionType = "Completed"
 	// JobSetFailed means the job has failed its execution.
@@ -134,6 +136,10 @@ type JobSetStatus struct {
 	// RestartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// Phase of the JobSet.
+	// +kubebuilder:default="Running"
+	Phase string `json:"phase,omitempty"`
+
 	// ReplicatedJobsStatus track the number of JobsReady for each replicatedJob.
 	// +optional
 	// +listType=map
@@ -169,6 +175,7 @@ type ReplicatedJobStatus struct {
 // +k8s:openapi-gen=true
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
+// +kubebuilder:printcolumn:name="Phase",JSONPath=".status.phase",type=string,description="Phase of the JobSet"
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
 // +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="JobSet suspended"
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -149,7 +149,7 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 
 	// Calculate JobsReady and update statuses for each ReplicatedJob.
 	rjobStatuses := r.calculateReplicatedJobStatuses(ctx, js, ownedJobs)
-	updateReplicatedJobsStatuses(ctx, js, rjobStatuses, updateStatusOpts)
+	updateReplicatedJobsStatuses(js, rjobStatuses, updateStatusOpts)
 
 	// If JobSet is already completed or failed, clean up active child jobs and requeue if TTLSecondsAfterFinished is set.
 	if jobSetFinished(js) {
@@ -185,7 +185,7 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 
 	// If any jobs have succeeded, execute the JobSet success policy.
 	if len(ownedJobs.successful) > 0 {
-		if completed := executeSuccessPolicy(ctx, js, ownedJobs, updateStatusOpts); completed {
+		if completed := executeSuccessPolicy(js, ownedJobs, updateStatusOpts); completed {
 			return ctrl.Result{}, nil
 		}
 	}
@@ -304,7 +304,7 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet) 
 }
 
 // updateReplicatedJobsStatuses updates the replicatedJob statuses if they have changed.
-func updateReplicatedJobsStatuses(ctx context.Context, js *jobset.JobSet, statuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) {
+func updateReplicatedJobsStatuses(js *jobset.JobSet, statuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) {
 	// If replicated job statuses haven't changed, there's nothing to do here.
 	if replicatedJobStatusesEqual(js.Status.ReplicatedJobsStatus, statuses) {
 		return
@@ -630,7 +630,7 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js 
 // executeSuccessPolicy checks the completed jobs against the jobset success policy
 // and updates the jobset status to completed if the success policy conditions are met.
 // Returns a boolean value indicating if the jobset was completed or not.
-func executeSuccessPolicy(_ context.Context, js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) bool {
+func executeSuccessPolicy(js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) bool {
 	if numJobsMatchingSuccessPolicy(js, ownedJobs.successful) >= numJobsExpectedToSucceed(js) {
 		setJobSetCompletedCondition(js, updateStatusOpts)
 		return true
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -630,7 +630,7 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js 
 // executeSuccessPolicy checks the completed jobs against the jobset success policy
 // and updates the jobset status to completed if the success policy conditions are met.
 // Returns a boolean value indicating if the jobset was completed or not.
-func executeSuccessPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) bool {
+func executeSuccessPolicy(_ context.Context, js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) bool {
 	if numJobsMatchingSuccessPolicy(js, ownedJobs.successful) >= numJobsExpectedToSucceed(js) {
 		setJobSetCompletedCondition(js, updateStatusOpts)
 		return true
@@ -867,14 +867,15 @@ func enqueueEvent(updateStatusOpts *statusUpdateOpts, event *eventParams) {
 // function parameters for setCondition
 type conditionOpts struct {
 	eventType string
+	phase     string
 	condition *metav1.Condition
 }
 
-// setCondition will add a new condition to the JobSet status (or update an existing one),
+// setCondition will add a new condition and phase to the JobSet status (or update an existing one),
 // and enqueue an event for emission if the status update succeeds at the end of the reconcile.
 func setCondition(js *jobset.JobSet, condOpts *conditionOpts, updateStatusOpts *statusUpdateOpts) {
-	// Return early if no status update is required for this condition.
-	if !updateCondition(js, condOpts) {
+	// Return early if no status update is required for this condition and phase.
+	if !updateConditionAndPhase(js, condOpts) {
 		return
 	}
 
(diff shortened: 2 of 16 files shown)


googs1025:
```bash 
root@VM-0-9-ubuntu:~/jobset# kubectl get jobset
NAME           STATUS    RESTARTS   COMPLETED   SUSPENDED   AGE
paralleljobs   Running                                      3s
```

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -134,6 +152,8 @@ type JobSetStatus struct {
 	// RestartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// Status of the JobSet.
+	Status string `json:"status,omitempty"`
```
ahg-g, line 156: We should use a different term, like Phase. Status is the name of the whole outer struct.

pkg/controllers/jobset_controller.go
```diff
@@ -314,6 +316,48 @@ func updateReplicatedJobsStatuses(ctx context.Context, js *jobset.JobSet, status
...
+		failedJobs += replicatedJobStatus.Failed
+		completedJobs += replicatedJobStatus.Succeeded
+		suspendedJobs += replicatedJobStatus.Suspended
+	}
+
+	status := jobset.Pending
+	// If all jobs are completed, the status is "successful".
+	// If one job are failed, the status is "failed".
+	// If one job is suspended, the status is "suspended".
+	// Otherwise, the status is "running".
```
ahg-g, line 336: We already have the conditions set, `Phase` should simply be a reflection of that. So the `Phase` should be set in the following place based on the condition  https://github.com/kubernetes-sigs/jobset/blob/390fac27fad21cc79c2f2e7af9e64b53c10f710b/pkg/controllers/jobset_controller.go#L875

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -62,6 +62,24 @@ const (
 	JobSetStartupPolicyCompleted JobSetConditionType = "StartupPolicyCompleted"
 )
 
+// the status of JobSet
+const (
+	// Succeed means the JobSet execution is successful.
```
ahg-g, line 67: We should use the conditions: https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L52

I think phase should take one of four values: completed, failed and suspended and if none of those are true, then set to running

pkg/controllers/jobset_controller.go
googs1025: thanks for reviewing!
I roughly understand the meaning, so I will use the following method to modify it.
like this
```go
// function parameters for setCondition
type conditionOpts struct {
	eventType string
        // new add phase field
        phase.        string
	condition *metav1.Condition
}

```

api/jobset/v1alpha2/jobset_types.go
googs1025: OK, I will modify it in the next few days.

pkg/controllers/jobset_controller.go
googs1025: I might need to think about this again. sorry

api/jobset/v1alpha2/jobset_types.go
googs1025: change to "Phase"

api/jobset/v1alpha2/jobset_types.go
googs1025: I followed the review's suggestion and set the phase to Running by default. Otherwise, the phase is changed according to the conditions.

googs1025:
```go
        // JobSetRunning means the job is running.
	JobSetRunning JobSetConditionType = "Running"
	// JobSetCompleted means the job has completed its execution.
	JobSetCompleted JobSetConditionType = "Completed"
	// JobSetFailed means the job has failed its execution.
	JobSetFailed JobSetConditionType = "Failed"
	// JobSetSuspended means the job is suspended.
	JobSetSuspended JobSetConditionType = "Suspended"
```

The phase is converted according to these four fields. The default is Running

googs1025:
/assign @danielvegamyhre  
Please also help review, thank you!

pkg/controllers/jobset_controller.go
```diff
@@ -897,12 +898,12 @@ func setCondition(js *jobset.JobSet, condOpts *conditionOpts, updateStatusOpts *
 	enqueueEvent(updateStatusOpts, event)
 }
 
-// updateCondition accepts a given condition and does one of the following:
+// updateConditionAndPhase accepts a given condition and does one of the following:
```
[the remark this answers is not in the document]
googs1025: done

shrinandj:
Thanks for doing this @googs1025. I will test this out..

pkg/controllers/jobset_controller.go
```diff
@@ -630,7 +630,7 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js
 // executeSuccessPolicy checks the completed jobs against the jobset success policy
 // and updates the jobset status to completed if the success policy conditions are met.
 // Returns a boolean value indicating if the jobset was completed or not.
-func executeSuccessPolicy(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) bool {
+func executeSuccessPolicy(_ context.Context, js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) bool {
```
danielvegamyhre, line 633: If context isn't needed in this function anymore, we can just remove the parameter

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -134,6 +136,10 @@ type JobSetStatus struct {
 	// RestartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// Phase of the JobSet.
+	// +kubebuilder:default="Running"
```
ahg-g, line 140: this should be in the status, not the spec, and we can't default it.

pkg/controllers/jobset_controller.go
```diff
@@ -867,14 +867,15 @@ func enqueueEvent(updateStatusOpts *statusUpdateOpts, event *eventParams) {
 // function parameters for setCondition
 type conditionOpts struct {
 	eventType string
+	phase     string
```
ahg-g, line 870: We don't need to pass this around, we should be able to compute the phase from the condition in the status update function.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -50,6 +50,8 @@ type JobSetConditionType string
 
 // These are built-in conditions of a JobSet.
 const (
+	// JobSetRunning means the job is running.
+	JobSetRunning JobSetConditionType = "Running"
```
ahg-g, line 54: I don't want to have a "running" phase because we may not be able to track it accurately and this may cause confusion. Users will think that running means all pods are created, scheduled and actually running, and I don't think this is how we are setting a running phase. 

I suggest to limit the phase to the most deterministic statuses: completed, failed and suspended.

@danielvegamyhre if you have any thoughts.

ahg-g: The other suggestion is to not track "Phase" but a "TerminalState", in which case this will have two values: failed and completed (suspended wouldn't be set here because it is not a terminal state).

api/jobset/v1alpha2/jobset_types.go
googs1025: Currently it is set in jobs status
```go
// JobSetStatus defines the observed state of JobSet
type JobSetStatus struct {
	// +optional
	// +listType=map
	// +listMapKey=type

         ....
	// Phase of the JobSet.
	// +kubebuilder:default="Running"
	Phase string `json:"phase,omitempty"`

	
}

```

api/jobset/v1alpha2/jobset_types.go
googs1025: @ahg-g Indeed, we can maintain the TerminalState field and display it only when the status changes to completed or failed, and keep it empty in other cases? Would this be more reasonable?

pkg/controllers/jobset_controller.go
googs1025: If we only use condition calculation, we may need to exclude other JobSetConditionType. I think it would be more convenient to use one more field such as phase or terminalState.

such as: These conditions are not what this field needs to care about
```go
// These are built-in conditions of a JobSet.
const (
	...
	// JobSetSuspended means the job is suspended.
	JobSetSuspended JobSetConditionType = "Suspended"
	// JobSetStartupPolicyInProgress means the StartupPolicy is in progress.
	JobSetStartupPolicyInProgress JobSetConditionType = "StartupPolicyInProgress"
	// JobSetStartupPolicyCompleted means the StartupPolicy has completed.
	JobSetStartupPolicyCompleted JobSetConditionType = "StartupPolicyCompleted"
)

```

api/jobset/v1alpha2/jobset_types.go
googs1025: Maybe we should decide between "TerminalState" or "Phase" @ahg-g @danielvegamyhre

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -134,6 +136,10 @@ type JobSetStatus struct {
 	// RestartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// Phase of the JobSet.
+	// +kubebuilder:default="Running"
+	Phase string `json:"phase,omitempty"`
```
tenzen-y, line 141: @ahg-g @danielvegamyhre @kannon92 Do we really need to introduce the phase field? Having the separate state machine opposed to conditions will bring us high maintenance costs, and they potential cause bugs.

Actually, this approach is not the recommended approach described in the API documentation. 
So, I would recommend adding new condition types instead of a new phase field.

> Some resources in the v1 API contain fields called phase, and associated message, reason, and other status fields. The pattern of using phase is deprecated. Newer API types should use conditions instead. Phase was essentially a state-machine enumeration field, that contradicted [system-design principles](https://git.k8s.io/design-proposals-archive/architecture/principles.md#control-logic) and hampered evolution, since [adding new enum values breaks backward compatibility](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md). Rather than encouraging clients to infer implicit properties from phases, we prefer to explicitly expose the individual conditions that clients need to monitor. Conditions also have the benefit that it is possible to create some conditions with uniform meaning across all resource types, while still exposing others that are unique to specific resource types. See [#7856](http://issues.k8s.io/7856) for more details and discussion.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#typical-status-properties

tenzen-y: Although, at a glance, the phase field is a simpler solution, I don't think so.
Because the phase field will bring us needs to maintain 2 dimensions state machine between conditions and phase, I think that maintaining 2 dimensions state machine is not easy.

ahg-g: The phase should be a direct translation of conditions, it shouldn't represent a different state. I suggest we limit it to the terminal state of the jobset, which I think is what the original ask was about. TerminalState: success|failed, if more details are needed, then one needs to look that up in the conditions.

shrinandj:
>... TerminalState: success|failed, if more details are needed, then one needs to look that up in the conditions.

This would work perfectly..

api/jobset/v1alpha2/jobset_types.go
ahg-g: I would move forward with a TerminalState field that is limited to two values for now: completed and failed

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: > The phase should be a direct translation of conditions, it shouldn't represent a different state. I suggest we limit it to the terminal state of the jobset, which I think is what the original ask was about. TerminalState: success|failed, if more details are needed, then one needs to look that up in the conditions.

+1, I think these terminal states will be easy to implement and maintain. The "running" phase would be much more ambiguous / hard to define and maintain.

api/jobset/v1alpha2/jobset_types.go
googs1025: And we keep it empty in other cases. right？ @ahg-g

shrinandj:
I was able to do the following testing on this PR:

- Built a docker image with the code in this PR and pushed it a local docker registry
- Installed the new CRD and the new controller. The controller came up fine
- Create various types of jobsets that would succeed/fail. Verified that the `status.phase` field in the jobset object was set correctly to `Completed` or `Failed`
- Also liked the `Phase` field being in the output of `kubectl get jobset`

api/jobset/v1alpha2/jobset_types.go
ahg-g: yes, it would only be set when the jobset reached a terminal state.

danielvegamyhre:
> I was able to do the following testing on this PR:
> 
> * Built a docker image with the code in this PR and pushed it a local docker registry
> * Installed the new CRD and the new controller. The controller came up fine
> * Create various types of jobsets that would succeed/fail. Verified that the `status.phase` field in the jobset object was set correctly to `Completed` or `Failed`
> * Also liked the `Phase` field being in the output of `kubectl get jobset`

This is great! Thanks for testing this.

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: > I don't want to have a "running" phase because we may not be able to track it accurately and this may cause confusion. Users will think that running means all pods are created, scheduled and actually running, and I don't think this is how we are setting a running phase.
> 
> I suggest to limit the phase to the most deterministic statuses: completed, failed and suspended.
> 
> @danielvegamyhre if you have any thoughts.

Yes, I think the "running" phase would be tricky to define, easy for users to misinterpret, and difficult to maintain. Sticking with terminal states seems like the right way to go imo, for now at least.

googs1025:
> I was able to do the following testing on this PR:
> 
> * Built a docker image with the code in this PR and pushed it a local docker registry
> * Installed the new CRD and the new controller. The controller came up fine
> * Create various types of jobsets that would succeed/fail. Verified that the `status.phase` field in the jobset object was set correctly to `Completed` or `Failed`
> * Also liked the `Phase` field being in the output of `kubectl get jobset`
Thanks for your test

Later, it will be changed to the TerminalState field and will only have complated and failed, and the others will be empty by default.
I will revise this PR in the next few days. Thank all of you for suggestions.

api/jobset/v1alpha2/jobset_types.go
googs1025: OK, I will collect the comments and make changes. 
If I change it to the TerminalState field, will there still be any problems according to your suggestion? @tenzen-y

googs1025:
I changed the field to terminalState. When the jobset reaches the final state, it will be set to Completed or Failed.
```bash
root@VM-0-16-ubuntu:/home/ubuntu# kubectl get jobset
NAME             TERMINALSTATE   RESTARTS   COMPLETED   SUSPENDED   AGE
max-restarts     Failed          3                                  89m
paralleljobs     Completed                  True                    89m
success-policy   Completed                  True                    89m

```

googs1025:
@ahg-g @danielvegamyhre /PTAL

shrinandj:
Could this be merged and new release be cut with this feature?

danielvegamyhre:
> Could this be merged and new release be cut with this feature?

Yep, will do final review shortly

pkg/controllers/jobset_controller.go
```diff
@@ -941,6 +944,13 @@ func updateCondition(js *jobset.JobSet, opts *conditionOpts) bool {
 		js.Status.Conditions = append(js.Status.Conditions, newCond)
 		shouldUpdate = true
 	}
+
+	// If the jobset is in a terminal state, set the terminal state on the jobset.
+	if opts.terminalState != "" && js.Status.TerminalState != opts.terminalState {
```
danielvegamyhre, line 949: Why did you choose to put the logic updating the terminal state in the function that is responsible for updating the condition, rather than put it in its own function?  The logic doesn't seem to depend on the condition updates here, it just depends on the `opts` passed in, so I don't see why this logic belongs here.

googs1025: My initial thought was that the TerminalState field is actually a subtype of the Conditions field, specifically the JobSetCompleted or JobSetFailed types.  TerminalState can be seen as a special case within Conditions. That's why I grouped these two together and updated them in the same method.

googs1025: If  separating them, we can distinguish them into updateCondition and updateTerminalState methods.
```go
// setCondition will add a new condition and terminalState to the JobSet status (or update an existing one),
// and enqueue an event for emission if the status update succeeds at the end of the reconcile.
func setCondition(js *jobset.JobSet, condOpts *conditionOpts, updateStatusOpts *statusUpdateOpts) {
	// Return early if no status update is required for this condition and terminalState.
	if !updateCondition(js, condOpts) && !updateTerminalState(js, condOpts) {
		return
	}

	if updateStatusOpts == nil {
		updateStatusOpts = &statusUpdateOpts{}
	}

	...
}

```

ahg-g: I suggest to set status.TerminalState in the same functions where we set the completed or failed condition. Meaning in `setJobSetFailedCondition` and `setJobSetCompletedCondition`

```
func setJobSetCompletedCondition(js *jobset.JobSet, updateStatusOpts *statusUpdateOpts) {
	setCondition(js, makeCompletedConditionsOpts(), updateStatusOpts)
        jobset.Status.TerminalState = JobSetCompleted
}
```

```
func setJobSetFailedCondition(ctx context.Context, js *jobset.JobSet, reason, msg string, updateStatusOpts *statusUpdateOpts) {
	setCondition(js, makeFailedConditionOpts(reason, msg), updateStatusOpts)
        jobset.Status.TerminalState = JobSetFailed
}
```

and that is all, we don't need to track anything in updateStatusOpts

googs1025: Thanks for the suggestion, this makes the code much simpler!  done

ahg-g: Please add a jobSetTerminalState function to https://github.com/kubernetes-sigs/jobset/blob/main/test/util/util.go :

```
func checkJobSetTerminalState(ctx context.Context, k8sClient client.Client, js *jobset.JobSet, state string) (bool, error) {
	var fetchedJS jobset.JobSet
	if err := k8sClient.Get(ctx, types.NamespacedName{Namespace: js.Namespace, Name: js.Name}, &fetchedJS); err != nil {
		return false, err
	}
        return fetchedJS.Status.TerminalState == state
}
```

and call it from [JobSetCompleted](https://github.com/kubernetes-sigs/jobset/blob/main/test/util/util.go#L55)  to and [JobSetFailed](https://github.com/kubernetes-sigs/jobset/blob/main/test/util/util.go#L66) to verify the terminalState during integration tests.

googs1025: ok，i will save it soon

googs1025: done

ahg-g:
/lgtm
/approve



This is great, thank you for your patience during the review process!

tenzen-y:
Thank you or moving this forward!

PR #609, opened 2024-06-28
feat: add component config

rainfd:
Fix: #55 
The basic configuration follows kueue.


danielvegamyhre:
Thanks for working on this @rainfd, I will review this soon

main.go
```diff
@@ -79,39 +88,53 @@ func main() {
...
+		kubeConfig.Burst = burst
+	} else {
+		opts, cfg, err := apply(configFile)
+		if err != nil {
+			setupLog.Error(err, "Unable to load the configuration")
+			os.Exit(1)
+		}
+		options = opts
+		kubeConfig.QPS = *cfg.ClientConnection.QPS
+		kubeConfig.Burst = int(*cfg.ClientConnection.Burst)
```
googs1025, line 134: If the user sets both the qps flag and the configFile flag, will the qps flag not take effect? ​​Have we considered this situation?

rainfd: It wouldn't take effect. Only configFile and featuregate flag is support in kueue currently. 
Considering supporting both configuration files and command-line arguments requires to add spf13/viper like dependencies.
I just want to keep it simple.

danielvegamyhre:
FYI I am OOO for the next week or so but can review when I get back.

main.go
googs1025: Thanks for your reply. so, config is the first priority. Once configured, flag will not take effect.

rainfd: I add this comment to the description.

api/config/v1alpha1/configuration_types.go
```diff
@@ -0,0 +1,133 @@
...
+	metav1.TypeMeta `json:",inline"`
+
+	// Namespace is the namespace in which jobset controller is deployed.
+	// It is used as part of DNSName of the webhook Service.
+	// If not set, the value is set from the file /var/run/secrets/kubernetes.io/serviceaccount/namespace
+	// If the file doesn't exist, default value is kueue-system.
+	Namespace *string `json:"namespace,omitempty"`
+
+	// ControllerManager returns the configurations for controllers
+	ControllerManager `json:",inline"`
```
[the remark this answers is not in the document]
rainfd: Your guess is right. The basic structure come from kueue, and I think it was clear, so I didn't change it

danielvegamyhre:
Looks good, I would prefer to include a larger change like this in 0.7 release, rather than as a last minute addition to 0.6, so unless there's any objections that's what we'll do

danielvegamyhre:
/lgtm

Will leave approval for @ahg-g.

main.go
```diff
@@ -80,41 +90,55 @@ func main() {
...
-		// the manager stops, so would be fine to enable this option. However,
-		// if you are doing or is intended to do any operation such as perform cleanups
-		// after the manager stops then its usage might be unsafe.
-		// LeaderElectionReleaseOnCancel: true,
-	})
+	var options manager.Options
+
+	kubeConfig := ctrl.GetConfigOrDie()
+	if configFile == "" {
+		options = ctrl.Options{
```
ahg-g, line 104: Why don't we get the defaults as defined in SetDefaults_Configuration and then override the couple of parameters that have corresponding cmd flags if set? This will unify all defaulting in one place.

pkg/config/config.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+
+	codecs := serializer.NewCodecFactory(scheme, serializer.EnableStrict)
+
+	// Regardless of if the bytes are of any external version,
+	// it will be read successfully and converted into the internal version
+	return runtime.DecodeInto(codecs.UniversalDecoder(), content, cfg)
+}
+
+func addTo(o *ctrl.Options, cfg *configapi.Configuration) {
```
ahg-g, line 31: please add a comment explaining what this does, this applies the configuration to the option struct, right?

pkg/config/config.go
```diff
@@ -0,0 +1,139 @@
...
+	content, err := os.ReadFile(path)
+	if err != nil {
+		return err
+	}
+
+	codecs := serializer.NewCodecFactory(scheme, serializer.EnableStrict)
+
+	// Regardless of if the bytes are of any external version,
+	// it will be read successfully and converted into the internal version
+	return runtime.DecodeInto(codecs.UniversalDecoder(), content, cfg)
```
ahg-g, line 28: Will this invoke SetDefaults_Configuration to populate the default values of the fields that are not set.

main.go
ahg-g: `config.Load` handles the case where the configFile is not set, so we just need to invoke it instead of defining defaults again here.

ahg-g (commented):
Thanks a lot for this!

pkg/config/config.go
rainfd: Yes, it will call defaulter before assigning value.

pkg/config/config.go
rainfd: already added

main.go
rainfd: Modify as you said, now check each flag individually and set the value.

ahg-g:
Thanks, looks good, we need to fix the tests though.

kannon92:
/hold cancel

I think this is ready to go now.

Issue #617, opened 2024-07-15
Support for defining a global coordinator pod in the JobSet spec


danielvegamyhre:
**What would you like to be added**:
Support for defining a global coordinator pod in the JobSet spec. 

**Why is this needed**:
We need to be able to build automation on top of JobSet which knows the stable network endpoint of the pod assigned to be the global coordinator distributed ML training / HPC workloads.

danielvegamyhre:
Adding coordinator field and controller changes: https://github.com/kubernetes-sigs/jobset/pull/618

Adding validation: https://github.com/kubernetes-sigs/jobset/pull/627

Adding runnable example: https://github.com/kubernetes-sigs/jobset/pull/628

PR #618, opened 2024-07-15
Add Coordinator field to JobSet spec

danielvegamyhre:
Part of #617. I plan on adding validation logic for this new field in a follow up PR, as well as more tests.

- Add coordinator field to JobSet spec, allowing the user to define the ReplicatedJob, JobIndex, and PodIndex of the pod they wish to be the global coordinator.
- Add a new label/annotation, `jobset.sigs.k8s.io/coordinator`, which will store the stable network endpoint where the coordinator pod can be reached, when the `jobset.spec.coordinator` field is defined.
- Minor refactoring of test code so it is more flexible, allowing us to do a "set or merge" operation when adding labels/annotations to Job or Pod wrappers.


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -44,6 +44,11 @@ const (
 	// JobSetControllerName is the reserved value for the managedBy field for the built-in
 	// JobSet controller.
 	JobSetControllerName = "jobset.sigs.k8s.io/jobset-controller"
+
+	// CoordinatorKey is used as an annotation and label on Jobs and Pods. If the JobSet spec
+	// defines the .spec.coordinator field, this annotation/label will be added to store a stable
```
googs1025, line 49: Hello, I have a small question, why do we need to put this tag on both the annotation and the label? Is it not enough to just use the label?

danielvegamyhre: Yes but inconsistency between labels vs annotations and Job labels/annotations vs Pod labels/annotations has led to bugs in the past, as well as confusion for users. So while the API is still in alpha I've just been added everything as both a label and annotation to both Jobs and Pods, so there is full consistency across Jobs, pods, labels, and annotations. Then for graduation to v1 I plan on reviewing these and assigning them more thoughtfully, and documenting them in the JobSet docsite.

googs1025: Understood, thanks for your reply

googs1025:
I have a small question: Does this feature require the use of featureGate?

danielvegamyhre:
> I have a small question: Does this feature require the use of featureGate?

i don't think we'll add a feature gate for this one since it is isn't a risky feature. It just sets a new annotation and label on jobs and pods if the field is set.

danielvegamyhre:
cc @kannon92 would you ming taking a look at this if you have time? Abdullah is OOO currently. 

Happy to discuss the feature further if you'd like.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -323,6 +336,22 @@ type StartupPolicy struct {
...
 
+// Coordinator defines which pod can be marked as the coordinator for the JobSet workload.
+type Coordinator struct {
+	// ReplicatedJob is the name of the ReplicatedJob which contains
+	// the coordinator pod.
+	ReplicatedJob string `json:"replicatedJob"`
+
+	// JobIndex is the index of Job which contains the coordinator pod
+	// (i.e., for a ReplicatedJob with N replicas, there are Job indexes 0 to N-1).
+	// Defaults to 0 if unset.
```
kannon92, line 347: Let's drop the comment about "Defaults". Undefined ints are 0 by default so its not necessary to include that.

pkg/controllers/jobset_controller.go
```diff
@@ -746,6 +746,12 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
 	annotations[jobset.JobIndexKey] = strconv.Itoa(jobIdx)
 	annotations[jobset.JobKey] = jobHashKey(js.Namespace, jobName)
 
+	// Apply coordinator annotation/label if a coordinator is defined in the JobSet spec.
+	if js.Spec.Coordinator != nil {
+		labels[jobset.CoordinatorKey] = coordinatorEndpoint(js)
```
kannon92, line 751: One subtle bug here would be that if someone specifies these labels we would overwrite.

Do we want a warning if someone uses these labels?

sdk/python/docs/JobsetV1alpha2Coordinator.md
```diff
@@ -0,0 +1,13 @@
+# JobsetV1alpha2Coordinator
+
+Coordinator defines which pod can be marked as the coordinator for the JobSet workload.
+## Properties
+Name | Type | Description | Notes
+------------ | ------------- | ------------- | -------------
+**job_index** | **int** | JobIndex is the index of Job which contains the coordinator pod (i.e., for a ReplicatedJob with N replicas, there are Job indexes 0 to N-1). Defaults to 0 if unset. | [optional] 
+**pod_index** | **int** | PodIndex is the Job completion index of the coordinator pod. Defaults to 0 if unset. | [optional] 
+**replicated_job** | **str** | ReplicatedJob is the name of the ReplicatedJob which contains the coordinator pod. | [default to '']
```
kannon92, line 9: Where do we get "default to ''" from?

kannon92:
/hold

/lgtm

Code looks good. I have some minor nits. Feel free to unhold.

api/jobset/v1alpha2/jobset_types.go
kannon92: I’m a little confused about the string field and its message in the python api about the null string.

I don’t think this is pressing.

pkg/controllers/jobset_controller.go
danielvegamyhre: Yep, I have a follow up PR with webhook changes ready to go after this is merged. Will tag you on it shortly.

api/jobset/v1alpha2/jobset_types.go
danielvegamyhre: Removed comment about defaults.

danielvegamyhre:
@kannon92 can you reapply LGTM when you have a sec please? I made changes to address the comments and it was removed

PR #620, opened 2024-07-22
WIP: Used fixed API for Job Pod Failure Policy

mimowo:
Bump the k8s/api to use the cherry-picks of the fix in https://github.com/kubernetes/kubernetes/pull/126046 - it is going to released in 1.30.4 and 1.29.8, so for now using hash pointing to https://github.com/kubernetes/api/commit/7c71f3c0b7a73274bcbe375b1450eea5c79ad530, which includes https://github.com/kubernetes/api/commit/a63486cfc05ed9b883d9fa0c4c95dc026042b24c.

WIP because we probably want to wait until full release.


kannon92:
LGTM but I think it would be best to wait for official patch release also.

@danielvegamyhre any objections?

danielvegamyhre:
> LGTM but I think it would be best to wait for official patch release also.
> 
> @danielvegamyhre any objections?

I would prefer to watch for an official patch release as well, using a feature branch is a bit hacky and I want to prioritize production readiness and good engineering practices. The target date for 1.30.4 is 08-13-2024 which isn't an unreasonably long time. This bug is blocking the main use case of JobSet's new configurable failure policy (i.e., pairing JobSet failure policy rules with Pod failure policy rules to define more targeted restart/failure handling behavior), but I think we can wait until mid August for 0.6.0 release, and in the meantime perhaps include other important changes in 0.6.0, such as #618

danielvegamyhre:
Closing this in favor of waiting for proper patch release, see https://github.com/kubernetes-sigs/jobset/pull/620#issuecomment-2251114708

mimowo:
> Closing this in favor of waiting for proper patch release, see [#620 (comment)](https://github.com/kubernetes-sigs/jobset/pull/620#issuecomment-2251114708)

Sure, makes sense. I opened it mostly to demonstrate e2e that the coming fix works. For the fix you can either use patch release of 1.30.x or 1.29.x.

Issue #579, opened 2024-05-24
Relax validation rules enforcing immutability of replicated jobs


danielvegamyhre:
**What would you like to be added**:
Relax validation rules enforcing immutability of replicated jobs. Specifically:
- `.spec.replicatedJobs[*].spec.replicas` (to support elastic jobsets https://github.com/kubernetes-sigs/jobset/issues/463 )
- `.spec.replicatedJobs[*].spec.template.spec.template` (job podTemplateSpec, to add annotations, taint tolerations, etc)

**Why is this needed**:
DWS compatibility with JobSet. cc @alculquicondor for more context

danielvegamyhre:
Planning on including this in a patch release early next week. cc @ahg-g

kannon92:
I would vote to put it in a new release. It’s not really something I would think to be included in a patch release.

danielvegamyhre:
Reduced scope of bug specifically to replicatedJob pod template since #529 already exists

danielvegamyhre:
> I would vote to put it in a new release. It’s not really something I would think to be included in a patch release.

I reduced the scope of the changes to be the validation fix only, and postponed the elastic jobset support to be included in v0.6.0. So I think we can go ahead and do a patch release for this.

Issue #482, opened 2024-03-27
Investigate using Elastic Index Job in exclusive placement implementation


danielvegamyhre:
Right now, the implementation of exclusive job placement per topology domain (node pool, zone, etc) relies on a pod webhook which allows leader pods (index 0) to be admitted, created, and scheduled, but blocks follower pods (all other indexes) until the leader pod for that child Job is scheduled. 

The leader pods have pod affinity/anti-affinity constraints ensuring each leader pod lands in a different topology. Follower pods have nodeSelectors injected by the pod mutating webhook ensuring the land on the same topology as their leader.

This is an improvement over the original implementation of using pod affinity/anti-affinity constraints on *all* pods (which did not scale well due to the pod affinity rule computation time scaling linearly with the number of pods and nodes). However, the repeated wasted follower pod creation attempts are putting unnecessary pressure on the apiserver.

One possible option is to use [Elastic Indexed Jobs](https://github.com/kubernetes/enhancements/issues/3715) to first create every Job with completions == parallelism == 1 (so this will only create index 0 / leader pods). The pod webhook will still inject pod affinity/anti-affinity constraints into the leader pods as it currently does.

Once a leader pod for a given Job is scheduled, resize the Job to have completions == parallelism == <desired number in the spec>. The follower pods will then be created and have the nodeSelector injected to follow the leader, as it currently does. This will minimize pressure on the apiserver by avoiding unnecessary pod creation attempts.

kannon92:
https://github.com/kubernetes-sigs/jobset/issues/463

I did look into supporting elastic jobs with jobset. I opened up a PR but I found out that we enforce immutability on all replicated jobs.

I'm still waiting for some comments on what kind of validation we would want for replicated jobs.

danielvegamyhre:
> #463
> 
> I did look into supporting elastic jobs with jobset. I opened up a PR but I found out that we enforce immutability on all replicated jobs.
> 
> I'm still waiting for some comments on what kind of validation we would want for replicated jobs.

We don't need to mutate replicatedJobs for this issue, the controller would be performing this logic on individual Jobs.

The thing you're exploring is different (elastic replicated jobs, enabling number of replicas to scale up or down). This issue is scaling the number of Job pods up after creation, which doesn't require mutating the replicatedJob.

danielvegamyhre:
> Yikes.. Thats a good distinction but I'm not sure we want to support that in that way..
> 
[...]

PR #622, opened 2024-07-22
support elastic jobset

kannon92:
Fixes https://github.com/kubernetes-sigs/jobset/issues/463

Add elastic jobset support (downscale/upscaling replicas of a Replicated Job).

Before this change, we treated ReplicateJobs (in JobSet.Spec) as immutable. This relaxes that restriction.

I choose to validate that names cannot be changed and the job templates must not change during updates. I don't see a reason for names to change and we also don't allow order change of the replicated job list so we can detect differences easily.

I am reopening this to resolve the rebase labels..

### DIFF
--- a/pkg/controllers/elastic_jobset.go
+++ b/pkg/controllers/elastic_jobset.go
@@ -15,6 +15,7 @@ package controllers
 
 import (
 	"fmt"
+	"slices"
 	"strconv"
 
 	batchv1 "k8s.io/api/batch/v1"
@@ -22,29 +23,64 @@ import (
 	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
 )
 
-// jobsToDeleteDownScale gathers the excess jobs during a downscale
+func indexFunc(a, b batchv1.Job) int {
+	jobIndexA, errA := strconv.Atoi(a.Labels[jobset.JobIndexKey])
+	jobIndexB, errB := strconv.Atoi(b.Labels[jobset.JobIndexKey])
+	if errA != nil {
+		return 0
+	}
+	if errB != nil {
+		return 0
+	}
+	if jobIndexA > jobIndexB {
+		return 1
+	} else if jobIndexA < jobIndexB {
+		return -1
+	} else {
+		return 0
+	}
+}
+
+// jobsToDeleteForDownScale gathers the excess jobs during a downscale
 // and deletes the jobs
-func jobsToDeleteDownScale(replicatedJobs []jobset.ReplicatedJob, replicatedJobStatus []jobset.ReplicatedJobStatus, jobItems []batchv1.Job) ([]*batchv1.Job, error) {
+func jobsToDeleteForDownScale(replicatedJobs []jobset.ReplicatedJob, replicatedJobStatuses []jobset.ReplicatedJobStatus, jobItems []batchv1.Job) ([]*batchv1.Job, error) {
 	jobsToDelete := []*batchv1.Job{}
+	type payload struct {
+		batchJobs []batchv1.Job
+		rjStatus  jobset.ReplicatedJobStatus
+		replicas  int32
+	}
+	replicatedJobToBatchJobMap := map[string]payload{}
 	for _, replicatedJob := range replicatedJobs {
-		status := findReplicatedJobStatus(replicatedJobStatus, replicatedJob.Name)
-		countOfJobsToDelete := status.Ready - replicatedJob.Replicas
+		status := findReplicatedJobStatus(replicatedJobStatuses, replicatedJob.Name)
+		newPayload := &payload{}
+		newPayload.rjStatus = status
+		newPayload.replicas = replicatedJob.Replicas
+		for _, val := range jobItems {
+			if val.Labels[jobset.ReplicatedJobNameKey] != replicatedJob.Name {
+				continue
+			}
+			newPayload.batchJobs = append(newPayload.batchJobs, val)
+		}
+		slices.SortFunc(newPayload.batchJobs, indexFunc)
+		replicatedJobToBatchJobMap[replicatedJob.Name] = *newPayload
+	}
+	for _, jobAndStatus := range replicatedJobToBatchJobMap {
+		countOfJobsToDelete := jobAndStatus.rjStatus.Ready - jobAndStatus.replicas
 		if countOfJobsToDelete > 0 {
 			jobsWeDeleted := 0
-			for _, val := range jobItems {
-				if val.Labels[jobset.ReplicatedJobNameKey] != replicatedJob.Name {
-					continue
-				}
-				jobIndex, err := strconv.Atoi(val.Labels[jobset.JobIndexKey])
+			for i := len(jobAndStatus.batchJobs) - 1; i >= 0; i-- {
+
+				jobIndex, err := strconv.Atoi(jobAndStatus.batchJobs[i].Labels[jobset.JobIndexKey])
 				if err != nil {
 					return nil, fmt.Errorf("unable get integer from job index key")
 				}
 				if jobIndex >= int(countOfJobsToDelete) {
 					jobsWeDeleted = jobsWeDeleted + 1
-					jobsToDelete = append(jobsToDelete, &val)
+					jobsToDelete = append(jobsToDelete, &jobAndStatus.batchJobs[i])
 				}
 				if jobsWeDeleted == int(countOfJobsToDelete) {
-					continue
+					break
 				}
 			}
 		}
--- a/pkg/controllers/elastic_jobset.go
+++ b/pkg/controllers/elastic_jobset.go
@@ -1,0 +1,53 @@
+/*
+Copyright 2024 The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"fmt"
+	"strconv"
+
+	batchv1 "k8s.io/api/batch/v1"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+)
+
+// jobsToDeleteDownScale gathers the excess jobs during a downscale
+// and deletes the jobs
+func jobsToDeleteDownScale(replicatedJobs []jobset.ReplicatedJob, replicatedJobStatus []jobset.ReplicatedJobStatus, jobItems []batchv1.Job) ([]*batchv1.Job, error) {
+	jobsToDelete := []*batchv1.Job{}
+	for _, replicatedJob := range replicatedJobs {
+		status := findReplicatedJobStatus(replicatedJobStatus, replicatedJob.Name)
+		countOfJobsToDelete := status.Ready - replicatedJob.Replicas
+		if countOfJobsToDelete > 0 {
+			jobsWeDeleted := 0
+			for _, val := range jobItems {
+				if val.Labels[jobset.ReplicatedJobNameKey] != replicatedJob.Name {
+					continue
+				}
+				jobIndex, err := strconv.Atoi(val.Labels[jobset.JobIndexKey])
+				if err != nil {
+					return nil, fmt.Errorf("unable get integer from job index key")
+				}
+				if jobIndex >= int(countOfJobsToDelete) {
+					jobsWeDeleted = jobsWeDeleted + 1
+					jobsToDelete = append(jobsToDelete, &val)
+				}
+				if jobsWeDeleted == int(countOfJobsToDelete) {
+					continue
+				}
+			}
+		}
+	}
+	return jobsToDelete, nil
+}
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -536,7 +536,7 @@ func (r *JobSetReconciler) downscaleElasticJobs(ctx context.Context, js *jobset.
 		return err
 	}
 
-	jobsToDelete, err := jobsToDeleteDownScale(js.Spec.ReplicatedJobs, replicatedJobStatus, childJobList.Items)
+	jobsToDelete, err := jobsToDeleteForDownScale(js.Spec.ReplicatedJobs, replicatedJobStatus, childJobList.Items)
 	if err != nil {
 		return err
 	}
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -203,6 +203,13 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 		return ctrl.Result{}, err
 	}
 
+	// ElasticJobSet can downscale and upscale Jobs of a given replicated job
+	// On downscale, we need to gather the jobs to delete.
+	if err := r.downscaleElasticJobs(ctx, js, rjobStatuses); err != nil {
+		log.Error(err, "unable to downscale elastic jobs")
+		return ctrl.Result{}, err
+	}
+
 	// Handle suspending a jobset or resuming a suspended jobset.
 	jobsetSuspended := jobSetSuspended(js)
 	if jobsetSuspended {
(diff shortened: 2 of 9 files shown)


kannon92:
/hold

Missed the conflicts..

mimowo:
> @mimowo wouls you be open for a review also?

Ack, I will try this week, but don't wait for my review unnecessarily if you get enough input from others.

kannon92:
@danielvegamyhre do you know how to remove the rebase label? I did resolve conflicts in my branch and pushes those up. I think the prow removal of it is not working.

danielvegamyhre:
> @danielvegamyhre do you know how to remove the rebase label? I did resolve conflicts in my branch and pushes those up. I think the prow removal of it is not working.
> 
> ref: https://kubernetes.slack.com/archives/C09QZ4DQB/p1708018595891739
> 
> Looks like rebase label can have issues. So I think removing is fine. It also doesn't block the merge so this is ready for review.

I removed it via the github UI, the "labels" button on the top right.

kannon92:
I think since you have admin rights to the repo you have access to that.

pkg/webhooks/jobset_webhook.go
```diff
@@ -328,11 +329,16 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
...
-
-	// Note that SucccessPolicy and failurePolicy are made immutable via CEL.
-	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
-	errs = append(errs, apivalidation.ValidateImmutableField(mungedSpec.ManagedBy, oldJS.Spec.ManagedBy, field.NewPath("spec").Child("managedBy"))...)
-	return nil, errs.ToAggregate()
+	// Note that SuccessPolicy and failurePolicy are made immutable via CEL.
+	// Comparing job templates can be slow
+	// Only do it if we detect a difference.
+	if !equality.Semantic.DeepEqual(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs) {
+		if err := validateReplicatedJobsUpdate(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs); err != nil {
```
mimowo, line 336: It seems this allows to update the number of replicas for an unsuspended JobSet. Is this desired? 
I this would not work if there is no mechanism to recreate the Jobs. Maybe it is better to only support mutability of replicas when the JobSet is suspended?

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1599,7 +1599,7 @@ func TestValidateUpdate(t *testing.T) {
 			}.ToAggregate(),
 		},
 		{
-			name: "replicated job pod template can be updated for suspended jobset",
+			name: "replicated jobs updates can change replicas",
```
mimowo, line 1602: Make sure there are two versions of the test: for suspended JobSet and unsuspended.

pkg/webhooks/jobset_webhook.go
kannon92: I was thinking that one would want to add more replicated jobs as a major use case for Elastic JobSet.

We debated about whether or not we should modify the indexed job or the JobSet. I think we went with scaling up/down on the replicas of a ReplicatedJob.

Are you thinking that elasticity should only occur on a suspended object?

mimowo: Sorry, I have probably mixed a single Job mutability and the number of Job replicas.

In this PR you only let to update the Job replicas count, right? 

Still, it is not clear to me if we already have code to wcale up and down the number of replicas.

kannon92: I did a e2e test and this works.

I change the replicated jobs. I should check on scaling down actually.

pkg/webhooks/jobset_webhook_test.go
kannon92: Would that be relevant once #625 merges?

kannon92: sorry, I see what you mean.

kannon92: pushed up a change.

pkg/webhooks/jobset_webhook.go
mimowo: >  I create a JobSet and then I edit the replicated jobs of a worker jobset and I can confirm that new jobs are created.

Great!

> I change the replicated jobs. I should check on scaling down actually.

Let me know the results of the testing.

test/integration/controller/jobset_controller_test.go
```diff
@@ -1563,6 +1563,53 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+			},
+			steps: []*step{
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						setReplicasReplicatedJob(js, "replicated-job-a", 4)
+					},
+				},
+				{
+					checkJobCreation: func(js *jobset.JobSet) {
+						expectedStarts := 7
```
mimowo, line 1596: It will be good to add a comment why 7 is expected. It is a bit magical number here.

test/integration/controller/jobset_controller_test.go
```diff
@@ -1563,6 +1563,53 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+					},
+				},
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						setReplicasReplicatedJob(js, "replicated-job-a", 1)
+					},
+				},
+				{
+					checkJobCreation: func(js *jobset.JobSet) {
+						expectedStarts := 2
```
mimowo, line 1607: Similarly here. Please add a comment why 2 is expected.

pkg/webhooks/jobset_webhook.go
kannon92: Downscaling doesn’t seem to work.

I think I’ll add some e2e tests for this as it “works” for integration tests so there may be some subtleties

kannon92:
Pushed up some e2e tests for this. I need to work on the downscale case.

What should be the behavior for elastic jobs on a downscale?

If more jobs exists, do we delete the extra jobs only? Or should we delete and recreate all the jobs?

On upscale, the extra jobs are created and I think the original ones are left alone.

On downscale, I guess we could delete the extra replicas.

danielvegamyhre:
@kannon92 Looks like e2e tests are failing, can you take a look?

kannon92:
Yes, my hope is to get some clarity on the design/code.

Downscaling doesn't work for sure. What should we do on a downscale of replicas in ReplicatedJobs? Do we delete all the jobs and recreate? Or do we terminate the jobs that are not in the list of replicas?

pkg/webhooks/jobset_webhook.go
```diff
@@ -390,3 +396,21 @@ func replicatedJobByName(js *jobset.JobSet, replicatedJob string) *jobset.Replic
...
+func validateReplicatedJobsUpdate(currentRepJobs, oldRepJobs []jobset.ReplicatedJob) error {
+	// Changing length of replicated jobs on updates is forbidden
+	if len(currentRepJobs) != len(oldRepJobs) {
+		return fmt.Errorf("updates can not change the length of replicated jobs")
+	}
+	for i := range currentRepJobs {
+		if currentRepJobs[i].Name != oldRepJobs[i].Name {
+			return fmt.Errorf("updates can not change job names or reorder the jobs")
+		}
+		if !equality.Semantic.DeepEqual(currentRepJobs[i].Template, oldRepJobs[i].Template) {
```
danielvegamyhre, line 411: Will this conflict with the changes in #579 allowing mutability of pod template specs if the JobSet is suspended?

danielvegamyhre:
> Yes, my hope is to get some clarity on the design/code.
> 
> Downscaling doesn't work for sure. What should we do on a downscale of replicas in ReplicatedJobs? Do we delete all the jobs and recreate? Or do we terminate the jobs that are not in the list of replicas?

Hmm, I think if replicatedJob.replicas is downscaled from N to N-2, we should just delete 2 jobs so the state matches the spec.

pkg/webhooks/jobset_webhook.go
kannon92: I don’t think so but if we want the Kueue integration I’m happy to let @mimowo work go first and I’ll rebase.

kannon92:
> > Yes, my hope is to get some clarity on the design/code.
> > Downscaling doesn't work for sure. What should we do on a downscale of replicas in ReplicatedJobs? Do we delete all the jobs and recreate? Or do we terminate the jobs that are not in the list of replicas?
> 
> Hmm, I think if replicatedJob.replicas is downscaled from N to N-2, we should just delete 2 jobs so the state matches the spec.

Maybe I’m overthinking it.

Say I have 4 jobs and I downscale to 2. Should I only delete the 3rd and 4th replicated job via index? Say I want to keep the 1st and 2nd job running and I remove the excess ones?

danielvegamyhre:
> > > Yes, my hope is to get some clarity on the design/code.
> > > Downscaling doesn't work for sure. What should we do on a downscale of replicas in ReplicatedJobs? Do we delete all the jobs and recreate? Or do we terminate the jobs that are not in the list of replicas?
> > 
> > 
> > Hmm, I think if replicatedJob.replicas is downscaled from N to N-2, we should just delete 2 jobs so the state matches the spec.
> 
> Maybe I’m overthinking it.
> 
> Say I have 4 jobs and I downscale to 2. Should I only delete the 3rd and 4th replicated job via index? Say I want to keep the 1st and 2nd job running and I remove the excess ones?

If I recall correctly the Job controller works this way, removing indexes from largest to smallest until there are no more excess pods, so it would make sense to follow that pattern here for consistency I think.

test/integration/controller/jobset_controller_test.go
kannon92: Updated.

test/integration/controller/jobset_controller_test.go
kannon92: Updated.

pkg/webhooks/jobset_webhook.go
kannon92: Didn't see any issue here.

kannon92:
@danielvegamyhre @mimowo 

I got the downscale to work so I think this PR is ready for review.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1853,12 +1922,143 @@ func TestValidateUpdate(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			want: fmt.Errorf("updates can not change job names or reorder the jobs"),
+		},
+		{
+			name: "updates on replicated job templates are not allowed",
```
danielvegamyhre, line 2010: Why do we want to disallow job template updates? This would prevent us from using Elastic Indexed Jobs, which I'm hoping could be used in a more efficient implementation of exclusive job placement (see #482)

kannon92: See this comment: https://github.com/kubernetes-sigs/jobset/issues/463#issuecomment-2013121170.

I wasn’t sure if there was a use case for this so I omitted support for it. 

@ahg-g would this be a strong enough usecase to add support for ElasticIndexedJobs.

pkg/controllers/elastic_jobset.go
```diff
@@ -0,0 +1,53 @@
...
+	"strconv"
+
+	batchv1 "k8s.io/api/batch/v1"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+)
+
+// jobsToDeleteDownScale gathers the excess jobs during a downscale
+// and deletes the jobs
+func jobsToDeleteDownScale(replicatedJobs []jobset.ReplicatedJob, replicatedJobStatus []jobset.ReplicatedJobStatus, jobItems []batchv1.Job) ([]*batchv1.Job, error) {
```
danielvegamyhre, line 27: can we name this `jobsToDeleteForDownScale`? Only 3 characters longer and is more clear imo.

danielvegamyhre, line 27: can we rename `replicatedJobStatus` to `replicatedJobStatuses` since it is a slice of replicatedJob statuses?

pkg/controllers/jobset_controller.go
```diff
@@ -520,6 +527,22 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
 	return nil
 }
 
+// We need to check if the replicas of a replicated job are mismatching
+// If they are, we should delete the extra jobs
+func (r *JobSetReconciler) downscaleElasticJobs(ctx context.Context, js *jobset.JobSet, replicatedJobStatus []jobset.ReplicatedJobStatus) error {
```
danielvegamyhre, line 532: This can also be moved to the elastic_jobset.go file

pkg/controllers/elastic_jobset.go
```diff
@@ -0,0 +1,53 @@
...
+			jobsWeDeleted := 0
+			for _, val := range jobItems {
+				if val.Labels[jobset.ReplicatedJobNameKey] != replicatedJob.Name {
+					continue
+				}
+				jobIndex, err := strconv.Atoi(val.Labels[jobset.JobIndexKey])
+				if err != nil {
+					return nil, fmt.Errorf("unable get integer from job index key")
+				}
+				if jobIndex >= int(countOfJobsToDelete) {
```
danielvegamyhre, line 42: So if we have 4 jobs (indexes 0,1,2,3) and we want to delete 1 job, then `countOfJobsToDelete=1` and we may end up deleting any of job indexes 1,2,3 - so it is non-deterministic and may leave us with an non-contiguous range of job indexes (remaining jobs indexes 0,2,3).

Rather than this, I think we should delete jobs from largest job index to smallest, so the remaining jobs still occupy a contiguous index range - following the same pattern as how the Job controller behaves for elastic indexed jobs, deleting pods from largest to smallest completion indexes).

pkg/controllers/elastic_jobset.go
```diff
@@ -0,0 +1,53 @@
...
+// jobsToDeleteDownScale gathers the excess jobs during a downscale
+// and deletes the jobs
+func jobsToDeleteDownScale(replicatedJobs []jobset.ReplicatedJob, replicatedJobStatus []jobset.ReplicatedJobStatus, jobItems []batchv1.Job) ([]*batchv1.Job, error) {
+	jobsToDelete := []*batchv1.Job{}
+	for _, replicatedJob := range replicatedJobs {
+		status := findReplicatedJobStatus(replicatedJobStatus, replicatedJob.Name)
+		countOfJobsToDelete := status.Ready - replicatedJob.Replicas
+		if countOfJobsToDelete > 0 {
+			jobsWeDeleted := 0
+			for _, val := range jobItems {
```
danielvegamyhre, line 34: Ii think it would be more optimal (although perhaps not provide a big perf improvement in practice) to store the jobs in a map of "replicatedJobName -> [slice of jobs for that replicatedJob]".  That way we don't need to iterate over every job in the jobset, for every replicatedJob we are downscaling. 

The O(jobs) cost of constructing the map is preferable to the O(replicatedJobs * jobs) cost of this approach.

pkg/controllers/elastic_jobset.go
```diff
@@ -0,0 +1,53 @@
...
+				jobIndex, err := strconv.Atoi(val.Labels[jobset.JobIndexKey])
+				if err != nil {
+					return nil, fmt.Errorf("unable get integer from job index key")
+				}
+				if jobIndex >= int(countOfJobsToDelete) {
+					jobsWeDeleted = jobsWeDeleted + 1
+					jobsToDelete = append(jobsToDelete, &val)
+				}
+				if jobsWeDeleted == int(countOfJobsToDelete) {
+					continue
```
danielvegamyhre, line 47: we can break here rather than continue right, since there's no more jobs to delete in this inner loop?

pkg/controllers/jobset_controller.go
kannon92: I don't think it can. We keep anything related to the JobSetReconciler in jobset_controller.

ElasticJob and other utilites seem to be pure functions that can easily be tested with units. This does rely on the reconcile class to talk to the client.

pkg/controllers/elastic_jobset.go
kannon92: done.

pkg/controllers/elastic_jobset.go
kannon92: Good point. I went ahead and added this change.

pkg/controllers/elastic_jobset.go
kannon92: yes. done.

pkg/controllers/jobset_controller.go
kannon92: Yea, in theory that sounds like a good idea. I tend to like separating code from controller logic as unit testing controllers is quite painful.

I will eagerly await your PR.

pkg/controllers/elastic_jobset.go
ultrons: @danielvegamyhre , keeping the indices unchanged could be desirable in cases where we don't want to restart the entire jobset right?

danielvegamyhre: > @danielvegamyhre , keeping the indices unchanged could be desirable in cases where we don't want to restart the entire jobset right?

Currently we always restart the entire jobset (recreating all child jobs), we don't yet support restarting only a subset of jobs, since there hasn't been a clear use case for that (due to the group-startup semantics required for distributed training, everything must restart, reload from latest checkpoint, do distributed initialization, etc).

What cases are you envisioning where we only want to restart a subset of jobs?

kannon92:
@ahg-g id thinking we should clear use cases with this in mind. I am not a ML engineer so I’m not really able to give clear concrete cases for this. Let me ask around and see. Otherwise I may close this and wait until we have a customer ask for this.

kannon92:
/close

We need more design/use cases on this. Going to close for now.

PR #580, opened 2024-05-24
Relax validation on replicated job pod templates

danielvegamyhre:
Fixes #579 

I also moved the validateFailurePolicy helper below the main exported ValidateX functions.

cc @alculquicondor @ahg-g

PR #623, opened 2024-07-24
Allow mutating schedulingGates when the Jobset is suspended

mimowo:
First, schedulingGates can be mutated for suspeded batch Jobs: https://github.com/kubernetes/kubernetes/blob/ceb58a4dbc671b9d0a2de6d73a1616bc0c299863/pkg/apis/batch/validation/validation.go#L662

We are considering to use schedulingGates in Kueue for fine-grained scheduling where a subset of pods may get an additional set of node labels, so we don't update the entire PodTemplate, but specific pods. In that case Kueue would have a controller which ungates the pods once adjusted.

This is an extension to https://github.com/kubernetes-sigs/jobset/pull/580/.


pkg/webhooks/jobset_webhook.go
```diff
@@ -318,6 +318,9 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Labels = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Labels
 			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
 			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.Tolerations = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.Tolerations
+
+			// Pod Scheduling Gates can be updated for batch/v1 Job: https://github.com/kubernetes/kubernetes/blob/ceb58a4dbc671b9d0a2de6d73a1616bc0c299863/pkg/apis/batch/validation/validation.go#L662
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates
```
tenzen-y, line 323: In the batch/job, we can mutate the schedulingGates only when the Job didn't start: https://github.com/kubernetes/kubernetes/blob/ceb58a4dbc671b9d0a2de6d73a1616bc0c299863/pkg/registry/batch/job/strategy.go#L194

So, shouldn't we introduce the same criteria for the mutable scheduling directive here, right?

mimowo: For JobSet we generally just check if the Job is suspended. This is done in the entry if above `if ptr.Deref(oldJS.Spec.Suspend, false) {`. 

The other part of the check is based on the `status.startTime`, but this field is not in the JobSet API: https://github.com/kubernetes-sigs/jobset/blob/56c77da8303562b0bb034ee237a4264b59a22902/api/jobset/v1alpha2/jobset_types.go#L125-L146.

The difference between Job and JobSet is related to all of the fields (Annotations, Labels, NodeSelector, Tolerations too),.

mimowo: Actually, the check above only verifies that `oldJS` is suspended. Which means that, IIUC, there is a bug that we cannot modify the template on suspending. I think this could render failures in Kueue for the suspend, I will double-check and open a separate PR / Issue for it.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1492,6 +1492,13 @@ func TestValidateUpdate(t *testing.T) {
 										ObjectMeta: metav1.ObjectMeta{
 											Annotations: map[string]string{"key": "value"},
 										},
+										Spec: corev1.PodSpec{
```
kannon92, line 1495: Can you add a new test case for scheduling gates?

pkg/webhooks/jobset_webhook.go
tenzen-y: > The other part of the check is based on the status.startTime, but this field is not in the JobSet API:

Oh, I didn't know that! Thank you for the explanation!

> I think this could render failures in Kueue for the suspend, I will double-check and open a separate PR / Issue for it.

I also feel that the behavior is a bug/regression. But, I agree with treating the regression as a separate issue.

tenzen-y (commented):
Thank you!
I think that this would be useful feature!

pkg/webhooks/jobset_webhook_test.go
mimowo: Sure, I pushed the change. PTAL

pkg/webhooks/jobset_webhook.go
mimowo: Actually, I have confirmed that this currently makes the suspend fail for JobSet in Kueue: https://github.com/kubernetes-sigs/kueue/issues/2691. I have opened an issue for JobSet https://github.com/kubernetes-sigs/jobset/issues/624, and the PR: https://github.com/kubernetes-sigs/jobset/pull/625. PTAL.

mimowo: I think this PRs can be merged independently. 

This PR already will allow us to use schedulingGates in Kueue, even if the restore still does not work.

mimowo: I think this change is rather straightforward (and follows the patterns of other fields) so I wasn't planning an e2e test here. 

I think once this PR lands, we could have an extended e2e test scenario added for the other PR, where we:
- unsuspend a JobSet adding scheduling gates, and other PodTemplate fields
- suspend the JobSet reverting the scheduling gates and other PodTemplate fields (as Kueue would do)
- unsuspend again the JobSet with different values of the PodTemplate fields and check that the pods executes (JobSet completes)

WDYT?

danielvegamyhre: Unit + integration tests should be sufficient here.

kannon92: @mimowo that sounds good. Maybe we create an issue documenting the goal of the e2e test so we can reference it.

mimowo: Would it work to just reference the proposal from the issue here: https://github.com/kubernetes-sigs/jobset/issues/624?

danielvegamyhre:
/lgtm

Will approve once test finish running and I confirm they all are passing.

danielvegamyhre:
/approve

Thanks!

PR #625, opened 2024-07-25
Fix resuming Jobset after restoring PodTemplate (by deleting Jobs)

### LINKED ISSUE #535: Do not create Jobs and the service when JobSet is suspended
To reproduce just create a simple JobSet with `spec.suspend=true`. In that case Jobs and the service are created.

This is wasteful in case of Jobs which are queued by Kueue, and may stay in the queue for a long time, potentially.

### LINKED ISSUE #624: Allow to mutate PodTemplate when suspending a JobSet and support resuming such JobSet
There are currently two related issues which prevent JobSet - Kueue integration:
1. JobSet rejects mutation of PodTemplate on suspend

When Kueue evicts a workload (represented by JobSet) it stops the JobSet and tries to restore the PodTemplate to enable re-admitting the same JobSet to another ResourceFlavor (with potentially different nodeSelectors).
For example, the following e2e test for Job shows how Kueue can preempt a workload and re-admit with another nodeSelector: [link](https://github.com/kubernetes-sigs/kueue/blob/4b90a76a424be5456165f40a304dbe557ab4d422/test/e2e/singlecluster/e2e_test.go#L253-L287).

However, the integration with Kueue does not work currently, because the Kueue request to suspend
the JobSet fails if it also wants to update the PodTemaplte.

mimowo:
#### What type of PR is this?

/kind bug 

Fixes #624
Fixes #535

This approach relies on updating the Jobs on resume, rather the deleting Jobs on suspend.
Alternative implementation was done in: https://github.com/kubernetes-sigs/jobset/pull/625

```release-note
Allow restoring PodTemplate on suspend, and fix resuming JobSet after restoring 
the PodTemplate. This fixes the integration with Kueue to support eviction of workloads
and re-admitting in another ResourceFlavor (with other nodeSelectors, or without nodeSelectors).
It is achieved by deleting the Jobs while the JobSet is suspended.
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -206,10 +206,11 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 	// Handle suspending a jobset or resuming a suspended jobset.
 	jobsetSuspended := jobSetSuspended(js)
 	if jobsetSuspended {
-		if err := r.suspendJobs(ctx, js, ownedJobs.active, updateStatusOpts); err != nil {
-			log.Error(err, "suspending jobset")
+		if err := r.deleteJobs(ctx, ownedJobs.active); err != nil {
+			log.Error(err, "deleting jobs")
 			return ctrl.Result{}, err
 		}
+		setJobSetSuspendedCondition(js, updateStatusOpts)
 	} else {
 		if err := r.resumeJobsIfNecessary(ctx, js, ownedJobs.active, rjobStatuses, updateStatusOpts); err != nil {
 			log.Error(err, "resuming jobset")
@@ -379,19 +380,6 @@ func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, j
 	return rjStatus
 }
 
-func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job, updateStatusOpts *statusUpdateOpts) error {
-	for _, job := range activeJobs {
-		if !jobSuspended(job) {
-			job.Spec.Suspend = ptr.To(true)
-			if err := r.Update(ctx, job); err != nil {
-				return err
-			}
-		}
-	}
-	setJobSetSuspendedCondition(js, updateStatusOpts)
-	return nil
-}
-
 // resumeJobsIfNecessary iterates through each replicatedJob, resuming any suspended jobs if the JobSet
 // is not suspended.
 func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job, replicatedJobStatuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) error {
@@ -492,6 +480,11 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
 		jobs, err := constructJobsFromTemplate(js, &replicatedJob, ownedJobs)
 		if err != nil {
 			return err
+		}
+
+		// Don't create child Jobs if the JobSet is suspended
+		if jobSetSuspended(js) {
+			continue
 		}
 
 		status := findReplicatedJobStatus(replicatedJobStatus, replicatedJob.Name)
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -206,6 +206,10 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 	// Handle suspending a jobset or resuming a suspended jobset.
 	jobsetSuspended := jobSetSuspended(js)
 	if jobsetSuspended {
+		// We delete Jobs when the JobSet gets suspended. We do it because while
+		// the JobSet is suspended its PodTemplates can be mutated. In that case
+		// we need to recreate the Jobs so that they match the latest
+		// PodTemplates in the JobSet.
 		if err := r.deleteJobs(ctx, ownedJobs.active); err != nil {
 			log.Error(err, "deleting jobs")
 			return ctrl.Result{}, err
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -318,20 +318,7 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 	// This is needed for integration with Kueue/DWS.
 	if ptr.Deref(oldJS.Spec.Suspend, false) || ptr.Deref(js.Spec.Suspend, false) {
 		for index := range js.Spec.ReplicatedJobs {
-			munge := true
-			// Don't allow to unsuspend if there are still active Jobs with
-			// different PodTemplate. We do this to avoid a race condition when
-			// Jobs with an old PodTemplate (before suspension) are still
-			// runnable and would not be deleted if we unsuspend the JobSet.
-			if !ptr.Deref(js.Spec.Suspend, false) {
-				rStatus := js.Status.ReplicatedJobsStatus
-				// Don't allow to mutate PodTemplate on unsuspending if there
-				// are still active or suspended jobs from the previous run
-				if len(rStatus) > index && (rStatus[index].Active > 0 || rStatus[index].Suspended > 0) {
-					munge = false
-				}
-			}
-			if munge {
+			if allowToMutatePodTemplate(js, index) {
 				// Pod values which must be mutable for Kueue are defined here: https://github.com/kubernetes-sigs/kueue/blob/a50d395c36a2cb3965be5232162cf1fded1bdb08/apis/kueue/v1beta1/workload_types.go#L256-L260
 				mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Annotations = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Annotations
 				mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Labels = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Labels
@@ -348,6 +335,32 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
 	errs = append(errs, apivalidation.ValidateImmutableField(mungedSpec.ManagedBy, oldJS.Spec.ManagedBy, field.NewPath("spec").Child("managedBy"))...)
 	return nil, errs.ToAggregate()
+}
+
+// allowToMutatePodTemplate checks if a PodTemplate in a specifc ReplicatedJob
+// can be mutated.
+func allowToMutatePodTemplate(js *jobset.JobSet, index int) bool {
+	// We allow to mutate the PodTemplate when the JobSet is suspended already
+	// or is getting suspended.
+	if ptr.Deref(js.Spec.Suspend, false) {
+		return true
+	}
+	// We allow to mutate the PodTemplate if the ReplicatedJob status is not
+	// initialized yet, because in that case there are no Jobs from the previous
+	// run.
+	if len(js.Status.ReplicatedJobsStatus) < len(js.Spec.ReplicatedJobs) {
+		return true
+	}
+
+	// We don't allow to mutate the PodTemplate when the JobSet is resuming, and
+	// the Jobs from the previous run still exist. This gives time the JobSet
+	// controller to delete the Jobs from the previous run if they could
+	// conflict with the Job creation on resume.
+	rStatus := js.Status.ReplicatedJobsStatus
+	if rStatus[index].Active > 0 || rStatus[index].Suspended > 0 {
+		return false
+	}
+	return true
 }
 
 // ValidateDelete implements webhook.Validator so a webhook will be registered for the type
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -314,18 +314,33 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 	}
 	mungedSpec := js.Spec.DeepCopy()
 
-	// Allow pod template to be mutated for suspended JobSets.
+	// Allow pod template to be mutated for suspended JobSets, or JobSets getting suspended.
 	// This is needed for integration with Kueue/DWS.
-	if ptr.Deref(oldJS.Spec.Suspend, false) {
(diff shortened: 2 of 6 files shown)


pkg/webhooks/jobset_webhook.go
```diff
@@ -309,9 +309,9 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
 	}
 	mungedSpec := js.Spec.DeepCopy()
 
-	// Allow pod template to be mutated for suspended JobSets.
+	// Allow pod template to be mutated for suspended JobSets, or JobSets getting suspended.
 	// This is needed for integration with Kueue/DWS.
-	if ptr.Deref(oldJS.Spec.Suspend, false) {
+	if ptr.Deref(oldJS.Spec.Suspend, false) || ptr.Deref(js.Spec.Suspend, false) {
```
tenzen-y, line 314:
```suggestion
	if ptr.Deref(oldJS.Spec.Suspend, false) || (ptr.Deref(js.Spec.Suspend, false) && js.Spec.ReplicatedJobs[*].status.startTime == nil) {
```

Wouldn't it be helpful to validate if the jobs don't have startTime?
Because if any jobs have startTime, this operation should fail due to batch/job validation.
But, I'm not sure if we should add validations deeply depending on the batch/job specifications.

@danielvegamyhre @mimowo WDYT?

mimowo: I think the flow is that we first suspend the parent JobSet, then it suspends the Jobs. So, if we added this condition, the `startTime` on running child jobs would prevent the update.

I think I should have added an e2e test in JobSet to verify the suspending works.

kannon92: Yea I’m thinking integration and e2e test solidifying the requirements that JobSet needs for working with Kueue will be a great idea.

We obviously don’t want to bring in Kueue as a dependency so I think just verifying that the updates/patches work in integration/e2e will be welcome

danielvegamyhre:
Thanks for working on the fix for this @mimowo. Please take a look at my comment here https://github.com/kubernetes-sigs/jobset/issues/624#issuecomment-2250687273 as well, I'd like to improve our coordination to prevent these kinds of issues from occurring in the future.

mimowo:
/hold
for e2e or integration test inside the JobSet repo

pkg/webhooks/jobset_webhook.go
tenzen-y: > I think the flow is that we first suspend the parent JobSet, then it suspends the Jobs. So, if we added this condition, the startTime on running child jobs would prevent the update.

Uhm, that makes sense. 
But, I guess that users fall into the rabbit hole when they accidentally modify the scheduling directives against the ReplicatedJob with startTime because the jobset-controller just output errors in the controller logs, this is not directly user feedback.

Maybe we can improve these implied specifications for the JobSet users once we introduce https://github.com/kubernetes/kubernetes/issues/113221.
 
So, I'm ok with the current implementation for now.

mimowo: Actually, the e2e test would not work, because there is a deeper issue - JobSet does not delete the Jobs once suspended: https://github.com/kubernetes-sigs/jobset/issues/535. I'm trying to implement it by deleting the Jobs on suspend, and it looks promising. There are some integration tests for startupPolicy to adjust I yet need to look into.

mimowo: Let me know if you have some more context knowledge if traps I may encounter with this approach, but essentially this is the only I see, and it would mimic the Job behavior, where suspended Jobs deletes the pods.

danielvegamyhre:
Let me know when this is ready for review and I'll take a look. Thanks!

mimowo:
> Let me know when this is ready for review and I'll take a look. Thanks!

Thank you @danielvegamyhre ! I think it is ready now, let me also know if some parts require additional explanation.

mimowo:
/hold cancel
Two new e2e tests are added on suspending and resuming, one focuses on fast suspend -> resume -> suspend -> resume path to check for any race conditions. The other tracks deletion and creation of pods also.

mimowo:
@danielvegamyhre  I'm wondering about introducing `DeleteJobsOnSuspend` feature gate (true by default in Beta) that could be used in emergency (a bug found) to revert the behavior for a user without recompiling, but I didn't see any other feature gates in JobSet yet. WDYT?

pkg/webhooks/jobset_webhook.go
```diff
@@ -309,18 +309,33 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
...
+			munge := true
+			// Don't allow to unsuspend if there are still active Jobs with
+			// different PodTemplate. We do this to avoid a race condition when
+			// Jobs with an old PodTemplate (before suspension) are still
+			// runnable and would not be deleted if we unsuspend the JobSet.
+			if !ptr.Deref(js.Spec.Suspend, false) {
+				rStatus := js.Status.ReplicatedJobsStatus
+				// Don't allow to mutate PodTemplate on unsuspending if there
+				// are still active or suspended jobs from the previous run
+				if len(rStatus) > index && (rStatus[index].Active > 0 || rStatus[index].Suspended > 0) {
```
danielvegamyhre, line 325: Can we assign `len(rStatus) > index` to a well-named variable to make it more clear/readable why it matters if the number of replicatedJob statuses is greater than the current replicatedJob index?

Furthermore, could we abstract this block out into a helper function to make it clear at a glance what we are trying to check here, and make it more concise/readable?

mimowo: Sure, instead of using `len(rStatus) > index` I basically want to check if the ReplicatedJobsStatus is initialized. If not, then I assume the JobSet didn't run yet, so there are no Jobs. I have also extracted the code to a function and documented. PTAL.

test/e2e/e2e_test.go
```diff
@@ -131,6 +133,159 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+				gomega.Consistently(func() int32 {
+					gomega.Expect(k8sClient.Get(ctx, jsKey, js)).Should(gomega.Succeed())
+					if js.Status.ReplicatedJobsStatus == nil {
+						return 0
+					}
+					return js.Status.ReplicatedJobsStatus[0].Active
+				}).WithTimeout(time.Second).WithPolling(200 * time.Millisecond).Should(gomega.Equal(int32(0)))
+			})
+
+			ginkgo.By("Unsuspend the JobSet setting schedulingGates that prevent pods from being scheduled", func() {
```
kannon92, line 158: Reading this by, I read this to mean that when you set schedulingGates we are unsuspended the job.

kannon92: But then you are also setting suspend to false in below not not exactly sure..

pkg/webhooks/jobset_webhook.go
```diff
@@ -335,6 +337,32 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
...
+	if len(js.Status.ReplicatedJobsStatus) < len(js.Spec.ReplicatedJobs) {
+		return true
+	}
+
+	// We don't allow to mutate the PodTemplate when the JobSet is resuming, and
+	// the Jobs from the previous run still exist. This gives time the JobSet
+	// controller to delete the Jobs from the previous run if they could
+	// conflict with the Job creation on resume.
+	rStatus := js.Status.ReplicatedJobsStatus
+	if rStatus[index].Active > 0 || rStatus[index].Suspended > 0 {
```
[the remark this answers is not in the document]
mimowo: Indeed, currently we only count as "active" Jobs which have at least one "active" pods. This means that it is possible that some Jobs which have only terminating pods may still exist. 

One fixing idea would be to count Jobs as active if they have at least one active or terminating pod. Another is to introduce a new counter for "terminating" jobs - Jobs which have all pods terminating. WDYT? 

I guess this is a corner case for now, so maybe could be done in a follow up?

test/e2e/e2e_test.go
```diff
@@ -131,6 +133,159 @@ var _ = ginkgo.Describe("JobSet", func() {
...
 
+	ginkgo.When("job is unsuspended and suspend", func() {
+		ginkgo.It("should not create Jobs while suspended, and delete Jobs on suspend", func() {
+			ctx := context.Background()
+			js := shortSleepTestJobSet(ns).Obj()
+			jsKey := types.NamespacedName{Name: js.Name, Namespace: js.Namespace}
+
+			ginkgo.By("Create a suspended JobSet", func() {
+				js.Spec.Suspend = ptr.To(true)
+				js.Spec.TTLSecondsAfterFinished = ptr.To[int32](5)
```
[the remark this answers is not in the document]
mimowo: It's not necessary, I just added it so that the JobSets get removed eventually from the API server, but it would all work equally well without it. I can remove. WDYT?

test/e2e/e2e_test.go
mimowo: SchedulingGates are for Pods. The Jobs are unsuspended. This is just to show scheduling gates can be added on resume. Also, I want to make sure the Pods don't actually run, so that the Job does not complete on the first resume.

pkg/controllers/jobset_controller.go
```diff
@@ -206,10 +206,11 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 	// Handle suspending a jobset or resuming a suspended jobset.
 	jobsetSuspended := jobSetSuspended(js)
 	if jobsetSuspended {
-		if err := r.suspendJobs(ctx, js, ownedJobs.active, updateStatusOpts); err != nil {
-			log.Error(err, "suspending jobset")
+		if err := r.deleteJobs(ctx, ownedJobs.active); err != nil {
```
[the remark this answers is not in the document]
mimowo: Good idea. Added

pkg/webhooks/jobset_webhook.go
```diff
@@ -314,18 +314,20 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, old, newObj runtime.
...
-			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Annotations = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Annotations
-			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Labels = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Labels
-			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
-			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.Tolerations = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.Tolerations
-
-			// Pod Scheduling Gates can be updated for batch/v1 Job: https://github.com/kubernetes/kubernetes/blob/ceb58a4dbc671b9d0a2de6d73a1616bc0c299863/pkg/apis/batch/validation/validation.go#L662
-			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates = oldJS.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates
+			if allowToMutatePodTemplate(js, index) {
+				// Pod values which must be mutable for Kueue are defined here: https://github.com/kubernetes-sigs/kueue/blob/a50d395c36a2cb3965be5232162cf1fded1bdb08/apis/kueue/v1beta1/workload_types.go#L256-L260
```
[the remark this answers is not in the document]
mimowo: This "Pod values which must be mutable for Kueue..." comment wasn't actually added by in this PR, but I just indented it by adding the `allowToMutatePodTemplate` if.

danielvegamyhre:
> @danielvegamyhre I'm wondering about introducing `DeleteJobsOnSuspend` feature gate (true by default in Beta) that could be used in emergency (a bug found) to revert the behavior for a user without recompiling, but I didn't see any other feature gates in JobSet yet. OTOH, the amount of testing to cover both paths would be huge, plus it is rather a bug than a new feature, so I'm hesitant. WDYT?

I don't think we need a feature gate for this.

pkg/webhooks/jobset_webhook.go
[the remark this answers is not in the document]
mimowo: I think it is not too bad in this PR, because a terminating Job (all pods are terminating) is going to be deleted eventually. Yes, if we resume the JobSet while it exists, then it can block creating replacement Job for a while, but IIUC the JobSet controller will create replacement for such a Job as soon as it is fully gone, and the replacement Job will have a good PodTemplate.

In order to prevent resuming a JobSet while there is a terminating Job we could:
1. include such a Job as active, if `job.Status.Active > 0 || ptr.Deref(job.Status.Terminating, 0) > 0` [here](https://github.com/kubernetes-sigs/jobset/blob/9f3220a91905c7e42770cd5d880ab90a9e28882a/pkg/controllers/jobset_controller.go#L350)
2. introduce a new counter for `Terminating` Jobs. Terminating would be a fallback (so job.Status.Active == 0)  counter if `ptr.Deref(job.Status.Terminating, 0) > 0`

I'm leaning towards option (2.). I could implement it either in this PR or a follow up.

mimowo:
@danielvegamyhre @kannon92 thanks for reviewing. let me know what are the remaining things to address in this PR.
It seems this is the outstanding comment https://github.com/kubernetes-sigs/jobset/pull/625#discussion_r1697015563.

pkg/webhooks/jobset_webhook.go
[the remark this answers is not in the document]
kannon92: I think a follow up is fine. Terminating on Jobs is currently a beta feature so logic may be somewhat complex.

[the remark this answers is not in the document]
mimowo: > Terminating on Jobs is currently a beta feature so logic may be somewhat complex.

I think we don't need any code in JobSet depending on the feature-gate in k8s. If the feature gate is disabled in k8s the field is not set. So, the code in JobSet would see `nil`, we could prevent panics with `ptr.Deref(job.Status.Terminating, 0)`, but this should be enough, IIUC.

mimowo:
/hold
I have synced with @danielvegamyhre and we discussed that it would be good to see if we can achieve the functionality needed for Kueue without deleting the Jobs on suspend, reusing bits introduced in https://github.com/kubernetes-sigs/jobset/pull/590/files#diff-68677f8530f00fbbb399ac36268f928ee584948dd09091644a8fad81a8658153R457-R478. 

I agree this is worth exploring as this is closer to how JobSet operates currently, so the impact is more predictable. I will try this approach and update the PR or summarize if there are any blockers.

mimowo:
@danielvegamyhre @kannon92 
I have prepared another PR which fixes the issue by updating the Jobs on resume, rather than recreating them. I'm happy to move forward with that approach. PTAL: https://github.com/kubernetes-sigs/jobset/pull/640. I have also update the issue description to make it more clear.
Finally, there is a small PR with e2e test showing the issue: https://github.com/kubernetes-sigs/jobset/pull/635.

mimowo:
/close 
in favor of https://github.com/kubernetes-sigs/jobset/pull/640

PR #636, opened 2024-08-03
fix error output

googs1025:
#### What type of PR is this?
/kind cleanup
/kind bug

- fix error output 

Fixes # None

None

```release-note
None
```

### DIFF
--- a/pkg/webhooks/pod_admission_webhook.go
+++ b/pkg/webhooks/pod_admission_webhook.go
@@ -54,7 +54,7 @@ func (p *podWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (ad
 		return nil, fmt.Errorf("follower pod node selector not set")
 	}
 	if _, exists := pod.Spec.NodeSelector[topologyKey]; !exists {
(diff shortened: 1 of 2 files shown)


pkg/webhooks/pod_admission_webhook.go
```diff
@@ -54,7 +54,7 @@ func (p *podWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (ad
 		return nil, fmt.Errorf("follower pod node selector not set")
 	}
 	if _, exists := pod.Spec.NodeSelector[topologyKey]; !exists {
-		return nil, fmt.Errorf("follower pod node selector not set")
+		return nil, fmt.Errorf("follower pod node selector not found topology label: %s", topologyKey)
```
danielvegamyhre, line 57:
```suggestion
		return nil, fmt.Errorf("follower pod node selector for topology domain not found. missing selector: %s", topologyKey)
```

googs1025: done

PR #640, opened 2024-08-06
Fix resuming JobSet after restoring PodTemplate (by Jobs update)

### LINKED ISSUE #624: Allow to mutate PodTemplate when suspending a JobSet and support resuming such JobSet

mimowo:
#### What type of PR is this?

/kind bug 

Fixes #624

This approach relies on updating the Jobs on resume, rather the deleting Jobs on suspend.
Alternative implementation was done in: https://github.com/kubernetes-sigs/jobset/pull/625

```release-note
Allow restoring PodTemplate on suspend, and fix resuming JobSet after restoring 
the PodTemplate. This fixes the integration with Kueue to support eviction of workloads
and re-admitting in another ResourceFlavor (with other nodeSelectors, or without nodeSelectors).
It is achieved by updating the Jobs on resuming the JobSet.
```

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -51,6 +51,11 @@ const (
 	CoordinatorKey = "jobset.sigs.k8s.io/coordinator"
 )
 
+var (
+	// the legacy names are no longer defined in the api, only in k/2/apis/batch
+	JobManagedLabels = []string{"job-name", "controller-uid", batchv1.JobNameLabel, batchv1.ControllerUidLabel}
+)
+
 type JobSetConditionType string
 
 // These are built-in conditions of a JobSet.
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -395,12 +395,6 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
 // resumeJobsIfNecessary iterates through each replicatedJob, resuming any suspended jobs if the JobSet
 // is not suspended.
 func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job, replicatedJobStatuses []jobset.ReplicatedJobStatus, updateStatusOpts *statusUpdateOpts) error {
-	// Store pod template for each replicatedJob.
-	replicatedJobTemplateMap := map[string]corev1.PodTemplateSpec{}
-	for _, replicatedJob := range js.Spec.ReplicatedJobs {
-		replicatedJobTemplateMap[replicatedJob.Name] = replicatedJob.Template.Spec.Template
-	}
-
 	// Map each replicatedJob to a list of its active jobs.
 	replicatedJobToActiveJobs := map[string][]*batchv1.Job{}
 	for _, job := range activeJobs {
@@ -422,7 +416,7 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 			if !jobSuspended(job) {
 				continue
 			}
-			if err := r.resumeJob(ctx, job, replicatedJobTemplateMap); err != nil {
+			if err := r.resumeJob(ctx, job, js, &replicatedJob); err != nil {
 				return err
 			}
 		}
@@ -440,7 +434,7 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 	return nil
 }
 
-func (r *JobSetReconciler) resumeJob(ctx context.Context, job *batchv1.Job, replicatedJobTemplateMap map[string]corev1.PodTemplateSpec) error {
+func (r *JobSetReconciler) resumeJob(ctx context.Context, job *batchv1.Job, js *jobset.JobSet, rjob *jobset.ReplicatedJob) error {
 	log := ctrl.LoggerFrom(ctx)
 	// Kubernetes validates that a job template is immutable
 	// so if the job has started i.e., startTime != nil), we must set it to nil first.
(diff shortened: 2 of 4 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -46,6 +46,16 @@ import (
 	"sigs.k8s.io/jobset/pkg/util/placement"
 )
 
+const (
+	legacyJobNameLabel       = "job-name"
```
kannon92, line 50: These labels should be defined in the Jobset api. I think they already exist there.

mimowo: sgtm, updated.

danielvegamyhre:
Can #625 be closed now in favor of this

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -51,6 +52,11 @@ const (
 	CoordinatorKey = "jobset.sigs.k8s.io/coordinator"
 )
 
+var (
+	// the legacy names are no longer defined in the api, only in k/2/apis/batch
```
kannon92, line 56: Can we not introduce the legacy labels now? I think we can remove the original ones now that 1.27 is out of support.

mimowo: I think it is better to copy them, because Job controller still sets them. 

Here is an example Pod template created on cluster 1.30.2:
```
        batch.kubernetes.io/controller-uid: c46c3b0d-76d8-427b-bda6-a9cb06c77cfd
        batch.kubernetes.io/job-name: sample-job-a-lv6wd
        controller-uid: c46c3b0d-76d8-427b-bda6-a9cb06c77cfd
        job-name: sample-job-a-lv6wd
```
If they are not present on the original Job template (null) I wouldn't copy them to the newJob template.

mimowo: Moreover we still expect them in the Job validation: https://github.com/kubernetes/kubernetes/blob/60c4c2b2521fb454ce69dee737e3eb91a25e0535/pkg/apis/batch/validation/validation.go#L139-L146

mimowo: Hm, but if I don't copy them here, then we would update Job template not to contain them, I also not sure this will pass validation, let me test.

kannon92: I opened up https://github.com/kubernetes-sigs/jobset/pull/642 to drop JobNameKey as it wasn't actually used at all in JobSet.

test/e2e/e2e_test.go
```diff
@@ -246,3 +323,25 @@ func sleepTestJobSet(ns *corev1.Namespace) *testing.JobSetWrapper {
 			Replicas(int32(replicas)).
 			Obj())
 }
+
+func shortSleepTestJobSet(ns *corev1.Namespace) *testing.JobSetWrapper {
```
kannon92, line 327: I don't see the point of two functions that do sleep with different times.

pkg/controllers/jobset_controller.go
```diff
@@ -452,30 +446,44 @@ func (r *JobSetReconciler) resumeJob(ctx context.Context, job *batchv1.Job, repl
 	}
 
 	// Get name of parent replicated job and use it to look up the pod template.
-	replicatedJobName := job.Labels[jobset.ReplicatedJobNameKey]
-	replicatedJobPodTemplate := replicatedJobTemplateMap[replicatedJobName]
-	if job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] != "" {
+	if job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] != "" && job.Labels[jobset.JobIndexKey] != "" {
```
kannon92, line 449: Maybe I am missing it but `job.Labels[jobset.ReplicatedJobNameKey] != ""` seems to not be needed now.

kannon92:
LGTM.

pkg/controllers/jobset_controller.go
mimowo: That's right, simplified.

api/jobset/v1alpha2/jobset_types.go
kannon92: I think maybe we should just copy all labels and not just the managed ones.

If a user has user specified labels, why would we drop those in this case?

test/e2e/e2e_test.go
mimowo: Done, I parametrized the helper function with duration. I keep 20s for the old test, but we could consider reducing the time too. I tested that it passes anyway.

api/jobset/v1alpha2/jobset_types.go
kannon92: Ah I see. I missed that this is resume.

pkg/controllers/jobset_controller.go
kannon92: Please push up the change before resolving.

api/jobset/v1alpha2/jobset_types.go
kannon92:
```
		// Copy the Job labels managed by the Job controller
		for _, labelKey := range jobset.JobManagedLabels {
			newJob.Spec.Template.Labels[labelKey] = job.Spec.Template.Labels[labelKey]
		}
		job.Spec.Template.Labels = collections.MergeMaps(
			newJob.Spec.Template.Labels,
			rjob.Template.Spec.Template.Labels,
		)
```

I think that we are already copying all labels. I guess my point is why don't we just copy all labels from job.Spec.Templates.Labels[labelKey]?

We should not need to expose the details of those labels and copy all labels.

kannon92: > If a user has user specified labels, why would we drop those in this case?

This seems like a bug to me. A user creates a JobSet and sets labels in their template. I don't expect that they should ever be dropped?

mimowo: The code you quote here:
```
		job.Spec.Template.Labels = collections.MergeMaps(
			newJob.Spec.Template.Labels,
			rjob.Template.Spec.Template.Labels,
		)
```
merges labels from `newJob` and `rjob` (Job template in JobSet). None of the two places contain the labels in question. The labels are only present on the old Job, and so I copy from them.

kannon92: So could we copy all the labels from the old Job without looping over the special managedByJob ones?

mimowo: > We should not need to expose the details of those labels and copy all labels.

For now I don't see another way of fixing the bug, either:
1. copy over the Job labels, 
2. delete the Jobs and recreate the Jobs, letting API server default all the fields / labels

This is partly why I prefer (2.) - and implemented initially. However, I understand that updating the Jobs might be preferred as deleting and recreating is a bigger change.

kannon92: My point is that 1 could be done just by copying ALL labels from the previous job.

pkg/controllers/jobset_controller.go
```diff
@@ -452,30 +446,44 @@ func (r *JobSetReconciler) resumeJob(ctx context.Context, job *batchv1.Job, repl
...
+		}
+		// Copy the Job labels managed by the Job controller
+		for _, labelKey := range jobset.JobManagedLabels {
+			newJob.Spec.Template.Labels[labelKey] = job.Spec.Template.Labels[labelKey]
+		}
 		job.Spec.Template.Labels = collections.MergeMaps(
-			job.Spec.Template.Labels,
-			replicatedJobPodTemplate.Labels,
+			newJob.Spec.Template.Labels,
+			rjob.Template.Spec.Template.Labels,
```
kannon92, line 470: What if we put job.Spec.Template.Labels[labelKey] in the MergeMaps?

api/jobset/v1alpha2/jobset_types.go
mimowo: > So could we copy all the labels from the old Job without looping over the special managedByJob ones?

Consider the following transitions by Kueue: resume (RF1) -> suspend -> resume (RF2).

It is possible that resuming on RF2 does not add the same labels as for RF1. If we copied all labels, then we would always trail the old ones from RF1.

Indeed, this is mostly problem for `nodeSelectors`, where Kueue may want to choose a new set of nodeSelectors for RF2. I would be ok to accept this issue for labels as it does not have much of a practical issue for Kueue currently, but it would be a "known issue", and may hit us one day.

pkg/controllers/jobset_controller.go
kannon92: I just don't understand why we should selectively only copy the JobManagedLabels.

Wouldn't that mean that we dropping the original job labels and only preserving the JobManagedLabels?

If I was building a custom CRD using JobSet and I had custom labels and they were dropped, that would be a bug to me.

mimowo: > I just don't understand why we should selectively only copy the JobManagedLabels.

I feel this is a very similar question as already being discussed above. Take a look at the most recent answer there: https://github.com/kubernetes-sigs/jobset/pull/640/files/4b9625725820a015d29bbcb65b114a5baa032d29#r1711840934. Basically, I merge labels from two sources: JobSet template, and oldJob (but only limited to `JobManagedLabels`, because those are added by API server.

api/jobset/v1alpha2/jobset_types.go
mimowo: > My point is that 1 could be done just by copying ALL labels from the previous job.

Seems like a race condition :). Consider the scenario I describe here: https://github.com/kubernetes-sigs/jobset/pull/640#discussion_r1711840934.  The code will then remain problematic if RF1 adds a label, which is not expected in RF2, because with this approach we would never get rid of it. 

It will remain problematic in some corner cases and hard to debug. It will also be very surprising why we restore nodeSelectors, but not labels.

test/e2e/e2e_test.go
```diff
@@ -131,6 +132,82 @@ var _ = ginkgo.Describe("JobSet", func() {
 		})
 	})
 
+	ginkgo.When("job is suspended and resumed", func() {
```
kannon92, line 135:
```suggestion
	ginkgo.When("jobset is suspended and resumed", func() {
```

api/jobset/v1alpha2/jobset_types.go
mimowo: I synced with @kannon92 on slack and there is a simpler solution which I implement here: https://github.com/kubernetes-sigs/jobset/pull/644. It does not cover fully the case of "restoring" a PodTemplate, but it should cover most cases where the admission to RF2 overrides the previous values.

This is needed anyway as the first step, and part of this PR. We can return to these bits if really shown they are needed.

test/e2e/e2e_test.go
mimowo: Applied in the other PR

mimowo:
/close
After syncing with @kannon92 we decided it is better to start with something simple, which already covers most use cases, so I open another PR: https://github.com/kubernetes-sigs/jobset/pull/644

PR #642, opened 2024-08-09
drop job-name from labels as it is not used

### LINKED ISSUE #26: Headless service should use fully qualified job label name.
https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L244

```
					// TODO: Migrate to the fully qualified label name.
					"job-name": job.Name,
```

Generally, we should use names that are fully qualified for labels.

Not sure if that means we should reject JobSpec names that aren't valid label names.

kannon92:
Drop use of legacy label. It seems that this was no longer used.

Fixes #26 

```release-note
NONE
```


kannon92:
/hold

https://github.com/kubernetes-sigs/jobset/pull/640 is introducing usage of this so it may be used.

danielvegamyhre:
/lgtm
/approve

Thanks!

PR #627, opened 2024-07-26
Validation for Coordinator field

danielvegamyhre:
Part of #618

PR #643, opened 2024-08-09
add the unparam linter to golangci and fix those issues flagged

kannon92:
#### What type of PR is this?
/kind cleanup

add linter for unused parameters.

Fixes #

```release-note
NONE
```

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -16,3 +16,4 @@ linters-settings:
 linters:
   enable:
     - goimports
+    - unparam
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -489,11 +489,7 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
 	startupPolicy := js.Spec.StartupPolicy
 
 	for _, replicatedJob := range js.Spec.ReplicatedJobs {
-		jobs, err := constructJobsFromTemplate(js, &replicatedJob, ownedJobs)
-		if err != nil {
-			return err
-		}
-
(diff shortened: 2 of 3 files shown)


pkg/webhooks/jobset_webhook.go
```diff
@@ -382,7 +382,7 @@ func validateCoordinator(js *jobset.JobSet) error {
 
 // replicatedJobByName fetches the replicatedJob spec from the JobSet by name.
 // Returns nil if no replicatedJob with the given name exists.
-func replicatedJobByName(js *jobset.JobSet, replicatedJob string) *jobset.ReplicatedJob {
+func replicatedJobByName(js *jobset.JobSet) *jobset.ReplicatedJob {
```
danielvegamyhre, line 385: Looks like in https://github.com/kubernetes-sigs/jobset/pull/627 the `replicatedJobByName` function had a bug that was missed. We should keep the `replicatedJob` string param for this function, but on line 387 it should say:

```go
		if rjob.Name == replicatedJob { 
                    ...
                 }

```

rather than hardcoding the rjob.Name == js.spec.coordinator.replicatedJob. The coordinator replicatedJob is being passed in as a param by the caller.

kannon92: kk. Feel free to propose a PR to fix that bug. I'll add this in after.

danielvegamyhre: sounds good, just created #645 with the fix

kannon92: updated with your fix.

danielvegamyhre:
/lgtm
/approve

Thanks for adding this!

PR #644, opened 2024-08-09
Allow to update JobSets on suspend

mimowo:
#### What type of PR is this?

/kind bug

To fix the main scenarios of integration with Kueue, where Kueue may want to suspend a workload, 
and re-admit in another ResourceFlavor (with different nodeSelectors)

Fixes https://github.com/kubernetes-sigs/jobset/issues/624

This is a minimal fix for the main scenarios needed for integration with Kueue - it does not
allow to "restore" fully PodTemplate, but it allows to overwrite it, which is enough in most cases
See previous approach: https://github.com/kubernetes-sigs/jobset/pull/640

There are two commits:
1. shows that the e2e demonstrates the issue, see example failure: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_jobset/644/pull-jobset-test-e2e-main-1-27/1821972758104379392
2. the change which fixes the issue

```release-note
Allow updating PodTemplate on suspend. This fixes the main scenarios for integration with Kueue
to support eviction of workloads and re-admitting in another ResourceFlavor (with other
nodeSelectors).
```

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -133,7 +133,7 @@ var _ = ginkgo.Describe("JobSet", func() {
 	})
 
 	// This test is added to test the JobSet transitions as Kueue would when:
-	// doing: resume in RF1 -> suspend -> resume in RF2.
+	// doing: resume in ResourceFlavor1 -> suspend -> resume in ResourceFlavor2.
 	// In particular, Kueue updates the PodTemplate on suspending and resuming
 	// the JobSet.
 	ginkgo.When("JobSet is suspended and resumed", func() {
@@ -145,7 +145,6 @@ var _ = ginkgo.Describe("JobSet", func() {
 
 			ginkgo.By("Create a suspended JobSet", func() {
 				js.Spec.Suspend = ptr.To(true)
-				js.Spec.TTLSecondsAfterFinished = ptr.To[int32](5)
 				gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
 			})
 
@@ -163,6 +162,9 @@ var _ = ginkgo.Describe("JobSet", func() {
 			})
 
 			ginkgo.By("Await for all Jobs to be active", func() {
+				// In this step the Pods remain Pending due to the nodeSelector
+				// which does not match any nodes. Still, JobSet considers as
+				// active any Jobs which have at least one Pending or Running Pod.
 				gomega.Eventually(func() int32 {
 					gomega.Expect(k8sClient.Get(ctx, jsKey, js)).Should(gomega.Succeed())
 					if js.Status.ReplicatedJobsStatus == nil {
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -24,6 +24,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/ptr"
 
 	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
 	"sigs.k8s.io/jobset/pkg/util/testing"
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -131,6 +132,66 @@ var _ = ginkgo.Describe("JobSet", func() {
 		})
 	})
 
+	ginkgo.When("job is suspended and resumed", func() {
```
kannon92, line 135:
```suggestion
	ginkgo.When("jobset is suspended and resumed", func() {
```

test/e2e/e2e_test.go
```diff
@@ -131,6 +132,66 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+
```
kannon92, line 136: Maybe you could add a comment here about this e2e test being more kueue integration.

mimowo: Done

test/e2e/e2e_test.go
mimowo: Done

kannon92 (commented):
/lgtm

/assign @ahg-g @danielvegamyhre

test/e2e/e2e_test.go
```diff
@@ -131,6 +132,70 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+	ginkgo.When("JobSet is suspended and resumed", func() {
+
+		ginkgo.It("should allow to resume JobSet after updating PodTemplate", func() {
+			ctx := context.Background()
+			js := sleepTestJobSet(ns, 1).Obj()
+			jsKey := types.NamespacedName{Name: js.Name, Namespace: js.Namespace}
+
+			ginkgo.By("Create a suspended JobSet", func() {
+				js.Spec.Suspend = ptr.To(true)
+				js.Spec.TTLSecondsAfterFinished = ptr.To[int32](5)
```
danielvegamyhre, line 148: why do we need to set TTL for this test?

test/e2e/e2e_test.go
```diff
@@ -131,6 +132,70 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+				}, timeout, interval).Should(gomega.Succeed())
+			})
+
+			ginkgo.By("Await for all Jobs to be active", func() {
+				gomega.Eventually(func() int32 {
+					gomega.Expect(k8sClient.Get(ctx, jsKey, js)).Should(gomega.Succeed())
+					if js.Status.ReplicatedJobsStatus == nil {
+						return 0
+					}
+					return js.Status.ReplicatedJobsStatus[0].Active
```
danielvegamyhre, line 171: I think we should add a comment here explaining a Job is considered "active" by the JobSet controller if the `job.Status.Active > 0`, meaning if the Job's pending + running pods > 0. 

Since for other developers it is probably confusing and unintuitive that in the previous step we deliberately made the pods unschedulable yet now we expect the jobs to be "active." I know personally I've always found the Job API counting a Job's pending pods as "active" a bit odd, I feel like the name could be better, although naming things can be deceptively tricky :)

test/e2e/e2e_test.go
mimowo: Not needed really, used it just to make sure the Job is deleted at some point, but we don't need to. Reverted.

test/e2e/e2e_test.go
mimowo: Makes sense, added a comment to clarify that.

test/e2e/e2e_test.go
```diff
@@ -131,6 +132,70 @@ var _ = ginkgo.Describe("JobSet", func() {
 		})
 	})
 
+	// This test is added to test the JobSet transitions as Kueue would when:
+	// doing: resume in RF1 -> suspend -> resume in RF2.
```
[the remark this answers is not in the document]
mimowo: Makes sense. Done

danielvegamyhre:
/lgtm
/approve
/hold

I'll remove the hold once CI tests pass.

mimowo:
@kannon92 @danielvegamyhre could we cherry-pick this?

danielvegamyhre:
> @kannon92 @danielvegamyhre could we cherry-pick this?

v0.6.0 will be released next week, I think that would be easier. Is it urgent to cherry pick this into v0.5.x?

mimowo:
I think 0.6 is fine.

kannon92:
@danielvegamyhre I think you said there was some issues with kicking off v0.6.0 so we are a holding pattern for this.

Is it worth cherry-picking this to unblock Kueue?

kannon92:
opened up https://github.com/kubernetes-sigs/jobset/pull/651#issuecomment-2294873274

PR #650, opened 2024-08-13
Add new job-id annotation to assign globally unique job index to each job

### LINKED ISSUE #649: Add Job ID label/annotation to provide a global index for each job across the entire JobSet
Add a label and annotation `jobset.sigs.k8s.io/job-id` which contains an integer value from 0 to N-1 where N=total number of jobs in the JobSet, to assign each Job a globally unique index within the JobSet.

**Why is this needed**:
Currently the `jobset.sigs.k8s.io/job-index` label contains the local job index within its parent replicatedJob (values range from 0 to N-1 where N=replicatedJob.replicas).

This means for a JobSet with multiple replicatedJobs, multiple jobs may have the same job index (for example, two replicated jobs of 1 replica each will result in 2 Jobs each with job-index of 0).

In [TPU multislice training](https://cloud.google.com/kubernetes-engine/docs/how-to/tpu-multislice) we have used a JobSet with a single replicated job and exclusive job placement per slice (node pool), to assign 1 job replica exclusive usage of each TPU Slice. The job-index is then a natural and convenient way of assigning a unique TPU slice ID at the TPU runtime layer, which is required by TPU driver/runtime libraries for multislice training.

However, some users want to run multislice training workloads using a JobSet multiple replicated jobs with different templates - however, this is currently not possible because the job-index annotations from multiple different replicatedJobs are not unique (as described above), and TPU runtime requires unique slice IDs.

Therefore, we can add a new annotation, "jobset.sigs.k8s.io/job-id" which sets a globally job index that is unique across the JobSet.

danielvegamyhre:
Fixes #649

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -28,9 +28,9 @@ const (
 	// For each replicatedJob, this value will range from 0 to replicas-1, where `replicas`
 	// is equal to jobset.spec.replicatedJobs[*].replicas.
 	JobIndexKey string = "jobset.sigs.k8s.io/job-index"
-	// JobIDKey is a label/annotation set to an integer that is unique across the entire JobSet.
+	// JobGlobalIndexKey is a label/annotation set to an integer that is unique across the entire JobSet.
 	// For each JobSet, this value will range from 0 to N-1, where N=total number of jobs in the jobset.
-	JobIDKey string = "jobset.sigs.k8s.io/job-id"
+	JobGlobalIndexKey string = "jobset.sigs.k8s.io/job-global-index"
 	// JobKey holds the SHA256 hash of the namespaced job name, which can be used to uniquely identify the job.
 	JobKey string = "jobset.sigs.k8s.io/job-key"
 	// ExclusiveKey is an annotation that can be set on the JobSet or on a ReplicatedJob template.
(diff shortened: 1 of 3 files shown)


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -24,8 +24,15 @@ const (
...
 	ReplicatedJobNameKey string = "jobset.sigs.k8s.io/replicatedjob-name"
-	JobIndexKey          string = "jobset.sigs.k8s.io/job-index"
-	JobKey               string = "jobset.sigs.k8s.io/job-key"
+	// JobIndexKey is a label/annotation set to the index of the Job replica within its parent replicatedJob.
+	// For each replicatedJob, this value will range from 0 to replicas-1, where `replicas`
+	// is equal to jobset.spec.replicatedJobs[*].replicas.
+	JobIndexKey string = "jobset.sigs.k8s.io/job-index"
+	// JobIDKey is a label/annotation set to an integer that is unique across the entire JobSet.
+	// For each JobSet, this value will range from 0 to N-1, where N=total number of jobs in the jobset.
+	JobIDKey string = "jobset.sigs.k8s.io/job-id"
```
ahg-g, line 33: i would call it `jobset.sigs.k8s.io/job-global-index`

danielvegamyhre: Makes sense, updated PR accordingly.

PR #667, opened 2024-08-27
fix: delete active jobs right away when job finishes even when TTLSecondsAfterFinished is set

### LINKED ISSUE #666: Active jobs are not deleted when the job finishes if TTLSecondsAfterFinished is set
**What happened**:

When `TTLSecondsAfterFinished` is not set, the active jobs get deleted right away to free up resources. However, when `TTLSecondsAfterFinished` is set, the active jobs and the completed jobs all get deleted after the TTL expires when the jobset itself gets deleted.

`TTLSecondsAfterFinished` should control jobset cleanup but active jobs should be deleted regardless when a jobset is failed or complete to release resources.

**Environment**:
- Install tools:
- Others:

CecileRobertMichon:
#### What type of PR is this?

/kind bug

When `TTLSecondsAfterFinished` is not set, the active jobs get deleted right away to free up resources. However, when `TTLSecondsAfterFinished` is set, the active jobs and the completed jobs all get deleted after the TTL expires when the jobset itself gets deleted. We should always delete active jobs to free up resources when a jobset is failed or completed, then optionally cleanup the jobset itself if `TTLSecondsAfterFinished` is set.

Fixes #666

```release-note
Delete active jobs right away when job finishes even when TTLSecondsAfterFinished is set
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -154,6 +154,10 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 
 	// If JobSet is already completed or failed, clean up active child jobs and requeue if TTLSecondsAfterFinished is set.
 	if jobSetFinished(js) {
+		if err := r.deleteJobs(ctx, ownedJobs.active); err != nil {
+			log.Error(err, "deleting jobs")
+			return ctrl.Result{}, err
+		}
 		requeueAfter, err := executeTTLAfterFinishedPolicy(ctx, r.Client, r.clock, js)
 		if err != nil {
 			log.Error(err, "executing ttl after finished policy")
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -154,6 +154,10 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 
 	// If JobSet is already completed or failed, clean up active child jobs and requeue if TTLSecondsAfterFinished is set.
 	if jobSetFinished(js) {
+		if err := r.deleteJobs(ctx, ownedJobs.active); err != nil {
```
danielvegamyhre, line 157: I think this change makes sense, although I think we originally put deleteJobs after the TTL check because users wanted to use TTL to allow time to inspect logs via kubectl before everything is deleted, to debug more easily without having to craft queries in cloud logging systems.

Quickly freeing up resources that may still be being used by a completed/failed JobSet is  important as well, but for this purpose the user can just tune their TTL accordingly. 

@CecileRobertMichon @kannon92 @ahg-g any thoughts on this?

kannon92: Sgtm.

ahg-g: This is only deleting active jobs, which I think makes sense.

/lgtm
/approve

PR #677, opened 2024-09-16
Add global-job-replicas label/annotation

### LINKED ISSUE #676: Add global replica count label/annotation to support multislice training workloads with different templates

Add a label and annotation `jobset.sigs.k8s.io/global-job-replicas` which will contain the total number of replicas in the JobSet.

**Why is this needed**:

Some users want to run multislice training workloads using replicated jobs with different templates. #617 and #649 added important features to make this possible, but the use case is still not supported because the workloads need to know the total number of replicas. The new label and annotation will finally address this.

GiuseppeTT:
/kind feature

Fixes #676

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -22,6 +22,10 @@ import (
 const (
 	JobSetNameKey         string = "jobset.sigs.k8s.io/jobset-name"
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
+	// GlobalJobReplicasKey is a label/annotation set to the total number of replicatedJob replicas.
+	// For each JobSet, this value will be equal to the sum of `replicas`, where `replicas`
+	// is equal to jobset.spec.replicatedJobs[*].replicas.
+	GlobalJobReplicasKey string = "jobset.sigs.k8s.io/global-job-replicas"
 	// ReplicatedJobNameKey is used to index into a Jobs labels and retrieve the name of the parent ReplicatedJob
 	ReplicatedJobNameKey string = "jobset.sigs.k8s.io/replicatedjob-name"
 	// JobIndexKey is a label/annotation set to the index of the Job replica within its parent replicatedJob.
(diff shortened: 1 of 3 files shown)


googs1025:
/ok-to-test
@GiuseppeTT 
can you fix the EasyCLA? :)

GiuseppeTT:
> /ok-to-test @GiuseppeTT can you fix the EasyCLA? :)

Done : )

danielvegamyhre:
/lgtm

@GiuseppeTT is this going to be used to set TPU_NUM_SLICES (or similar) env var?

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -22,6 +22,10 @@ import (
 const (
 	JobSetNameKey         string = "jobset.sigs.k8s.io/jobset-name"
 	ReplicatedJobReplicas string = "jobset.sigs.k8s.io/replicatedjob-replicas"
+	// GlobalJobReplicasKey is a label/annotation set to the total number of replicatedJob replicas.
+	// For each JobSet, this value will be equal to the sum of `replicas`, where `replicas`
+	// is equal to jobset.spec.replicatedJobs[*].replicas.
+	GlobalJobReplicasKey string = "jobset.sigs.k8s.io/global-job-replicas"
```
ahg-g, line 28: `jobset.sigs.k8s.io/global-job-count`

ahg-g: Actually, we used `jobset.sigs.k8s.io/replicatedjob-replicas` above, so the equivalent here would be `jobset.sigs.k8s.io/global-replicas` perhaps.

GiuseppeTT:
> @GiuseppeTT is this going to be used to set TPU_NUM_SLICES (or similar) env var?

Yes! It is going to be used to set the env variable `MEGASCALE_NUM_SLICES`.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: Hi Abdullah.

Just to be extra sure. Are you suggesting to rename `jobset.sigs.k8s.io/global-job-replicas` to `jobset.sigs.k8s.io/global-replicas`? If so, I agree, it's a better name and I can change it : )

ahg-g: Yes, to align with the other annotation we have for number of replicas within a replicatedJob

GiuseppeTT: Done!

Issue #672, opened 2024-09-06
Support for the execution policy API in JobSet


andreyvelich:
**What would you like to be added**:

The new `executionPolicy` API which allows to submit replicated Jobs in order.
When the first replicated Jobs are the reached required condition, the next replicated Jobs are created.

**Note**. The complex DAG workflow capability is out of scope of this API, since we don't want to implement workflow functionality as part of this KEP. Users should consider to use Argo Workflows or Tekton Pipelines if they need it.

The initial API design:

```go
type JobSetSpec struct {
	ExecutionPolicy *ExecutionPolicy `json:"executionPolicy,omitempty"`
}

type ExecutionPolicy struct {
	// Order in which Jobs will be created. The default is AnyOrder.
	ExecutionPolicyOrder ExecutionPolicyOption `json:"executionPolicyOrder"`

	// After all replicated Jobs reach this status, the JobSet will create the next replicated Jobs.
	ReplicatedJobsStatus ReplicatedJobsStatusOption `json:"replicatedJobsStatus"`
}

type ExecutionPolicyOption string

const (
	AnyOrder ExecutionPolicyOption = "AnyOrder"

	InOrder ExecutionPolicyOption = "InOrder"
)

type ReplicatedJobsStatusOption string

// We don't add Ready condition here, since users can use the `startupPolicy` API for that.
const (
	ReadyStatus ReplicatedJobsStatusOption = "Succeeded"

	FailedStatus ReplicatedJobsStatusOption = "Failed"

	ActiveStatus ReplicatedJobsStatusOption = "Active"

	SuspendedStatus ReplicatedJobsStatusOption = "Suspended"
)

```

**Why is this needed**:

More context in this Kubernetes `wg-batch` thread: https://kubernetes.slack.com/archives/C032ZE66A2X/p1725400839102729

As part of the Kubeflow Training V2 APIs, we want to implement the LLM runtimes for LLMs fine-tuning: https://github.com/kubeflow/training-operator/issues/2170
That will require JobSet to orchestrate the sequence of 2-3 Jobs: Initializer -> Trainer -> Post-Processor.
The capacity management for such workload should be allocated for all Jobs combined and be controlled by Kueue.
When `TrainJob` is suspended, we will suspend all underlying Jobs.

I think, we might have more use-cases from the HPC side. Any thoughts @vsoch  @alculquicondor ?

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

[...]

PR #680, opened 2024-09-26
KEP-672: Serial Job Execution with DependsOn API

andreyvelich:
#### What type of PR is this?

Part of: https://github.com/kubernetes-sigs/jobset/issues/672.

/kind feature
/kind api-change

I created initial KEP for the ExecutionPolicy API.

A few points:

1. I was thinking a bit more on the `ExecutionPolicyOption` and we might want to introduce the `ExecutionPolicyRule` API, since the ReplicatedJob status can be different in the Job sequence.
For example, when MPIJob is created, the Initializer Job should be in `Succeeded` state, but the Launcher should be in the `Ready` state before running the `mpirun` command. WDYT about this API @tenzen-y @alculquicondor ?

2. @vsoch Please can you help us with the HPC use-case when Job sequence can be useful ?
3. Quota management via Kueue is still under discussion. What do you think we should add to this KEP ?

/assign @tenzen-y @johnugeorge @ahg-g @kannon92 @danielvegamyhre @googs1025 @akshaychitneni @shravan-achar


keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,383 @@
...
+                  persistentVolumeClaim:
+                    claimName: dataset-initializer
+                - name: model-initializer
+                  persistentVolumeClaim:
+                    claimName: model-initializer
+```
+
+#### Story 3
+
+TODO: Add HPC use-case with Job sequence
```
vsoch, line 247: The HPC use case could be similar to the already mentioned ML ones - e.g., "As an HPC user, I want to run a set of simulations with MPI that complete successfully (or some percentage) before using the data for a next step analysis. For the workload manager - "as a workload manager developer I want to ensure my lead broker is up and Ready before starting the workers to connect to it." Or a third case - "As an HPC user I want to scale up my simulation job only after a small set have completed successfully." And for that case, I could imagine essentially the same replicatedJob done twice, just with a larger size. If the outcome of the first isn't success you wouldn't launch the larger bulk of work.

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,383 @@
...
+
+	// Status in which the next replicated Jobs will be created.
+	ReplicatedJobsStatus ReplicatedJobsStatusOption `json:"replicatedJobsStatus"`
+}
+
+type ReplicatedJobsStatusOption string
+
+// TODO: What statuses do we want to support in the first version ?
+const (
+	ReadyStatus ReplicatedJobsStatusOption = "Ready"
```
danielvegamyhre, line 297: If we use an execution policy which specifies prior replicatedJobs must have a "Ready" status before subsequent ones can execute, isn't this effectively the same thing as using a StartupPolicy?

If so, perhaps we could eventually deprecate StartupPolicy in favor of this more flexible ExecutionPolicy over a couple releases

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,383 @@
...
+                    - mountPath: /workspace/dataset
+                      name: dataset-initializer
+              volumes:
+                - name: dataset-initializer
+                  persistentVolumeClaim:
+                    claimName: dataset-initializer
+                - name: model-initializer
+                  persistentVolumeClaim:
+                    claimName: model-initializer
+    - name: Node
```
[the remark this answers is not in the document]
andreyvelich: I think, we make the Node name here on purpose, since from the distributed training perspective this ReplicatedJob represents the single Training Node.
Here is the table which explains it: https://github.com/kubeflow/training-operator/tree/master/docs/proposals/2170-kubeflow-training-v2#worker-and-node-definition.
cc @tenzen-y

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,383 @@
...
+
+	FailedStatus ReplicatedJobsStatusOption = "Failed"
+)
+```
+
+### Implementation
+
+Open Questions:
+
+1. How we should calculate quota for JobSet with execution policy ?
```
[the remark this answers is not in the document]
andreyvelich: That is correct, the tricky part if we allow Kueue to admit JobSet if resources are available for the Initializer Job, but not available for the Trainer Job.
@tenzen-y How do you think we should implement the ExecutionPolicy initially in terms of Kueue integrations ?

keps/672-execution-policy/README.md
andreyvelich: >If we use an execution policy which specifies prior replicatedJobs must have a "Ready" status before subsequent ones can execute, isn't this effectively the same thing as using a StartupPolicy?

That is correct. Eventually, we just run the StartupPolicy functionality here.

>If so, perhaps we could eventually deprecate StartupPolicy in favor of this more flexible ExecutionPolicy over a couple releases

I am fine with that, but want to hear @kannon92 @ahg-g @vsoch @tenzen-y opinion on that.

vsoch: `StartupPolicy` seems more intuitive (from a user perspective) than an `ExecutionPolicy` -> Ready state, but I do think the latter is more flexible. I don't have a strong opinion either way, but if something is deprecated it should be well documented, etc.

kannon92: I think we need to phase out StartupPolicy over a few releases.

So I think this KEP should comment on compatibility between StartupPolicy and ExecutionPolicy.

If I specify StartupPolicy and execution Policy together, what would happen?  Should we enforce validation so that these are "either or".

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,389 @@
...
+desired replicated Jobs reach the status defined in replicatedJobsStatus, the controller creates
+the next set of replicated Jobs.
+
+If JobSet is suspended the all ReplicatedJobs will be suspended and the Job sequence starts again.
+
+### Defaulting/Validation
+
+- ExecutionPolicy is immutable.
+- ExecutionPolicyOrder of `AnyOrder` is default setting.
+- StartupPolicy should be empty if ExecutionPolicy is set
```
kannon92, line 328: I think StartupPolicy uses AnyOrder actually. So you would need to validate that StartupPolicy is not InOrder.

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
andreyvelich: We discussed today with @tenzen-y that we probably name this Job template as TrainerNode to make it clearer for user.

keps/672-execution-policy/README.md
andreyvelich: Good point.

keps/672-execution-policy/README.md
danielvegamyhre: >If I specify StartupPolicy and execution Policy together, what would happen? Should we enforce validation so that these are "either or".

Yes, I think we should enforce "either or" on the first release then deprecate startup policy over the next few releases.

andreyvelich: Sure, @kannon92 @danielvegamyhre do we need separate section to explain it ?
Currently, I explained it in the validation section:
```
StartupPolicy should be equal to AnyOrder when ExecutionPolicy is used.
The StartupPolicy API will be deprecated over the next few JobSet releases, since
ExecutionPolicy can be used with Ready status of ReplicatedJobs.
```

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+	// Names of the replicated Jobs that applied the status.
+	TargetReplicatedJobs []string `json:"targetReplicatedJobs"`
+
+	// Status the target ReplicatedJobs must reach before subsequent ReplicatedJobs begin executing.
+	ReplicatedJobsStatus ReplicatedJobsStatusOption `json:"replicatedJobsStatus"`
+}
+
+type ReplicatedJobsStatusOption string
+
+// TODO: What statuses do we want to support in the first version ?
```
kannon92, line 302: Ready and Succeeded I think are the only ones we should support.

kannon92: Though I would like to hear @ahg-g thoughts on ready..

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+### Manage quota for Job sequence
+
+It is hard to manage quota for sequence of Job within single JobSet. Usually, users don't want to
+lock resources (e.g. GPUs/TPUs) when the first Job is running.
+
+### Support complex DAGs with JobSet
+
+If users want to create complex DAG workflows, they should not use JobSet for such purpose.
+
+## Alternatives
```
kannon92, line 393: I still would like to hear why moving this type of workflow to a designated workflow engine is not a valid option?

Why would one want to use ExecutionPolicy without using a dag workflow?

For Kubeflow, why would not folding this kind of workflow into a Kubeflow pipeline not be sufficient for your purpose?

TrainingOperator is used for training while prep or post work could be specified as part of a workflow engine.

It is also not clear to me why one would not want to use this to for data prep-blend (cleaning data, transforming data) which to me is really when you should use a workflow engine..

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+kind: JobSet
+metadata:
+  name: fine-tune-llm
+spec:
+  executionPolicy:
+    executionPolicyOrder: InOrder
+    rules:
+      - targetReplicatedJobs:
+          - initializer
+        replicatedJobsStatus: Succeeded
```
danielvegamyhre, line 95: As a suggestion, I think naming this field `waitForReplicatedJobsStatus` would make the semantics of this clearer.

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+
+- `controllers`: `01/19/2024` - `30.2%`
+
+We will create a new file called execution_policy for the functionality.
+
+#### Integration tests
+
+- Having 2 Jobs running in sequence with Succeeded condition.
+- Suspend JobSet with execution policy being set.
+- Restart the failed Job when execution policy is set in JobSet.
```
danielvegamyhre, line 358: What would be the intended behavior for this test case? Wouldn't this depend on the JobSet's failure policy?

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+
```
danielvegamyhre, line 359: I think at a high level we should have tests for:

1. Happy path for using "ready" replicatedJob status for JobSet with 2+ replicatedJobs
2. Happy path for using "succeeded" replicatedJob status for JobSet with 2+ replicatedJobs
3. Case where first replicatedJob fails, we validate the second replicatedJob does not execute
4. Case where replicatedJob A is "ready" so replicatedJob B starts, then replicatedJob A fails and is no longer ready (what should the desired behavior here be? restart the whole jobset and begin executionPolicy from the beginning? or only restart replicatedJob A?)

As a follow up it would be great to test the case where the JobSet controller restarts in the middle of the execution policy (I'm not sure how feasible that is it to implement in an integration test, but this would be a great test for many of our features actually).

ahg-g:
Thanks @andreyvelich  I will take a look this weekend!

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+- Allowing for a percentage of Jobs in a ReplicatedJob to be ready to consider the
+  whole ReplicatedJob to ready.
+
+## Proposal
+
+### User Stories (Optional)
+
+#### Story 1
+
+As a user, I want to fine-tune LLM using multi-node training with JobSet. I have the first
```
kannon92, line 78: This story brings up some vagueness with the API. So you will start itiailizer and waiti for succeeded. Then I assume trainer job will start.

If I have a post-train replicated job, then do I also start the post-train replicated job once the initializer is complete.

We have executionPolicy InOrder but we don't specify what to do if the replicated job is not in the list.

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+                    claimName: dataset-initializer
+                - name: model-initializer
+                  persistentVolumeClaim:
+                    claimName: model-initializer
+```
+
+#### Story 3
+
+As an HPC user, I want to run a set of simulations with MPI that complete successfully
+(or some percentage) before using the data for a next step analysis.
```
kannon92, line 249: Some percentage is out of scope for this..

keps/672-execution-policy/README.md
andreyvelich: I agree with that. Support for Failed could be tricky given the failurePolicy API.

keps/672-execution-policy/README.md
andreyvelich: Let me add some points about it in the alternative section.

keps/672-execution-policy/README.md
andreyvelich: It will, but we need to make sure that if JobSet fails, we restart the every ReplicatedJobs in the job sequence.

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,401 @@
...
+}
+
+// ExecutionPolicyRule represents the execution policy rule for Job sequence.
+type ExecutionPolicyRule struct {
+
+	// Names of the replicated Jobs that applied the status.
+	TargetReplicatedJobs []string `json:"targetReplicatedJobs"`
+
+	// Status the target ReplicatedJobs must reach before subsequent ReplicatedJobs begin executing.
+	ReplicatedJobsStatus ReplicatedJobsStatusOption `json:"replicatedJobsStatus"`
```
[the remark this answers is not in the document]
andreyvelich: I think, making this as a list makes sense to me. Any thoughts @danielvegamyhre @kannon92 @vsoch ?

keps/672-execution-policy/README.md
andreyvelich: >If I have a post-train replicated job, then do I also start the post-train replicated job once the initializer is complete. We have executionPolicy InOrder but we don't specify what to do if the replicated job is not in the list.

If you have the post-train replicated job, it will start once the trainer job is complete, if you have this configuration:
```yaml
    - targetReplicatedJobs:
        - initializer
      waitForReplicatedJobsStatus: Succeeded
    - targetReplicatedJobs:
        - trainer-node
      waitForReplicatedJobsStatus: Succeeded
```
If you want to start trainer and post-processor job at the same time, after the initializer job is finished, just use this:
```yaml
    - targetReplicatedJobs:
        - initializer
      waitForReplicatedJobsStatus: Succeeded
```

keps/672-execution-policy/README.md
andreyvelich: Good point, let me remove it.

keps/672-execution-policy/README.md
kannon92: > If you want to start trainer and post-processor job at the same time, after the initializer job is finished, just use this:

This is a bit confusing to me because we only specify this with ExecutionPolicy: InOrder but if you don't specify any target replicated jobs than I guess it will start in any order?

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,398 @@
...
+}
+
+type ExecutionPolicyOption string
+
+const (
+	// This is the default settings.
+	// AnyOrder means that Jobs will be started in any order.
+	AnyOrder ExecutionPolicyOption = "AnyOrder"
+
+	// InOrder starts the ReplicatedJobs in order that they are listed. Jobs within a ReplicatedJob
```
shravan-achar, line 277: curious if this depends on the data structure used in the controller? Is it easy to establish order without a serial number in the spec? 

Could the order be changed while persisting to etcd?

keps/672-execution-policy/README.md
shravan-achar: To support the InOrder policy, would having an explicit serial number (ID or similar) in the replicatedJobsSpec indicating the index of the jobs be more intuitive?

andreyvelich: >This is a bit confusing to me because we only specify this with ExecutionPolicy: InOrder but if you don't specify any target replicated jobs than I guess it will start in any order?

That is valid point. I think, if the InOrder execution policy is used, by default we should wait until ReplicatedJobs will be in the Ready status before create the next one. That will just replicate the existing StartupPolicy feature.
In that case, if executionPolicy is set for JobSet with 3 replicatedJobs the following settings will be used as a default one:
```yaml
executionPolicy:
  executionPolicyOrder: InOrder
  rules:
    - targetReplicatedJobs:
        - job-1
        - job-2
        - job-3
      waitForReplicatedJobsStatus: Ready
```

Any thoughts ?

keps/672-execution-policy/README.md
andreyvelich: I've added my initial thoughts on it. From my point of view, as long as we suggest users to leverage this functionality to just initialization and preparing data for training, we should be fine.

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
andreyvelich: I guess, if we only support Ready + Succeeded it doesn't make sense to put list here, since Succeeded always comes after Ready status.
The list might be useful if in the future, we also support the Failed condition.
In that case, user can say that I want my ReplicatedJobs to be Succeeded or Failed before processing to the next ReplicatedJobs.

keps/672-execution-policy/README.md
andreyvelich: I think, this is how currently StartupPolicy API works. If I remember correctly, the InitContainers and Containers also executed in the order how the listed in the YAML. @tenzen-y @kannon92 do you have any thoughts on this ?

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
danielvegamyhre: >In that case, user can say that I want my ReplicatedJobs to be Succeeded or Failed before processing to the next ReplicatedJobs.

Rather than representing this as a list of ["Succeeded OR Failed"], we could represent this as a "Completed" status (succeeded or failed) and just keep the field as a single specific status to wait for.

keps/672-execution-policy/README.md
danielvegamyhre: > curious if this depends on the data structure used in the controller? Is it easy to establish order without a serial number in the spec?
> 
> Could the order be changed while persisting to etcd?

[ReplicatedJobs is represented as a slice data structure](https://github.com/kubernetes-sigs/jobset/blob/b92cbdcbe77f168a82cf12c9f349bdd053d2ae5c/api/jobset/v1alpha2/jobset_types.go#L82C1-L85C66) so the order is preserved, even through serialization, persistence to etcd, and deserialization. 

So operating on replicatedJobs in the order they are listed is straightforward and consistent.

keps/672-execution-policy/README.md
ahg-g: Can we list a few alternatives side-by-side to compare? 

I can also think of not having the rules at all in this struct, and have a `InOrderStartupRule` parameter in `ReplicatedJob` that takes values: `Succeeded|Ready`. This has the disadvantage of spreading the API though.

keps/672-execution-policy/README.md
andreyvelich: Yes, I mentioned it in one of the alternatives: https://github.com/kubernetes-sigs/jobset/blob/4a77dab7aabbc99fee259f61f6b4e4c7efb51470/keps/672-execution-policy/README.md#add-executionpolicyrule-parameter-into-the-startuppolicy-api.

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
andreyvelich: Completed status looks good to me.

keps/672-execution-policy/README.md
andreyvelich: Yes, this could be one of the options.
However, I tried to be consistent with other APIs, like [FailurePolicy](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L313) or [SuccessPolicy](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L325), where we indicate to which ReplicatedJobs we want to apply the rule.

@ahg-g Do you want me to add those API alternatives to the KEP ?

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,421 @@
...
+It is hard to manage quota for sequence of Job within single JobSet. Usually, users don't want to
+lock resources (e.g. GPUs/TPUs) when the first Job is running.
+
+### Support complex DAGs with JobSet
+
+If users want to create complex DAG workflows, they should not use JobSet for such purpose.
+
+## Alternatives
+
+### Add ExecutionPolicyRule parameter into the StartupPolicy API
```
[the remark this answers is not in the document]
andreyvelich: I think, both ExecutionPolicy and StartupPolicy makes sense to me for this feature.
Using different API would just make it simpler to deprecate the legacy API.
Any thoughts @kannon92 @tenzen-y @danielvegamyhre ?

keps/672-execution-policy/README.md
```diff
@@ -0,0 +1,421 @@
...
+type ExecutionPolicyRule struct {
+
+	// Names of the replicated Jobs that applied the status.
+	TargetReplicatedJobs []string `json:"targetReplicatedJobs"`
+
+	// Status the target ReplicatedJobs must reach before subsequent ReplicatedJobs begin executing.
+	WaitForReplicatedJobsStatus ReplicatedJobsStatusOption `json:"waitForReplicatedJobsStatus"`
+}
+
+type ReplicatedJobsStatusOption string
```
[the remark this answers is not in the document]
andreyvelich: Sure, let me add it.
I am wondering, do we want to introduce any new JobSet condition to improve UX ?
Similar to what we've done with StartupPolicy: https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L511

keps/672-execution-policy/README.md
```diff
@@ -264,45 +265,49 @@ The goal is to only focus on Job sequence to cover model training/HPC use-cases.
...
 type ReplicatedJobsStatusOption string
 
 const (
-	ReadyStatus ReplicatedJobsStatusOption = "Ready"
+ // Ready status means the Ready counter equals the number of child Jobs.
+ // .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].ready
+ ReadyStatus ReplicatedJobsStatusOption = "Ready"
 
-	SucceededStatus ReplicatedJobsStatusOption = "Succeeded"
+ // Succeeded status means the Ready counter equals the number of child Jobs.
```
mimowo, line 308:
```suggestion
 // Succeeded status means the Succeeded counter equals the number of child Jobs.
```

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
andreyvelich: I think, for ExecutionPolicy we might want to introduce another condition. For the StartupPolicy, currently user can see the following:
```yaml
  status:
    conditions:
    - lastTransitionTime: "2024-10-07T14:55:37Z"
      message: in order startup policy is in progress
      reason: InOrderStartupPolicyInProgress
      status: "False"
      type: StartupPolicyInProgress
    - lastTransitionTime: "2024-10-07T14:55:38Z"
      message: in order startup policy has completed
      reason: InOrderStartupPolicyCompleted
      status: "True"
      type: StartupPolicyCompleted
    - lastTransitionTime: "2024-10-07T14:55:38Z"
      message: jobset completed successfully
      reason: AllJobsCompleted
      status: "True"
      type: Completed
```

I am not sure if we should introduce another condition which shows that ReplicatedJob-X is completed.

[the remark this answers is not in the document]
danielvegamyhre: > I think, for ExecutionPolicy we might want to introduce another condition. For the StartupPolicy, currently user can see the following:
> 
> ```yaml
>   status:
>     conditions:
>     - lastTransitionTime: "2024-10-07T14:55:37Z"
>       message: in order startup policy is in progress
>       reason: InOrderStartupPolicyInProgress
>       status: "False"
>       type: StartupPolicyInProgress
>     - lastTransitionTime: "2024-10-07T14:55:38Z"
>       message: in order startup policy has completed
>       reason: InOrderStartupPolicyCompleted
>       status: "True"
>       type: StartupPolicyCompleted
>     - lastTransitionTime: "2024-10-07T14:55:38Z"
>       message: jobset completed successfully
>       reason: AllJobsCompleted
>       status: "True"
>       type: Completed
> ```
> 
> I am not sure if we should introduce another condition which shows that ReplicatedJob-X is completed.

Hmm, thinking about it more I think it would be simpler to extend StartupPolicy to support this behavior rather than create a new ExecutionPolicy API. All of the conditions would be reusable, and the code changes would be more minimal I believe.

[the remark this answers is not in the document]
andreyvelich: >All of the conditions would be reusable, and the code changes would be more minimal I believe.

I think, the status should look a little bit different.
Do we want to introduce the same condition if user sets sequence of ReplicatedJobs ?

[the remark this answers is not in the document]
kannon92: I think I tried this but we run into a problem with conditions. Conditions cannot be duplicated (either status is True or False) and adding particular "messages" to the StartupPolicy condition will cause an error.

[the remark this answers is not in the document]
andreyvelich: Yes, that is correct, they should be unique. I am not sure if adding that condition gives a lot of value for the user:
`type: StartupPolicyInProgress, status: True` or `type: ExecutionPolicyInProgress, status: True`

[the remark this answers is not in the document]
danielvegamyhre: If we extend StartupPolicy rather than adding a new ExecutionPolicy API, we should be able to continue using the same condition (e.g., `type: StartupPolicyInProgress, status: True`) regardless of whether the startupPolicy defined by the user waits for replicatedJob status Ready or Succeeded. Let me know if I'm misunderstanding the point you're discussing.

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
danielvegamyhre: Initially I thought the name "StartupPolicy" was too different to be applicable to this use case of serial execution (rather than serial startup), but after thinking more about the implementation details and maintenance of these 2 similar code paths, I think extending the StartupPolicy API to support serial execution would be preferable to building and maintaining a new, similar API. 

The name is close enough to be applicable, and I suppose we can always rename it to a more generic "ExecutionPolicy" if/when we graduate the API to beta/v1.

keps/672-execution-policy/README.md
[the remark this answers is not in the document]
andreyvelich: Sure, that makes sense.

andreyvelich:
I refactored KEP to address discussion about re-using the StartupPolicy API for that feature.
If we have time, I will be happy to share this KEP on upcoming Batch WG call today

kannon92:
I will not be able to attend that meeting.

kannon92 (commented):
I read your changes over again.

I think we should not change the behavior of the startupPolicy that exists.

There is still ambiguous aspects of your user stories. I think that we should validate that all replicatedJobs have a corresponding status if we have a new policy for StartupPolicy.

Maybe we use `inOrder` to mean the original startupPolicy intention where every replicated job must match ready.

So in case of a jobSet with replicatedJobA, replicatedJobB.

StartupPolicy: inOrder would actually mean:

```yaml
  startupPolicy:
    startupPolicyOrder: InOrder
    rules:
      - targetReplicatedJobs:
          - replicatedJobA
        waitForReplicatedJobsStatus: Ready
        - targetReplicatedJobs:
          - replicatedJobB
        waitForReplicatedJobsStatus: Ready
```
 We should aim to keep the existing APIs and not break existing users.
 
 I also think we should have validation that when using the new policy `InOrderWithStatus` (name tbd), we validate that the user specified all replicated jobs so we don't have undefined behavior with our API.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,448 @@
...
+
+- Add support for serial Job execution for the StartupPolicy API.
+
+### Non-Goals
+
+- Support workflow management like DAGs outside of Job sequence.
+  - Users should consider to use Argo Workflows or Tekton Pipelines for such use-cases.
+- Allowing for a percentage of Jobs in a ReplicatedJob to be ready to consider the
+  whole ReplicatedJob to ready.
+
```
tenzen-y, line 73: Let's say that we do not aim to introduce any other status other than "Succeeded" and "Ready".

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,448 @@
...
+As an HPC user, I want to run a set of simulations with MPI that complete successfully before
+using the data for a next step analysis.
+
+As an HPC user I want to scale up my simulation job only after a small set have completed
+successfully. For that case, I could imagine essentially the same replicatedJob done twice,
+just with a larger size. If the outcome of the first isn't success you wouldn't launch
+the larger bulk of work.
+
+### Risks and Mitigations
```
tenzen-y, line 258: Could you mention the collaboration of Kueue?
I guess that this will take a longer time until the main Job will be created. That indicates that reserving a whole of JobSet resources for a while. But we can mitigate the risks to enqueue each Jobs instead of enqueuing a whole of JobSet.

I believe that improving collaboration between JobSet and Kueue in the future.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,448 @@
...
+)
+
+// StartupPolicyRule represents the startup policy rule for Job sequence.
+type StartupPolicyRule struct {
+
+ // Names of the replicated Jobs that applied the status.
+ TargetReplicatedJobs []string `json:"targetReplicatedJobs"`
+
+ // Status the target ReplicatedJobs must reach before subsequent ReplicatedJobs begin executing.
+ WaitForReplicatedJobsStatus ReplicatedJobsStatusOption `json:"waitForReplicatedJobsStatus"`
```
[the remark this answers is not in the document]
danielvegamyhre: Hmm, I think `waitForReplicatedJobsStatus` is more straightforward and clear.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,448 @@
...
+ SucceededStatus ReplicatedJobsStatusOption = "Succeeded"
+)
+
+// StartupPolicyRule represents the startup policy rule for Job sequence.
+type StartupPolicyRule struct {
+
+ // Names of the replicated Jobs that applied the status.
+ TargetReplicatedJobs []string `json:"targetReplicatedJobs"`
+
+ // Status the target ReplicatedJobs must reach before subsequent ReplicatedJobs begin executing.
```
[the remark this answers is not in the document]
andreyvelich: Yeah, let me add it.

keps/672-serial-job-execution/README.md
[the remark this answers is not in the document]
andreyvelich: Yeah, I think `waitForReplicatedJobsStatus` is more declarative.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,448 @@
...
+ // Succeeded status means the Succeeded counter equals the number of child Jobs.
+ // .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].succeeded
+ SucceededStatus ReplicatedJobsStatusOption = "Succeeded"
+)
+
+// StartupPolicyRule represents the startup policy rule for Job sequence.
+type StartupPolicyRule struct {
+
+ // Names of the replicated Jobs that applied the status.
+ TargetReplicatedJobs []string `json:"targetReplicatedJobs"`
```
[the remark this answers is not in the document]
andreyvelich: No, I don't think order matters, we just need to make sure that all replicated job names are unique and persists in the targetReplicatedJobs lists.
I will add it in the validation.

keps/672-serial-job-execution/README.md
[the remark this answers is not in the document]
tenzen-y: This is not my strong opinion. So, I'm ok with `waitForReplicatedJobsStatus`.

keps/672-serial-job-execution/README.md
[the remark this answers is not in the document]
tenzen-y: That makes sense. So, could you mention in the API comment that the order is not considered?

andreyvelich:
> I read your changes over again.
> 
> I think we should not change the behavior of the startupPolicy that exists.
> 
> There is still ambiguous aspects of your user stories. I think that we should validate that all replicatedJobs have a corresponding status if we have a new policy for StartupPolicy.
> 
> Maybe we use `inOrder` to mean the original startupPolicy intention where every replicated job must match ready.
> 
> So in case of a jobSet with replicatedJobA, replicatedJobB.
> 
> StartupPolicy: inOrder would actually mean:
> 
> ```yaml
>   startupPolicy:
>     startupPolicyOrder: InOrder
>     rules:
>       - targetReplicatedJobs:
>           - replicatedJobA
>         waitForReplicatedJobsStatus: Ready
>         - targetReplicatedJobs:
>           - replicatedJobB
>         waitForReplicatedJobsStatus: Ready
> ```
> 
> We should aim to keep the existing APIs and not break existing users.
> 
> I also think we should have validation that when using the new policy `InOrderWithStatus` (name tbd), we validate that the user specified all replicated jobs so we don't have undefined behavior with our API.

@kannon92 as we discussed yesterday it might be fine to re-use the existing enum. [The existing implementation](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L497-L498) just looks for Ready condition for each ReplicatedJob.
If the rules are empty for `startupPolicyOrder: InOrder`, can we just inject the following spec:
```
kind: JobSet
metadata:
  name: startup-policy
spec:
  startupPolicy:
    startupPolicyOrder: InOrder
    rules:
      - targetReplicatedJobs:
          - job-1
          - job-2
        waitForReplicatedJobsStatus: Ready
  replicatedJobs:
    - name: job-1
    - name: job-2
    - name: job-3
```

andreyvelich:
As we discussed yesterday, I added alternative option to add the WaitForStatus API under ReplicatedJob.
@ahg-g @kannon92 @tenzen-y @danielvegamyhre If you have any other ideas on how we can improve that API, let me know.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,512 @@
...
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: deepspeed-mpi
+spec:
+  startupPolicy:
+    startupPolicyOrder: InOrder
+    rules:
```
ahg-g, line 167: this is going to be specific to InOrder, so what about calling `InOrderStartupRules`

I also suggest to be more explicit here in the way the user describes the rules:

```
InOrderStartupRules
  replicatedJobsWaitingOnSucceededStatus
  - initializer
  replicatedJobsWaitingOnReadyStatus
  - launcer 
```

I am not too sure about the exact name of the `replicatedJobsWaitingOnSucceededStatus|replicatedJobsWaitingOnReadyStatus` fields, please suggest others

andreyvelich: I like this idea, what do we think about this name: `replicatedJobsWaitForSucceededStatus` @danielvegamyhre @kannon92 @tenzen-y @ahg-g @mimowo ?

ahg-g: sgtm

danielvegamyhre: > this is going to be specific to InOrder, so what about calling `InOrderStartupRules`
> 
> I also suggest to be more explicit here in the way the user describes the rules:
> 
> ```
> InOrderStartupRules
>   replicatedJobsWaitingOnSucceededStatus
>   - initializer
>   replicatedJobsWaitingOnReadyStatus
>   - launcer 
> ```
> 
> I am not too sure about the exact name of the `replicatedJobsWaitingOnSucceededStatus|replicatedJobsWaitingOnReadyStatus` fields, please suggest others

I like this, it also simplifies the API imo.

>I like this idea, what do we think about this name: replicatedJobsWaitForSucceededStatus @danielvegamyhre @kannon92 @tenzen-y @ahg-g @mimowo ?

Personally I prefer `replicatedJobsWaitingOnSucceededStatus` since it is close to a real, grammatically correct phrase, so interpreting the meaning/purpose of the variable is effortless. 

`replicatedJobsWaitForSucceededStatus` is a bit jarring / confusing to me.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,499 @@
...
+metadata:
+  name: deepspeed-mpi
+spec:
+  startupPolicy:
+    startupPolicyOrder: InOrder
+    inOrderStartupRules:
+      replicatedJobsWaitForSucceededStatus:
+        - initializer
+      replicatedJobsWaitForReadyStatus:
+        - launcher
```
alculquicondor, line 171: Wouldn't users have to list here all the replicated jobs that aren't listed in `replicatedJobsWaitForSucceededStatus`?

I would prefer to see something more like this:

```yaml
inOrderStartupRules:
  steps:
  - replicatedJobs: ["initializer"]
  - replicatedJobs: ["launcher", trainer-node"]
```

But names need more work.

alculquicondor: It could be a good exercise to forget that we ever added `startupPolicyOrder`.

Then, the full API could look something like this:

```yaml
startupPolicy:
  steps:
  - replicatedJobs: ["initializer"]
  - replicatedJobs: ["launcher", "trainer-node"]
    waitForReadiness: ["launcher"]
```

keps/672-serial-job-execution/README.md
andreyvelich: Sure, I am ok with that name too.

keps/672-serial-job-execution/README.md
andreyvelich: >Wouldn't users have to list here all the replicated jobs that aren't listed in replicatedJobsWaitForSucceededStatus?

@alculquicondor No, since we don't need to list `trainer-node` replicatedJob in that list.

>Then, the full API could look something like this:

How can we make sure that steps API are following the order of ReplicatedJobs ?
For example, user can't do this:
```
startupPolicy:
  steps:
  - replicatedJobs: ["launcher"]
  - replicatedJobs: ["initializer", "trainer-node"]
```

ahg-g: I am ok with the steps approach, but can one propose a full API please? I don't fully understand the semantics of the proposal in https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1805221427

What we want to capture for each replicatedJob is what status we should be blocking on before proceeding to create the next one, the options are: noWait|Ready|Finished (finished means succeeded or failed)

andreyvelich: >Clarify something for me: would it be possible to start multiple replicated jobs in parallel?

I think, we haven't discussed it. Basically, what if we want to run Job 1 and Job 2 in parallel before start a Job 3 and Job 4
```
Job 1 -- Succeeded --\
                      Job 3 ---- Ready ---> Job 4
Job 2 -- Succeeded --/
```

@alculquicondor @tenzen-y @ahg-g @kannon92 @danielvegamyhre @vsoch  Do you see an use-case like the above one ?

vsoch: Yes, you'd typically have two models that are going into a higher level analysis. However, I probably wouldn't put this on the level of JobSet but a workflow manager or state machine equivalent.

alculquicondor: Yeah, I agree that it's worth having such use case in a workflow manager.

But I have heard of customers running workers in two machine types.

So, more like:

```
initializer -> (head, workers type 1 + workers type2)
```

In that case, you might want the head to be ready before the workers, but both types of workers can start together.

IMO, this breaks the original intention of `startupPolicyOrder`. So I would prefer we deprecate it when using "steps" or "barriers"

tenzen-y: > But I have heard of customers running workers in two machine types.

I think that this is generic use cases. We often use similar machine types (especially GPUs) to improve GPU obtainability, like NVIDIA A100 40GB + A100 80GB (actual distributed Job processes use only 40GB VRAM), when we prioritize GPU obtainability rather than costs (prices).

> IMO, this breaks the original intention of startupPolicyOrder. So I would prefer we deprecate it or not allow to set it when using "steps" or "barriers"

I would recommend deprecating the `startupPolicyOrder` and making `statupPolicyOrder` and `inOrderStartupRules` mutually exclusive, then in the next API version (v1beta1 or something) we remove the `inOrderStartupOrder`.

@andreyvelich @danielvegamyhre @ahg-g What is your opinion?

andreyvelich: > In that case, you might want the head to be ready before the workers, but both types of workers can start together.

@alculquicondor I think, this use case can be solved with this API:
```
startupPolicy:
  startupPolicyOrder: InOrder
  inOrderStartupRules:
    replicatedJobsWaitForSucceededStatus:
      - initializer
    replicatedJobsWaitForReadyStatus:
      - driver
replicatedJobs:
  - name: initializer
  - name: driver
  - name: trainer-node-nvidia-a100
  - name: trainer-node-nvidia-h100
```


>I would recommend deprecating the startupPolicyOrder and making statupPolicyOrder and inOrderStartupRules mutually exclusive, then in the next API version (v1beta1 or something) we remove the inOrderStartupOrder.

@tenzen-y I think, we should discuss whether we want to support use-case that I introduced above: https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1809273523
For example, if we want to run several Jobs in parallel and wait until they succeeded or become ready.
What if we have TensorFlow distributed training job which has 1 chief, 2 parameter servers, and workers. E.g.:
```
replicatedJobs:
  - name: chief
  - name: ps-1
  - name: ps-2
  - name: trainer-node
```
Do we need to wait until ps-1 and ps-2 is ready before start the trainer-node Job ?

danielvegamyhre: >What we want to capture for each replicatedJob is what status we should be blocking on before proceeding to create the next one, the options are: noWait|Ready|Finished (finished means succeeded or failed)

This may not be feasible unless we deprecate the existing startup policy API entirely, but I think an API capturing these semantics could potentially be much simpler and clearer if we embed each ReplicatedJob's dependencies locally into its spec, rather than trying to capture all the dependencies for all replicated jobs in one global JobSet level policy.

For example:

```yaml
replicatedJobs:
  - name: initializer
  - name: driver
     waitForCompleted:
     - initializer
  - name: trainer-node-nvidia-a100
     waitForReady:
     - driver
  - name: trainer-node-nvidia-h100
     waitForReady:
     - driver
```

Instead of:

```yaml
startupPolicy:
  startupPolicyOrder: InOrder
  inOrderStartupRules:
    replicatedJobsWaitForSucceededStatus:
      - initializer
    replicatedJobsWaitForReadyStatus:
      - driver
replicatedJobs:
  - name: initializer
  - name: driver
  - name: trainer-node-nvidia-a100
  - name: trainer-node-nvidia-h100
```

I think packing all these rules into one global policy is going to make the API confusing.

vsoch: I'm still worried this is packing workflow logic into a unit of work, JobSet. It would be akin to kueue not existing and putting scheduling logic all here.

alculquicondor: >  I think an API capturing these semantics could potentially be much simpler and clearer if we embed each ReplicatedJob's dependencies locally into its spec, rather than trying to capture all the dependencies for all replicated jobs in one global JobSet level policy.

I agree with this. Although having it in a "startupPolicy" field makes it visible at a single glance, as you don't have the podspecs in between. But any option so far has pros and cons.

Going into the details of your proposal, I agree with @vsoch that it might be overlapping too much with a workflow manager. So maybe we can change a little where to put the "hold", like this:

```yaml
replicatedJobs:
  - name: initializer
    waitForStatusBeforeContinuing: Completed
  - name: driver
     waitForStatusBeforeContinuing: Ready
  - name: trainer-node-nvidia-a100
  - name: trainer-node-nvidia-h100
```

tenzen-y: > @tenzen-y I think, we should discuss whether we want to support use-case that I introduced above: https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1809273523

The JobSet project is not only for ML Traininig, this is used for HPC, CI/CD, and more Job systems.
So, I believe that we should consider the possible use cases during design. Additionally, once we introduce something to the APIs, changing the APIs loses the backward compatibility. So, I would like to evaluate possible designs here.

danielvegamyhre: >The JobSet project is not only for ML Traininig, this is used for HPC, CI/CD, and more Job systems.

I think JobSet should focus on ML training and HPC support as the north star - CI/CD systems and other use cases are not really the goal of this project, there's other, more specialized software for that. IMO we should try to do one thing really well, rather than support a bunch of different things less robustly. 

>Do you assume that the waitForStatusBeforeContinuing can be specified only when we specify the startupPolicyOrder: InOrder?

Here we are discussing removing StartupPolicy in favor of embedding the ReplicateJob's dependencies (if any) into the ReplicatedJob spec itself, rather than capturing the dependencies in a global, JobSet level policy, as proposed in my original comment [here](https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1809945813). 

@alculquicondor the change you suggested would be fine with me I think. Curious to get @ahg-g's thoughts as well.

tenzen-y: > I think JobSet should focus on ML training and HPC support as the north star - CI/CD systems and other use cases are not really the goal of this project

I thought that the JobSet is just a set of Jobs. So I assumed that JobSet covers the user case which Job covers.
But, as you say, JobSet focus on the ML/HPC 

> JobSet: a k8s native API for distributed ML training and HPC workloads

tenzen-y: > Here we are discussing removing StartupPolicy in favor of embedding the ReplicateJob's dependencies (if any) into the ReplicatedJob spec itself, rather than capturing the dependencies in a global, JobSet level policy, as proposed in my original comment https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1809945813.

I see. In that case, the `AnyOrder` is not supported by the new `waitForStatusBeforeContinuing`, and then the JobSet operator creates Jobs based on the order of replicatedJobs, right?

tenzen-y: So, after graduating the JobSet API to beta or GA, we will not provide the possibility for the AnyOrder Jobs creation, right?

tenzen-y: > AnyOrder will still be supported. To achieve AnyOrder startup, the user would simply not specify any dependencies for their ReplicatedJobs (i.e., waitForStatusBeforeContinuing is nil):

That makes sense. Thank you for expanding the API description. +1 on the `waitForStatusBeforeContinuing` approach if @ahg-g does not have any objections.

andreyvelich: @danielvegamyhre @alculquicondor @tenzen-y  can we support use case that i described here: https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1809570498 If we introduce `waitForStatusBeforeContinuing` ?
E.g. the JobSet should do this:
```
                         PS-1 -- Ready --\
Initializer --Succeeded-->                Trainer
                         PS-2 -- Ready --/
```

alculquicondor: No, that wouldn't be supported. Is this a common use case or can we apply YAGNI?

andreyvelich: Actually, it might become common for large scale distributed training use-cases to have 2 parameter servers to reduce load on a single communicator.
I just want to make sure that our API will be future proof.

@alculquicondor @tenzen-y @kannon92 @danielvegamyhre @ahg-g @vsoch  What do you think about this API:
```yaml
replicatedJobs:
  - name: initializer
  - name: ps-1
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: ps-2
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: trainer-node
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
      - name: ps-1
        status: Ready
      - name: ps-2
        status: Ready
```

We can validate that ReplicatedJobs can only reference the previously listed Jobs (e.g. `ps-2` can only reference `ps-1` or `initializer`)

vsoch: Is there no way to represent this more simply - instead of hardcoding "waitForPreviousJobStatusBeforeContinuing" have an generic pattern where there is a rule, and a rule has a trigger and action? For our ensemble-operator (that uses ensemble-python under the hood) I designed a pretty simple pattern, and it's event driven based on events from the workload manager queue ([example](https://github.com/converged-computing/ensemble-python/blob/80f5c095c4483d5e29402097443eb0dedbab62cc/examples/grow-shrink-example.yaml#L21-L41)). Could a replicatedJob not have a similar list of rules that respond to events (or conditions) with some action?

andreyvelich: I might be very hard to manage quota for such Job sequence with rules.
E.g. Job dispatcher needs to analyse set of rules for each ReplicatedJob to understand how to allocate quota.

@vsoch Do you have any JobSet API example in mind that we can discuss here ?

vsoch: I was planning on doing exactly that, but JobSet is immutable in every way so I couldn’t put it on this context. My plan is to put together a demo/talk for the group to teach about ensembles and the use case, and then hopefully be able to provide that.

ahg-g: > to have 2 parameter servers to reduce load on a single communicator.

Wouldn't those be represented as two jobs in a single replicatedJob?

I don't want to dismiss the use case, but also the API described in https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1818283828 starts to become complicated to the most common cases where this will be used.

If we consider https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1814857024 as out of scope, I feel the API currently described in the KEP is simple and it is able to clearly describe the targeted use cases, or did I miss something?

danielvegamyhre: > to have 2 parameter servers to reduce load on a single communicator.
>> Wouldn't those be represented as two jobs in a single replicatedJob?

I think @ahg-g is correct about this.

>If we consider https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1814857024 as out of scope, I feel the API currently described in the KEP is simple and it is able to clearly describe the targeted use cases, or did I miss something?

@ahg-g I agree we should consider that out of scope. However, I actually think the API currently described in the KEP is more complicated, and the alternative above is a better fit. This thread is getting long so let's do a side-by-side comparison here.

Option 1: JobSet level API currently proposed in the KEP (global set of startup policy rules which capture the local dependencies of all replicatedJobs): 
  ```yaml
  
   startupPolicy:
    startupPolicyOrder: InOrder
    inOrderStartupRules:
      replicatedJobsWaitingOnReadyStatus:
      - job-2
      replicatedJobsWaitingOnSucceededStatus:
      - job-1
  replicatedJobs:
    - name: job-1
    - name: job-2
    - name: job-3
```

Option 2: ReplicatedJob level API (each replicatedJob's local condition to continue is embedded in the replicatedJob spec itself):

```yaml
replicatedJobs:
  - name: initializer
    waitForStatusBeforeContinuing: Completed
  - name: driver
    waitForStatusBeforeContinuing: Ready
  - name: trainer-node-nvidia-a100
  - name: trainer-node-nvidia-h100
  ```

Personally I think the status that a given replicatedJob must reach before the subsequent replicatedJob begins execution is an inherently local attribute of that replicatedJob, so it is a natural fit for an API local to the replicatedJob (option 2). 

In addition, the global rules-based API is separating the startup order (as defined in the `replicatedJobs[]`order) from the startup rules (as defined in the list of `inOrderStartupPolicyRules[]`), when I think they really belong together. Separating them requires the user to repeatedly cross-reference different parts of the manifest to understand the full execution model of the JobSet. This becomes especially cumbersome for larger, more complicated JobSets.

In contrast, the local API captures both startup order and startup rules in `replicatedJobs[]`, so you can simply read the manifest from top to bottom in an intuitive fashion.

In my opinion, the current API described in the KEP feels a bit like we are trying force something with inherently local semantics into a global, rules-based API because it more closely follows the same rules-based pattern we have in other parts of the JobSet API (failure policy rules, etc) and other upstream k8s APIs (pod failure policy rules, etc). Having said that, the consistency with other APIs is an attractive property, so there is a trade-off.

@ahg-g @andreyvelich what are your thoughts on this?

Comment (ahg-g) [reply keps/672-serial-job-execution/README.md]: I think for the most part the number of replicated jobs will be limited, and so expressing the policy in one place should not be too bad.

> In contrast, the local API captures both startup order and startup rules in replicatedJobs[], so you can simply read the manifest from top to bottom in an intuitive fashion.

It is spread out though with full job/pod specs in between.

I see both approaches to have pros/cons, but I lean towards the existing format because it expresses the policy in one place. But does it express everything we want? @andreyvelich?

Comment (andreyvelich) [reply keps/672-serial-job-execution/README.md]: >Wouldn't those be represented as two jobs in a single replicatedJob?

Yes, that could work as long as we don't want to have 2 different PodSpec templates for those Parameter Servers. For example, to use different Node affinity. Since ReplicatedJob only has single PodTemplateSpec.

>@ahg-g @andreyvelich what are your thoughts on this?

Personally, I prefer the Option 2. Given that the FailurePolicy and SuccessPolicy has a different meaning and we don't need to define FailurePolicy for every ReplicatedJob.

@danielvegamyhre @ahg-g What do we think about API that I proposed here: https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1818283828?

Comment (ahg-g) [reply keps/672-serial-job-execution/README.md]: > @danielvegamyhre @ahg-g What do we think about API that I proposed here: [#680 (comment)](https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1818283828)?

What is the behavior for the setups?

Setup 1
```
replicatedJobs:
  - name: initializer
  - name: ps-1
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: ps-2
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: trainer-node
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Ready            <------------ changed this from Succeeded to Ready
      - name: ps-1
        status: Ready
      - name: ps-2
        status: Ready
```

Setup 2

```
replicatedJobs:
  - name: initializer
  - name: ps-1
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: ps-2
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: trainer-node
    waitForPreviousJobStatusBeforeContinuing:   
      - name: ps-2                     <------------ removed the first two items
        status: Ready
```

Setup 3

```
replicatedJobs:
  - name: initializer
  - name: ps-1
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: ps-2
    waitForPreviousJobStatusBeforeContinuing:
      - name: initializer
        status: Succeeded
  - name: trainer-node. <------------ removed all items
```

Comment (ahg-g) [reply keps/672-serial-job-execution/README.md]: I am concerned that this will not be easy to reason about compared to simple sequences. I am still not sure we need this added complexity. Can we have the API initially support strict sequences and extend it in the future to support this case? Basically not have the list of replicatedJobs and assume that the waitForPreviousJobStatusBeforeContinuing refers to the previous replicatedJob.

Comment (andreyvelich) [reply keps/672-serial-job-execution/README.md]: Sure, I am ok with that, so we just not introduce `name` parameter into API, right ?
Do you prefer Option 1 or Option 2 ?
Option 1
```
replicatedJobs:
  - name: initializer
  - name: ps-1
    waitForPreviousJobStatusBeforeContinuing:
      status: Completed
  - name: trainer-node
    waitForPreviousJobStatusBeforeContinuing:
      status: Ready
```

Option 2

```
replicatedJobs:
  - name: initializer
    waitForStatusBeforeContinuing: Completed
  - name: ps-1
    waitForStatusBeforeContinuing: Ready
  - name:  trainer-node
```

Comment (andreyvelich) [reply keps/672-serial-job-execution/README.md]: @ahg-g I think, it would be breaking change, since for Option 1, we should add the slice:
```
  - name: ps-1
    waitForPreviousJobStatusBeforeContinuing:
      - status: Completed
        name: initializer
```
For option 2, I don't think we can extend it.

**Comment by andreyvelich** [comment]
Hi @ahg-g @tenzen-y @vsoch @kannon92 @alculquicondor @danielvegamyhre @mimowo @shravan-achar @akshaychitneni.

As we discussed with @kannon92 at the last Batch WG call, I drafted this Google doc with all API ideas that we discussed in this PR, so we can compare them.

https://docs.google.com/document/d/1_xZ4UJ6FwJ8sG3i7XnBi1kG4mglNzxji94fweNH4YnE/edit?usp=sharing

Did I miss something ?

**Comment by vsoch** [comment]
@andreyvelich I didn't get to it today, but I should be able to take a look in the next few days!

File: keps/672-serial-job-execution/README.md [hunk 25]
```diff
@@ -0,0 +1,499 @@
...
+                      name: dataset-initializer
+              volumes:
+                - name: dataset-initializer
+                  persistentVolumeClaim:
+                    claimName: dataset-initializer
+                - name: model-initializer
+                  persistentVolumeClaim:
+                    claimName: model-initializer
+    - name: trainer-node
+      dependsOn:
```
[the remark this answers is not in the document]
andreyvelich: @ahg-g Do you mean add the `startupPolicy` under ReplicatedJob level ?
```yaml
replicatedJobs:
  - name: ps-a
    startupPolicy:
      dependsOn:
        - name: initializer
          status: Completed
```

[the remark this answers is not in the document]
ahg-g: yeah, under ReplicatedJob; fixed the formatting for the comment above.

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,500 @@
...
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+#### Unit Tests
+
+- `controllers`: `01/19/2024` - `30.2%`
+
+We will add this functionality to the startup_policy API for the functionality.
```
kannon92, line 415: If we are going to deprecate startup policy I would probably put the execution policy elsewhere.

kannon92 (commented):
LGTM! Thank you for your diligent follow through!

keps/672-serial-job-execution/README.md
```diff
@@ -0,0 +1,500 @@
...
+### Deprecation of StartupPolicy
+
+Since the new DependsOn API covers use-cases for the `startupPolicy: InOrder`, we will deprecate the
+StartupPolicy API in the future version of JobSet. Before deprecation, we will make
+StartupPolicy and DependsOn API mutually exclusive. Which means, user can use only
+StartupPolicy or DependsOn API.
+
+User should configure JobSet as follows to use the existing functionality of StartupPolicy:
+
+```yaml
```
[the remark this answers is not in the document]
andreyvelich: No, startupPolicy and dependsOn APIs are mutually exclusive, which means only one of them can be set.
I mentioned it in this section: https://github.com/kubernetes-sigs/jobset/pull/680/files/ea622a6d6ba8cfd876a7d662da2e141c977ecddc#diff-760e1d37e0293ac3b2c0170c72e6ac764d7abd930c1f746f1764db1abe6a12b7R403

keps/672-serial-job-execution/README.md
andreyvelich: What do you mean by that ?
Similar to what @ahg-g proposed above, but with the executionPolicy field name: https://github.com/kubernetes-sigs/jobset/pull/680#discussion_r1875924692 ?

kannon92: We will add this functionality to the startup_policy API for the functionality.

You are going to put the code in startup_policy.go but we are going to deprecate startup_policy. That seems odd to me.

keps/672-serial-job-execution/README.md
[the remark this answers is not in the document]
kannon92: Yea I understand. Reading this text it was not clear to me.

keps/672-serial-job-execution/README.md
andreyvelich: Oh, I think, we should put the code under `depends_on.go` or `execution_policy.go` file, but I am open to the various options.

keps/672-serial-job-execution/README.md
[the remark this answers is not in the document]
kannon92: Maybe it was pre coffee brain. Anyway I think this is fine.

keps/672-serial-job-execution/README.md
kannon92: Yea I think depends_on.go is fine but that is mostly an implementation detail.

Issue #665, opened 2024-08-26
ForegroundDeletion of Jobs is not always enforced before recreation


nstogner:
**What happened**:
 
I noticed this issue when I was testing what would happen if I were to trigger a failure in a Job when multiple `.replicatedJobs[]` are specified.

And observed that the JobSet controller did not wait for all Jobs to be deleted before creating a new one.

```bash
k get jobs -w -o custom-columns=NAME:.metadata.name,UID:.metadata.uid,CREATION_TIME:.metadata.creationTimestamp
NAME                            UID                                    CREATION_TIME
coordinator-example-workers-0   7ed23398-a07b-4a4d-9f72-a14c082567e9   2024-08-26T16:04:13Z
coordinator-example-driver-0    464b7733-9a37-4847-ac12-26be2468bf25   2024-08-26T16:04:13Z
coordinator-example-driver-0    464b7733-9a37-4847-ac12-26be2468bf25   2024-08-26T16:04:13Z
coordinator-example-workers-0   7ed23398-a07b-4a4d-9f72-a14c082567e9   2024-08-26T16:04:13Z
coordinator-example-driver-0    464b7733-9a37-4847-ac12-26be2468bf25   2024-08-26T16:04:13Z
coordinator-example-workers-0   7ed23398-a07b-4a4d-9f72-a14c082567e9   2024-08-26T16:04:13Z
coordinator-example-workers-0   7ed23398-a07b-4a4d-9f72-a14c082567e9   2024-08-26T16:04:13Z
coordinator-example-workers-0   7ed23398-a07b-4a4d-9f72-a14c082567e9   2024-08-26T16:04:13Z
coordinator-example-driver-0    464b7733-9a37-4847-ac12-26be2468bf25   2024-08-26T16:04:13Z
coordinator-example-driver-0    464b7733-9a37-4847-ac12-26be2468bf25   2024-08-26T16:04:13Z
coordinator-example-workers-0   7ed23398-a07b-4a4d-9f72-a14c082567e9   2024-08-26T16:04:13Z
# NEW workers-0 Job created HERE (notice new UID)
coordinator-example-workers-0   fa59cbba-37a4-4152-8d94-af6837f19e9b   2024-08-26T16:04:20Z
coordinator-example-workers-0   fa59cbba-37a4-4152-8d94-af6837f19e9b   2024-08-26T16:04:20Z
[...]

PR #686, opened 2024-10-18
Add restart strategy

### LINKED ISSUE #684: Blocking restart behavior

A blocking restart behavior where child jobs are created only after all jobs from the previous attempt are fully deleted.

**Why is this needed**:
Currently when restarting a JobSet, the child Jobs are recreated in parallel, and so one child job may get recreated before another is fully deleted. This can cause connection problems for training workloads where the new job tries to connect to a soon to be deleted pod, which will trigger another failure.

We can do this by adding a "JobSetRestartPolicy" to the FailurePolicy struct. The two values would be `Recreate|BlockingRecreate`, where the default is `Recreate`

- [ ] Design doc
- [x] API change
- [x] Docs update

nstogner:
#### What type of PR is this?

/kind feature

Fixes #684 

```release-note
Add configurable restart strategy.
```


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -306,13 +306,30 @@ type FailurePolicy struct {
 	// A restart is achieved by recreating all active child jobs.
 	MaxRestarts int32 `json:"maxRestarts,omitempty"`
 
+	// RestartStrategy defines the strategy to use when restarting the JobSet.
+	// Defaults to Recreate.
+	// +optional
+	// +kubebuilder:default=Recreate
```
ahg-g, line 312: Can we do validation here as follows:

// +kubebuilder:validation:Enum:=

Also, if the user was running a jobset and at the same time upgraded the controller, I suspect that this field will not exist for this job. I think this needs to be a reference that we explicitly default in the webhook and in the code, we also default it if not set to ensure we don't break running jobsets.

pkg/controllers/jobset_controller.go
```diff
@@ -637,6 +638,13 @@ func executeSuccessPolicy(js *jobset.JobSet, ownedJobs *childJobs, updateStatusO
 
 func constructJobsFromTemplate(js *jobset.JobSet, rjob *jobset.ReplicatedJob, ownedJobs *childJobs) []*batchv1.Job {
 	var jobs []*batchv1.Job
+	// If the JobSet is using the BlockingRecreate failure policy, we should not create any new jobs until
+	// all the jobs slated for deletion (i.e. from the last restart index) have been deleted.
+	if len(ownedJobs.previous) > 0 &&
+		js.Spec.FailurePolicy != nil && js.Spec.FailurePolicy.RestartStrategy == jobset.BlockingRecreate {
```
danielvegamyhre, line 644: nit: can we store `js.Spec.FailurePolicy != nil && js.Spec.FailurePolicy.RestartStrategy == jobset.BlockingRecreate ` in a variable to make this big if-statement more concise and readable?

test/integration/controller/jobset_controller_test.go
```diff
@@ -545,6 +545,44 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+						exp := map[int]int{
+							0: 1,
+							1: testutil.NumExpectedJobs(js) - 1,
+						}
+						gomega.Eventually(testutil.NumJobsByRestartAttempt, timeout, interval).WithArguments(ctx, k8sClient, js).Should(gomega.Equal(exp))
+					},
+				},
+			},
+		}),
+		// TODO(nstogner): Add test case for BlockingRecreate!!!
```
danielvegamyhre, line 585: We should add this test case for blockingRecreate before merging

test/integration/controller/jobset_controller_test.go
```diff
@@ -545,6 +545,44 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+				},
+				{
+					jobSetUpdateFn: func(js *jobset.JobSet) {
+						removeForegroundDeletionFinalizers(js, testutil.NumExpectedJobs(js)-1)
+					},
+				},
+				{
+					checkJobSetState: func(js *jobset.JobSet) {
+						ginkgo.By("checking 3/4 jobs were recreated")
+						exp := map[int]int{
```
danielvegamyhre, line 576: Can we rename `exp` to something a little more descriptive which makes it clear what the keys and values represent, so people reading the code don't need to reverse-engineer it by finding where/how it's used, etc.

danielvegamyhre (commented):
Looks like a useful change, the idea is to resolve the issue described [here](https://github.com/kubernetes-sigs/jobset/issues/665#issuecomment-2420961182) correct? 

Added a few comments.

test/integration/controller/jobset_controller_test.go
nstogner: Updated

test/integration/controller/jobset_controller_test.go
nstogner: Done

pkg/controllers/jobset_controller.go
nstogner: Updated

api/jobset/v1alpha2/jobset_types.go
nstogner: Added the enum validation. Here is the behavior as-is when the CRD is updated:

```bash
(base) ➜  jobset.blocking-recreate git:(restart-strategy) kind create cluster
using podman due to KIND_EXPERIMENTAL_PROVIDER
enabling experimental podman provider
Creating cluster "kind" ...
 ✓ Ensuring node image (kindest/node:v1.29.2) 🖼 
 ✓ Preparing nodes 📦  
 ✓ Writing configuration 📜 
 ✓ Starting control-plane 🕹️ 
 ✓ Installing CNI 🔌 
 ✓ Installing StorageClass 💾 
Set kubectl context to "kind-kind"
You can now use your cluster with:

kubectl cluster-info --context kind-kind

Have a nice day! 👋
(base) ➜  jobset.blocking-recreate git:(restart-strategy) git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
(base) ➜  jobset.blocking-recreate git:(main) kubectl apply --server-side -k ./config/components/crd
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
customresourcedefinition.apiextensions.k8s.io/jobsets.jobset.x-k8s.io serverside-applied
(base) ➜  jobset.blocking-recreate git:(main) kubectl apply -f ./examples/simple/max-restarts.yaml
jobset.jobset.x-k8s.io/max-restarts created
(base) ➜  jobset.blocking-recreate git:(main) kubectl get jobsets -oyaml | grep -A 3 failurePolicy
    failurePolicy:
      maxRestarts: 3
    replicatedJobs:
    - name: leader
(base) ➜  jobset.blocking-recreate git:(main) git checkout -
Switched to branch 'restart-strategy'
Your branch is up to date with 'nstogner/restart-strategy'.
(base) ➜  jobset.blocking-recreate git:(restart-strategy) kubectl apply --server-side -k ./config/components/crd
# Warning: 'patchesStrategicMerge' is deprecated. Please use 'patches' instead. Run 'kustomize edit fix' to update your Kustomization automatically.
customresourcedefinition.apiextensions.k8s.io/jobsets.jobset.x-k8s.io serverside-applied
(base) ➜  jobset.blocking-recreate git:(restart-strategy) kubectl get jobsets -oyaml | grep -A 3 failurePolicy
    failurePolicy:
      maxRestarts: 3
      restartStrategy: Recreate # <-------------- Default is applied
    replicatedJobs:
```

nstogner: Seems like this will not break existing JobSets... WDYT?

nstogner:
> Looks like a useful change, the idea is to resolve the issue described [here](https://github.com/kubernetes-sigs/jobset/issues/665#issuecomment-2420961182) correct?
> 
> Added a few comments.

Yep, thats exactly what this is targeting.

danielvegamyhre:
@nstogner you probably need to run `make generate` to fix the CI error

api/jobset/v1alpha2/jobset_types.go
ahg-g: sg, can you add the validation, it is not on the last commit.

ahg-g: oh, I see you added it on the enum definition itself, make sense

ahg-g:
/lgtm
/approve

Great, thanks!

Issue #682, opened 2024-10-05
Support for JobSet Preemption


ahg-g:
**What would you like to be added**:

Preemption at the whole JobSet level. 

The user would like to run a training workload using JobSet on one or more accelerator island (e.g., TPU slices). To do this, the user creates a JobSet with a replicatedJob of one or more replicas, and uses exclusive placement to ensure that each child Job lands on one accelerator island.

Consider the case where the user would like to run multiple training workloads as the ones described above with different priorities, and would like to ensure that the high priority training job preempts the low priority ones when there is not enough capacity to run both.

Currently this doesn't work because of the anti-affinity rules that implement exclusive placement: 
1) Low priority workload is running
2) High priority workload comes in, the leader pod is created first. 
3) The leader pod of the high priority workload can't schedule because it has anti-affinity against all the pods on the island, and cross-node preemption is not supported: https://kubernetes.io/docs/concepts/scheduling-eviction/pod-priority-preemption/#cross-node-preemption

Exclusivity is currently implemented against any pod created by Job that doesn't belong to the same JobSet, specifically it has the following anti-affinity constraint:

```
      podAntiAffinity: # ensures only this job lands on the rack
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: job-name
                operator: NotIn
                values:
                - my-job
              - key: job-name
                operator: Exists
            namespaceSelector: {}
            topologyKey: rack
```

The solution to the above problem is to limit exclusivity to the same priority level, and let pod preemption address race conditions if two jobs from different priority levels race to the same slice.

If exclusivity is limited to the same priority level, then in the above example, the leader pod of the higher priority workload will be able to preempt any of the pods of the lower priority workload, and once it does, the worker pods of the higher priority workload will be created and assigned to the same slice and will preempt the rest of the lower priority workers (the low priority worker may also not exist anymore if the low priority workload is restarting because of the initial preemption caused by the leader pod of the high priority workload).

To do this, we need to allow injecting another selector for the priority to the anti-affinity term that JobSet injects automatically.

```
      podAntiAffinity: # ensures only this job lands on the rack
          requiredDuringSchedulingIgnoredDuringExecution:
          - labelSelector:
              matchExpressions:
              - key: job-name
                operator: NotIn
                values:
                - my-job
              - key: job-name
                operator: Exists
            mismatchLabelKeys
            - priority
            namespaceSelector: {}
            topologyKey: rack
```

[...]

PR #687, opened 2024-10-20
Priority-based exclusive placement

ahg-g:
/kind feature

Limit exclusive placement to the same priority level.

Fixes #https://github.com/kubernetes-sigs/jobset/issues/682

```release-note
Limit exclusive placement to jobs within the same priority level.
```

### DIFF
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -28,6 +28,10 @@ const (
 	// the JobSet is currently on.
 	RestartsKey = "jobset.sigs.k8s.io/restart-attempt"
 
+	// PriorityKey is a label key to record the pod priority. This is needed to enfroce exclusive placement
+	// only among jobs within the same priority.
+	PriorityKey = "jobset.sigs.k8s.io/priority"
+
 	// MaxParallelism defines the maximum number of parallel Job creations/deltions that
 	// the JobSet controller can perform.
 	MaxParallelism = 50
--- a/pkg/constants/constants.go
+++ b/pkg/constants/constants.go
@@ -28,7 +28,7 @@ const (
 	// the JobSet is currently on.
 	RestartsKey = "jobset.sigs.k8s.io/restart-attempt"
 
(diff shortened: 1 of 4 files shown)


danielvegamyhre:
By limiting exclusive placement to jobs within the same priority level, couldn't this cause partial scheduling of jobs with different priority levels onto the same TPU slice?

ahg-g:
> By limiting exclusive placement to jobs within the same priority level, couldn't this cause partial scheduling of jobs with different priority levels onto the same TPU slice?

One the same slice, only one workload should be running, the assumption is that this workload will occupy all resources on that slice. If a race happens, then the higher priority slice will eventually preempt the lower priority one.

We should move to a proper API for exclusive placement in the next release, and have this configured as a knob, call it `ExclusivityScope` which defines the set of labels against which anti-affinity should be implemented.

ahg-g:
@danielvegamyhre the test is failing due to golang version, I updated it in https://github.com/kubernetes/test-infra/pull/33689, can you please take a look?

danielvegamyhre:
> @danielvegamyhre the test is failing due to golang version, I updated it in [kubernetes/test-infra#33689](https://github.com/kubernetes/test-infra/pull/33689), can you please take a look?

Done

pkg/constants/constants.go
```diff
@@ -28,6 +28,10 @@ const (
 	// the JobSet is currently on.
 	RestartsKey = "jobset.sigs.k8s.io/restart-attempt"
 
+	// PriorityKey is a label key to record the pod priority. This is needed to enfroce exclusive placement
```
danielvegamyhre, line 31:
```suggestion
	// PriorityKey is a label key to record the pod priority. This is needed to enforce exclusive placement
```

danielvegamyhre (commented):
LGTM overall, just one minor comment

ahg-g:
> LGTM overall, just one minor comment

done, thanks

PR #696, opened 2024-11-02
Introduce Go std slices and maps lib

tenzen-y:
#### What type of PR is this?

/kind cleanup

The Go std library has the same functions as the collections pkg below. 
So, I replace self-defined collections with Go std libs.

- Concat -> slices.Concat
- CloneMap -> maps.Clone
- Contains -> slices.Contains
- IndexOf -> slices.Index

Fixes #

```release-note
NONE
```


pkg/controllers/jobset_controller.go
```diff
@@ -731,7 +733,10 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
 	jobName := placement.GenJobName(js.Name, rjob.Name, jobIdx)
 
 	// Set labels on the object.
-	labels := collections.CloneMap(obj.GetLabels())
+	labels := maps.Clone(obj.GetLabels())
```
ahg-g, line 736: Why do we need to clone the map here? similar question to the annotations clone below

tenzen-y: That is a good question. 
TBH, I have the same question since we already set the constructed labels and annotations to object the below: https://github.com/kubernetes-sigs/jobset/blob/4b1eddd94402905260342791be255044d58d8a55/pkg/controllers/jobset_controller.go#L786-L787

I thought this redundant clone came from historical reasons.
Anyway, let me remove these clones.

tenzen-y: Done.
If you want to keep these clones, let me know.

pkg/controllers/jobset_controller.go
```diff
@@ -731,7 +733,7 @@ func labelAndAnnotateObject(obj metav1.Object, js *jobset.JobSet, rjob *jobset.R
 	jobName := placement.GenJobName(js.Name, rjob.Name, jobIdx)
 
 	// Set labels on the object.
-	labels := collections.CloneMap(obj.GetLabels())
+	labels := make(map[string]string)
```
ahg-g, line 736: this caused a regression, see https://github.com/kubernetes-sigs/jobset/issues/729

PR #697, opened 2024-11-02
Remove duplicated condition judgement

tenzen-y:
#### What type of PR is this?

/kind cleanup

Fixes #

```release-note
NONE
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -335,7 +335,7 @@ func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, j
 
 	// Calculate jobsReady for each Replicated Job
 	for _, job := range jobs.active {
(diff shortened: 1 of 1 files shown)


pkg/controllers/jobset_controller.go
```diff
@@ -335,7 +335,7 @@ func (r *JobSetReconciler) calculateReplicatedJobStatuses(ctx context.Context, j
 
 	// Calculate jobsReady for each Replicated Job
 	for _, job := range jobs.active {
-		if job.Labels == nil || (job.Labels != nil && job.Labels[jobset.ReplicatedJobNameKey] == "") {
+		if job.Labels == nil || job.Labels[jobset.ReplicatedJobNameKey] == "" {
```
tenzen-y, line 338: The second condition is evaluated only when the `job.Labels` is not nil.
Hence, the `job.Labels != nil` in the second condition are not needed.

danielvegamyhre:
/lgtm
/approve

Thanks!

PR #719, opened 2024-12-02
Allow for one to install jobset in a different namespace

### LINKED ISSUE #713: Internal cert manager relies on jobset-system namespace
**What happened**:
If you try to install jobset in a custom namespace, the cert rotation code fails to grab the secret.
Internal cert manager finds the right secret and everything works
Install jobset outside of jobset-system.

**Environment**:
- Install tools:
- Others:

kannon92:
#### What type of PR is this?
/kind bug

Allow for one to change the namespace of jobset.

Fixes #713 

```release-note
Allow for jobset to be installed in a different namespace from jobset-system
```

### DIFF
--- a/pkg/util/cert/cert.go
+++ b/pkg/util/cert/cert.go
@@ -19,12 +19,11 @@ import (
 	cert "github.com/open-policy-agent/cert-controller/pkg/rotator"
 	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
+
+	config "sigs.k8s.io/jobset/api/config/v1alpha1"
 )
 
 const (
-	serviceName             = "jobset-webhook-service"
-	secretName              = "jobset-webhook-server-cert"
-	secretNamespace         = "jobset-system"
 	certDir                 = "/tmp/k8s-webhook-server/serving-certs"
 	validateWebhookConfName = "jobset-validating-webhook-configuration"
 	mutatingWebhookConfName = "jobset-mutating-webhook-configuration"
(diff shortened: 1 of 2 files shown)


pkg/util/cert/cert.go
```diff
@@ -19,32 +19,30 @@ import (
...
 // CertsManager creates certs for webhooks.
-func CertsManager(mgr ctrl.Manager, setupFinish chan struct{}) error {
+func CertsManager(mgr ctrl.Manager, cfg config.Configuration, setupFinish chan struct{}) error {
+	// DNSName is <service name>.<namespace>.svc
+	var dnsName = fmt.Sprintf("%s.%s.svc", *cfg.InternalCertManagement.WebhookServiceName, *cfg.Namespace)
 	return cert.AddRotator(mgr, &cert.CertRotator{
 		SecretKey: types.NamespacedName{
-			Namespace: secretNamespace,
-			Name:      secretName,
+			Namespace: *cfg.Namespace,
```
andreyvelich, line 44: @tenzen-y @kannon92 I am wondering is there any specific reason that we want [to configure namespace in the Config ](https://github.com/kubernetes-sigs/jobset/blob/74e8cd088906e72d01841c5c9badbce978abdc19/api/config/v1alpha1/configuration_types.go#L31-L35)rather than taking it from the Pod's `/var/run/secrets/kubernetes.io/serviceaccount/namespace` ?

Our internal cert manager should always create certificates for the service in the same namespace where manager is deployed.

andreyvelich (commented):
Thanks @kannon92! 
It will help us to unblock this TODO: https://github.com/kubeflow/training-operator/blob/master/manifests/v2/overlays/standalone/kustomization.yaml#L10

pkg/util/cert/cert.go
kannon92: We are technically doing that.

https://github.com/kubernetes-sigs/jobset/blob/main/api/config/v1alpha1/defaults.go#L58

This is setting the namespace based on the serviceAccount. And then we use that value for certRotation.

andreyvelich: Yeah, but why do we allow users to change the namespace in the Config ?

kannon92: https://github.com/kubernetes-sigs/kueue/blob/main/apis/config/v1beta1/configuration_types.go#L34

andreyvelich: Yeah, Kueue has similar config. I think, we should remove it, if user should not set other namespace there.
For instance, in Katib we don't allow to set that value for CertGenerator config: https://github.com/kubeflow/katib/blob/master/pkg/apis/config/v1beta1/types.go#L86-L98.
Thoughts @tenzen-y @ahg-g @danielvegamyhre ?

kannon92: IMO this is not really related to this issue/PR. 

We have a real problem where folks can't run jobset if it is installed outside of jobset-system. I'm happy to discuss this but I think it belongs as a separate issue.

kannon92: Opened #720

PR #722, opened 2024-12-04
Remove kube-rbac-proxy

### LINKED ISSUE #721: Action Required: Replace Deprecated gcr.io/kubebuilder/kube-rbac-proxy
See https://github.com/kubernetes-sigs/kubebuilder/discussions/3907 for context.

related: https://github.com/kubernetes-sigs/kueue/issues/3737

kannon92:
#### What type of PR is this?
/kind cleanup

Fixes #721 

```release-note
NONE
```


kannon92:
I was able to confirm this works via:

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: metrics-consumer
  namespace: jobset-system
spec:
  # Use the scaffolded service account name to allow authn/authz
  serviceAccountName: jobset-controller-manager
  containers:
  - name: metrics-consumer
    image: curlimages/curl:7.78.0
    command: ["/bin/sh"]
    args:
      - "-c"
      - >
        while true;
        do
          # Note here that we are passing the token obtained from the ServiceAccount to curl the metrics endpoint
          curl -s -k -H "Authorization: Bearer $(cat /var/run/secrets/kubernetes.io/serviceaccount/token)" https://jobset-controller-manager-metrics-service.jobset-system.svc.cluster.local:8443/metrics;
          sleep 60;
        done
```

When I run this pod, I am able to see that the metrics server is correctly secured.

ahg-g:
@camilamacedo86 thanks a lot for reviewing https://github.com/kubernetes-sigs/lws/pull/284; do you mind taking a look at this PR? It is already merged, but we can open a followup if there is something here that needs tweaking.

main.go
```diff
@@ -127,7 +129,16 @@ func main() {
...
 	}
-
+	// Metrics endpoint is enabled in 'config/default/kustomization.yaml'. The Metrics options configure the server.
+	// More info:
+	// - https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.19.1/pkg/metrics/server
+	// - https://book.kubebuilder.io/reference/metrics.html
+	metricsServerOptions := metricsserver.Options{
+		BindAddress:    metricsAddr,
+		SecureServing:  true,
+		FilterProvider: filters.WithAuthenticationAndAuthorization,
```
ahg-g, line 139: @kannon92 it seems that we should disable http/2: https://github.com/kubernetes-sigs/lws/pull/284#discussion_r1888126368

config/default/kustomization.yaml
```diff
@@ -23,14 +23,10 @@ resources:
...
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
 #- ../prometheus
+# [METRICS] Expose the controller manager metrics service.
+- manager_metrics_service.yaml
 
 patchesStrategicMerge:
-# Protect the /metrics endpoint by putting it behind auth.
-# If you want your controller-manager to expose the /metrics
-# endpoint w/o any authn/z, please comment the following line.
-- manager_auth_proxy_patch.yaml
```
kannon92, line 31: @googs1025 this is probably causing #836 

We don't patch our deployment with the metrics ports.

PR #735, opened 2024-12-19
update golang net to 0.33 to satisfy security alerts

kannon92:
#### What type of PR is this?
/kind bug

https://pkg.go.dev/vuln/GO-2024-3333

Update golang/net

Fixes #

```release-note
NONE
```


ahg-g:
curious why was this not triggered by dependent bot?

kannon92:
> curious why was this not triggered by dependent bot?

I don't think dependabot updates transitive dependencies. I think the real fix would be a patch in client-go.

PR #738, opened 2024-12-20
update k8s to 0.32 apis

kannon92:
#### What type of PR is this?
/kind dependency

Bump k8s api to 0.32.

Fixes #

Kueue usually waits until the first patch release of 0.32. We can do the same if we want.

```release-note
NONE
```


kannon92:
Kueue PR: https://github.com/kubernetes-sigs/kueue/pull/3820

ahg-g:
why not wait for the depbot?

kannon92:
> why not wait for the depbot?

We turn off dependabot for K8s release changes. It’s enabled for patches for a version.

Besides I think K8s bumps require one to generate the manifests in case API changes so it’s better for a manually process.

client-go/informers/externalversions/jobset/v1alpha2/jobset.go
```diff
@@ -15,24 +15,24 @@ limitations under the License.
 package v1alpha2
 
 import (
-	"context"
+	context "context"
```
ahg-g, line 18: can you please revert all those files?

kannon92: If we disagree with this I think we gotta fix this upstream. Otherwise everything someone needs to revert this whenever they do generate.

Issue #723, opened 2024-12-06
Rename JobSet status to Complete


andreyvelich:
**What would you like to be added**:

@tenzen-y and I noticed that JobSet uses the [`Completed` status rather than `Complete`](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L70).

However, Kubernetes Job uses [the `Complete` status](https://github.com/kubernetes/api/blob/fed59eeeda185ddcf67ea79840f6df9f3285303a/batch/v1/types.go#L622-L623). 
We also use [the `Complete` status in the TrainJob](https://github.com/kubeflow/training-operator/blob/master/pkg/apis/kubeflow.org/v2alpha1/trainjob_types.go#L55-L56C2) to be consistent with the Batch/Job.

Additionally, I would propose that we rename [`Succeeded` ReplicatedJobStatus to `Complete`,](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L182-L183) since there is no such status as `Succeeded` on the Job level. E.g.
```yaml
replicatedJobStatus:
 - name: Job-A
  complete: 1
  failed: 0
    ...
```

Any thoughts @tenzen-y @kannon92 @danielvegamyhre @ahg-g ?

**Why is this needed**:

Being consistent across various Kubernetes resources.

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

googs1025:
If it makes sense to me, I am willing to make this change. 🤔

kannon92:
I tend to prefer "Succeeded" as it matches terminal states like 'Suspended' and 'Failed'. I do think it unfortunate that Job uses complete.

I don't really have a strong feeling about the renaming but I think this should be considered a breaking change so we should probably introduce "Complete" and leave "Succeeded" as deprecated. And then maybe in a release or two we can drop "Succeeded".

But I don't really feel that this renaming is worth the effort.

andreyvelich:
> I tend to prefer "Succeeded"

We discussed before with @tenzen-y, that it is tricky to define what does Succeeded mean for Job or JobSet given all of the Success and Failure policies that we have: https://github.com/kubeflow/training-operator/pull/2298.

> But I don't really feel that this renaming is worth the effort.

[...]

PR #740, opened 2024-12-27
KEP-672: Implement the DependsOn API

andreyvelich:
#### What type of PR is this?

/kind feature
/kind api-change

I added implementation for the DependsOn API. The majority of validations are implemented using CEL.

TODO list:
- [x] Add unit tests for the `depends_on.go`
- [x] Add controller integration tests to verify CEL validation.
- [x] Add controller integration tests to verify Job creation.

/cc @ahg-g @kannon92 @danielvegamyhre @tenzen-y @vsoch 

Fixes: https://github.com/kubernetes-sigs/jobset/issues/672

```release-note
Add the DependsOn API to support serial execution of ReplicatedJobs.
```


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -230,8 +233,44 @@ type ReplicatedJob struct {
...
+type DependsOnStatus string
+
+const (
+	// Ready status means the Ready counter equals the number of child Jobs.
+	// .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].ready
+	ReadyStatus DependsOnStatus = "Ready"
+
+	// Complete status means the Succeeded counter equals the number of child Jobs.
+	// .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].succeeded
+	CompleteStatus DependsOnStatus = "Complete"
```
andreyvelich, line 271: What do we think about `Complete` status here, given the discussion in: https://github.com/kubernetes-sigs/jobset/issues/723 ?

andreyvelich:
I've added integration and unit tests. Let me know if that looks good to you!
/assign @ahg-g @kannon92 @danielvegamyhre @tenzen-y @vsoch @shravan-achar @akshaychitneni

api/jobset/v1alpha2/jobset_types.go
vsoch: We probably want to match the Kubernetes Job and use Complete, so keep as is.

vsoch (approved):

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -417,5 +417,126 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 			},
 			updateShouldFail: true,
 		}),
+		ginkgo.Entry("DependsOn and StartupPolicy can't be set together", &testCase{
```
kannon92, line 420: Please add a test case for StartupPolicy set to AnyOrder and allowing DependsOn

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -230,8 +233,44 @@ type ReplicatedJob struct {
 	// Jobs names will be in the format: <jobSet.name>-<spec.replicatedJob.name>-<job-index>
 	// +kubebuilder:default=1
 	Replicas int32 `json:"replicas,omitempty"`
+
+	// DependsOn is an optional list that specifies the preceding ReplicatedJobs upon which
+	// the current ReplicatedJob depends. If specified, the ReplicatedJob will be created
+	// only after the referenced ReplicatedJobs reach their desired state.
+	// The Order of ReplicatedJobs is defined by their enumeration in the slice.
+	// Note, that the first ReplicatedJob in the slice cannot use the DependsOn API.
+	// TODO (andreyvelich): Currently, only a single item is supported in the DependsOn list.
```
kannon92, line 242: I don't think we should have TODO in user facing APIs.

kannon92, line 242:
```suggestion
	// Currently, only a single item is supported in the DependsOn list.
```

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -230,8 +233,44 @@ type ReplicatedJob struct {
...
 }
 
+type DependsOnStatus string
+
+const (
+	// Ready status means the Ready counter equals the number of child Jobs.
+	// .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].ready
+	ReadyStatus DependsOnStatus = "Ready"
+
+	// Complete status means the Succeeded counter equals the number of child Jobs.
```
kannon92, line 269: I'm not sure if this is actually true. A ReplicatedJob can be mark as succeeded if the success policy passes which doesn't necessarly mean that all the replicated jobs finished.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -230,8 +233,44 @@ type ReplicatedJob struct {
...
+	// Status defines the condition for the ReplicatedJob. Only Ready or Complete status can be set.
+	// +kubebuilder:validation:Enum=Ready;Complete
+	Status DependsOnStatus `json:"status"`
 }
 
+type DependsOnStatus string
+
+const (
+	// Ready status means the Ready counter equals the number of child Jobs.
+	// .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].ready
```
kannon92, line 266: I don't think the boolean expression is necessary.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -230,8 +233,44 @@ type ReplicatedJob struct {
...
+	ReadyStatus DependsOnStatus = "Ready"
+
+	// Complete status means the Succeeded counter equals the number of child Jobs.
+	// .spec.replicatedJobs["name==<JOB_NAME>"].replicas == .status.replicatedJobsStatus.name["name==<JOB_NAME>"].succeeded
```
kannon92, line 270: I don't think the boolean expression is necessary.

pkg/webhooks/jobset_webhook.go
```diff
@@ -191,39 +191,50 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, obj runtime.Object)
...
 				if strings.Contains(errMessage, dns1035MaxLengthExceededErrorMsg) {
 					errMessage = podNameTooLongErrorMsg
 				}
 				allErrs = append(allErrs, errors.New(errMessage))
 			}
 		}
+		replicatedJobNames[rJob.Name] = rIdx
+		// Check that DependsOn references the previous ReplicatedJob.
+		if rIdx > 0 && rJob.DependsOn != nil {
+			dependsOnIdx, ok := replicatedJobNames[rJob.DependsOn[0].Name]
```
kannon92, line 233:
```suggestion
			dependsOnIdx, ok := replicatedJobNames[rJob.DependsOn[rIdx - 1].Name]
```

I think you want to look at the previous job and not just the 0th element.

api/jobset/v1alpha2/jobset_types.go
kannon92: Yea I think that is fine.

pkg/controllers/jobset_controller.go
```diff
@@ -412,14 +412,27 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
...
+
 	// If JobSpec is unsuspended, ensure all active child Jobs are also
 	// unsuspended and update the suspend condition to true.
 	for _, replicatedJob := range js.Spec.ReplicatedJobs {
 		replicatedJobStatus := findReplicatedJobStatus(replicatedJobStatuses, replicatedJob.Name)
+
+		rJobsReplicas[replicatedJob.Name] = replicatedJob.Replicas
+
+		// For depends on, the Job is created only after the previous replicatedJob reached the status.
+		if replicatedJob.DependsOn != nil && !isDependsOnJobReachedStatus(replicatedJob.DependsOn[0], rJobsReplicas[replicatedJob.DependsOn[0].Name], replicatedJobStatuses) {
```
kannon92, line 427: I think this is a bug. You probably want the idx - 1 rather than 0.

kannon92 (commented):
Given how important this is for kubeflow, I am wondering if a e2e test would be useful.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1487,10 +1487,152 @@ func TestValidateCreate(t *testing.T) {
 		},
 	}
 
+	dependsOnTests := []validationTestCase{
+		{
+			name: "dependsOn is valid",
```
kannon92, line 1492: Can we add a case where job-2 waits for job-1 and job-3 waits for job-2?

api/jobset/v1alpha2/jobset_types.go
andreyvelich: Sure, I will remove it.

api/jobset/v1alpha2/jobset_types.go
andreyvelich: Why ?

Should we modify this as follows:
```
.spec.replicatedJobs["name==<JOB_NAME>"].replicas == 
.status.replicatedJobsStatus.name["name==<JOB_NAME>"].ready + 
.status.replicatedJobsStatus.name["name==<JOB_NAME>"].succeeded + 
.status.replicatedJobsStatus.name["name==<JOB_NAME>"].failed
```

api/jobset/v1alpha2/jobset_types.go
andreyvelich: Oh, you mean for user-facing docs. Sure, I think we can remove it.

pkg/webhooks/jobset_webhook.go
andreyvelich: Actually, DependsOn is an array which always has a single element, as you can see here:
https://github.com/kubernetes-sigs/jobset/blob/d751c13ed9835648fe39a2753040b6ba9fae8415/api/jobset/v1alpha2/jobset_types.go#L245
Does it make sense @kannon92 ?

api/jobset/v1alpha2/jobset_types.go
kannon92: Ah I see. You are correct.

api/jobset/v1alpha2/jobset_types.go
kannon92: https://github.com/kubernetes-sigs/jobset/pull/740#discussion_r1904278954

Yes.

pkg/webhooks/jobset_webhook.go
kannon92: Ah yes. I got my streams crossed.

andreyvelich:
>Given how important this is for kubeflow, I am wondering if a e2e test would be useful.

Yeah, maybe we could add one e2e test as part of JobSet: https://github.com/kubernetes-sigs/jobset/blob/main/test/e2e/e2e_test.go.
Will add it soon.

kannon92:
Can we make sure to add some examples and maybe call this out on our site?

kannon92:
> Sure, I can add simple example with sleep.
Actually, once we have implementation of MPI Runtime in Kubeflow Training V2, I am planning to add example of MPI job that uses DependsOn API. We can link those examples in the JobSet docs as well.

We can have sleep for now. I'd prefer to have standalone examples without requiring Kubeflow in our user facing docs.

pkg/controllers/jobset_controller.go
danielvegamyhre: @andreyvelich there are some comments in this PR confused about the hardcoded index 0, it surprised me as well at first until I remembered the design discussion about single vs multi-dependency and future proofing. 

I think hard coding the index of a specific dependency makes the code a bit surprising/confusing as compared to simply looping through the slice checking all dependencies, and using the CEL validation you've added to enforce the single dependency only.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -230,8 +233,42 @@ type ReplicatedJob struct {
...
+	// Status defines the condition for the ReplicatedJob. Only Ready or Complete status can be set.
+	// +kubebuilder:validation:Enum=Ready;Complete
+	Status DependsOnStatus `json:"status"`
 }
 
+type DependsOnStatus string
+
+const (
+	// Ready status means the Ready + Succeeded + Failed counter equals the number of child Jobs.
+	ReadyStatus DependsOnStatus = "Ready"
```
danielvegamyhre, line 266: I think we should rename these status variables something like `ReplicatedJobReady` and `ReplicatedJobComplete`. 

With the current naming, in the other places we're importing the api as `jobset` and using these variables, they sound like they are referring to the status of the JobSet itself (e.g. `jobset.ReadyStatus`, `jobset.CompletedStatus`, etc), which is confusing/misleading.

pkg/controllers/depends_on.go
```diff
@@ -0,0 +1,26 @@
+package controllers
+
+import (
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+)
+
+// isDependsOnJobReachedStatus checks if depends on ReplicatedJob reaches Ready or Complete status.
+func isDependsOnJobReachedStatus(dependsOnJob jobset.DependsOn, dependsOnJobReplicas int32, rJobsStatuses []jobset.ReplicatedJobStatus) bool {
```
danielvegamyhre, line 8: The function name is a bit confusing to me, maybe we can name it something like `dependencyReachedStatus` or `replicatedJobReachedStatus`?

pkg/controllers/jobset_controller.go
andreyvelich: I see, @danielvegamyhre do you suggest that in **this PR** we implement the `dependencyReachedStatus` function in a way that it can accept array and verify that all dependent ReplicatedJobs reach status ?

api/jobset/v1alpha2/jobset_types.go
andreyvelich: Yeah, that is good point!

andreyvelich:
@danielvegamyhre @kannon92 What do you think about example that I added ?

site/static/examples/depends-on/depends-on.yaml
```diff
@@ -0,0 +1,61 @@
...
+          template:
+            spec:
+              containers:
+                - name: initializer
+                  image: busybox
+                  command:
+                    - /bin/sh
+                    - -c
+                    - "echo 'initializer runs for 10 seconds' && sleep 10"
+    - name: launcher
```
kannon92, line 23: For the launcher maybe we should have some kind of readiness probe that can explain how this would be used for ready?

Right now it would go to active/running very quick due to no readiness probe.

andreyvelich: Sure, I am changed to startupProbe for launcher.

>Right now it would go to active/running very quick due to no readiness probe.

Actually, the Job will be Ready only after all InitContainers are complete: https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#:~:text=Ready%3A%20the%20Pod%20is%20able%20to%20serve%20requests%20and%20should%20be%20added%20to%20the%20load%20balancing%20pools%20of%20all%20matching%20Services.

kannon92: > Ready: the Pod is able to serve requests and should be added to the load balancing pools of all matching Services.

All init containers finished would have the condition Initialized.

site/content/en/docs/tasks/_index.md
```diff
@@ -20,17 +19,31 @@ no_list: true
...
 
 - [Parallel Jobs](https://github.com/kubernetes-sigs/jobset/blob/release-0.5/examples/simple/paralleljobs.yaml) demonstrates how we can submit multiple replicated jobs in a jobset.
 
-- [Startup Policy](https://github.com/kubernetes-sigs/jobset/blob/release-0.5/examples/startup-policy/startup-driver-ready.yaml) demonstrates how we can define a startup order for ReplicatedJobs in order to ensure a "leader"
-pod is running before the "workers" are created. This is important for enabling the leader-worker paradigm in distributed ML training, where the workers will attempt to register with the leader as soon as they spawn.
+- [Depends On](https://github.com/kubernetes-sigs/jobset/tree/main/examples/depends-on/depends-on.yaml)
+  demonstrates how we can define dependencies between ReplicatedJobs, ensuring they are executed in
```
kannon92, line 32: This section doesn't really read like the other ones. I would keep it brief here as we did in the other examples.

andreyvelich: Sure, do you want to remove the DependsOn validation rules from the docs @kannon92 ?

kannon92: I'm not sure if we have a place to put more user guide items.

In this area I don't think it fits. We mostly point to examples

andreyvelich: Sounds good. I the future, we can create dedicated user guide for this API if that is required.
Also, we have an API reference for it: https://jobset.sigs.k8s.io/docs/reference/jobset.v1alpha2/

andreyvelich:
Are there any remaining suggestions for this PR besides the @danielvegamyhre's comment: https://github.com/kubernetes-sigs/jobset/pull/740#discussion_r1908169546 ?
If not, I believe we can merge it.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -79,6 +79,8 @@ const (
 )
 
 // JobSetSpec defines the desired state of JobSet
+// +kubebuilder:validation:XValidation:rule="!(has(self.startupPolicy) && self.startupPolicy.startupPolicyOrder == 'InOrder' && self.replicatedJobs.exists(x, has(x.dependsOn)))",message="StartupPolicy and DependsOn APIs are mutually exclusive"
```
kannon92, line 82: Can we get some kind of test that confirms this works?

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -79,6 +79,8 @@ const (
...
+// +kubebuilder:validation:XValidation:rule="!(has(self.replicatedJobs[0].dependsOn))",message="DependsOn can't be set for the first ReplicatedJob"
```
kannon92, line 83: Same here.

pkg/controllers/depends_on.go
```diff
@@ -0,0 +1,26 @@
...
+	actualStatus := findReplicatedJobStatus(rJobsStatuses, dependsOnJob.Name)
+	if actualStatus == nil {
+		return false
+	}
+
+	// For Complete status, number of replicas must be equal to number of succeeded Jobs.
+	if dependsOnJob.Status == jobset.DependencyComplete && dependsOnJobReplicas == actualStatus.Succeeded {
+		return true
+	}
+
```
kannon92, line 19:
```suggestion
	if dependsOnJob.Status == jobset.DependencyComplete && dependsOnJobReplicas == actualStatus.Failed {
		return false
}
```

If we hit complete but the replicated job fail, what should we do?

test/e2e/e2e_test.go
```diff
@@ -270,6 +270,74 @@ var _ = ginkgo.Describe("JobSet", func() {
 		})
 	})
 
+	// This test runs JobSet with the DependsOn API.
```
kannon92, line 273: I would maybe add a use case that kubeflow uses?

We don't need to use the same images but it would be nice to make sure that we cover the trainingoperator v2 case in our e2e tests.

We followed a similar pattern for Kueue so we would know that the kueue integration broke in our e2e tests rather than wait for kueue to consume jobset.

ahg-g:
/hold

Putting a hold just so this doesn't get merged by mistake. I plan to review it this week.

api/jobset/v1alpha2/jobset_types.go
andreyvelich: We test it here: https://github.com/kubernetes-sigs/jobset/blob/c0bf2320b45a0ece4a480e9438760763e813f9b2/test/integration/webhook/jobset_webhook_test.go#L420

api/jobset/v1alpha2/jobset_types.go
andreyvelich: Here is the test case: 
https://github.com/kubernetes-sigs/jobset/blob/c0bf2320b45a0ece4a480e9438760763e813f9b2/test/integration/webhook/jobset_webhook_test.go#L470

pkg/controllers/depends_on.go
andreyvelich: We won't hit complete here if ReplicatedJob fails.
Complete here means that [`.replicatedJobStatus.succeeded`](https://github.com/kubernetes-sigs/jobset/blob/c0bf2320b45a0ece4a480e9438760763e813f9b2/api/jobset/v1alpha2/jobset_types.go#L186) ==  [`.replicatedJob.replicas`](https://github.com/kubernetes-sigs/jobset/blob/c0bf2320b45a0ece4a480e9438760763e813f9b2/api/jobset/v1alpha2/jobset_types.go#L235)
Does it make sense @kannon92 ?

test/e2e/e2e_test.go
andreyvelich: Good point.
Maybe we could add two test cases for now:
Initializer (Complete) -> Trainer
Launcher (Ready) -> Trainer

That would cover the LLM Runtime with PyTorch and MPI use-cases.

andreyvelich: Added the two E2Es for 2 usecases.

pkg/controllers/jobset_controller.go
```diff
@@ -418,14 +418,27 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
...
+
 	// If JobSpec is unsuspended, ensure all active child Jobs are also
 	// unsuspended and update the suspend condition to true.
 	for _, replicatedJob := range js.Spec.ReplicatedJobs {
 		replicatedJobStatus := findReplicatedJobStatus(replicatedJobStatuses, replicatedJob.Name)
+
+		rJobsReplicas[replicatedJob.Name] = replicatedJob.Replicas
+
+		// For depends on, the Job is created only after the dependent ReplicatedJob reaches the status.
+		if replicatedJob.DependsOn != nil && !dependencyReachedStatus(replicatedJob.DependsOn[0], rJobsReplicas[replicatedJob.DependsOn[0].Name], replicatedJobStatuses) {
```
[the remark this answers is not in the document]
andreyvelich: Sure, that sound good.

pkg/webhooks/jobset_webhook.go
andreyvelich: Actually, good point. Since we populate the map sequentially, we can just check if element exists in the map.

test/e2e/e2e_test.go
```diff
@@ -270,6 +270,138 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+						if rJobStatus.Name == initializerJob {
+							return rJobStatus.Ready
+						}
+					}
+					return 0
+				}, timeout, interval).Should(gomega.Equal(int32(numReplicas)))
+			})
+
+			ginkgo.By("Verify that only Initializer is created", func() {
+				gomega.Eventually(util.NumJobs, timeout, interval).WithArguments(ctx, k8sClient, jobSet).
```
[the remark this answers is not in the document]
andreyvelich: The goal of this check is to verify that Trainer Node Job is not started at this time.
Don't you think it is required here?

test/e2e/e2e_test.go
```diff
@@ -270,6 +270,138 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+					for _, rJobStatus := range jobSet.Status.ReplicatedJobsStatus {
+						if rJobStatus.Name == initializerJob {
+							return rJobStatus.Ready
+						}
+					}
+					return 0
+				}, timeout, interval).Should(gomega.Equal(int32(numReplicas)))
+			})
+
+			ginkgo.By("Verify that only Initializer is created", func() {
```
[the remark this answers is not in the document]
andreyvelich: Alright, let me add this comment.

test/e2e/e2e_test.go
```diff
@@ -270,6 +270,138 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+					for _, rJobStatus := range jobSet.Status.ReplicatedJobsStatus {
+						if rJobStatus.Name == initializerJob {
+							return rJobStatus.Succeeded
+						}
+					}
+					return 0
+				}, timeout, interval).Should(gomega.Equal(int32(numReplicas)))
+			})
+
+			ginkgo.By("Verify that Initializer and Trainer Node is created", func() {
```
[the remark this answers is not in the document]
andreyvelich: Yeah, I understand. The Trainer Node here references to name of ReplicatedJob.
E.g. in the ML training use-cases naming ReplicatedJob as `trainer-node` make sense since every Kubernetes Pod represents PyTorch, Jax, or TF Node.
So ML Engineers and Data Scientists see components of JobSet as:
```
trainer-node-0-1-xxx 
trainer-node-0-2-xxx
```
I think, we discussed it before as part of Kubeflow Trainer V2 work: https://github.com/kubeflow/training-operator/tree/master/docs/proposals/2170-kubeflow-training-v2.
 
Does it make sense @ahg-g ?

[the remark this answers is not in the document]
ahg-g: yes, I guess my point is to update the text to say the "training job" rather than the "trainer node", but it is a nit.

test/e2e/e2e_test.go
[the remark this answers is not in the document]
ahg-g: got it, ok makes sense.

test/integration/controller/jobset_controller_test.go
```diff
@@ -1672,6 +1672,315 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+					},
+				},
+				{
+					// Update the Replicated-Job-A to the ready status.
+					jobUpdateFn: func(jobList *batchv1.JobList) {
+						readyReplicatedJob(jobList, "rjob-a")
+					},
+					// Only the Replicated-Job-A must be created.
+					checkJobCreation: func(js *jobset.JobSet) {
+						expectedStarts := 1
```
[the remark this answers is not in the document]
andreyvelich: Sure, updated integration tests.
Also, I noticed that if we run `jobUpdateFn` in the integration test step, we should have `checkJobCreation` in the same step. Otherwise, we wait until all ReplicatedJob is created:
```
gomega.Eventually(testutil.NumJobs, timeout, interval).WithArguments(ctx, k8sClient, js).Should(gomega.Equal(testutil.NumExpectedJobs(js)))
```

E.g. we execute `checkJobCreation` before we run `jobUpdateFn`.

This is due to:
https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go#L183-L187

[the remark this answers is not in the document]
ahg-g: Sounds good, we could have a noop check function if necessary

ahg-g:
/lgtm
/approve

Thanks @andreyvelich, this is a great feature, the implementation is clean and the testing is thorough! Feel free to remove the hold if others don't have anymore feedback.

andreyvelich:
Thanks everyone for the review 🎉 
/hold cancel

PR #751, opened 2025-01-14
[Release 0.7]: cherry-pick allow for one to install jobset in a different namespace

kannon92:
#### What type of PR is this?
/kind bug

Kubeflow is hoping to install jobset into kubeflow-jobset namespace.

As an openshift user, we also would like to install jobset into a separate namespace.

Fixes #

@ahg-g I am a little confused by our branch names. I cherry-picked this PR into release-7.1 branch which I hope is sufficient.

```release-note
Fix internal certmanager bug where jobset-system namespace is used for cert rotations
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -135,7 +135,7 @@ func main() {
 	}
 
 	certsReady := make(chan struct{})
-	if err = cert.CertsManager(mgr, certsReady); err != nil {
+	if err = cert.CertsManager(mgr, cfg, certsReady); err != nil {
 		setupLog.Error(err, "unable to setup cert rotation")
 		os.Exit(1)
 	}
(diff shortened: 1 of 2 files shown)


kannon92:
cc @andreyvelich

andreyvelich (commented):
Thanks @kannon92, lgtm.
I am wondering why we decided to make separate branches for every JobSet patch release ? E.g. why we don't just have single `release-0.7` branch, like in Kubernetes: https://github.com/kubernetes/kubernetes/tree/release-1.32

kannon92:
Usually we have a release branch (ie release-0.5). In this case, I think there were a few issues that caused some problems. If I remember right there was a bad branch that tagged the wrong commit causing a regression so a new branch was created from release-7.1. 

I think for release-0.8, it will be more sane.

ahg-g:
yeah, that was a one off.

/lgtm
/approve

Issue #720, opened 2024-12-03
Drop namespace from config option


kannon92:
**What would you like to be added**:

We should drop namespace from the config option and take the namespace that the serviceaccount for JobSet lives under.

**Why is this needed**:
https://github.com/kubernetes-sigs/jobset/pull/719#discussion_r1866469895

It is unclear the behavior that would happen if someone installs jobset into namespace x and the configuration is using namespace y. In the thread we discussed potentially dropping the namespace field from the config and using the namespace in serviceAccount.

kannon92:
cc @andreyvelich @ahg-g @danielvegamyhre

kannon92:
/close

#752 achieved this.

PR #752, opened 2025-01-14
Remove Namespace from the JobSet Config

andreyvelich:
#### What type of PR is this?

I removed namespace from the JobSet config as we discussed here: https://github.com/kubernetes-sigs/jobset/pull/719#discussion_r1866469895

/kind cleanup
/assign @kannon92 @ahg-g @danielvegamyhre @tenzen-y 

The webhook and JobSet controller always have the same namespace, so user should not override this value in the Config.
The namespace is taken from the Pod's service account: `/var/run/secrets/kubernetes.io/serviceaccount/namespace`

Fixes # https://github.com/kubernetes-sigs/jobset/issues/720

```release-note
The namespace parameter has been removed from the JobSet manager config.
```

### DIFF
--- a/api/config/v1alpha1/configuration_types.go
+++ b/api/config/v1alpha1/configuration_types.go
@@ -28,12 +28,6 @@ import (
 type Configuration struct {
 	metav1.TypeMeta `json:",inline"`
 
-	// Namespace is the namespace in which jobset controller is deployed.
-	// It is used as part of DNSName of the webhook Service.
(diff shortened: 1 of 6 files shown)


kannon92:
/assign @ahg-g 

We probably should not drop fields from an API without a deprecation process.

I would add a release-note saying that we are dropping this at least.

kannon92:
/hold

I want to wait for comments from @danielvegamyhre and @ahg-g.

tenzen-y (commented):
/lgtm

PR #756, opened 2025-01-16
Use config metrics binding address if flag is not set

ardaguclu:
#### What type of PR is this?
/kind bug

Metrics in options is directly [initialized by using the value](https://github.com/kubernetes-sigs/jobset/blob/ab3881469b6c5fbd5e4a8e979bf041013eb1add0/main.go#L160-L166) in `metricAddr` regardless of the value in config file (which should not happen). This PR overwrites the `metricAddr` by the value in config file unless flag is explicitly set.

```release-note
The BindAddress value in config file will be used unless flag is explicitly set
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -115,10 +115,8 @@ func main() {
 	kubeConfig.QPS = *cfg.ClientConnection.QPS
 	kubeConfig.Burst = int(*cfg.ClientConnection.Burst)
 
-	if flagsSet["metrics-bind-address"] {
-		options.Metrics.BindAddress = metricsAddr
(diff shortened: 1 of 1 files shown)


ahg-g:
@kannon92 what about this one? I think we can cherrypick this one too since it is a bug.

kannon92:
I don't think this one makes sense to cherry-pick. This became a bug when we dropped kube-rbac-proxy.

But I actually don't see us using this in 0.7 release branch. So not sure if this is worth carrying.

PR #760, opened 2025-01-21
[Helm] add helm chart for jobset

### LINKED ISSUE #726: Helm chart for JobSet

Add a helm chart for the JobSet installation.

**Why is this needed**:

Customers have expressed a preference for helm vs kubectl applying a github URL.  Example chart from Kueue: https://github.com/kubernetes-sigs/kueue/blob/1b7d3f3f71d3c49b5b02ec0766abe4c3df7cfdfd/charts/kueue/README.md

- [ ] Design doc
- [ ] API change
- [x] Docs update

slin1237:
#### What type of PR is this?
/kind regression
Adds helm chart for jobset

Some organizations are fully dependent on helm installation, thus jobset should provide an official helm chart for installation

Fixes #726 

```release-note
None
```

### DIFF
--- a/charts/jobset/README.md
+++ b/charts/jobset/README.md
@@ -1,0 +1,74 @@
+# JobSet Helm Chart
+![Version: 0.1.0](https://img.shields.io/badge/Version-0.1.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.1.0](https://img.shields.io/badge/AppVersion-0.1.0-informational?style=flat-square)
+
+This Helm chart installs the JobSet Controller in your Kubernetes cluster. JobSet is a Kubernetes controller that manages groups of related Jobs as a single unit.
+
+## Installing the Chart
+
+To install the chart with the release name `jobset`:
+
+```bash
+helm install jobset ./charts/jobset
+```
+
+## Configuration
+
+| Key                                       | Type   | Default                      | Description |
+|-------------------------------------------|--------|------------------------------|-------------|
+| affinity                                  | object | `{}`                         |             |
+| certManager.certificate.duration          | string | `"8760h"`                    |             |
+| certManager.certificate.renewBefore       | string | `"720h"`                     |             |
+| certManager.enabled                       | bool   | `true`                       |             |
+| crds.enabled                              | bool   | `true`                       |             |
+| crds.install                              | bool   | `true`                       |             |
+| fullnameOverride                          | string | `""`                         |             |
+| image.pullPolicy                          | string | `"IfNotPresent"`             |             |
+| image.repository                          | string | `"jobset-controller"`        |             |
+| image.tag                                 | string | `""`                         |             |
+| imagePullSecrets                          | list   | `[]`                         |             |
+| leaderElection.enabled                    | bool   | `true`                       |             |
+| leaderElection.resourceName               | string | `"jobset-leader-election"`   |             |
+| manager.healthProbe.livenessInitialDelay  | int    | `15`                         |             |
+| manager.healthProbe.livenessPath          | string | `"/healthz"`                 |             |
+| manager.healthProbe.livenessTimeout       | int    | `30`                         |             |
+| manager.healthProbe.port                  | int    | `8081`                       |             |
+| manager.healthProbe.readinessInitialDelay | int    | `5`                          |             |
+| manager.healthProbe.readinessPath         | string | `"/readyz"`                  |             |
+| manager.healthProbe.readinessTimeout      | int    | `30`                         |             |
+| metrics.enabled                           | bool   | `true`                       |             |
+| metrics.service.port                      | int    | `8443`                       |             |
+| metrics.service.type                      | string | `"ClusterIP"`                |             |
+| metrics.serviceMonitor.enabled            | bool   | `false`                      |             |
+| metrics.serviceMonitor.interval           | string | `"30s"`                      |             |
+| metrics.serviceMonitor.labels             | object | `{}`                         |             |
+| metrics.serviceMonitor.scrapeTimeout      | string | `"10s"`                      |             |
+| nameOverride                              | string | `""`                         |             |
+| nodeSelector                              | object | `{}`                         |             |
+| podAnnotations                            | object | `{}`                         |             |
+| podSecurityContext.runAsNonRoot           | bool   | `true`                       |             |
+| podSecurityContext.seccompProfile.type    | string | `"RuntimeDefault"`           |             |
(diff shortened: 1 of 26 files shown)


charts/jobset/README.md
```diff
@@ -0,0 +1,81 @@
...
+| certManager.certificate.renewBefore       | string | `"720h"`                               |             |
+| certManager.enabled                       | bool   | `true`                                 |             |
+| crds.enabled                              | bool   | `true`                                 |             |
+| crds.install                              | bool   | `true`                                 |             |
+| fullnameOverride                          | string | `""`                                   |             |
+| image.pullPolicy                          | string | `"IfNotPresent"`                       |             |
+| image.repository                          | string | `"jobset-controller"`                  |             |
+| image.tag                                 | string | `""`                                   |             |
+| imagePullSecrets                          | list   | `[]`                                   |             |
+| kubeRbacProxy.image.pullPolicy            | string | `"IfNotPresent"`                       |             |
```
kannon92, line 29: We dropped kubeRbacPolicy and it will be dropped in 0.8.0. Let's focus on that for now.

andreyvelich (commented):
Thank you for doing this @slin1237!
Do we want to explore @tenzen-y's script that converts kustomize manifests to the Helm charts ?
https://github.com/kubernetes-sigs/kueue/blob/main/hack/update-helm.sh

slin1237:
> Thank you for doing this @slin1237! Do we want to explore @tenzen-y's script that converts kustomize manifests to the Helm charts ? https://github.com/kubernetes-sigs/kueue/blob/main/hack/update-helm.sh

i have not tried it yet
i looked through the code, its tailored for kueue, its something we can add as a follow up?
it would be nice to automate this, we also have a pretty big operator that requires this config and helm chart sync

kannon92:
cc @mimowo @tenzen-y 

I see that Kueue has helm support. Is there anything you suggest to help with maintainability?

mimowo:
we have the `update-helm` goal which is using [hack/update-helm.sh](https://github.com/kubernetes-sigs/kueue/blob/main/hack/update-helm.sh), which generates some of the helm files based on the configs is great for keeping them in-sync. Otherwise you risk the two sets of configuration diverging a lot over time.

charts/jobset/templates/_helpers.tpl
```diff
@@ -0,0 +1,60 @@
...
+{{- define "jobset.chart" -}}
+{{- printf "%s-%s" .Chart.Name .Chart.Version | replace "+" "_" | trunc 63 | trimSuffix "-" }}
+{{- end }}
+
+{{/*
+Common labels
+*/}}
+{{- define "jobset.labels" -}}
+helm.sh/chart: {{ include "jobset.chart" . }}
+{{ include "jobset.selectorLabels" . }}
```
Edwinhr716, line 36: This caused duplicate labels in LWS https://github.com/kubernetes-sigs/lws/issues/353, and looks like `/instance` is set in the resources here as well. Will be good to double check that the same issue doesn't arise here.

ahg-g:
I think we can merge this and follow up with the script, see https://github.com/kubernetes-sigs/lws/pull/364 for example

wdyt @kannon92 ?

charts/jobset/templates/_helpers.tpl
kannon92: Could you address this comment?

kannon92:
> I think we can merge this and follow up with the script, see [kubernetes-sigs/lws#364](https://github.com/kubernetes-sigs/lws/pull/364) for example
> 
> wdyt @kannon92 ?

I see.. Yea if that is all we need. The kueue script is much more of a beast.

There is an outstanding comment [here](https://github.com/kubernetes-sigs/jobset/pull/760#discussion_r1947088062).

ahg-g:
Perhaps the contributor doesn't have time, wdyt about merging and follow up with a fix like https://github.com/kubernetes-sigs/lws/pull/354

kannon92:
> Perhaps the contributor doesn't have time, wdyt about merging and follow up with a fix like [kubernetes-sigs/lws#354](https://github.com/kubernetes-sigs/lws/pull/354)

Let me do some testing with this PR and if it works then we can merge and iterate. I’d prefer more automated in place but it is what it is.

kannon92:
Got around to testing this:

```bash
kehannon@kehannon-thinkpadp1gen4i:~/Work/jobset$ helm install jobset ./charts/jobset
WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /home/kehannon/Work/openshift/clusters/kueue-operator-feb10/auth/kubeconfig
Error: INSTALLATION FAILED: 1 error occurred:
	* serviceaccounts "jobset-controller-manager" already exists

```

Looks like serviceAccounts duplicated.

I also see that we are missing namespace guards around namespace resources like Deployments/ServiceAccounts so these are getting installed into the default namespace.

charts/jobset/templates/manager/deployment.yaml
```diff
@@ -0,0 +1,91 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ include "jobset.fullname" . }}-controller-manager
+  labels:
+    {{- include "jobset.labels" . | nindent 4 }}
```
kannon92, line 6: I think you need:

```yaml
  namespace: {{.Release.Namespace }}
```

charts/jobset/templates/manager/serviceaccount.yaml
```diff
@@ -0,0 +1,12 @@
+{{- if .Values.serviceAccount.create -}}
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: {{ include "jobset.serviceAccountName" . }}
```
kannon92, line 5: I think you need:

```yaml
  namespace: {{.Release.Namespace }}
```

kannon92:
https://github.com/kubernetes-sigs/jobset/pull/760#issuecomment-2649005493

/hold

let’s address these comments before merge.

ahg-g:
@imreddy13 perhaps can help address the remaining comments to help unblock the release.

charts/jobset/templates/manager/deployment.yaml
ChenYi015: For namespaced scoped resources, I think it is recommended to make clarity by adding `namespace: {{ .Release.Namespace }}` to the metadata field. But this is not a necessity, as we can add `-n <namespace>` or `--namespace <namespace>` to the `helm install` command to create the resources in the specified namespace.

ahg-g:
Thanks a lot @slin1237, this is now superseded by https://github.com/kubernetes-sigs/jobset/pull/785

PR #764, opened 2025-01-24
vendor golang dependencies

### LINKED ISSUE #763: Vendor dependencies
I notice that Kueue vendors their dependencies. They found that it helped speed up the build system.

I'd like to propose that we vendor also so that we are consistent with kueue and we can gain some benefits of builds.

https://github.com/kubernetes-sigs/kueue/issues/2374

kannon92:
#### What type of PR is this?
/kind feature

Vendor our dependencies.

Fixes #763 

```release-note
NONE
```


kannon92:
/hold

I know the reason for these failures. Will work on them.

We need to exclude vendor from fmt-verify and the dockerfile needs to include vendor.

andreyvelich:
Does it really significantly speed up the process compared to `go mod`, especially for a lightweight Kubernetes controller like JobSet? 
One concern that I have with using vendor is that it significantly increases the size of the GitHub repository.

For large projects like `kubernetes/kubernetes`, the benefits of vendor are clear, but for a small-scale project like JobSet, the value seems less compelling.
cc @tenzen-y @alculquicondor

kannon92:
> Does it really significantly speed up the process compared to `go mod`, especially for a lightweight Kubernetes controller like JobSet? One concern that I have with using `vendor` is that it significantly increases the size of the GitHub repository. Which slows down GitOps, etc.
> 
> For large projects like `kubernetes/kubernetes`, the benefits of `vendor` are clear, but for a small-scale project like JobSet, the value seems less compelling. cc @tenzen-y @alculquicondor

Let's discuss on #763.

kannon92:
/close

Seems to be controversial so we can discuss on issue.

PR #771, opened 2025-01-26
update apiref to 1.32

kannon92:
#### What type of PR is this?
/kind documentation

APIREF was using 1.28 kubernetes.

Updating this to be 1.32.

Fixes #

I don't think there was much changes for this so this isn't a big deal.

```release-note

```


kannon92:
It looks like the reference docs are using 0.28 anyway. I don't think we need this change yet.

PR #775, opened 2025-01-30
Set user agent for requests coming from the jobset controller to "jobset"

### LINKED ISSUE #768: Set the user agent
Currently the api-server logs show that requests coming from the jobset controller are coming from an agent named `manager`, which is the command name.

See example: https://github.com/kubernetes-sigs/kueue/pull/339

imreddy13:

Set a meaningful user agent for requests coming from the jobset controller eg: jobset/v0.2.0 (linux/amd64)

Fixes #768

```release-note
Modified user agent for requests coming from the jobset controller. Action required: None.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -44,6 +44,8 @@ import (
 	"sigs.k8s.io/jobset/pkg/controllers"
 	"sigs.k8s.io/jobset/pkg/metrics"
 	"sigs.k8s.io/jobset/pkg/util/cert"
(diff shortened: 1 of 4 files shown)


kannon92:
I'm a bit unfamilar with this.

Can you deploy on a cluster and share that this behaves as you expect?

ahg-g:
/lgtm
/approve

I tested it, this works but needs a couple of tweaks to the make and dockerfile that I will submit.

PR #780, opened 2025-02-10
Set image tag and commit version at build time

ahg-g:
Follow up to https://github.com/kubernetes-sigs/jobset/pull/775


ahg-g:
image

Issue #777, opened 2025-01-31
Cleanup config CRDs


ahg-g:
The config API CRDs are not needed and should be removed to avoid confusion.

PR #781, opened 2025-02-10
Remove the Configuration API as a CRD

ahg-g:
Fixes https://github.com/kubernetes-sigs/jobset/issues/777


ardaguclu:
We discussed this in LWS https://github.com/kubernetes-sigs/lws/pull/325#discussion_r1923139687 and indeed there is no need for a configuration for this resource in kustomize, because it will not be stored in API server.

PR #782, opened 2025-02-10
Add missing external types to apply configurations

astefanutti:
#### What type of PR is this?

/kind bug

/kind api-change

This PR adds the external types the JobSet API depends on to the code-generator command so the generated client apply configurations are complete and do not rely on "non-apply" types. 

```release-note
Add missing external types to apply configurations
```

### DIFF
--- a/client-go/applyconfiguration/jobset/v1alpha2/replicatedjob.go
+++ b/client-go/applyconfiguration/jobset/v1alpha2/replicatedjob.go
@@ -15,16 +15,16 @@ limitations under the License.
 package v1alpha2
 
 import (
-	v1 "k8s.io/api/batch/v1"
+	v1 "k8s.io/client-go/applyconfigurations/batch/v1"
 )
 
 // ReplicatedJobApplyConfiguration represents a declarative configuration of the ReplicatedJob type for use
 // with apply.
 type ReplicatedJobApplyConfiguration struct {
-	Name      *string                       `json:"name,omitempty"`
-	Template  *v1.JobTemplateSpec           `json:"template,omitempty"`
-	Replicas  *int32                        `json:"replicas,omitempty"`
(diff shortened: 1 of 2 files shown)


astefanutti:
@ahg-g @kannon92 could you please give this a review?

astefanutti:
@tenzen-y FYI

kannon92:
Can you explain a bit more about this? What was the motivating factor behind this? I did look at kubeflow and here and I don't really understand what this gives us.

(Code looks good, but I'm unclear on motivation)

astefanutti:
For Kubeflow Trainer, we are considering moving to using SSA, and for this we need apply configuration for the owned JobSet resources to be "valid".

In general, the apply configurations should not be depending on non-apply configuration structs, otherwise that defeats the purpose of SSA.

tenzen-y:
> For Kubeflow Trainer, we are considering moving to using SSA, and for this we need apply configuration for the owned JobSet resources to be "valid".
> 
> In general, the apply configurations should not be depending on non-apply configuration structs, otherwise that defeats the purpose of SSA.

Let me put background here: https://github.com/kubeflow/training-operator/issues/2297

kannon92:
/lgtm

@ahg-g could you take a look also? I'm less sure of SSA but I think the change is fine.

andreyvelich (commented):
@tenzen-y @astefanutti Do we need to make sure that all objects that we create have external types in apply configs ? For example, PodGroup for Gang-Scheduling with co-scheduling.

astefanutti:
@andreyvelich you're right. I've checked this and the latest version of the apply configurations for the co-scheduling APIs are already OK as far as I can tell.

ahg-g:
@jpbetz any thoughts on this?

PR #810, opened 2025-03-03
fix(sdk): Add Kubernetes refs to the JobSet OpenAPI swagger

andreyvelich:
#### What type of PR is this?

/kind bug

After we migrated to the latest OpenAPI generator: `openapi-generator-cli:v7.11.0`, it generates JobSet models using the Pydantic library. It requires for all external models to be presented, otherwise serialization will fail:
```
from_dict()
to_dict()
```

As you can see, the `JobsetV1alpha2JobSet` class depends on [`V1ObjectMeta` class](https://github.com/kubernetes-sigs/jobset/blob/main/sdk/python/jobset/models/jobset_v1alpha2_job_set.py#L29-L36) which we don't import in the `jobset_v1alpha2_job_set.py` file.

To solve this problem, I've updated JobSet swagger to include external Kubernetes models, for instance
```
"$ref": "https://raw.githubusercontent.com/kubernetes/kubernetes/refs/tags/v1.32.2/api/openapi-spec/swagger.json#/definitions/io.k8s.apimachinery.pkg.apis.meta.v1.ObjectMeta"
```

cc @tenzen-y @kannon92 @ahg-g @Electronic-Waste @danielvegamyhre @epicseven-cup @astefanutti

We need this fix for Kubeflow Trainer integration: https://github.com/kubeflow/trainer/pull/2466

```release-note
NONE
```


ahg-g:
Is this embedding all jobset api dependencies in jobset's python sdk? is this a common practice?

kannon92:
/hold

@andreyvelich is working on this for kubeflow. I don't have a strong stance on this but I'd like to get buy in from kubeflow's PR until we merge this one.

sdk/python/README.md
```diff
@@ -67,6 +67,121 @@ Class | Method | HTTP request | Description
 
 ## Documentation For Models
```
kannon92, line 68: Is there a way to turn off the generation of documentation for these?

andreyvelich:
> Is this embedding all jobset api dependencies in jobset's python sdk?

Yes, otherwise the serialization will fail.
As I said, currently JobSet models uses Kubernetes models that don't exist in JobSet SDK package: https://github.com/kubernetes-sigs/jobset/blob/main/sdk/python/jobset/models/jobset_v1alpha2_replicated_job.py#L33.

> is this a common practice?

I noticed that Argo Workflows also import all Kubernetes dependencies in their SDK models: https://github.com/argoproj/argo-workflows/blob/main/sdks/python/client/argo_workflows/model/object_meta.py.

Argo Workflows has the valid OpenAPI spec swagger with all definitions: https://github.com/argoproj/argo-workflows/blob/main/api/openapi-spec/swagger.json

sdk/python/README.md
andreyvelich: Yeah, let me actually follow what we do for Kubeflow Trainer to remove it.

andreyvelich:
@kannon92 @ahg-g FYI, this fix also blocks us to use v0.8.0 JobSet Python package in our Kubeflow SDK.
I had to downgrade the JobSet version to avoid error in e2e: https://github.com/kubeflow/trainer/pull/2470/files#diff-271810b5660339392a509b5e1e83a9272497a323dabc5eb82910292c2f4ce87aR34-R36

The same error user reported in Slack: 
https://cloud-native.slack.com/archives/C0742LDFZ4K/p1741077959344629?thread_ts=1740708722.833979&cid=C0742LDFZ4K

andreyvelich:
Do we want to consider to cherry-pick it for v0.8.1 if that makes sense?

ahg-g:
> Do we want to consider to cherry-pick it for v0.8.1 if that makes sense?

I am ok to cherry-pick if this unblocks kubeflow trainer

sdk/python/README.md
```diff
@@ -59,36 +65,15 @@ from pprint import pprint
 
 ## Documentation for API Endpoints
 
-All URIs are relative to *http://localhost*
```
kannon92, line 62: @ahg-g is there a case to keep the jobset models?

I'm not sure what value these docs provide but we do have the documentation for jobset.

ahg-g: I don't think so, but I wouldn't wrestle with it if automatically generated

sdk/python/README.md
kannon92: This looks good!

kannon92 (approved):
/lgtm

Have we got consensus on the KubeFlow PR for this?

andreyvelich:
>Have we got consensus on the KubeFlow PR for this?

I think, so. 
@tenzen-y @Electronic-Waste @astefanutti  @saileshd1402  @deepanker13 @shravan-achar @akshaychitneni Any concerns to merge this PR?

Electronic-Waste (commented):
I'm ok for this PR:)

tenzen-y (commented):
/lgtm

Thank you.

kannon92:
@tenzen-y what are we missing to do this? Some kind of test-infra plugin?

tenzen-y:
> @tenzen-y what are we missing to do this? Some kind of test-infra plugin?

This one: https://github.com/kubernetes/test-infra/pull/34463

tenzen-y:
> status code 406 not one of [200], body: {"message":"Sorry, the diff exceeded the maximum number of lines (20000)","errors":[{"resource":"PullRequest","field":"diff","code":"too_large"}],"documentation_url":"https://docs.github.com/rest/pulls/pulls#get-a-pull-request","status":"406"}

It seems that diff size exceeds the limits 😓

kannon92:
@andreyvelich can you create a manually cherry-pick for this into 0.8?

PR #815, opened 2025-03-04
fix wrong cmd of output service

Monokaix:
#### What type of PR is this?
/kind cleanup
/kind documentation

The service created by jobset controller has  no label, so label-selector will not take effect and no output will show.

Fixes #

```release-note
NONE
```


ahg-g:
/lgtm
/approve

Thanks!

PR #819, opened 2025-03-06
bug fix: fix depends on validation if there are no replicated jobs

### LINKED ISSUE #816: JobSet errors due to dependsOn if replicated jobs are empty
**What happened**:

If you try and create a JobSet with no replicated jobs, we are hitting an out of bounds error due 
to the dependsOnValidation.

```yaml
kind: JobSet
metadata:
  name: paralleljobs
spec:
  replicatedJobs: []
```

We get the following error:

```
k create -f examples/simple/no-rp.yaml 
The JobSet "paralleljobs" is invalid: spec: Invalid value: "object": index out of bounds: 0 evaluating rule: DependsOn can't be set for the first ReplicatedJob
```

The cause of this:

```golang
// +kubebuilder:validation:XValidation:rule="!(has(self.replicatedJobs[0].dependsOn))",message="DependsOn can't be set for the first ReplicatedJob"
```

If replicated jobs are empty we are hitting a validation error and it states DependsOn can't be set.

I don't know if we should allow an empty JobSet. Kueue discovered this as they were testing with no replicated jobs for webhooks.

Create the above jobset on 0.8.0

Option 1 Fix:

Fix the validation to correctly only apply this rule if there are replicated jobs.

Option 2: Add validation to make sure there are no empty ReplicatedJobs.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.32
- JobSet version (use `git describe --tags --dirty --always`): v0.8.0
- Cloud provider or hardware configuration: local
- Install tools:
- Others:

kannon92:
#### What type of PR is this?
/kind bug

DependsOn validation hits an exception if there are no replicated jobs.

We should only check dependsOn if there exists replicated jobs.

Fixes #816

I think in the next release we will validate the minium length of ReplicatedJObs 
but that would be a breaking change.

I want to cherry-pick this to 0.8 so I want to fix this bug first.

```release-note
bug fix: fix depends on validation if there are no replicated jobs
```

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -80,7 +80,6 @@ const (
 
 // JobSetSpec defines the desired state of JobSet
 // +kubebuilder:validation:XValidation:rule="!(has(self.startupPolicy) && self.startupPolicy.startupPolicyOrder == 'InOrder' && self.replicatedJobs.exists(x, has(x.dependsOn)))",message="StartupPolicy and DependsOn APIs are mutually exclusive"
-// +kubebuilder:validation:XValidation:rule="!(has(self.replicatedJobs[0].dependsOn))",message="DependsOn can't be set for the first ReplicatedJob"
 type JobSetSpec struct {
 	// ReplicatedJobs is the group of jobs that will form the set.
 	// +listType=map
--- a/charts/jobset/crds/jobset.x-k8s.io_jobsets.yaml
+++ b/charts/jobset/crds/jobset.x-k8s.io_jobsets.yaml
@@ -1,3 +1,4 @@
+---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
@@ -9080,8 +9081,6 @@ spec:
             - message: StartupPolicy and DependsOn APIs are mutually exclusive
               rule: '!(has(self.startupPolicy) && self.startupPolicy.startupPolicyOrder
                 == ''InOrder'' && self.replicatedJobs.exists(x, has(x.dependsOn)))'
-            - message: DependsOn can't be set for the first ReplicatedJob
-              rule: '!(has(self.replicatedJobs[0].dependsOn))'
           status:
             description: JobSetStatus defines the observed state of JobSet
             properties:
--- a/config/components/crd/bases/jobset.x-k8s.io_jobsets.yaml
+++ b/config/components/crd/bases/jobset.x-k8s.io_jobsets.yaml
@@ -9081,8 +9081,6 @@ spec:
             - message: StartupPolicy and DependsOn APIs are mutually exclusive
               rule: '!(has(self.startupPolicy) && self.startupPolicy.startupPolicyOrder
                 == ''InOrder'' && self.replicatedJobs.exists(x, has(x.dependsOn)))'
-            - message: DependsOn can't be set for the first ReplicatedJob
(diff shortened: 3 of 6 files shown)


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -80,7 +80,6 @@ const (
 
 // JobSetSpec defines the desired state of JobSet
 // +kubebuilder:validation:XValidation:rule="!(has(self.startupPolicy) && self.startupPolicy.startupPolicyOrder == 'InOrder' && self.replicatedJobs.exists(x, has(x.dependsOn)))",message="StartupPolicy and DependsOn APIs are mutually exclusive"
-// +kubebuilder:validation:XValidation:rule="!(has(self.replicatedJobs[0].dependsOn))",message="DependsOn can't be set for the first ReplicatedJob"
```
ahg-g, line 83: so this now allows users to set it on the first replicatedJob, what happens if they do?

ahg-g:
> DependsOn validation hits an exception if there are no replicated jobs.

Oh, we allow creating jobSets with no replicatedJob set? what would such a jobset do?

api/jobset/v1alpha2/jobset_types.go
kannon92: https://github.com/kubernetes-sigs/jobset/pull/819/files#diff-abd102e840ad3a94481819c05b8f6b1e8eda15c7b18483739bc9066460c9d2b8R162

I removed the CEL validation and moved it to an actual webhook.

I couldn't find any examples of using CEL to validate slice length so I gave up.

kannon92:
> > DependsOn validation hits an exception if there are no replicated jobs.
> 
> Oh, we allow creating jobSets with no replicatedJob set? what would such a jobset do?

It would create fine once this rule is relaxed. Right now it fails due to "DependsOn set for first replicated job".

ReplicatedJobs are not mutable so it is technically possible that someone could patch JobSet to add replicatedJobs.

I am thinking for 0.9 (today) I will put up a patch to force JobSets to have at least one replicated job. But I don't think this should be cherry-picked.

ahg-g:
> ReplicatedJobs are not mutable so it is technically possible that someone could patch JobSet to add replicatedJobs.

So we have an indirect validation on it that we can improve by making it explicit. I still don't get what problem this PR is trying to address.

kannon92:
> > ReplicatedJobs are not mutable so it is technically possible that someone could patch JobSet to add replicatedJobs.
> 
> So we have an indirect validation on it that we can improve by making it explicit. I still don't get what problem this PR is trying to address.

I added the bug to the description. https://github.com/kubernetes-sigs/jobset/issues/816

We saw this in the kueue integration tests.

ahg-g:
I think the solution is to fix the Kueue integration tests and continue to validate the JobSet should have at least one replicatedJob.

kannon92:
> I think the solution is to fix the Kueue integration tests and continue to validate the JobSet should have at least one replicatedJob.

Yea but without this change, the depends on validation fails because we don't have a replicated job.

Hence my issue filed.

We get an error saying we hit an out of bounds error trying to do a kubebuilder validation and it gives a silly error message.

ahg-g:
/lgtm


I am trying to make sure we have most validation done using CEL ...

kannon92:
Yea I wanted CEL to only check depends on if the replicatedJobs have at least 1 element. But I could not find an example for that.

Something like (len(x) > 0 && dependsOn)? But I could not find anything close to that..

PR #824, opened 2025-03-07
Add bash script to sync manifests from Helm chart to Kustomize

### LINKED ISSUE #820: Automated way to generate helm templates
We need to have some kind of sync or check to make sure that the helm templates do not differ from kustomize.

We can do this in phases but it is currently required that any CRD change must be copied to the helm crd directory because we don't autosync this.

This will easily be a bug in the future.

ChenYi015:
#### What type of PR is this?

/kind cleanup

Fixes #820

```release-note

```


kannon92:
@ChenYi015 feel free to mark this ready if you want the CI to run?

I guess draft PRs do not run the CI.

ahg-g:
I am starting to lean towards making helm as the source of truth, and use `helm template` to auto-generate a kubectl deployable manifest. wdyt?

kannon92:
@ChenYi015 this would be really great to have.

Do you have any cycles to spare on this?

PR #828, opened 2025-03-08
bump envtest

kannon92:
#### What type of PR is this?
/kind cleanup

- Bump envtest to 1.31.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,5 @@
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.30
+ENVTEST_K8S_VERSION = 1.31
 
 # Get the currently used golang install path (in GOPATH/bin, unless GOBIN is set)
 ifeq (,$(shell go env GOBIN))


kannon92:
cc @mimowo @tenzen-y 

Bumping controller-tools seem to unlease loads of Job failures. Any ideas on the fix?

Makefile
```diff
@@ -1,5 +1,5 @@
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
-ENVTEST_K8S_VERSION = 1.30
+ENVTEST_K8S_VERSION = 1.31
```
andreyvelich, line 2: Do we want to bump it to 1.32 since we support it ?
https://github.com/kubernetes-sigs/jobset/blob/main/go.mod#L12

kannon92: I tried that but 1.32 seems to not work yet for us.

kannon92: I can bump it here and try to fix it. probably best.  

Bumping from 1.30 to 1.32 seemed like a big jump.

andreyvelich: Do you want to do that as a followup ?
I am fine migrating to 1.31 first.

PR #834, opened 2025-03-12
address snyk vulnerabilites in downstream fork

kannon92:
#### What type of PR is this?
/kind cleanup

My downstream fork flags these as vulnerabilities. 

These are minor but going to address them anyway.

- https://github.com/openshift/kubernetes-sigs-jobset/pull/32
- https://github.com/openshift/kubernetes-sigs-jobset/pull/31

Fixes #

```release-note
NONE
```


kannon92:
@andreyvelich @tenzen-y 

could you guide here?

My internal scanners flag these test dependencies as a viper ability. I tried adding the pins to the requirements.txt files but those are autogenerated. I’m not exactly sure how to get the requirements pinned with openapi gen code.

tenzen-y:
> @andreyvelich @tenzen-y
> 
> could you guide here?
> 
> My internal scanners flag these test dependencies as a vulnerability . I tried adding the pins to the requirements.txt files but those are autogenerated. I’m not exactly sure how to get the requirements pinned with openapi gen code.

Can I access Snyc Dashboard or can you provide full error message? I could not find what exactly the problems were.

andreyvelich:
Should we remove the tests from OpenAPI and only keep actual JobSet modules since we don't use those tests?

kannon92:
@tenzen-y i can check tomorrow. The errors were mostly related to those dependencies so the automatic PR says that you can just make sure those versions are greater than a certain value. If they are I assume the “vulnerability” is resolved.

do in this case I mostly want to preserve the pinning through auto generation.

PR #841, opened 2025-03-18
update to golang 1.24

kannon92:
#### What type of PR is this?
/kind cleanup

Upgrade golang to 1.24.

Fixes #

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -315,7 +315,7 @@ openapi-gen:
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) $(GO_CMD) install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240813183042-b901db121e1f
+	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) $(GO_CMD) install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -315,7 +315,7 @@ openapi-gen:
 .PHONY: envtest
 envtest: $(ENVTEST) ## Download envtest-setup locally if necessary.
 $(ENVTEST): $(LOCALBIN)
-	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) $(GO_CMD) install sigs.k8s.io/controller-runtime/tools/setup-envtest@v0.0.0-20240813183042-b901db121e1f
+	test -s $(LOCALBIN)/setup-envtest || GOBIN=$(LOCALBIN) $(GO_CMD) install sigs.k8s.io/controller-runtime/tools/setup-envtest@latest
```
ardaguclu, line 318: In my opinion, always using a tagged versions is better like https://github.com/kubernetes-sigs/lws/blob/8c57c6dc6768cb571a013ab59e84cf60968b9909/Makefile#L246. Otherwise, this will start failing again in each golang bump of setup-envtest.

Edwinhr716: Don't we have to update them on each golang bump anyway? What are the drawbacks of having a tagged version instead of `latest`?

kannon92: @ahg-g WDYT here?

If you are happy with what we have, this can be approved.

I don't think it matters either way what we pick here. If we pin a version we have to remember to update it.

If we keep it at latest then we know what happens if there is a go update.

so I lean towards leaving this as is for now.

ahg-g: lets use latest and if it fails we can update

Issue #762, opened 2025-01-24
Support for Job-level ResourceClaimTemplate


johnbelamaric:
**What would you like to be added**:
I would like to be able to associate a ResourceClaimTemplate with a Job in the JobSet, such that each time the Job is instantiated, a new ResourceClaim is created based on the template, and all Pods in the Job are pointed to that ResourceClaim.

**Why is this needed**:
ResourceClaimTemplates in DRA are used to create ResourceClaims that follow a Pod's lifecycle. For example, it allows you to use a Deployment, and when the Deployment creates a Pod, a new ResourceClaim is also created for that Pod in order to get it a GPU.

That makes sense for ResourceClaims that are specific to a Pod. However, there can also be ResourceClaims that are used *by a set of pods in a workload*. This is because ResourceClaims can be shared, and can represent a multi-node resource.

For example, @klueska is building a DRA driver that uses a ResourceSlice to represent IMEX channels within an NVLink domain. A single ResourceClaim in a given namespace is created to allocate an IMEX channel. All Pods that need to participate in that channel then reference that ResourceClaim. The Pods would *additionally* have a reference to a ResourceClaimTemplate for obtaining their GPUs. But there is only a *single* ResourceClaim across all participating Pods for the IMEX channel.

Similarly, @mortent and I are working on a representation of multi-node TPU slices in DRA. These would work much the same way. A single ResourceClaim would be shared amongst all Pods that need to run on that TPU slice.

So, the idea is that when you instantiate a Job from a JobSet, you will need to allocate, say, a TPU slice or an IMEX channel that is shared amongst all Pods in the Job.

This enhancement requires the following artifacts:

- [X] Design doc
- [X] API change
- [X] Docs update

The artifacts should be linked in subsequent comments.

**Note**

This may also require - or be done better - with changes to the core Job API. Even if we don't need that, it very well may require changes to the ResourceClaim controller, if we want to use that to generate these ResourceClaims (we could generate them ourselves).

This issue is to open up the discussion on the best way to solve these use cases.

cc @acpana @kannon92 @pohly @klueska @mortent

kannon92:
> This may also require - or be done better - with changes to the core Job API. Even if we don't need that, it very well may require changes to the ResourceClaim controller, if we want to use that to generate these ResourceClaims (we could generate them ourselves).

My first glance of this I think having the ability in tree for kubernetes would be the long term plan.

[...]

Issue #572, opened 2024-05-14
Support Stateful JobSet


tenzen-y:
**What would you like to be added**:
I would like to support features to create a single PVC and mount the PV to some replicatedJobs like this:

```yaml
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata:
  name: volume-sample
spec:
  volumePolicy:
    replicatedJobs:
     - name: A
     - name: B
    volumeClaimTemplates:
    - metadata:
        name: pretrained-model
      spec:
        accessModes: [ "ReadWriteMany" ]
        storageClassName: "my-storage-class"
        resources:
          requests:
            storage: 1000Gi
  replicatedJobs:
  - name: workers
[...]
```
In this example, JobSet creates a PVC, "pretrained-model" and then the created PV is mounted to replicatedJobs specified in the `.spec.volumePolicy.replicatedJobs`

This feature is similar to https://github.com/kubernetes/kubernetes/issues/115066

**Why is this needed**:
In large distributed training, we often store the base model, and then we want to share the pre-trained model with all workers so that we can avoid downloading the pre-trained model many times.

tenzen-y:
cc: @ahg-g @andreyvelich

danielvegamyhre:
Is this shared volume containing the pretrained model not possible to do using a regular PV, and specifying a PVC in the JobTemplate? Or is the purpose of this just to provide some automation and lifecycle management for the PV so the user doesn't need to specify the PV manifest separately?

andreyvelich:
>Or is the purpose of this just to provide some automation and lifecycle management for the PV so the user doesn't need to specify the PV manifest separately?

That's right, we want to manage lifecycle of the storage on the controller side, not on the client side.
This should simplify the ability to fine-tune models where [each worker should share pre-trained model and dataset](https://deploy-preview-3718--competent-brattain-de2d6d.netlify.app/docs/components/training/reference/fine-tuning/#architecture) that we download in the master pod.

ahg-g:
I like this feature, we should track it under 0.7 release.

danielvegamyhre:
@tenzen-y I'm tentatively labeling the issue to be marked as part of the v0.7.0 release, under the assumption you plan to work on this - let me know if that's not the case.

andreyvelich:
@danielvegamyhre I can try to take this feature in the mid Q4 - Q1.
Do we know when are you planning to cut the v0.7.0 for JobSet ?

[...]

PR #853, opened 2025-03-24
docs: small kep for DRA

acpana:
#### What type of PR is this?

/kind feature
/kind api-change

Small KEP for https://github.com/kubernetes-sigs/jobset/issues/762

Adresses https://github.com/kubernetes-sigs/jobset/issues/762

n/a

### DIFF
--- a/keps/762-ResourceClaimTemplate/README.md
+++ b/keps/762-ResourceClaimTemplate/README.md
@@ -1,0 +1,405 @@
+# KEP-762: Support ResourceClaimTemplate
+
+<!--
+This is the title of your KEP. Keep it short, simple, and descriptive. A good
+title can help communicate what the KEP is and should be considered as part of
+any review.
+-->
+
+<!--
+A table of contents is helpful for quickly jumping to sections of a KEP and for
+highlighting any additional information provided beyond the standard KEP
+template.
+
+Ensure the TOC is wrapped with
+  <code>&lt;!-- toc --&rt;&lt;!-- /toc --&rt;</code>
+tags, and then generate with `hack/update-toc.sh`.
+-->
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Design Details](#design-details)
+    - [API Changes](#api-changes)
+    - [Behavior Changes](#behavior-changes)
+  - [User Stories](#user-stories)
+    - [IMEX User Story](#imex-user-story)
+    - [TPU Slice User Story](#tpu-slice-user-story)
+    - [Kubeflow User Story](#kubeflow-user-story)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+  - [Test Plan](#test-plan)
+      - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Alternatives](#alternatives)
+  - [Manual ResourceClaim management](#manual-resourceclaim-management)
+  - [Modifying CoreJob API](#modifying-corejob-api)
+  - [PodSpec Patching](#podspec-patching)
+<!-- /toc -->
+
+## Summary
+
+<!--
+This section is incredibly important for producing high-quality, user-focused
+documentation such as release notes or a development roadmap. It should be
+possible to collect this information before implementation begins, in order to
+avoid requiring implementors to split their attention between writing release
+notes and implementing the feature itself. KEP editors and SIG Docs
+should help to ensure that the tone and content of the `Summary` section is
+useful for a wide audience.
+
+A good summary is probably at least a paragraph in length.
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP proposes adding a `resourceClaimTemplates` field to the JobSet CRD
+(specifically within `spec.replicatedJobs[].resourceClaimTemplates`) to allow users to define
+ResourceClaimTemplates that the JobSet controller will use to dynamically provision shared ResourceClaims
+for each Job created by the JobSet. All Pods within a given Job will be
+configured to use the same set of created ResourceClaims, enabling access
+to shared resources like multi-node accelerator slices.
+
+## Motivation
+
+<!--
+This section is for explicitly listing the motivation, goals, and non-goals of
+this KEP.  Describe why the change is important and the benefits to users. The
+motivation section can optionally provide links to [experience reports] to
+demonstrate the interest in a KEP within the wider Kubernetes community.
+
+[experience reports]: https://github.com/golang/go/wiki/ExperienceReports
+-->
+
+Currently, Kubernetes' Dynamic Resource Allocation (DRA) primarily focuses on ResourceClaims
+that are tightly coupled to the lifecycle of individual Pods (via `ResourceClaimTemplate` in the Pod spec).
+This model works well for per-pod resources like GPUs. However, some distributed workloads
+require resources that are shared across all Pods within a Job, such as a multi-node accelerator slice
+(e.g., IMEX channels within an NVLink domain, or multi-node TPU slices). The existing `ResourceClaimTemplate`
+mechanism within the Pod spec doesn't support this shared resource allocation
+pattern for Jobs within a JobSet. This enhancement will fill that gap.
+
+### Goals
+
+<!--
+List the specific goals of the KEP. What is it trying to achieve? How will we
+know that this has succeeded?
+-->
+
+* Enable `JobSet` controller to create and manage shared `ResourceClaims` (using `ResourceClaimTemplates`)
+* Allow users to define `ResourceClaimTemplates` at the JobSet level.
+* Enable the JobSet controller to create a `ResourceClaim` (based on the template) for each Job created.
+* Allow optional targeting of specific containers within the Pod to receive access to the shared ResourceClaim.
+* Manage the lifecycle of the Job-scoped ResourceClaims, ensuring they are
+created when the Job starts and deleted when the Job completes or is deleted.
+
+### Non-Goals
+
+<!--
+What is out of scope for this KEP? Listing non-goals helps to focus discussion
+and make progress.
+-->
+
+* Modifying the core Kubernetes Job/ PodSpec API.
+* Supporting retroactive application of new or modified `resourceClaimTemplates` to already running Jobs.
+
+## Proposal
+
+<!--
+This is where we get down to the specifics of what the proposal actually is.
+This should have enough detail that reviewers can understand exactly what
+you're proposing, but should not include things like API designs or
+implementation. What is the desired outcome and how do we measure success?.
+The "Design Details" section below is for the real
+nitty-gritty.
+-->
+
+Introduce a new field `resourceClaimTemplates` within the `JobSet.spec`. This
+field will be a list of `ResourceClaimTemplate` objects, similar to how they are defined
+within a Pod's spec, but at the JobSet level.
+
+
+### Design Details
+
+#### API Changes
+
+<!--
+This section should contain enough information that the specifics of your
+change are understandable. This may include API specs (though not always
+required) or even code snippets. If there's any ambiguity about HOW your
+proposal will be implemented, this is the place to discuss them.
+-->
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: shared-resource-jobset
+spec:
+  replicatedJobs:
+    - name: worker-group
+      replicas: 3
+      template: # This is the JobTemplateSpec
+        spec: # This is the JobSpec
+          parallelism: 1 # JobSpec fields
+          completions: 1 # JobSpec fields
+          template: # This is the PodTemplateSpec
+            spec: # This is the PodSpec
+              containers:
+                - name: worker
+                  image: your-training-image:latest
+                  command: ["python", "train.py"]
+                  resources:
+                    requests:
+                      cpu: "2"
+                      memory: "4Gi"
+                  resourceClaims: # Pod-level DRA
+                    - name: gpu-claim
+                      source:
+                        resourceClaimTemplateName: pod-gpu-template
+                - name: helper
+                  image: helper-image:latest
+                  command: ["sleep", "infinity"]
+              restartPolicy: OnFailure
+              # Example Pod-level ResourceClaimTemplate (for comparison)
+              # resourceClaimTemplates:
+              # - metadata:
+              #     name: pod-gpu-template
+              #   spec:
+              #     resourceClassName: nvidia-gpu
+          resourceClaimTemplates:  # NEW: JobSet-level ResourceClaimTemplates (added inside JobSpec)
+            - metadata:
+                name: imex-channel
+              spec:
+                resourceClassName: imex-nvlink-domain
+                containers:
+                  - worker
+            - metadata:
+                name: shared-data
+              spec:
+                resourceClassName: shared-data-resource
+```
+
+#### Behavior Changes
+1. For each `Job` created from the `replicatedJobs`, the controller creates a `ResourceClaim` per
+`resourceClaimTemplates` entry (e.g., `gpu-training-jobset-job-0-imex-channel`).
+
+2. An `ownerReference` is added from the `ResourceClaim` to the `Job`.
+
+3. The controller modifies the Job's Pod spec. For the imex-channel claim,
+only the worker container gets a reference. For the shared-data claim, all containers get a reference.
+
+4. `ResourceClaim`s are deleted when the `Job` completes (default policy).
+
+### User Stories
+
+<!--
+Detail the things that people will be able to do if this KEP is implemented.
+Include as much detail as possible so that people can understand the "how" of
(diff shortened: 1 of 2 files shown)


acpana:
cc @johnbelamaric

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+                - name: helper
+                  image: helper-image:latest
+                  command: ["sleep", "infinity"]
+              restartPolicy: OnFailure
+              # Example Pod-level ResourceClaimTemplate (for comparison)
+              # resourceClaimTemplates:
+              # - metadata:
+              #     name: pod-gpu-template
+              #   spec:
+              #     resourceClassName: nvidia-gpu
```
johnbelamaric, line 179: This comment should be *before* the entire `JobSet`, as a separate top-level resource. Also, this is the older API, what you want instead is:

```yaml
apiVersion: resource.k8s.io/v1beta1
kind: ResourceClaimTemplate
metadata:
  name: gpu
spec:
  devices:
    requests:
    - name: gpu
      deviceClassName: gpu.nvidia.com
```

I don't think you need to comment it out, it's useful to show the per-pod and the per-job level RCTs.

For the per-job RCT, it would look something like:

```yaml
- metadata:
    name: imex-channel
  spec:
    devices:
      requests:
      - name: imex-channel
        deviceClassName: imex.nvidia.com
```

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+
+<!--
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+* The containers field within `resourceClaimTemplates` is optional. If omitted,
+the `ResourceClaim` will be made available to all containers in the Pod.
```
johnbelamaric, line 242: Hmm. Probably this should be that it does not add it to any containers. Pod-level claims are valid for NICs, for example. I don't know though of multi-node pod-level claims. But still, I think less magic is better.

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+### Manual ResourceClaim management
+
+Users could manually create the shared ResourceClaim and then reference it in their
+Pod specs. This requires manual cleanup, and doesn't integrate well with the
+JobSet's lifecycle management.
+
+### Modifying CoreJob API
+
+We could add a similar feature to the core Kubernetes Job API. However, this has a
+much  barrier to entry, wider impact, and longer development cycle compared to
```
johnbelamaric, line 374: s/much/much higher/

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+enhancing an out-of-tree controller like JobSet.
+
+### PodSpec Patching
+
+Instead of the `JobSet` controller injecting resourceClaims entries into the Pod
+spec based on `resourceClaimTemplates`, we can consider an alternative approach.
```
johnbelamaric, line 380: `resourceClaimTemplates[].spec.containers`

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+containers:
+  - name: worker
+    resourceClaims:
+      - name: shared-claim-ref # Name for the reference within the container
+        source:
+          # Uses the name from JobSet's resourceClaimTemplates.metadata.name
+          resourceClaimName: imex-channel # Placeholder
+```
+
+1. Users would define a `resourceClaim`s entry with the containr spec in the `Job`'s
```
johnbelamaric, line 393: s/containr/container/

johnbelamaric (commented):
awesome, thanks

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+-->
+
+Introduce a new field `resourceClaimTemplates` within the `JobSet.spec`. This
+field will be a list of `ResourceClaimTemplate` objects, similar to how they are defined
+within a Pod's spec, but at the JobSet level.
+
+
+### Design Details
+
+#### API Changes
```
kannon92, line 134: I like the example but I’d also like to see the API you are proposing.

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,405 @@
...
+shared by all Pods of a Job, ensuring my training job has the necessary multi-node resources.
+
+#### Kubeflow User Story
+
+As a Kubeflow user leveraging JobSet through common interfaces like the [Training Operator](https://github.com/kubeflow/trainer?tab=readme-ov-file#kubeflow-training-operator-v1),
+I should be able to request per-job shared resources in addition to my per-pod resources in a declarative manner,
+so that the platform can manage resources for complex multi-node, multi-pod jobs without manual intervention.
+
+
+### Notes/Constraints/Caveats
```
kannon92, line 232: A few things I'll ask you to answer here.

1) How does your proposal handle that DRA has various api levels between 1.30-1.32?

2) How do you plan to test this functionality? 
  a) Is there any way to test this cross node functionality with dra-example-driver?

kannon92: If we eventually get this work into batch/Job in upstream, what does that mean about this feature? Can you see a world where this exists and Job ResourceClaimTemplates exist together?

acpana: @johnbelamaric please keep me honest/ on track here but here's my thoughts:

> How does your proposal handle that DRA has various api levels between 1.30-1.32?

One advantage of use adding the RCT as a top level JobSet change is that we can "decouple" from the core API changes. That said, we should encourage that people use the beta/ latest DRA API.

> How do you plan to test this functionality?
a) Is there any way to test this cross node functionality with dra-example-driver?

AFAICT, the `dra-example-driver` doesn't have cross node support yet; Do you have any suggestions beyond manual testing?

>  Can you see a world where this exists and Job ResourceClaimTemplates exist together?

I'll defer to folks that have more insight into the future of DRA/ JobSet but my 2c are that having a semantic for making per JobSet device claims will continue to be useful.

kannon92: > One advantage of use adding the RCT as a top level JobSet change is that we can "decouple" from the core API changes. That said, we should encourage that people use the beta/ latest DRA API.

At some point, I imagine the RCT will create ResourceClaims via DRA, right? So in this case, we will eventually create DRA APIs where the API level for these differns from Kubernetes versions.

I'm trying to understand why you will use this feature for 1.30, 1.31 and 1.32 (where DRA is alpha, alpha and beta). JobSet runs on the latest three releases of Kubernetes.

>AFAICT, the dra-example-driver doesn't have cross node support yet; Do you have any suggestions beyond manual testing?

If that is what we have, I guess that is fine. Do you have an environment where you can test this feature and confirm it works before merge?

johnbelamaric: @bg-chun is working on beefing up the example driver to demonstrate various device modeling features (e.g., partitionable). We might though want a separate test driver?

The version is an interesting point. In 1.33 we have v1beta2, which we expect will be the shape of the API that we GA. I don't think we should support the alpha versions, frankly. Is it a problem for this particular feature to only work on 1.32+? Since that's when it went to beta? If we do that, then we can shape JobSet's embedded template based on the v1beta2 RCT, but actually generate v1beta1 RCTs (they can roundtrip, so this is do-able). That way, it works on 1.32 and 1.33. Eventually (when 1.34 is the oldest release), we change the RCTs we create to v1 not v1beta1.

Alternatively, we reference a separate RCT by name in the JobSet spec, and so the shape of that API no longer includes the template for the RCT, but pretty much just the "containers" bit that we have to layer in.

bg-chun: I’m currently working on adding a reference implementation for the Partitionable Device API to the example DRA driver. This API covers two scenarios: NVIDIA MIG and Multi-Host TPU Scheduling. However, since the current example DRA driver mimics NVIDIA GPUs, I’ll start by implementing a simple version of the MIG feature.

After adding this MIG-like implementation of the Partitionable Device API, I think we should have a discussion about the overall direction of the example device driver project. I believe the example DRA driver should ultimately demonstrate various device modeling scenarios.

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,417 @@
...
+                      memory: "4Gi"
+                  resourceClaims: # Pod-level DRA
+                    - name: gpu-claim
+                      source:
+                        resourceClaimTemplateName: pod-gpu-template
+                - name: helper
+                  image: helper-image:latest
+                  command: ["sleep", "infinity"]
+              restartPolicy: OnFailure
+          resourceClaimTemplates:  # NEW: JobSet-level ResourceClaimTemplates (added inside JobSpec)
```
andreyvelich, line 187: Why this API is under `replicatedJob[].template.spec.resourceClaimTemplates` ?

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+#### TPU Slice User Story
+
+As a researcher running distributed training on a multi-node TPU slice, I want
+to define a JobSet that automatically provisions a TPU slice ResourceClaim
+shared by all Pods of a Job, ensuring my training job has the necessary multi-node resources.
+
+#### Kubeflow User Story
+
+As a Kubeflow user leveraging JobSet through common interfaces like the [Training Operator](https://github.com/kubeflow/trainer?tab=readme-ov-file#kubeflow-training-operator-v1),
```
andreyvelich, line 240:
```suggestion
As a Kubeflow user leveraging JobSet through common interfaces like the [Kubeflow Trainer](https://www.kubeflow.org/docs/components/trainer/overview/#who-is-this-for),
```

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,417 @@
...
+
+Users could manually create the shared ResourceClaim and then reference it in their
+Pod specs. This requires manual cleanup, and doesn't integrate well with the
+JobSet's lifecycle management.
+
+### Modifying CoreJob API
+
+We could add a similar feature to the core Kubernetes Job API. However, this has a
+much higher barrier to entry, wider impact, and longer development cycle compared to
+enhancing an out-of-tree controller like JobSet.
```
andreyvelich, line 387: What is our strategy here @ahg-g @kannon92 @acpana @tenzen-y @johnbelamaric ?
We have similar features that we want to implement on the Job layer first: https://github.com/kubernetes-sigs/jobset/issues/572.

I am not sure if it is right for the long-term to push more features to the JobSet controller rather than Job controller.

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,417 @@
...
+            - metadata:
+                name: shared-data
+              spec:
+                devices:
+                  requests:
+                  - name: shared-data
+                    deviceClassName: shared-data-resource
+```
+
+#### Behavior Changes
```
andreyvelich, line 206: This is similar to PVC Template for StatefulSet, right?
https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#volume-claim-templates

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,417 @@
...
+What are the caveats to the proposal?
+What are some important details that didn't come across above?
+Go in to as much detail as necessary here.
+This might be a good place to talk about core concepts and how they relate.
+-->
+
+* If `spec.replicatedJobs[].template.spec.resourceClaimTemplates[].spec.containers` is omitted, the controller should still create the ResourceClaim for the Job and associate it with the Pod (e.g., by adding it to pod.spec.resourceClaims), but it should not automatically add a reference to it in spec.containers[].resourceClaims for any container.
+
+* Pod-level `resourceClaim`s with the same name as a JobSet-templated
+claim will override the JobSet-level claim for that container, and a warning will be emitted.
```
andreyvelich, line 257: Can we solve it with webhook validation rather than warning ?

andreyvelich (commented):
Thank you for this @johnbelamaric @acpana!
Maybe we could also speak about it during one of the following Batch WG calls.

keps/762-ResourceClaimTemplate/README.md
kannon92: Thank you @johnbelamaric!

That is the main items I am thinking through on this.

How do we handle different version for DRA?

I am okay with supporting 1.32+ but we should make it clear.

What happens with 1.30 and 1.31 if this API is used? Will there be a feature gate for JobSet?

Can we discover what kind of API is available in the APIServer and convert?

keps/762-ResourceClaimTemplate/README.md
kannon92: xref: https://github.com/kubernetes-sigs/lws/issues/444

kannon92: xref: https://github.com/kubernetes-sigs/jobset/issues/762#issuecomment-2613969432

We did meet and discuss this feature at wg-device-management a few months back. I think there is a lot of interest in this and it is not clear if we put all of this in Job controller.  If we want each replicated job to have a different resourceClaim, then I think it would make sense to have this just in Job API.

If we want multiple replicated jobs to share the same resourceClaim it may make sense to have it at the JobSet leve.

For LWS, @ahg-g mentions that even if there was support for this in StatefulSet it still may be necessary to have this for LWS to help tune the behavior LWS wants.

I also am open to experimenting more in out-of-tree controllers. 

There are downsides to putting this in an out-of-tree controller though. We don't test/verify DRA in any capacity in our CI so we would need some kind of testing to verify that this works with DRA across different releases. We also inherit a potentially unstable API and if upstream adds suport for this feature maybe we have to deprecate.

Edwinhr716: For LWS, we want both the Leader and Worker StatefulSet to share a resource claim. Support at the StatefulSet level only would support the all the workers sharing one, and the leader having a separate resource claim.

keps/762-ResourceClaimTemplate/README.md
acpana: ( just chiming in here, I reckon the thinking is to go with:

> I am okay with supporting 1.32+ but we should make it clear.

& feature gate it. We can error our if the gate is on but we don't meet the minimum API server requirement?)

keps/762-ResourceClaimTemplate/README.md
```diff
@@ -0,0 +1,417 @@
...
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP proposes adding a `resourceClaimTemplates` field to the JobSet CRD
+(specifically within `spec.replicatedJobs[].resourceClaimTemplates`) to allow users to define
```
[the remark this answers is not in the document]
acpana: I think this thread addresses the comment -- https://github.com/kubernetes-sigs/jobset/pull/853#discussion_r2035802009

keps/762-ResourceClaimTemplate/README.md
acpana: Ah good catch! This should be one level higher 😛

keps/762-ResourceClaimTemplate/README.md
acpana: I think so but I am not super familiar with PVCs at this point

keps/762-ResourceClaimTemplate/README.md
acpana: We could! But I'd prefer not to rely on webhooks given the complexity they introduce 🤔 . I can add a small blurb about that.

andreyvelich: JobSet already maintains Validation Webhook, so we can just add additional check for `resourceClaim`

keps/762-ResourceClaimTemplate/README.md
mortent: So my understanding from this discussion is:
* We think that DRA eventually will be added to the in-tree workload controllers.
* We think there are scenarios where it will not be sufficient to only rely on support in the in-tree controllers (like the LWS example), meaning that there needs to be some level of support for DRA out-of-tree.

What is not clear to me, is whether the implementation in JobSet would look different if there we had support in-tree, i.e. are we creating long-term challenges or inconsistencies if this is implemented in JobSet and we later also have support in Job/StatefulSet. Maybe it would be worth thinking about what that would look like?

As for the question around CI and unstable APIs, we are planning to get DRA to GA in the next release. That should make it pretty straightforward to test with DRA, although we need to handle the versioning for a few releases. And with the v1beta2 API that was added in 1.33, we don't expect any significant changes between v1beta2 and v1.

keps/762-ResourceClaimTemplate/README.md
acpana: Added a note about using a validating webhook.

keps/762-ResourceClaimTemplate/README.md
acpana: added types spell out.

acpana:
hey @kannon92 @andreyvelich are you the right folks to sign off on this? I think y'all had a fair amount of feedback here (in addition to john)

kannon92:
I am having a hard time following DRA and this work.

Have you implemented this idea? It seems there is a lot of work called this out as a motivating use case. Is it actually possible to implement this with current DRA?

I'd love a POC and a demo to show that this works in reality before we commit to design/implementation.

johnbelamaric:
I am wondering if we should close this. Our current thinking is to instead tie RC/RCT support to the `Workload` resource being developed as part of https://github.com/kubernetes/enhancements/issues/4671

kannon92:
+1 on that
/close

Thank you for your work @acpana!

PR #863, opened 2025-04-03
Enable TLS metrics

### LINKED ISSUE #793: Secure metrics with Cert manager

Right now, our service monitor template assumes no tls options.

We want to use cert manager and we would have to use tls no verify.

Kubebuilder has a series of examples on how to use cert manager to secure your metrics endpoint and require TLS. As of now, we at least have https on our metrics endpoint but it is insecure.

I think one needs to expose a few metric options in our config.

**Why is this needed**:

Secure metrics and allows one to use certificates to manage metrics.

atiratree:
#### What type of PR is this?

/kind feature

Add an option `PROMETHEUS WITH CERTMANAGER` to use a certificate generated by the cert manager in metrics serving instead of a localhost self-signed certificate. Optionally another certifiicate can be configured in `metrics.certDir` when `PROMETHEUS` is used. ServiceMonitor can then be secured with this certificate CA and  `insecureSkipVerify: false` can now be used when collecting metrics.

Fixes #793 

```release-note
Add an option `PROMETHEUS WITH CERTMANAGER` to use a certificate generated by the cert manager in metrics serving instead of a localhost self-signed certificate. Optionally another certifiicate can be configured in `metrics.certDir` when `PROMETHEUS` is used. ServiceMonitor can then be secured with this certificate CA and  `insecureSkipVerify: false` can now be used when collecting metrics. Prometheus support in helm can now also be enabled.
```

### DIFF
--- a/api/config/v1alpha1/configuration_types.go
+++ b/api/config/v1alpha1/configuration_types.go
@@ -83,6 +83,12 @@ type ControllerMetrics struct {
 	// It can be set to "0" to disable the metrics serving.
 	// +optional
 	BindAddress string `json:"bindAddress,omitempty"`
+
+	// CertDir is the directory that contains the server key and certificate.
+	// If not set, a self-signed certificate will be used. The server key and
+	// certificate must be named tls.key and tls.crt, respectively.
+	// +optional
+	CertDir string `json:"certDir,omitempty"`
 }
 
 // ControllerHealth defines the health configs.
--- a/charts/jobset/templates/certmanager/_helpers.tpl
+++ b/charts/jobset/templates/certmanager/_helpers.tpl
@@ -28,3 +28,11 @@ Create the name of the jobset webhook certificate.
 {{- define "jobset.certManager.certificate.name" -}}
 {{ include "jobset.name" . }}-cert
 {{- end -}}
+
+
+{{/*
+Create the name of the jobset metrics certificate.
+*/}}
+{{- define "jobset.certManager.metricsCertificate.name" -}}
+{{ include "jobset.name" . }}-metrics-cert
+{{- end -}}
--- a/charts/jobset/templates/controller/deployment.yaml
+++ b/charts/jobset/templates/controller/deployment.yaml
@@ -62,7 +62,7 @@ spec:
         - name: webhook-cert
           mountPath: /tmp/k8s-webhook-server/serving-certs
           readOnly: true
-        {{- if and .Values.prometheus.enable .Values.certManager.enable }}
+        {{- if .Values.certManager.enable }}
         - mountPath: /tmp/k8s-metrics-server/serving-certs
           name: metrics-certs
           readOnly: true
@@ -116,7 +116,7 @@ spec:
         secret:
           secretName: {{ include "jobset.webhook.secret.name" . }}
           defaultMode: 420
-      {{- if and .Values.prometheus.enable .Values.certManager.enable }}
+      {{- if .Values.certManager.enable }}
       - name: metrics-certs
         secret:
           defaultMode: 420
--- a/charts/jobset/templates/controller/deployment.yaml
+++ b/charts/jobset/templates/controller/deployment.yaml
@@ -62,6 +62,11 @@ spec:
         - name: webhook-cert
           mountPath: /tmp/k8s-webhook-server/serving-certs
           readOnly: true
+        {{- if and .Values.prometheus.enable .Values.certManager.enable }}
+        - mountPath: /tmp/k8s-metrics-server/serving-certs
+          name: metrics-certs
+          readOnly: true
+        {{- end }}
         ports:
         - name: health-probe
           containerPort: 8081
@@ -111,6 +116,12 @@ spec:
         secret:
           secretName: {{ include "jobset.webhook.secret.name" . }}
           defaultMode: 420
+      {{- if and .Values.prometheus.enable .Values.certManager.enable }}
+      - name: metrics-certs
+        secret:
+          defaultMode: 420
+          secretName: {{ include "jobset.metrics.secret.name" . }}
+      {{- end }}
       {{- with .Values.controller.nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/charts/jobset/templates/controller/deployment.yaml
+++ b/charts/jobset/templates/controller/deployment.yaml
@@ -74,11 +74,9 @@ spec:
         - name: webhook
           containerPort: 9443
           protocol: TCP
-        {{- if .Values.prometheus.enable }}
         - name: metrics
           containerPort: 8443
           protocol: TCP
-        {{- end }}
         livenessProbe:
           httpGet:
             port: 8081
--- a/charts/jobset/templates/prometheus/service_monitor.yaml
+++ b/charts/jobset/templates/prometheus/service_monitor.yaml
@@ -30,5 +30,27 @@ spec:
     matchLabels:
       {{- include "jobset.controller.selectorLabels" . | nindent 6 }}
   endpoints:
-  - port: metrics
+  - port: https
+    path: /metrics
+    scheme: https
+    bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+{{- if .Values.certManager.enable }}
+    tlsConfig:
+      ca:
+        secret:
+          key: ca.crt
+          name: {{ include "jobset.metrics.secret.name" . }}
+      cert:
+        secret:
+          key: tls.crt
+          name: {{ include "jobset.metrics.secret.name" . }}
+      insecureSkipVerify: false
+      keySecret:
+        key: tls.key
+        name: {{ include "jobset.metrics.secret.name" . }}
+      serverName: {{ include "jobset.metrics.service.name" . }}.{{ .Release.Namespace }}.svc
+{{- else }}
+    tlsConfig:
+      insecureSkipVerify: true
+{{- end }}
 {{- end }}
--- a/charts/jobset/templates/prometheus/service_monitor.yaml
+++ b/charts/jobset/templates/prometheus/service_monitor.yaml
@@ -43,11 +43,11 @@ spec:
       cert:
         secret:
           key: tls.crt
-          name: {{ include "jobset.metrics.secret.name" . }}
+          name: {{ .Values.prometheus.prometheusClientCertSecretName }}
       insecureSkipVerify: false
       keySecret:
         key: tls.key
-        name: {{ include "jobset.metrics.secret.name" . }}
+        name: {{ .Values.prometheus.prometheusClientCertSecretName }}
       serverName: {{ include "jobset.metrics.service.name" . }}.{{ .Release.Namespace }}.svc
 {{- else }}
     tlsConfig:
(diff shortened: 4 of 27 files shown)


config/default/kustomization.yaml
```diff
@@ -42,6 +42,11 @@ patchesStrategicMerge:
 # 'CERTMANAGER' needs to be enabled to use ca injection
 #- webhookcainjection_patch.yaml
 
+#  Uncomment the patches line if you enable Metrics and CertManager
```
atiratree, line 45: Bunch of deprecated fields like patchesStrategicMerge and vars are currently used by the kustomize and should be later updated.

config/default/certmanager_metrics_manager_patch.yaml
```diff
@@ -0,0 +1,19 @@
...
+  template:
+    spec:
+      containers:
+        - name: manager
+          volumeMounts:
+            - mountPath: /tmp/k8s-metrics-server/metrics-certs
+              name: metrics-certs
+              readOnly: true
+      volumes:
+        - name: metrics-certs
```
ardaguclu, line 16: `metrics-certs` name in Certificate will be prefixed to `lws-metrics-certs`. Does the prefixing mechanism will prepend `lws` to here as well?

config/default/kustomization.yaml
ardaguclu: yeah, agree with you.

config/components/prometheus/monitor_tls_patch.yaml
```diff
@@ -0,0 +1,27 @@
...
+      tlsConfig:
+        insecureSkipVerify: false
+        serverName: $(METRICS_SERVICE_NAME).$(METRICS_SERVICE_NAMESPACE).svc
+        ca:
+          secret:
+            name: metrics-server-cert
+            key: ca.crt
+        cert:
+          secret:
+            name: $(CLIENT_CERT_NAME)
```
ardaguclu, line 23: Where is the default value defined?

pkg/config/config.go
```diff
@@ -37,6 +38,17 @@ func addTo(o *ctrl.Options, cfg *configapi.Configuration) {
 		o.Metrics.BindAddress = cfg.Metrics.BindAddress
 	}
 
+	internalCertManagementEnabled := ptr.Deref(cfg.InternalCertManagement.Enable, false)
+	if o.Metrics.CertDir == "" && (cfg.Metrics.CertDir != "" || !internalCertManagementEnabled) {
```
ardaguclu, line 42: internal cert management is enabled by default and nobody creates or injects these certificates (tls.crt, tls.key). I'd expect that CI should fail by throwing `certificates not found`, etc. But it didn't fail.

ardaguclu (commented):
I think, you need to configure helm charts?

api/config/v1alpha1/configuration_types.go
```diff
@@ -83,6 +83,12 @@ type ControllerMetrics struct {
 	// It can be set to "0" to disable the metrics serving.
 	// +optional
 	BindAddress string `json:"bindAddress,omitempty"`
+
+	// CertDir is the directory that contains the server key and certificate.
```
kannon92, line 87: For kueue, we weren't really sure if there was any need to have this Cert configurable.

Any reason why we can't just default to a hardcoded location?

ardaguclu: If we mount the default hardcoded locations `/tmp/k8s-metrics-server/serving-certs`, `tls.crt`, `tls.key`, there won't be any code changes needed in main.go.

kannon92: I think there would be one minor change in setting the right tlsOpts for metrics server. No having this CertDir does simply the code.

But if there is a reason for it I think leaving it in is fine.

kannon92:
Looks like metrics were never functionally with helm.. :(

Can you still look into enabling them as you are doing this?

config/default/certmanager_metrics_manager_patch.yaml
atiratree: nope, no prefix here

config/components/prometheus/monitor_tls_patch.yaml
atiratree: I have added documentation to this file and to kustomize.yaml on what to set. There are many manual steps already and I believe it is better for the user to decide how the prometheus authenticates.

Similar approach is done in the helm charts.

pkg/config/config.go
atiratree: self signed certificate is generated by the metrics server instead

api/config/v1alpha1/configuration_types.go
atiratree: In kustomize and helm we offer prometheus support without cert manager. This gives the users a way to inject their certificates for the metrics server to the deployment.

We also offer the same thing for the webhook.

atiratree:
@kannon92 @ardaguclu Yes, I was missing helm and there were so many things broken that it took me a fw days to get it working. 

Now the metrics collection works - tested with cert manager and prometheus - with both helm and kustomize.

The commits are all over the place, but a bit interwoven, so not sure if it is a good idea to split the the PR into multiple ones. Please let me know if you have any thoughts on this.

pkg/config/config.go
kannon92, line 42: Could we default to /tmp/k8s-metrics-server if internalCertManagement is false?

That way we can remove some of this setting and checking for nil?

kannon92: forget this. I thought I was looking elsewhere.

pkg/config/config_test.go
```diff
@@ -141,6 +141,28 @@ clientConnection:
 		t.Fatal(err)
 	}
 
+	metricsCertConfig := filepath.Join(tmpDir, "metrics-cert.yaml")
```
kannon92, line 144: Can we have a test where internalCertManagement is false and you specify a different value for certDir?

charts/jobset/templates/certmanager/_helpers.tpl
```diff
@@ -28,3 +28,11 @@ Create the name of the jobset webhook certificate.
 {{- define "jobset.certManager.certificate.name" -}}
 {{ include "jobset.name" . }}-cert
 {{- end -}}
+
+
+{{/*
+Create the name of the jobset metrics certificate.
+*/}}
+{{- define "jobset.certManager.metricscertificate.name" -}}
```
kannon92, line 36:
```suggestion
{{- define "jobset.certManager.metricsCertificate.name" -}}
```

charts/jobset/templates/controller/deployment.yaml
```diff
@@ -62,18 +62,21 @@ spec:
         - name: webhook-cert
           mountPath: /tmp/k8s-webhook-server/serving-certs
           readOnly: true
+        {{- if .Values.certManager.enable }}
+        - mountPath: /tmp/k8s-metrics-server/metrics-certs
```
kannon92, line 66: Would this work if someone specifies a different location than this in CertDir?

charts/jobset/templates/prometheus/service_monitor.yaml
```diff
@@ -30,5 +30,24 @@ spec:
     matchLabels:
       {{- include "jobset.controller.selectorLabels" . | nindent 6 }}
   endpoints:
-  - port: metrics
+  - port: https
+{{- if .Values.certManager.enable }}
```
ardaguclu, line 34: I think, there should be `else` if cert manager is not enabled and it should be `insecureSkipVerify: true`

pkg/config/config.go
atiratree: btw, I had an incorrect reference there, changed to `/tmp/k8s-metrics-server/serving-certs`

pkg/config/config_test.go
atiratree: +1 added a `metrics cert config with external certs` test

charts/jobset/templates/controller/deployment.yaml
atiratree: no it would not, they would have to modify the deployment

charts/jobset/templates/prometheus/service_monitor.yaml
atiratree: +1, falling back to insecure

atiratree:
> Looks like our prometheus tests caught something!

@kannon92 I think it is useful to have the service present even without the prometheus. I have updated the `service_test.yaml` to reflect that.

charts/jobset/templates/controller/deployment.yaml
kannon92: So we should maybe mention this in our documentation.

Otherwise I worry this will be a support issue.

kannon92:
just one item.

/approve

I leave final lgtm for @ardaguclu

pkg/config/config.go
ardaguclu: I'm not sure about the logic in here. `o.Metrics.CertDir == ""` must be true, but we are also checking `cfg.Metrics.CertDir != ""` (this is always true already?)

Isn't this simpler?
```go
if !internalCertManagementEnabled {
	o.Metrics.CertName = "tls.crt"
	o.Metrics.KeyName = "tls.key"
        o.Metrics.CertDir = "/tmp/k8s-metrics-server/serving-certs"
        if cfg.Metrics.CertDir != "" {
           o.Metrics.CertDir = cfg.Metrics.CertDir
        }
}
```

config/components/prometheus/monitor_tls_patch.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: controller-manager-metrics-monitor
+  namespace: system
+spec:
+  endpoints:
+    - path: /metrics
+      port: https
+      scheme: https
+      bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+      tlsConfig:
+        insecureSkipVerify: false
```
ardaguclu, line 22: I think, this should be `insecureSkipVerify: true` and there will be a "op: replace" command to replace with these values when cert-manager is enabled, in order to be compatible with https://github.com/kubernetes-sigs/jobset/pull/863/files#diff-89204119394a9cad711f1fdf90f9dd71be4f7ae8ef5bca4cd545661002a419bdR37-R55

ardaguclu (commented):
Overall looks good to me. Dropped a few comments.

pkg/config/config.go
atiratree: - `o.Metrics.CertDir == ""` this is just a defense against overwriting the value in case it is already defined
-  `cfg.Metrics.CertDir` we can still pass empty certDir to have the self signed certs be generated by the jobset controller unless you disable the cert management

I have tried to simplify the branching a bit.

charts/jobset/templates/controller/deployment.yaml
atiratree: This is the same approach we use with the webhook cert dir. Do we have to support such changes (i.e. custom deployments)? And we have documentation in the config API.

config/components/prometheus/monitor_tls_patch.yaml
atiratree: This is applied only when used in `PROMETHEUS WITH CERTMANAGER` configuration. 

We do use `insecureSkipVerify: true` otherwise in https://github.com/kubernetes-sigs/jobset/blob/789481a770af28fd94bf7724c54bd11d85759aea/config/components/prometheus/monitor.yaml#L23

PR #868, opened 2025-04-07
fix: certificate fails to find issuer

### LINKED ISSUE #857: Helm chart certificate fails to find issuer
**What happened**:
I deployed the JobSet helm chart v0.8.1 to my cluster with the value `certManager.enable=true` and the certificate failed to reference the provided issuer due to the group setting in the issuerRef.  The group setting at https://github.com/kubernetes-sigs/jobset/blob/v0.8.1/charts/jobset/templates/certmanager/certificate.yaml#L31 includes the API version of cert-manager.io and that prevents the reference from resolving.  Removing the `/v1` part of the group setting allows the issuer to be referenced.  I'm not sure how to configure the group setting to include a version but appending it seems to not work.

The helm chart should deploy a certificate that issues properly and allows the JobSet controller to start.

1. Create a Kubernetes cluster
2. Deploy cert-manager v1.17.1
3. Deploy the JobSet helm chart with the command `helm install jobset oci://registry.k8s.io/jobset/charts/jobset --namespace jobset-system --create-namespace --version v0.8.1 --set certManager.enable=true`

**Environment**:
- Kubernetes version (use `kubectl version`): v1.31.7
- JobSet version (use `git describe --tags --dirty --always`): v0.8.1
- Cloud provider or hardware configuration: Minikube/Bare Metal
- Install tools: Helm v3.17.2
- Others:

ChenYi015:
#### What type of PR is this?

/kind bug

CertManager should work as expected when setting `certManager.enable` to `true`. 

Fixes #857

```release-note
NONE
```

### DIFF
--- a/charts/jobset/templates/certmanager/certificate.yaml
+++ b/charts/jobset/templates/certmanager/certificate.yaml
@@ -28,7 +28,7 @@ spec:
   {{- with .Values.certManager.issuerRef }}
     {{ toYaml . | nindent 4 }}
   {{- else }}
-    group: cert-manager.io/v1
+    group: cert-manager.io
     kind: Issuer
     name: {{ include "jobset.certManager.issuer.name" . }}
   {{- end }}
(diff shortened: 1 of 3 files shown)


kannon92:
Can you update the release note?

This is a user facing bug so it'd be worth calling this out as such.

kannon92:
/hold

/lgtm
/approve

I did pull and test this. Once you update the release note I think we can see about including this.

ChenYi015:
> Can you update the release note?

Done.

atiratree:
yup lgtm, just rebased https://github.com/kubernetes-sigs/jobset/pull/863 on it

PR #873, opened 2025-04-10
chore: add integration tests for Helm chart

ChenYi015:
#### What type of PR is this?

/kind regression

Add integration tests for testing Helm chart (e.g. helm lint, helm unittest, helm docs).

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,9 +131,12 @@ toc-verify:
 	./hack/verify-toc.sh
 
 .PHONY: helm-verify
-helm-verify: helm-unittest helm-lint
-	${HELM} template charts/jobset
-
+helm-verify: helm-unittest helm-lint helm-docs
+	@if ! git diff --quiet -- $(JOBSET_CHART_DIR)/README.md; then \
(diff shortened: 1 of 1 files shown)


kannon92:
/hold
I don't think this is needed.

We have added this to `verify` step which already runs on our PRs.

https://github.com/kubernetes-sigs/jobset/blob/main/Makefile#L149

kannon92:
> This workflow contains extra integration tests (e.g. check whether Helm chart README is updated), use [ct](https://github.com/helm/chart-testing) lint rather than `helm lint` to lint the charts. Besides, we can also extend this workflow to use ct to test whether the Helm chart can be installed successfully within a kind cluster.

This is all great!

But can we add it to an existing job that runs with Prow.  

I'd like to not introduce github actions and have to manage two infra for running CI.

ChenYi015:
> But can we add it to an existing job that runs with Prow.
>
> I'd like to not introduce github actions and have to manage two infra for running CI.

Have updated Makefile target `make verify-helm` to reuse the existed Prow CI.

PR #892, opened 2025-04-24
Update kustomize version to v5.2.1

ardaguclu:
#### What type of PR is this?
/kind cleanup

JobSet uses kustomize `v3.8.7` version which is old. This PR updates the kustomize version to `v5.2.1` with the required changes to at least be compatible with LWS (ref: https://github.com/kubernetes-sigs/lws/blob/80af0086dd54432509eea06a80eeefe5714ec8b2/Makefile#L259)

```release-note
None
```


atiratree (commented):
Thanks!

config/components/crd/kustomization.yaml
```diff
@@ -5,15 +5,15 @@ resources:
 - bases/jobset.x-k8s.io_jobsets.yaml
 #+kubebuilder:scaffold:crdkustomizeresource
 
-patchesStrategicMerge:
```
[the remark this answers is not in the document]
ardaguclu: Thanks. Updated.

config/default/kustomization.yaml
```diff
@@ -27,67 +27,184 @@ resources:
...
-#    version: v1
-#    name: controller-manager-metrics-service
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: METRICS_SERVICE_NAME
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: controller-manager-metrics-service
+#replacements:
```
kannon92, line 54: Our CI does not test this flow.

Can you do a manual test to confirm this works?

ardaguclu: After uncommented the all cert-manager related configurations, I ran `kustomize build` and saw that the output is expected. If you still want me to test it on a live cluster, I can do it and let you know the results?.

kannon92: Yea, I think having an e2e test execute cert manager would be useful but we don't have that yet so I just want to make sure these configs will actually generate a functionally cert manager + jobset deployment.

ardaguclu: Cert-manager enabled e2e testing (i.e. pull-jobset-test-e2e-main-cert-manager) is on my TODO list in both JobSet and LWS. I think, we can execute the same e2e test with the only difference that cert-manager is enabled instead of internal cert management. But I don't know where the configuration locates in upstream?.

kannon92: I don't fully understand.

https://github.com/kubernetes-sigs/jobset/blob/main/config/components/manager/controller_manager_config.yaml#L7

I think you could toggle that to false and that should use CertManager for kustomize.

ardaguclu: Would it make sense if we have a new CI job that runs the same e2e tests but instead of using the internal cert management, it installs jobset with cert manager and all related patches?

kannon92: yea, that was something we did for Kueue.

ardaguclu: Thanks. I will test this pr tomorrow and let you know the results. Independently I am going to work on a pr for the new job.

ardaguclu: @kannon92 I've tested this on a live cluster. It works without any issue.

PR #878, opened 2025-04-13
feat: support multiple items in DependsOn API

Electronic-Waste:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it:

This PR adds the support for multiple items in DependsOn API, so that we can unblock the implementation of Kubeflow LLM Trainer V2.

/cc @andreyvelich @tenzen-y @kannon92 @ahg-g 

#### Which issue(s) this PR fixes:

Fixes #874

#### Does this PR introduce a user-facing change?

```release-note

```

PR #895, opened 2025-04-27
feat: add example for multiple DependsOn items.

Electronic-Waste:
#### What type of PR is this?

/kind feature

This is an example for multiple DependsOn items, as described in https://github.com/kubernetes-sigs/jobset/pull/878#issuecomment-2832273164

/cc @andreyvelich @tenzen-y @kannon92 @ahg-g 

Fixes #

```release-note
NONE
```

### DIFF
--- a/site/static/examples/depends-on/multiple-depends-on.yaml
+++ b/site/static/examples/depends-on/multiple-depends-on.yaml
@@ -1,0 +1,51 @@
+# This example creates three ReplicatedJobs.
+# Then node ReplicatedJob waits until the dataset-intializer and model-initializer ReplicatedJob are Complete.
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: multiple-depends-on
+spec:
+  replicatedJobs:
+    - name: dataset-initializer
+      template:
+        spec:
+          template:
+            spec:
+              containers:
+                - name: initializer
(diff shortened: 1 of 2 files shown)


site/static/examples/depends-on/multiple-depends-on.yaml
```diff
@@ -0,0 +1,51 @@
+# This example creates three ReplicatedJobs.
```
kannon92, line 1: Example looks good. I think it would be worth adding this to site also.

Electronic-Waste: Could you please kindly point to me where the site is?

Electronic-Waste: Do I need to change the url from the specific file to the `depends-on` dir? Or rephrase sentences concerned with DependsOn?

andreyvelich: @Electronic-Waste Maybe you could just link your example in additional to existing example to say that your example shows how to use DependsOn API to wait for completion of two previous ReplicatedJob

andreyvelich (commented):
Thank you @Electronic-Waste!
/lgtm
/assign @kannon92

kannon92 (commented):
/lgtm

PR #901, opened 2025-05-03
Use correct links to examples

lchrzaszcz:
#### What type of PR is this?

/kind documentation

All links to examples in documentation used path like: `./site/static/examples/simple/<yaml-file>`, however static files are exposed via `./examples/simple/<yaml-file>`.

E.g.:
- https://jobset.sigs.k8s.io/docs/overview/site/static/examples/simple/exclusive-placement.yaml - Example linked in docs that results in 404.
- https://jobset.sigs.k8s.io/examples/simple/exclusive-placement.yaml - Actual URL to the aforementioned example. This results in YAML file being downloaded.

Given that a correct link to a yaml might result in file being downloaded instead of displaying it in browser which breaks the flow of reading documentation, I have decided to follow other example URLs in documentation that lead to Github instead.

Fixes #

```release-note
NONE
```


site/content/en/docs/overview/_index.md
```diff
@@ -21,19 +21,19 @@ Take a look at the [concepts](../concepts/) page for a brief description of how
...
+- **Automatic headless service configuration and lifecycle management**: ML and HPC frameworks require a stable network endpoint for each worker in the distributed workload, and since pod IPs are dynamically assigned and can change between restarts, stable pod hostnames are required for distributed training on k8s, By default, JobSet uses [IndexedJobs](https://kubernetes.io/blog/2021/04/19/introducing-indexed-jobs/) to establish stable pod hostnames, and does automatic configuration and lifecycle management of the headless service to trigger DNS record creations and establish network connectivity via pod hostnames. These networking configurations are defaulted automatically to enable stable network endpoints and pod-to-pod communication via hostnames; however, they can be customized in the JobSet spec: see this [example](https://github.com/kubernetes-sigs/jobset/blob/main/site/static/examples/simple/jobset-with-network.yaml) of using a custom subdomain your JobSet's network configuration.
```
gabesaba, line 24: I think it would be worth investigating why ./site is not working for these static files, since (assuming that it works like Kueue) the website is versioned, and referencing `main` could result in some inconsistencies/skew as the site is referencing repo at a certain commit, while the static files are always `main`

lchrzaszcz: It does work, but it is exposed on a different path. We would need to change the path to be something like: `/examples/simple/exclusive-placement.yaml `. The drawback is that the example would be downloaded as a yaml file instead of being displayed, but it would be for sure consistent.

We could also prepare a separate page for each example (or one page with all examples and anchors). Use `{{< include "examples/simple/example-file.yaml" "yaml" >}}` there to display yaml in nice format and link to that page from other places in docs.

So I guess we just need to make a decision what we prefer:
- Display file in Github but risk discrepancy between site version and main branch
- Provide direct link that downloads a file
- Prepare "example" pages to have consistent version of examples and display them in a user-friendly way.

kannon92: > We could also prepare a separate page for each example (or one page with all examples and anchors). Use {{< include "examples/simple/example-file.yaml" "yaml" >}} there to display yaml in nice format and link to that page from other places in docs.

I'd prefer to go this route eventually.

But I would like the links valid for now.

WDYT @ahg-g?

PR #920, opened 2025-06-17
Implement RecreateJob failure policy action

carreter:
#### What type of PR is this?

/kind feature
/kind api-change

#### What this PR does / why we need it:
In some instances, failure of a specific Job does not necessitate a restart of the entire JobSet; rather, recreating the Job or ReplicatedJob that failed is sufficient for recovery. [KEP #262](https://github.com/kubernetes-sigs/jobset/tree/main/keps/262-ConfigurableFailurePolicy#future-work) suggests adding a RecreateJob failure policy action to support this use case.

#### Which issue(s) this PR fixes:

Fixes #918 

#### Does this PR introduce a user-facing change?

```release-note
Add RecreateJob and RecreateReplicatedJob to FailurePolicyActions
```

PR #925, opened 2025-06-27
Update failure policy KEP with proposed RecreateJob behavior

### LINKED ISSUE #918: Feature proposal - Implement RecreateJob failure policy action
## Current behavior
Currently, the only available actions upon failure of a child Job of a JobSet are to:
- fail the entire JobSet (`FailJobSet`)
- restart the entire JobSet (`RestartJobSet`)
- restart the entire JobSet without incrementing the reset attempt counter (`RestartJobSetAndIgnoreMaxRestarts`)

## Desired behavior
In some instances, failure of a specific Job does not necessitate a restart of the entire JobSet; rather, restarting the Job that failed is sufficient for recovery. [KEP #262](https://github.com/kubernetes-sigs/jobset/tree/main/keps/262-ConfigurableFailurePolicy#future-work) suggests adding a `RestartJob`  failure policy action to support this use case.

## How does this relate to #876 ?
In place Pod restarts (see #876) accomplish a similar purpose of reducing JobSet recovery time on failure, but as it is currently being proposed (@GiuseppeTT please correct me if I'm wrong) it still restarts _all_ of the child Pods of a JobSet even upon failure of a single Pod.

EDIT: New policy will be name `RecreateJob` rather than `RestartJob` to better indicate behavior.

(cc @imreddy13 @kannon92 @GiuseppeTT )

carreter:
#### What type of PR is this?

/kind documentation
/kind api-change

KEP proposing addition of new RecreateJob and RecreateReplicatedJob failure policies, see #918 and #920 

Fixes #918

```release-note
NONE
```

### DIFF
--- a/keps/262-ConfigurableFailurePolicy/README.md
+++ b/keps/262-ConfigurableFailurePolicy/README.md
@@ -278,11 +278,11 @@ spec:
                 python3 train.py
 ```
 
-#### Story 5: Recreating replicated jobs on failure rather than failing JobSet
+#### Story 5: Recreating individual jobs on failure rather than failing JobSet (`RecreateJob`)
 
-As a user, I have a JobSet with 2 replicated jobs: one which runs distributed training processes across a pool of GPU
-nodes, and one which runs the driver/coordinator on a CPU pool. If a child job of the GPU worker ReplicatedJob crashes, I just want to recreate the GPU workers and not the driver, then resume training from the latest checkpoint. However, if
-the driver crashes, I want to restart the entire JobSet, then resume training from the latest checkpoint.
+If it is possible for individual worker Jobs within a ReplicatedJob to be restarted independently on failure
+without requiring a full restart of their parent ReplicatedJob or the entire JobSet, the `RecreateJob`
+failure policy can be used.
 
 **Example Failure Policy configuration for this use case**:
 
@@ -290,20 +290,18 @@ the driver crashes, I want to restart the entire JobSet, then resume training fr
 apiVersion: jobset.x-k8s.io/v1alpha2
 kind: JobSet
 metadata:
-  name: recreate-replicated-job-example
-  annotations:
-    alpha.jobset.sigs.k8s.io/exclusive-topology: {{topologyDomain}} # 1:1 job replica to topology domain assignment
+  name: recreate-job-example
 spec:
-  # Failure Policy to restart the child jobs of the target ReplicatedJob (gpu-workers) if any fail, but fall
-  # back to the default behavior of restarting the entire JobSet if the driver fails.
+  # Failure Policy to restart individual jobs of the target ReplicatedJob (recoverable-workers) if they fail,
+  # without restarting the entire JobSet or other jobs in recoverable-workers.
   failurePolicy:
     rules:
-    - action: RecreateReplicatedJob
+    - action: RecreateJob
       targetReplicatedJobs:
-      - gpu-workers
+      - recoverable-workers
     maxRestarts: 10
   replicatedJobs:
-  - name: driver
+  - name: recoverable-workers
     replicas: 1
     template:
       spec:
@@ -317,24 +315,7 @@ spec:
             - name: main
               image: python:3.10
               command: ["..."]
-  - name: gpu-workers
-    replicas: 4 # number of node pools
-    template:
-      spec:
-        parallelism: 2
-        completions: 2
-        backoffLimit: 0
-        template:
-          spec:
-            containers:
-            - name: main
-              image: pytorch:latest
-              command: ["..."]
-            resources:
-              limits:
-                nvidia.com/gpu: 1
-```
-
+``
 
 ### Notes/Constraints/Caveats (Optional)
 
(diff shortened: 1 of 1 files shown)


keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -322,13 +379,20 @@ const (
 
   // Don't count the failure against maxRestarts.
   RestartJobSetAndIgnoreMaxRestarts FailurePolicyAction = "RestartJobSetAndIgnoreMaxRestarts"
+
+  // Recreate the failed Job without restarting the entire JobSet.
+  RecreateJob FailurePolicyAction = "RecreateJob"
```
kannon92, line 365: Thanks for adding the KEP.

I would still like to see a use case for RecreateJob.

imreddy13: @kannon92 the use case for RecreateJob is we do not need the whole JobSet to be restarted, but only the corresponding indexed job that the failed pod belonged to. We have a customer doing this today using (abusing?) the backoffLimit at the replicated job level to achieve this by setting it to something arbitrarily high. Their JobSet configuration is such that they do not need all the jobs to restart if a single pod fails i.e. the replicated jobs are independent.

imreddy13: We don't have a concrete ask for recreating the entire replicated job today but I could see it being useful based on how the JobSet is configured. Wdyt @ahg-g?

kannon92: I believe you. I would like to see a story 6 that at least demonstrates using this option also.

ahg-g: > We don't have a concrete ask for recreating the entire replicated job today but I could see it being useful based on how the JobSet is configured. Wdyt @ahg-g?

We can hold off adding recreate replicatedJob, but I wouldn't be surprised to see a request coming for it; for example, one may create two replicatedJobs, one for the control plane pods (e.g., driver pod) and one for the workers, and I can see a need to only restart the workers when any worker fail, but not the driver pod.

> the use case for RecreateJob is we do not need the whole JobSet to be restarted

Another case is Reinforcement learning; the jobset would include two replicatedJobs, one for the training workers and one for a group of multi-node inference servers; for the latter, each multi-host server is independent, when a pod fails within the multi-node server, we only need to restart that server, not the whole RL job.

kannon92: I'm a little confused.

What use case do we have asks for?

RecreateJob or RecreateReplicatedJob?

I mostly just want to see documentation as a user story on why someone would want to use each action.

carreter: Hey @kannon92 , sorry for the delay on getting back to you on this.

Current use case we have an ask for is RecreateJob. As @imreddy13 mentioned the customer is currently setting the backoffLimit for jobs to an arbitrarily high value to achieve similar behavior of restarting a single Job without restarting the whole JobSet.

We don't have a specific ask for RecreateReplicatedJob yet, but I will add a user story following the case @ahg-g envisioned where there is a ReplicatedJob for driver pods and one for worker pods, and that failures in worker pods should trigger a recreate of all workers but not the full JobSet.

ahg-g: I would limit this KEP to RecreateJob only, discussing RecreateReplicatedJob should be a separate KEP.

carreter: Ok, I'll split it into two PRs then!

ahg-g: My point is to just focus on RecreateJob, we can do RecreateReplicatedJob only when we have a concrete ask for it, what I mentioned above was a hypothetical case.

carreter: Got it, thank you for clarifying!

carreter:
@kannon92 @ahg-g @imreddy13 @GiuseppeTT This KEP now only includes `RecreateJob` and is ready for review!

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -322,13 +360,20 @@ const (
...
+
+  // Recreate all Jobs in a ReplicatedJob if any of them fail, without
```
kannon92: I think we should revert changes around RecreateReplicatedJob and leave it as a future item.

carreter: Oh whoops missed this, will revert

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -278,6 +278,44 @@ spec:
...
+  # without restarting the entire JobSet or other jobs in recoverable-workers.
+  failurePolicy:
+    rules:
+    - action: RecreateJob
+      targetReplicatedJobs:
+      - recoverable-workers
+    maxRestarts: 10
+  replicatedJobs:
+  - name: recoverable-workers
+    replicas: 1
```
GiuseppeTT: `replicas` should be `> 1` for this action to make sense. Or it should have a second replicatedJob. I would go with `replicas: 2`

GiuseppeTT: For the implementation. It might make sense to update the JobSet webhook to add this validation. That is, require `sum(.replicatedJobs[*].replicas) > 1` if there is an action `RecreateJob`

carreter: Done.

keps/262-ConfigurableFailurePolicy/README.md
carreter: Reverted.

kannon92:
/approve

leaving lgtm for @GiuseppeTT.

GiuseppeTT (approved):
LGTM

PR #927, opened 2025-06-30
Fix infinite reconciliation when patching JobSets

### LINKED ISSUE #926: Infinite reconciliation when patching JobSets with SSA
**What happened**:

When patching a JobSet with SSA, the operation is not idem-potent and the JobSet generation keeps incrementing even if the patch is exactly the same and the JobSet spec does not change. 

This causes infinite reconciliation for controllers that rely on patch / SSA.

Patching JobSet should be idem-potent and the JobSet generation should not be incremented when applying the same patch multiple times.

Create a JobSet with:

```
kubectl apply --server-side -f site/static/examples/simple/success-policy.yaml
```

Then re-run the command multiple times.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.31.7
- JobSet version (use `git describe --tags --dirty --always`): v0.8.0-devel-134-gecc8784

astefanutti:
#### What type of PR is this?

/kind bug

This changes the JobSet mutating webhook so it only applies to CREATE operations.

Fixes #926

```release-note
Fix infinite reconciliation when patching JobSets
```


astefanutti:
/assign @kannon92 @ahg-g

Could you please have a look?

kannon92:
could we add a test verifies that this fixes your problem?

andreyvelich:
@astefanutti Will this bug block the Kueue integration with TrainJob when it tries to patch the running job ?

pkg/webhooks/jobset_webhook.go
```diff
@@ -75,7 +75,7 @@ var validOnJobFailureReasons = []string{
 	batchv1.JobReasonPodFailurePolicy,
 }
 
-//+kubebuilder:webhook:path=/mutate-jobset-x-k8s-io-v1alpha2-jobset,mutating=true,failurePolicy=fail,sideEffects=None,groups=jobset.x-k8s.io,resources=jobsets,verbs=create;update,versions=v1alpha2,name=mjobset.kb.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/mutate-jobset-x-k8s-io-v1alpha2-jobset,mutating=true,failurePolicy=fail,sideEffects=None,groups=jobset.x-k8s.io,resources=jobsets,verbs=create,versions=v1alpha2,name=mjobset.kb.io,admissionReviewVersions=v1
```
kannon92:
```suggestion
//+kubebuilder:webhook:path=/mutate-jobset-x-k8s-io-v1alpha2-jobset,mutating=true,failurePolicy=fail,sideEffects=None,groups=jobset.x-k8s.io,resources=jobsets,verbs=create,versions=v1alpha2,name=mjobset.kb.io,admissionReviewVersions=v1
```

pkg/webhooks/jobset_webhook.go
```diff
@@ -151,7 +151,7 @@ func (j *jobSetWebhook) Default(ctx context.Context, obj runtime.Object) error {
 	return nil
 }
 
-//+kubebuilder:webhook:path=/validate-jobset-x-k8s-io-v1alpha2-jobset,mutating=false,failurePolicy=fail,sideEffects=None,groups=jobset.x-k8s.io,resources=jobsets,verbs=create;update,versions=v1alpha2,name=vjobset.kb.io,admissionReviewVersions=v1
+// +kubebuilder:webhook:path=/validate-jobset-x-k8s-io-v1alpha2-jobset,mutating=false,failurePolicy=fail,sideEffects=None,groups=jobset.x-k8s.io,resources=jobsets,verbs=create;update,versions=v1alpha2,name=vjobset.kb.io,admissionReviewVersions=v1
```
kannon92:
```suggestion
//+kubebuilder:webhook:path=/validate-jobset-x-k8s-io-v1alpha2-jobset,mutating=false,failurePolicy=fail,sideEffects=None,groups=jobset.x-k8s.io,resources=jobsets,verbs=create;update,versions=v1alpha2,name=vjobset.kb.io,admissionReviewVersions=v1
```

config/components/webhook/manifests.yaml
```diff
@@ -20,7 +20,6 @@ webhooks:
     - v1alpha2
     operations:
     - CREATE
```
kannon92, line 22: We don't have any way to automate the chart so can you make sure these changes are on the chart also?

astefanutti:
> @astefanutti Will this bug block the Kueue integration with TrainJob when it tries to patch the running job ?

Kueue is also affected by this problem as it watches for JobSets, basically all these 3 components enter an infinite reconciliation loop that leads to the JobSet reaching 1000 generations in few minutes.

config/components/webhook/manifests.yaml
astefanutti: Updated, thanks.

kannon92:
> It looks like the integration tests should be enough since the problematic point is not tied to real cluster operation. @kannon92 Do you still want to add tests for real cluster (e2e)?

I want a test case confirming this behavior. Integration or e2e is fine with me.

kannon92:
Code LGTM. I'll wait for CI though.

/assign @ahg-g 

Curious on if there was any reason why we had an Update for mutating webhook.

astefanutti:
> I want a test case confirming this behavior. Integration or e2e is fine with me.

@kannon92 added an e2e test.

I confirm running the same e2e test on the main branch fails.

kannon92:
/lgtm

I reached out to @ahg-g about this to see if I am missing any context on why Update was added for mutating webhooks.

I'll leave approve for him

ahg-g:
Makes sense, we did the same for lws: https://github.com/kubernetes-sigs/lws/pull/347
/approve

Issue #930, opened 2025-07-02
Release v0.8.2


kannon92:
Aiming for week of July 14th.

## Release Checklist

- [x] All [OWNERS](https://github.com/kubernetes-sigs/jobset/blob/main/OWNERS) must LGTM the release proposal
- [x] Verify that the changelog in this issue is up-to-date
- [x] For major or minor releases (v$MAJ.$MIN.0), create a new release branch.
  - [x] an OWNER creates a vanilla release branch with
        `git branch release-$MAJ.$MIN main`
  - [x] An OWNER pushes the new release branch with
        `git push release-$MAJ.$MIN`
- [x] Update things like README, deployment templates, docs, configuration, test/e2e flags.
      Submit a PR against the release branch: 
- [x] An OWNER [prepares a draft release](https://github.com/kubernetes-sigs/jobset/releases)
  - [x] Write the change log into the draft release.
  - [x] Run
      `make artifacts IMAGE_REGISTRY=registry.k8s.io/jobset GIT_TAG=$VERSION`
      to generate the artifacts and upload the files in the `artifacts` folder
      to the draft release.
- [x] An OWNER creates a signed tag running
     `git tag -s $VERSION`
      and inserts the changelog into the tag description.
      To perform this step, you need [a PGP key registered on github](https://docs.github.com/en/authentication/managing-commit-signature-verification/checking-for-existing-gpg-keys).
- [x] An OWNER pushes the tag with
      `git push $VERSION`
  - Triggers prow to build and publish a staging container image
      `gcr.io/k8s-staging-jobset/jobset:$VERSION`
- [x] Submit a PR against [k8s.io](https://github.com/kubernetes/k8s.io), 
      updating `k8s.gcr.io/images/k8s-staging-jobset/images.yaml` to
      [promote the container images](https://github.com/kubernetes/k8s.io/tree/main/k8s.gcr.io#image-promoter)
      to production: https://github.com/kubernetes/k8s.io/pull/8293
- [x] Wait for the PR to be merged and verify that the image `registry.k8s.io/jobset/jobset:$VERSION` is available.
- [x] Wait for the PR to be merged and verify that the chart `registry.k8s.io/jobset/charts/jobset` is available. 
      Try `helm template oci://registry.k8s.io/jobset/charts/jobset --version=v$VERSION.
- [x] Publish the draft release prepared at the [Github releases page](https://github.com/kubernetes-sigs/jobset/releases).
[...]

PR #937, opened 2025-07-11
update release branch with latest patch for 0.8

kannon92:
#### What type of PR is this?
/kind cleanup

Part of #930 

Fixes #

Update site and README.md to reference new release

```release-note
NONE
```

### DIFF
--- a/charts/jobset/Chart.yaml
+++ b/charts/jobset/Chart.yaml
@@ -21,7 +21,7 @@ description: A Helm chart for deploying JobSet controller and webhook on Kuberne
 
 type: application
 
-version: 0.7.3
+version: 0.8.2
 
 keywords:
 - kubernetes


andreyvelich (commented):
Thanks @kannon92!
Are we going to update Helm charts in the followup PR ?
https://github.com/kubernetes-sigs/jobset/blob/main/charts/jobset/Chart.yaml#L24

kannon92:
> Thanks @kannon92! Are we going to update Helm charts in the followup PR ? https://github.com/kubernetes-sigs/jobset/blob/main/charts/jobset/Chart.yaml#L24

cc @GiuseppeTT 

This is the one thing we should before we do a release..

charts/jobset/Chart.yaml
```diff
@@ -21,7 +21,7 @@ description: A Helm chart for deploying JobSet controller and webhook on Kuberne
 
 type: application
 
-version: 0.7.3
```
kannon92, line 24: @tenzen-y do you know if this is actually used for our helm repository?

kannon92: Looks like our release branch didn't even change this but I think its fine for 0.8.1..

Issue #786, opened 2025-02-18
Flaky Test: JobSet Depends on


kannon92:
https://testgrid.k8s.io/sig-apps#pull-jobset-test-e2e-main-1-30

I've seen this failure twice on the helm chart PR.

```
JobSet when DependsOn is enabled on JobSet [It] trainer-node Job depends on launcher Job ready status
/home/prow/go/src/sigs.k8s.io/jobset/test/e2e/e2e_test.go:345
  [FAILED] Timed out after 600.001s.
  Expected
      <int32>: 0
  to equal
      <int32>: 1
  In [It] at: /home/prow/go/src/sigs.k8s.io/jobset/test/e2e/e2e_test.go:397 @ 02/18/25 03:53:02.09
```

cc @andreyvelich @tenzen-y

tenzen-y:
Just confirmation for prioritizing.
Is this blocker for next release?

kannon92:
Yea, let's call it that for now. Just want to make sure there isn't anything wrong with the functionality.

andreyvelich:
I was just running the same test locally and it was working fine for me:
```
$ ./bin/ginkgo --focus "trainer-node Job depends on launcher Job ready status" ./test/e2e/...

...
Ginkgo ran 1 suite in 28.074639959s
Test Suite Passed
```

@kannon92 Did you see the same errors in other PRs ?

kannon92:
You can see the test grid. It failed on the helm chart PR twice this week.

tenzen-y:
As I checked these test cases local, I faced the different `dependsOn` failures after performing E2E 115 times:
I guess that the order of `dependsOn` is not guaranteed.

```shell
JobSet when DependsOn is enabled on JobSet trainer-node Job depends on launcher Job ready status
/Users/s14554/go/src/sigs.k8s.io/jobset/test/e2e/e2e_test.go:345
  STEP: Create a JobSet with DependsOn @ 02/23/25 04:49:20.757
  STEP: Verify that only Launcher is created @ 02/23/25 04:49:20.766
  STEP: Wait for Launcher to be in Ready status @ 02/23/25 04:49:20.769
  STEP: Verify that Launcher and Trainer Job is created @ 02/23/25 04:49:32.636
  STEP: Wait for JobSet to be Completed @ 02/23/25 04:49:32.64
  STEP: checking jobset status is: Completed @ 02/23/25 04:49:32.641
[...]

PR #938, opened 2025-07-14
cleanup: e2e tests: Don't sleep in trainer jobs

tchap:
#### What type of PR is this?

/kind cleanup

When testing `dependsOn`, it's not necessary to `sleep 10` in the trainer jobs as that is only postponing the jobset to be complete. Good 30 seconds saved, perhaps 🙂 

None, but I improved things a little bit when I was checking out #786 

N/A

N/A

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -18,6 +18,7 @@ package e2e
 import (
 	"context"
 	"fmt"
+	"strconv"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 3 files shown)


test/e2e/e2e_test.go
```diff
@@ -291,7 +292,7 @@ var _ = ginkgo.Describe("JobSet", func() {
 			// Every ReplicatedJob runs 1 container to sleep for 10 seconds.
 			rJobModelInitializer := dependsOnTestReplicatedJob(ns, modelInitializerJob, numReplicas, nil, nil)
 			rJobDatasetInitializer := dependsOnTestReplicatedJob(ns, datasetInitializerJob, numReplicas, nil, nil)
-			rJobTrainer := dependsOnTestReplicatedJob(ns, trainerJob, numReplicas, nil,
+			rJobTrainer := dependsOnTestReplicatedJobWithSleepDuration(ns, trainerJob, numReplicas, nil, 0,
```
andreyvelich: If we don't have sleep, how we can make sure that ReplicatedJobs are executed in order ?

tchap: This is only about the trainer jobs, i.e the jobs no other jobs depend on. The tests are written so that you start the dependencies, then check only they are started, then wait for all the jobs to exist and the jobset to complete. Nobody really checks the status of the trainer jobs.

andreyvelich: Oh, that sounds good!

PR #939, opened 2025-07-15
e2e tests: Dump namespace content on failure

tchap:
#### What type of PR is this?

/kind feature

#### What this PR does / why we need it

Currently it's basically impossible to understand the state of the testing namespace after a test fails since the namespace is immediately deleted. This patch changes that to print selected namespace state before deleting it, the selected state being all JobSets and Jobs right now.

A disadvantage is that the output is generated in AfterEach whether the test fails or not. This only happens 9 times currently and it doesn't seem to be much of an overhead.

The output is shown only for the failing test, though.

To enable this extended output, set `JOBSET_E2E_TESTS_DUMP_NAMESPACE` environment variable to `true`. This is set by default in `hacks/e2e-tests.sh` so that it's visible on CI.

Makes it possible to get some insight into https://github.com/kubernetes-sigs/jobset/issues/786 

N/A

N/A


kannon92:
Thank you!
/lgtm

PR #944, opened 2025-07-21
Setup kubeClient to drop invalid field warnings

tchap:
#### What type of PR is this?

/kind cleanup

Logs are now full of `invalid field` warnings, which can be seen in the linked issue.

This patch drops all invalid field log entries by implementing a custom warning handler.

It's not really a bullet-proof solution, but it improves the situation.

Workaround for https://github.com/kubernetes-sigs/jobset/issues/141

There is a unit test, I also tested this manually.

```release-note
NONE
```


main.go
```diff
@@ -108,6 +112,16 @@ func main() {
...
 
 	kubeConfig := ctrl.GetConfigOrDie()
+	// Setup warning logging so that unknown field warnings are not logged.
+	// The next handler is the usual default handler.
+	kubeConfig.WarningHandlerWithContext = logutil.NewFilteringWarningHandler(log.NewKubeAPIWarningLogger(
+		log.KubeAPIWarningLoggerOptions{
+			Deduplicate: false,
+		},
+	), func(ctx context.Context, code int, agent string, text string) bool {
+		return !strings.HasPrefix(text, `unknown field`)
```
carreter: Maybe we should do a regex instead to match _only_ `unknown field ... .metadata.creationTimestamp` ? That way other warnings about unknown fields are still emitted.

carreter:
```suggestion
		re := regexp.MustCompile("unknown field.*\.metadata\.creationTimestamp")
		return !re.Match(text)
```

pkg/util/logutil/warning_handler.go
```diff
@@ -0,0 +1,51 @@
...
+type FilterWarningFunc func(ctx context.Context, code int, agent string, text string) bool
+
+// FilteringWarningHandler is a warning handler that filters entries by the given filter function.
+// A log entry is accepted when the given filter function returns true.
+type FilteringWarningHandler struct {
+	next   rest.WarningHandlerWithContext
+	filter FilterWarningFunc
+}
+
+var _ rest.WarningHandlerWithContext = (*FilteringWarningHandler)(nil)
```
carreter: What does this do? Just ensure we satisfy the interface?

tchap: Yeah, but I am going to remove it actually.

main.go
tchap: See 0338c8b486d71103a43d32e1031d2db835d7919a

carreter: Should it be `&&` rather than `||` in that case?

carreter: D'oh nevermind, I was flipping value

tchap: Yeah, I was like, there is a boolean operator involved, I am going to add some tests just to be sure 😁

go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/open-policy-agent/cert-controller v0.13.0
 	github.com/prometheus/client_golang v1.22.0
 	github.com/stretchr/testify v1.10.0
+	github.com/tonglil/buflogr v1.1.1
```
kannon92: I don't really like taking a dependency on something non standard.

This project does not seem to be very active so would we have to resolve CVE/bugs in this project?

kannon92:
/hold

I think we should revisit that dependency we would take in.

go.mod
tchap: I considered that, but it's actually a rather simple package and only used for testing. But yeah, I can implement the logging interface for tests myself...

tchap: Alright, amended the last commit with my own mock implementation and removed the dependency.

tchap:
The dependency is no longer there.

/unhold
/retest

kannon92:
/hold

I am getting some help from @everettraven and @JoelSpeed on this.

everettraven:
Had a chat with @kannon92 offline about this.

The reason you are seeing this error is because of the use of the [`batchv1.JobTemplateSpec`](https://pkg.go.dev/k8s.io/api/batch/v1#JobTemplateSpec) type that includes the [`metav1.ObjectMeta`](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta) in the type. `metav1.ObjectMeta` has custom JSON marshalling behavior such that the `creationTimestamp` is always set to `null` when it is zero: https://github.com/kubernetes/apimachinery/blob/173776a0582da70432e19d16eca025c451600ae8/pkg/apis/meta/v1/time.go#L161-L173

This means that your structured Go client marshals the JSON request data to include the `creationTimestamp: null`.

There are a couple potential paths you could take to try resolving this without obfuscating the warning:
1. Write a custom `MarshalJSON` method for your type, giving you the ability to customize the serialization of the type and drop the inclusion of `creationTimestamp`.
    - This is likely low-level enough that it would not impact your users to make this change, but it _is_ technically a breaking change in the serialization of the current Go APIs. It would break anyone that for some reason relies on that `creationTimestamp` field being included, but it seems unlikely someone would be relying on that.
2. Drop the use of the `batchv1.JobTemplateSpec` and instead create your own type that consists of the metadata fields you care about and the [`batchv1.JobSpec`](https://pkg.go.dev/k8s.io/api/batch/v1#JobSpec) type. Something like:
    ```go
    type CustomJobTemplate struct {
        Metadata CustomMetadata `json:"metadata"`
        Spec batchv1.JobSpec `json:"spec"`
    }
    
    type CustomMetadata struct {
        // Fields you care about from metav1.ObjectMeta go here.
        // Should be done in a way such that the OpenAPI schema of your CRD remains unchanged.
    }
    ```
    - This would be a breaking change to importers of the Go pkg for your types. A mitigation here could be a new version,  `v1alpha3`, giving a migration path for Go-based clients. Alternatively, you just don't give a migration path. I took a look and the latest release of this project was major version `0`, which for Go packages means that the package is in development and has no compatibility or stability guarantees (see https://go.dev/doc/modules/version-numbers).
  
---

In my opinion, obfuscating this error sets a precedent that when a noisy warning is present the path to resolution is to obfuscate it - even if it is bringing attention to something legitimate. If it were me, I would attempt the two paths I highlighted before deciding that this obfuscation is necessary. Even then, I'd recommend trying to come up with some sort of criteria you use to justify future warning obfuscations.

At the end of the day, it is up to the project maintainers what path they are most comfortable with - just wanted to share my 2 cents after chatting with Kevin.

tchap:
Thank you very much for the comment @everettraven . I totally agree this is not a solution, that's why I mentioned this as a workaround, not marking it as closing the issue. I actually just wanted to get rid of all the spam quickly and then fix it properly, but wasn't a good strategy 🙂 Will look into the proposal more deeply when I have some spare time.

tchap:
I am going to close this now and let's continue the discussion in the issue feed.

Issue #946, opened 2025-07-21
Make ReplicatedJobs a custom resource with their own controller


carreter:
**What would you like to be added**:
Currently, both `JobSets` and `ReplicatedJobs` are managed by the same JobSet controller. `ReplicatedJobs` do not exist as their own CR; rather, they are an abstraction that exists only in the context of a `JobSet`.

I'd like to see `ReplicatedJobs` become their own CR managed by their own controller (which can itself still be part of this repo).

**Why is this needed**:
The current approach has several downsides:

1. The JobSet controller is rather complex, as it must contain logic to reconcile both `JobSets` and their child `ReplicatedJobs`,
2. The status of a `ReplicatedJob` is shoved into its parent JobSet's `.status.replicatedJobsStatus` field, which is currently recalculated on each JobSet controller reconciliation loop. This makes it impossible to persist state for a specific `ReplicatedJob` (example of why this might be desirable: https://github.com/kubernetes-sigs/jobset/pull/920/files#r2213959512),
3. Failure policies that actually act at the `ReplicatedJob` level (i.e. `RecreateJob`, see #920) must be configured at the `JobSet` level. This can be a little confusing.
4. When communicating with other JobSet developers, I've noticed that people at times use `ReplicatedJob` and `Job` somewhat interchangeably, even though they are distinct concepts.

Creating a new custom resource for `ReplicatedJobs` would alleviate these issues, especially 1). However, it would require a significant refactor of the codebase!

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

GiuseppeTT:
As much as I like making nice abstractions, I think that a change like this would require a lot of effort and careful examination. Realistically, I see this change making into >= v2

Besides, the current approach of JobSet is messy but it's more flexible, which makes it easier to handle new feature requests.

In any case, I'm happy to help review any proposals : )

carreter:
I'm happy to write a proposal on what the API would look like, I think it wouldn't be too big a change on that front.

Would like more thoughts from others on whether this is worth the effort before I start, though!

kannon92:
With all your work on scalability for JobSet, would adding a new CRD cause more apiserver issues?

kannon92:
I talked with @catblade about the challenges with using JobSet as is with gang scheduling. Right now its hard enough gang scheduling Jobs. I worry about this.

If I understand this right,

[...]

PR #947, opened 2025-07-22
Add conversion webhook for v1

imreddy13:

Fixes #

```release-note

```


api/jobset/v1/doc.go
```diff
@@ -0,0 +1,21 @@
+/*
+Copyright 2023 The Kubernetes Authors.
```
kannon92, line 2:
```suggestion
Copyright 2025 The Kubernetes Authors.
```

api/jobset/v1/jobset_types.go
```diff
@@ -0,0 +1,420 @@
+/*
```
kannon92, line 1: cc @andreyvelich

kannon92:
/hold

You mentioned this is still a draft offline.

kannon92:
cc @tchap @ardaguclu 

I mentioned this work to you all so just wanted you to be aware.

api/jobset/v1/jobset_types.go
andreyvelich: Thank you for driving this @imreddy13!
Can we discuss all of the breaking changes to the JobSet APIs before introducing the V1 please ?

I feel like we have a few open discussions that we need to resolve before the v1, for example:
- Remove StartupPolicy API in favour of DependsOn
- https://github.com/kubernetes-sigs/jobset/issues/723
- https://github.com/kubernetes-sigs/jobset/issues/946

cc @ahg-g @tenzen-y @Electronic-Waste @astefanutti

carreter: #946 Would involve a significant refactor/rewrite of much of the codebase, so it could probably wait til v2 as @GiuseppeTT pointed out.

andreyvelich: Shall we consider to have `v1alpha2` API before the `v1` API to deprecate several APIs ?

andreyvelich: cc @danielvegamyhre for visibility

kannon92: I think we would add this breaking API changes on the promotion to V1.

I'll be honest. I still don't feel that strongly on #723. 

#946 was closed so I think we are not going to tackle that for V1.

I am happy to support startup policy deprecation in favor of dependsOn.

andreyvelich: @kannon92 For the #946 as we discussed with @GiuseppeTT and @carreter, we can change the `status` API for ReplicatedJobs from `list` to `map`, similar to what we did for the `IndividualJobRecreates` API.
This change will allow the controller to update the count for ReplicatedJobs in O(1) time instead of O(n).
E.g.:
```go
ReplicatedJobStatus map[string]ReplicatedJobStatus `json:"replicatedJobStatus,omitempty"`

type ReplicatedJobStatus struct {
	Ready     int32 `json:"ready"`
	Succeeded int32 `json:"succeeded"`
	Failed    int32 `json:"failed"`
	Active    int32 `json:"active"`
	Suspended int32 `json:"suspended"`
}
```

Any thoughts @astefanutti @tenzen-y @imreddy13 @kannon92 @GiuseppeTT @carreter @ahg-g ?

tenzen-y: The `map[sting]int` should be fine since `int` does not have any schema.
But, `map[string]ReplicatedJobStatus` must not be selected since `ReplicatedJobStatus` has schema.
The map key is not stored in CRDs.

EDIT: Even if the map value is `int`, we should avoid the map as much as possible since we will lose the CRD scheme power against the field.

andreyvelich: Make sense, we can also do something like this:
```go
type JobSetStatus struct {
  ReplicatedJobsReady map[string]int32 `json:"replicatedJobsReady,omitempty"`
  ReplicatedJobsSucceeded map[string]int32 `json:"replicatedJobsSucceeded,omitempty"`
  ReplicatedJobsFailed map[string]int32 `json:"replicatedJobsFailed,omitempty"`
  ReplicatedJobsActive map[string]int32 `json:"replicatedJobsActive,omitempty"`
  ReplicatedJobsSuspended map[string]int32 `json:"replicatedJobsSuspended,omitempty"`
  IndividualJobRecreates map[string]int32 `json:"individualJobRecreates,omitempty"`
}

tenzen-y: I'm not the maintainer of this project. So, I don't have any preference here. Just as a technical fact, introducing a map (even if the value is int) means abandoning any kind of CRD depending features like CRD with CEL forever.

carreter: > ReplicatedJobs in O(1) time instead of O(n)

I think this isn't too big a difference. n here will likely be small, right? If it causes issues w/ CRD features maybe we should just leave as a list.

kannon92: The use of maps in kubernetes APIs is actually against API conventions.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#lists-of-named-subobjects-preferred-over-maps

I was talking with our api-review team (@JoelSpeed and @everettraven) about this and they sent me this link.

There is a [kube-api-linter](https://github.com/kubernetes-sigs/kube-api-linter/blob/main/docs/linters.md#nomaps) where the linter explictly calls out that maps other than map[string]string are actually not allowed in kube APIs.

andreyvelich: That makes sense.
My only recommendation is that we maintain consistency across JobSet status fields.
Which means if we maintain `ReplicatedJobsStatus` field, keep the `IndividualJobsStatus` field, where we define:
```go
type IndividualJobStatus struct {
	Name string `json:"name"`
	
	Recreates int32 `json:"recreates"`
}

kannon92: I agree with that. @carreter would you be up for making that change?

kannon92: I think we can do this in v1alpha2 since we did not yet release this change.

carreter: > I agree with that. @carreter would you be up for making that change?

Sure, I'll take a stab at it! Makes sense to try and keep it consistent.

carreter: https://github.com/kubernetes-sigs/jobset/pull/958

PR #956, opened 2025-07-25
chore(api): Remove Deprecated StartupPolicy API

andreyvelich:
#### What type of PR is this?

/kind api-change
/kind cleanup

I removed the deprecated StartupPolicy API, that was replaced with DependsOn API.
Let me know if we prefer to remove API only in the next version of **JobSet APIs (e.g. v1alpha3)**

/hold 

cc @ahg-g @kannon92 @astefanutti @tenzen-y @GiuseppeTT @carreter @imreddy13

Fixes #

```release-note
Remove StartupPolicy API
```


ahg-g:
We shouldn't be removing it from v1alpha2, we should remove it from v1

/hold

andreyvelich:
> We shouldn't be removing it from v1alpha2, we should remove it from v1
> 
> /hold

Sounds good, I am wondering do we want to make one more API iteration before we migrate to v1 ?
e.g. v1alpha3 or v1beta1.

GiuseppeTT:
> Sounds good, I am wondering do we want to make one more API iteration before we migrate to v1 ?
e.g. v1alpha3 or v1beta1.

@kannon92 @imreddy13

ahg-g:
I don't recommend that, customer migration is going to be a pain since v1alpha2 is pretty much a production ready api since the api is widely adopted already, so I would not recommend having an intermediary api. That is also why @imreddy13 is woking on the conversion webhook, it will allow the controller to support the two versions for a period of time to allow this migration.

andreyvelich:
I think, we should at least talk about breaking changes before we introduce v1, like I mentioned here: https://github.com/kubernetes-sigs/jobset/pull/947#discussion_r2225694864

For example, make changes to the JobSet status API:
- https://github.com/kubernetes-sigs/jobset/issues/946#issuecomment-3114183220
- https://github.com/kubernetes-sigs/jobset/issues/723

ahg-g:
> I think, we should at least talk about breaking changes before we introduce v1, like I mentioned here: [#947 (comment)](https://github.com/kubernetes-sigs/jobset/pull/947#discussion_r2225694864)
> 
> For example, make changes to the JobSet status API:
> 
> * [Make ReplicatedJobs a custom resource with their own controller #946 (comment)](https://github.com/kubernetes-sigs/jobset/issues/946#issuecomment-3114183220)
> * [Rename JobSet status to Complete #723](https://github.com/kubernetes-sigs/jobset/issues/723)

Yes, we have been holding of the graduation to v1 for a while, if there are specific status changes that needs to be made, can we propose them?

We already agreed on the deprecation of the StartupPolicy API and the change to the status from completed to complete. I don't think we should be doing any major changes to the api at this point, as I mentioned v1alpha2 is pretty much a v1 in the eyes of most users.

kannon92:
> We shouldn't be removing it from v1alpha2, we should remove it from v1
> 
> /hold

Agree!

PR #958, opened 2025-07-28
Change IndividualJobRecreates to IndividualJobStatus.Recreates

carreter:
#### What type of PR is this?

/kind api-change

See discussion over at https://github.com/kubernetes-sigs/jobset/pull/947#discussion_r2237543175

This makes `IndividualJobStatus` to be consistent with `ReplicatedJobsStatus`.

Fixes #

```release-note
Change how individual job recreates are stored to be consistent with ReplicatedJobStatus
```


GiuseppeTT:
@kannon92 Do you mind if I also review this one? : )

kannon92:
> @kannon92 Do you mind if I also review this one? : )

Your input is always encouraged!

pkg/controllers/failure_policy.go
```diff
@@ -165,7 +165,9 @@ func failurePolicyRecreateAll(ctx context.Context, js *jobset.JobSet, shouldCoun
 	js.Status.Restarts += 1
 
 	// Reset individual Job recreates counters.
-	js.Status.IndividualJobRecreates = map[string]int32{}
+	for _, individualJobStatus := range js.Status.IndividualJobStatus {
+		individualJobStatus.Recreates = 0
+	}
```
GiuseppeTT: This won't work, you should do something like

```go
for i := range js.Status.IndividualJobStatus {
    js.Status.IndividualJobStatus[i].Recreates = 0
}
```

pkg/controllers/failure_policy.go
```diff
@@ -248,11 +250,22 @@ var recreateJobActionApplier failurePolicyActionApplier = func(ctx context.Conte
...
+		}
+	}
+	if !found {
+		js.Status.IndividualJobStatus = append(js.Status.IndividualJobStatus, jobset.IndividualJobStatus{
+			Name:      matchingFailedJob.Name,
+			Recreates: newRecreates,
+		})
 	}
-
-	js.Status.IndividualJobRecreates[matchingFailedJob.Name] = getIndividualJobRecreates(js, matchingFailedJob.Name) + 1
```
GiuseppeTT: Can you extract this into a function similary to what you did to `getIndividualJobRecreates`? For instance, `setIndividualJobRecreates`

pkg/controllers/failure_policy.go
GiuseppeTT: See https://go.dev/tour/moretypes/16. The object from `range` is a copy

GiuseppeTT: Can you also extract this into a function? Something like `resetAllIndividualJobRecreates`

pkg/controllers/failure_policy.go
carreter: Done.

pkg/controllers/failure_policy.go
carreter: Thanks for catching that, done.

pkg/controllers/failure_policy.go
```diff
@@ -271,6 +274,26 @@ var recreateJobActionApplier failurePolicyActionApplier = func(ctx context.Conte
...
+			break
+		}
+	}
+	if !found {
+		js.Status.IndividualJobsStatus = append(js.Status.IndividualJobsStatus, jobset.IndividualJobStatus{
+			Name:      jobName,
+			Recreates: recreates,
+		})
+	}
+}
```
GiuseppeTT: nit: I would use an early return instead of a break and get rid of `found`. I would also do `js.Status.IndividualJobsStatus[i].Recreates = recreates` to simplify and make it consistent with `resetIndividualJobRecreates`

GiuseppeTT (commented):
/lgtm

pkg/controllers/failure_policy.go
carreter: Done!

pkg/controllers/failure_policy.go
```diff
@@ -164,8 +164,7 @@ func failurePolicyRecreateAll(ctx context.Context, js *jobset.JobSet, shouldCoun
 	// of old jobs not part of the current jobSet run.
 	js.Status.Restarts += 1
 
-	// Reset individual Job recreates counters.
-	js.Status.IndividualJobRecreates = map[string]int32{}
+	resetIndividualJobRecreates(js)
```
andreyvelich, line 167: Maybe we should make a note in the API that counter is reseted once JobSet is restarted ?
https://github.com/kubernetes-sigs/jobset/blob/9102778909de3bb291f877224792e6fb8f5e6f80/api/jobset/v1alpha2/jobset_types.go#L200

pkg/controllers/failure_policy.go
```diff
@@ -271,6 +274,21 @@ var recreateJobActionApplier failurePolicyActionApplier = func(ctx context.Conte
 	return nil
 }
 
+// setIndividualJobRecreates sets the Recreates value of the corresponding entry
+// in js.Status.IndividualJobsStatus, creating it if it does not exist.
```
andreyvelich, line 278: Shall we make recreate counter = 0, after we create Job from ReplicatedJobs ?
Like how we do this for ReplicatedJobsStatus: https://github.com/kubernetes-sigs/jobset/blob/9102778909de3bb291f877224792e6fb8f5e6f80/pkg/controllers/jobset_controller.go#L360

andreyvelich (commented):
Thanks for updating the API!
Do we want to add some unit tests to the `failure_policy_test.go` ?

pkg/controllers/failure_policy.go
carreter: I don't think this is necessary, as the counter is assumed to be 0 if it does not exist. I think making status entries for all Jobs will mostly just add noise to the status field, but I can make the change!

pkg/controllers/failure_policy.go
carreter: Done.

carreter:
> Thanks for updating the API @carreter! Do we want to add some unit tests to the `failure_policy_test.go` ?

Currently `failure_policy_test.go` has no unit tests at all for the failure policy actions themselves, just the logic that finds the first matching failed job. I'm happy to add the unit tests in a separate PR! Let me know if this is desired and I will create an issue for it.

carreter:
Whoops, forgot to regenerate the CRDs. 1sec

pkg/controllers/failure_policy.go
andreyvelich: @GiuseppeTT @kannon92 @ahg-g Any thoughts ?

andreyvelich:
>Let me know if this is desired and I will create an issue for it.

Sure, that would be fantastic! Thank you!

pkg/controllers/failure_policy.go
kannon92: I do think absence or 0 may be more clear.

id prefer the suggestion that @andreyvelich has.

Consistency is nice. If the other status field shows zero elements but the other field doesn’t that would be confusion for end user or an operator.

kannon92: When someone creates a JobSet and it doesn’t reconcile that could be no status whatsoever. And then once there is a reconcile it at least reflects the created jobs and their restart count.

carreter: Sounds good, I'll make this change (ie create `IndividualJobStatus` entries for ALL Jobs, even when they haven't been recreated yet, please correct if my understanding is incorrect).

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -185,10 +185,21 @@ type JobSetStatus struct {
...
 	ReplicatedJobsStatus []ReplicatedJobStatus `json:"replicatedJobsStatus,omitempty"`
 
-	// IndividualJobRecreates tracks the number of times an individual Job within
-	// the JobSet has been recreated (i.e. in case of RecreateJob failure policy).
+	// IndividualJobsStatus tracks the status of individual Jobs within ReplicatedJobs.
 	// +optional
-	IndividualJobRecreates map[string]int32 `json:"individualJobRecreates,omitempty"`
+	// +listType=map
+	// +listMapKey=name
+	IndividualJobsStatus []IndividualJobStatus `json:"individualJobsStatus,omitempty"`
```
kannon92, line 192: Do we not have any tests that verify this value?

kannon92: I'd expect some integration tests to verify this status field but it seems we are missing those.

carreter: No tests yet, will add!

pkg/controllers/failure_policy.go
carreter: https://github.com/kubernetes-sigs/jobset/pull/958/files#diff-68677f8530f00fbbb399ac36268f928ee584948dd09091644a8fad81a8658153R609-R614

Done!

api/jobset/v1alpha2/jobset_types.go
carreter: Added tests.

kannon92 (approved):
LGTM!

I'll leave final lgtm to @GiuseppeTT or @andreyvelich.

PR #965, opened 2025-07-31
jobset: Don't create jobs or the headless service immediately when a suspended JobSet is created

tchap:
#### What type of PR is this?

/kind feature

Currently when a JobSet is created suspended, the associated jobs and headless service are always created immediately, although the jobs too are suspended.

This patch changes the behavior to only create these child resources when the JobSet is unsuspended for the first time, hence saving resources then the JobSet spends a long time suspended before being started, which can happen with Kueue.

Fixes https://github.com/kubernetes-sigs/jobset/issues/535

There were actually no e2e tests that got broken, I am wondering whether we need to add some.

Also any ideas for more tests are welcome, but not saying they are really needed.

```release-note
Aligned JobSet controller not to create jobs or the headless service for initially suspended JobSets.
The child resources get created now as soon as the JobSet gets unsuspended for the first time.
```

### DIFF
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -29,6 +29,7 @@ import (
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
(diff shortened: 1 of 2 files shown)


test/integration/controller/jobset_controller_test.go
```diff
@@ -1326,76 +1318,16 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
-		ginkgo.Entry("startupPolicy with InOrder; suspend should keep jobs suspended", &testCase{
```
kannon92, line 1329: I'm really concerned with deleting tests.

This seems like we are losing coverage.

tchap: The thing is, the following tests testing resuming a jobset basically start with what this test would test, so I didn't have the feeling we need to keep these, but I can surely put them back.

kannon92: I see what you mean now. I am fine with this actually.

kannon92:
/hold

I'm not sure if this is really worth the effort at the moment.

I don't think creation of these resources is blocking anyone at the moment.

tchap:
Feel free to keep this open or close it, as you feel appropriate.

tchap:
Feel free to revisit later, I am closing this.

PR #974, opened 2025-08-04
chore(manifests): Set SeccompProfile in the default JobSet manifests

andreyvelich:
#### What type of PR is this?

/kind feature

As we discussed in this thread, we should add the `seccompProfile` to the default JobSet controller manifests: https://github.com/kubeflow/trainer/pull/2759

We already set this value in the JobSet Helm Charts manifests: https://github.com/kubernetes-sigs/jobset/blob/main/charts/jobset/values.yaml#L76

/cc @kannon92 @astefanutti @g-gaston @GiuseppeTT @ahg-g 

Fixes #

```release-note
SeccompProfile has been added to the default JobSet manifests
```

### DIFF
--- a/config/components/manager/manager.yaml
+++ b/config/components/manager/manager.yaml
@@ -61,8 +61,6 @@ spec:
         # TODO(user): For common cases that do not require escalating privileges
         # it is recommended to ensure that all your Pods/Containers are restrictive.
         # More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
-        # Please comment the following code if your project has to work on old Kubernetes
-        # versions < 1.19 or on vendors versions which do NOT support this field by default (i.e. Openshift < 4.11 ).
         seccompProfile:
           type: RuntimeDefault
       containers:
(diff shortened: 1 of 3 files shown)


config/components/manager/manager.yaml
```diff
@@ -61,10 +61,10 @@ spec:
         # TODO(user): For common cases that do not require escalating privileges
         # it is recommended to ensure that all your Pods/Containers are restrictive.
         # More info: https://kubernetes.io/docs/concepts/security/pod-security-standards/#restricted
-        # Please uncomment the following code if your project does NOT have to work on old Kubernetes
+        # Please comment the following code if your project has to work on old Kubernetes
```
kannon92: Maybe we should update these docs.

1.19 is so far out of support..

kannon92:
Can you add a release note?

config/components/manager/manager.yaml
andreyvelich: Do you prefer to remove this comment all together ?

kannon92: I'm not sure if we even support that old of Kubernetes anymore.

So yes.

hack/python-sdk/swagger.json
```diff
@@ -124,7 +124,7 @@
         },
```
kannon92: Could you do this as a separate PR?

This seems unrelated to this fix.

andreyvelich: Sure!

Issue #977, opened 2025-08-06
Add controller.hostNetwork (and matching dnsPolicy) to Helm chart


m-mamdouhi:
On EKS clusters that use an overlay CNI (Calico, Cilium, …) the control-plane cannot reach the webhook service because the endpoints have pod-CIDR addresses. Setting hostNetwork: true exposes 9443 on each node’s VPC IP, eliminating “failed calling webhook: Address is not allowed”. Other SIG-sponsored charts (cert-manager, opentelemetry-operator) already ship this 

This enhancement requires the following artifacts:

- [ ] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
This sounds reasonable. Are you interested in contributing?

m-mamdouhi:
yes @kannon92 I would be interested in contributing

kannon92:
Thank you @m-mamdouhi!

The PR merged so going to close this.

PR #978, opened 2025-08-08
chart(jobset): add controller.hostNetwork toggle

m-mamdouhi:
#### What type of PR is this?

/kind feature

discussed here https://github.com/kubernetes-sigs/jobset/issues/977

```release-note
Helm chart: add `controller.hostNetwork` (bool, default `false`).  When set to `true`, the controller Pod runs on the node network
```

### DIFF
--- a/charts/jobset/README.md
+++ b/charts/jobset/README.md
@@ -87,7 +87,8 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | controller.nodeSelector | object | `{}` | Node selector of the jobset controller pods. |
 | controller.affinity | object | `{}` | Affinity of the jobset controller pods. |
 | controller.tolerations | list | `[]` | Tolerations of the jobset controller pods. |
-| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of all jobset controller containers. |
+| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of the jobset controller containers. |
+| controller.hostNetwork | bool | `false` | Run the controller/webhook Pods on the node’s network namespace instead of the overlay CNI. |
 | certManager.enable | bool | `false` | Whether to use cert-manager to generate certificates for the jobset webhook. |
 | certManager.issuerRef | object | `{}` | The reference to the issuer. If empty, self-signed issuer will be created and used. |
 | prometheus.enable | bool | `false` | Whether to enable Prometheus metrics exporting. |
--- a/charts/jobset/README.md
+++ b/charts/jobset/README.md
@@ -87,7 +87,7 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | controller.nodeSelector | object | `{}` | Node selector of the jobset controller pods. |
 | controller.affinity | object | `{}` | Affinity of the jobset controller pods. |
 | controller.tolerations | list | `[]` | Tolerations of the jobset controller pods. |
-| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of the jobset controller containers. |
+| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of all jobset controller containers. |
 | controller.hostNetwork | bool | `false` | Run the controller/webhook Pods on the node’s network namespace instead of the overlay CNI. |
 | certManager.enable | bool | `false` | Whether to use cert-manager to generate certificates for the jobset webhook. |
 | certManager.issuerRef | object | `{}` | The reference to the issuer. If empty, self-signed issuer will be created and used. |
--- a/charts/jobset/templates/controller/deployment.yaml
+++ b/charts/jobset/templates/controller/deployment.yaml
@@ -137,3 +137,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "jobset.controller.serviceAccount.name" . }}
+      {{- if .Values.controller.hostNetwork }}
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
+      {{- end }}
--- a/charts/jobset/values.yaml
+++ b/charts/jobset/values.yaml
@@ -75,6 +75,7 @@ controller:
     runAsNonRoot: true
     seccompProfile:
       type: RuntimeDefault
+  # -- Run the controller/webhook Pods on the node’s network namespace instead of the overlay CNI.
   hostNetwork: false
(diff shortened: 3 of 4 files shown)


charts/jobset/README.md
```diff
@@ -87,12 +87,14 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
...
+| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of the jobset controller pods. |
 | certManager.enable | bool | `false` | Whether to use cert-manager to generate certificates for the jobset webhook. |
 | certManager.issuerRef | object | `{}` | The reference to the issuer. If empty, self-signed issuer will be created and used. |
 | prometheus.enable | bool | `false` | Whether to enable Prometheus metrics exporting. |
 | prometheus.prometheusNamespace | string | `"monitoring"` |  |
 | prometheus.prometheusClientCertSecretName | string | `"jobset-metrics-server-cert"` |  |
+| controller.hostNetwork | bool | `false` | Run the controller/webhook Pods on the node’s network namespace instead of the overlay CNI. |
+| controller.dnsPolicy | string | `"ClusterFirst"` | DNS policy for the controller Pods. When `controller.hostNetwork=true`, the chart automatically overrides this to `ClusterFirstWithHostNet` so in-cluster Service names still resolve. |
```
GiuseppeTT: You list `controller.dnsPolicy` here but it is not available in `charts/jobset/values.yaml`. Which one is right?

GiuseppeTT: Also, can you move the new controller key(s) closer to the other controller keys?

m-mamdouhi: I moved it closer to the other controllers,
I had added the dnsPolicy to clarify this https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy in the documentation
I removed it now to avoid confusion

GiuseppeTT: I understand the intention but the table should reflect the content of `charts/jobset/values.yaml` not the templates.

Anyway, thanks for the update!

GiuseppeTT:
@kannon92 do you know why this test is failing?

I checked the logs and it shows

```
git --no-pager diff --exit-code config api client-go sdk charts
diff --git a/charts/jobset/README.md b/charts/jobset/README.md
index 443ed1b..8c06e94 100644
--- a/charts/jobset/README.md
+++ b/charts/jobset/README.md
@@ -87,8 +87,8 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
 | controller.nodeSelector | object | `{}` | Node selector of the jobset controller pods. |
 | controller.affinity | object | `{}` | Affinity of the jobset controller pods. |
 | controller.tolerations | list | `[]` | Tolerations of the jobset controller pods. |
-| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of the jobset controller containers. |
-| controller.hostNetwork | bool | `false` | Run the controller/webhook Pods on the node’s network namespace instead of the overlay CNI. |
+| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of all jobset controller containers. |
+| controller.hostNetwork | bool | `false` |  |
 | certManager.enable | bool | `false` | Whether to use cert-manager to generate certificates for the jobset webhook. |
 | certManager.issuerRef | object | `{}` | The reference to the issuer. If empty, self-signed issuer will be created and used. |
 | prometheus.enable | bool | `false` | Whether to enable Prometheus metrics exporting. |
make: *** [Makefile:150: verify] Error 1
```

My guess is that `charts/jobset/README.md` was edited manually whereas it should be generated from `charts/jobset/values.yaml`. In other words, @m-mamdouhi should run a command like `make verify` to make sure the files are in sync and update the PR.

kannon92:
> @kannon92 do you know why this test is failing?
> 
> I checked the logs and it shows
> 
> ```
> git --no-pager diff --exit-code config api client-go sdk charts
> diff --git a/charts/jobset/README.md b/charts/jobset/README.md
> index 443ed1b..8c06e94 100644
> --- a/charts/jobset/README.md
> +++ b/charts/jobset/README.md
> @@ -87,8 +87,8 @@ See [helm uninstall](https://helm.sh/docs/helm/helm_uninstall) for command docum
>  | controller.nodeSelector | object | `{}` | Node selector of the jobset controller pods. |
>  | controller.affinity | object | `{}` | Affinity of the jobset controller pods. |
>  | controller.tolerations | list | `[]` | Tolerations of the jobset controller pods. |
> -| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of the jobset controller containers. |
> -| controller.hostNetwork | bool | `false` | Run the controller/webhook Pods on the node’s network namespace instead of the overlay CNI. |
> +| controller.podSecurityContext | object | `{"runAsNonRoot":true,"seccompProfile":{"type":"RuntimeDefault"}}` | Security context of all jobset controller containers. |
> +| controller.hostNetwork | bool | `false` |  |
>  | certManager.enable | bool | `false` | Whether to use cert-manager to generate certificates for the jobset webhook. |
>  | certManager.issuerRef | object | `{}` | The reference to the issuer. If empty, self-signed issuer will be created and used. |
>  | prometheus.enable | bool | `false` | Whether to enable Prometheus metrics exporting. |
> make: *** [Makefile:150: verify] Error 1
> ```
> 
> My guess is that `charts/jobset/README.md` was edited manually whereas it should be generated from `charts/jobset/values.yaml`. In other words, @m-mamdouhi should run a command like `make verify` to make sure the files are in sync and update the PR.

That is correct. I think we have a target that builds helm docs and this job looks to see if it’s different from that generates. Hence why it is failing.

m-mamdouhi:
Ran make verify to refresh generated files, this also updated the API reference (adds IndividualJobStatus) which is unrelated to this change but required for the verify job to pass. thank you @GiuseppeTT  for pointing out the make verify

charts/jobset/templates/controller/deployment.yaml
```diff
@@ -137,3 +137,7 @@ spec:
         {{- toYaml . | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "jobset.controller.serviceAccount.name" . }}
+      {{- if .Values.controller.hostNetwork }}
+      hostNetwork: true
+      dnsPolicy: ClusterFirstWithHostNet
```
kannon92, line 142: Do we need any configuration of the dnsPolicy?

m-mamdouhi: the reason I shipped them together is this in the documentation:
https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pod-s-dns-policy
it says: 
```
For Pods running with hostNetwork, you should explicitly set its DNS policy to "ClusterFirstWithHostNet". Otherwise, Pods running with hostNetwork and "ClusterFirst" will fallback to the behavior of the "Default" policy.
```

PR #999, opened 2025-09-02
use chart version and app version to match kueue tagging

kannon92:
#### What type of PR is this?
/kind cleanup

https://github.com/kubernetes-sigs/kueue/blob/main/hack/helm-chart-package.sh

Trying to follow tagging as Kueue does to resolve https://github.com/kubernetes/k8s.io/pull/8453

Fixes #

It turns out we were overwriting what we put in the chart and we should use this script to set the correct values based on the tag.

We were using the tag for both chart_version and app_version so that leads to the helm chart being tagged.

```release-note
NONE
```

### DIFF
--- a/hack/push-chart.sh
+++ b/hack/push-chart.sh
@@ -45,7 +45,7 @@ default_image_repo=$(${YQ} ".image.repository" charts/jobset/values.yaml)
 readonly default_image_repo
 
 # Update the image repo, tag and policy
-${YQ}  e  ".image.repository = \"${image_repository}\" | .image.tag = \"${chart_version}\" | .image.pullPolicy = \"IfNotPresent\"" -i charts/jobset/values.yaml
+${YQ}  e  ".image.repository = \"${image_repository}\" | .image.tag = \"${GIT_TAG}\" | .image.pullPolicy = \"IfNotPresent\"" -i charts/jobset/values.yaml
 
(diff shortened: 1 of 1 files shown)


andreyvelich (commented):
/hold

hack/push-chart.sh
```diff
@@ -35,7 +35,7 @@ readonly k8s_registry="registry.k8s.io/jobset"
 readonly semver_regex='^v([0-9]+)(\.[0-9]+){1,2}$'
 
 image_repository=${IMAGE_REPO}
-chart_version=${GIT_TAG}
+chart_version="${GIT_TAG/#v/}"
 if [[ ${EXTRA_TAG} =~ ${semver_regex} ]]
 then
 	image_repository=${k8s_registry}/jobset
```
andreyvelich, line 41: @kannon92 We might also want to remove L42, like in Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/hack/helm-chart-package.sh#L37-L42

kannon92:
> @kannon92 I also noticed that the image tag is incorrectly set here: https://github.com/kubernetes-sigs/jobset/blob/2f8db4c1735b7eae11351b7f0c4187980edfa5c5/hack/push-chart.sh#L48C75-L48C88
> 
> It should be:
> 
> ```
> .image.tag = \"${GIT_TAG}\"
> ```

fixed in https://github.com/kubernetes-sigs/jobset/pull/999/commits/85de10a531ed486c16f4ac24865d2c88bb9d83d4

PR #1001, opened 2025-09-02
refactor helm chart push for DRY

kannon92:
#### What type of PR is this?
/kind cleanup

Fix artifacts to use helm-chart-package so that is also correct.

Fixes #

When getting ready for release 0.9.0, I was running `make artifacts` and I saw that the old appversion was still being used.

Turns out we were using a slightly different script in `make artifacts` from `make helm-chart-push`. This should consolidate them.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -223,27 +223,26 @@ helm-lint: ## Run Helm chart lint test.
 helm-docs: helm-docs-plugin ## Generates markdown documentation for helm charts from requirements and values files.
 	$(HELM_DOCS) --sort-values-order=file
 
-.PHONY: helm-chart-push
-helm-chart-push: yq helm
-	EXTRA_TAG="$(EXTRA_TAG)" GIT_TAG="$(GIT_TAG)" IMAGE_REGISTRY="$(IMAGE_REGISTRY)" HELM_CHART_REPO="$(HELM_CHART_REPO)" IMAGE_REPO="$(IMAGE_REPO)" HELM="$(HELM)" YQ="$(YQ)" ./hack/push-chart.sh
+.PHONY: helm-chart-package
+helm-chart-package: yq helm ## Package a chart into a versioned chart archive file.
+	DEST_CHART_DIR=$(DEST_CHART_DIR) \
+	HELM="$(HELM)" YQ="$(YQ)" GIT_TAG="$(GIT_TAG)" IMAGE_REGISTRY="$(IMAGE_REGISTRY)" \
+	HELM_CHART_PUSH=$(HELM_CHART_PUSH) \
+	./hack/helm-chart-package.sh
 
+.PHONY: helm-chart-push
+helm-chart-push: HELM_CHART_PUSH=true
+helm-chart-push: helm-chart-package
 
 ##@ Release
 .PHONY: artifacts
-artifacts: kustomize helm yq
+artifacts: kustomize helm yq helm-chart-package
 	cd config/components/manager && $(KUSTOMIZE) edit set image controller=${IMAGE_TAG}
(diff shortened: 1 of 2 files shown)


kannon92:
cc @andreyvelich

Makefile
```diff
@@ -225,25 +225,28 @@ helm-docs: helm-docs-plugin ## Generates markdown documentation for helm charts
...
+.PHONY: helm-chart-package
+helm-chart-package: yq helm ## Package a chart into a versioned chart archive file.
+	DEST_CHART_DIR=$(DEST_CHART_DIR) \
+	HELM="$(HELM)" YQ="$(YQ)" GIT_TAG="$(GIT_TAG)" IMAGE_REGISTRY="$(IMAGE_REGISTRY)" \
+	HELM_CHART_PUSH=$(HELM_CHART_PUSH) \
+	./hack/helm-chart-package.sh
+
+.PHONY: helm-chart-push
+helm-chart-push: HELM_CHART_PUSH=true
+helm-chart-push: helm-chart-package
```
andreyvelich, line 235: Why did we use the same instruction here?
`helm-chart-push` already exists in L227.

kannon92: ah thanks.

I need to remove that definition.

kannon92: removed.

Makefile
```diff
@@ -223,27 +223,26 @@ helm-lint: ## Run Helm chart lint test.
...
+.PHONY: helm-chart-package
+helm-chart-package: yq helm ## Package a chart into a versioned chart archive file.
+	DEST_CHART_DIR=$(DEST_CHART_DIR) \
+	HELM="$(HELM)" YQ="$(YQ)" GIT_TAG="$(GIT_TAG)" IMAGE_REGISTRY="$(IMAGE_REGISTRY)" \
+	HELM_CHART_PUSH=$(HELM_CHART_PUSH) \
+	./hack/helm-chart-package.sh
 
+.PHONY: helm-chart-push
+helm-chart-push: HELM_CHART_PUSH=true
+helm-chart-push: helm-chart-package
```
andreyvelich, line 235: Shall we do something like 
```suggestion
.PHONY: helm-chart-push
helm-chart-push: HELM_CHART_PUSH=true helm-chart-package
```

hack/helm-chart-package.sh
```diff
@@ -51,5 +51,8 @@ ${HELM} package --version "${chart_version}" --app-version "${GIT_TAG}" charts/j
 
 # Revert the image changes
 ${YQ}  e  ".image.repository = \"${default_image_repo}\" | .image.tag = \"main\" | .image.pullPolicy = \"Always\"" -i charts/jobset/values.yaml
-echo "pushing chart to ${HELM_CHART_REPO}"
-${HELM} push "bin/jobset-${chart_version}.tgz" "oci://${HELM_CHART_REPO}"
+
+if [ "$HELM_CHART_PUSH" = "true" ]; then
+	echo "pushing chart to ${HELM_CHART_REPO}"
+  ${HELM} push "${DEST_CHART_DIR}/jobset-${chart_version}.tgz" "oci://${HELM_CHART_REPO}"
```
andreyvelich, line 57:
```suggestion
	echo "pushing chart to ${HELM_CHART_REPO}"
    ${HELM} push "${DEST_CHART_DIR}/jobset-${chart_version}.tgz" "oci://${HELM_CHART_REPO}"
```

Makefile
kannon92: I was following Kueue's code here so I would prefer to keep this change.
I never find Make user friendly so I prefer to follow Kueue's style here.

PR #1005, opened 2025-09-03
potentially fix date tag on helm push

kannon92:
#### What type of PR is this?
/kind cleanup

Match our script with Kueue.

Removing EXTRA_TAG and follow Kueue's script for pushing the helm chart.

Fixes #

cc @mimowo @tenzen-y 

We are struggling to figure out why our tag for the helm chart as the date rather than v0.9.0. I'm hoping this change will fix this but not entirely sure if it will.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -170,7 +170,7 @@ image-local-push: image-local-build
 
 .PHONY: image-build
 image-build:
-	$(IMAGE_BUILD_CMD) -t $(IMAGE_REPO):$(BRANCH_NAME) \
+	$(IMAGE_BUILD_CMD) -t $(IMAGE_TAG) -t $(IMAGE_REPO):$(BRANCH_NAME) \
 		--platform=$(PLATFORMS) \
 		--build-arg BASE_IMAGE=$(BASE_IMAGE) \
 		--build-arg BUILDER_IMAGE=$(BUILDER_IMAGE) \


mimowo:
Yeah, I think we simplified the release process and stabilized after dropping the EXTRA_TAG which is confusing because it makes the build variables different for local release builds on and CI, making them hard to debug.

cc @mbobrovskyi who actually implemented the changes in Kueue.

hack/helm-chart-package.sh
```diff
@@ -20,33 +20,35 @@ set -o pipefail
...
+image_repository=${IMAGE_REGISTRY}/jobset
+chart_version="${GIT_TAG/#v/}"
+
 default_image_repo=$(${YQ} ".image.repository" charts/jobset/values.yaml)
 readonly default_image_repo
 
 # Update the image repo, tag and policy
 ${YQ}  e  ".image.repository = \"${image_repository}\" | .image.tag = \"${GIT_TAG}\" | .image.pullPolicy = \"IfNotPresent\"" -i charts/jobset/values.yaml
 
+echo "helm package --version ${chart_version} --app-version ${GIT_TAG} charts/jobset -d "${DEST_CHART_DIR}""
```
mbobrovskyi: Do we need this?

mbobrovskyi:
It looks like this no need anymore, right?

https://github.com/kubernetes-sigs/jobset/blob/dcce7502f16e56b136ca910defa230f840def346/Makefile#L36-L41

hack/helm-chart-package.sh
kannon92: I wanted to leave it in to make sure its doing what I want. I can remove.

mbobrovskyi:
Without EXTRA_TAG you also need to add `-t $(IMAGE_REPO):$(RELEASE_BRANCH) \`:

https://github.com/kubernetes-sigs/jobset/blob/dcce7502f16e56b136ca910defa230f840def346/Makefile#L178-L184

Like we have on kueue:

https://github.com/kubernetes-sigs/kueue/blob/06cb5118a7a155c96cf358358d676fdccc7de19a/Makefile#L274-L276

hack/helm-chart-package.sh
kannon92: removed.

kannon92:
> Without EXTRA_TAG you also need to add `-t $(IMAGE_REPO):$(RELEASE_BRANCH) \`:
> 
> https://github.com/kubernetes-sigs/jobset/blob/dcce7502f16e56b136ca910defa230f840def346/Makefile#L178-L184
> 
> Like we have on kueue:
> 
> https://github.com/kubernetes-sigs/kueue/blob/06cb5118a7a155c96cf358358d676fdccc7de19a/Makefile#L274-L276

I don't think we need this atm. I don't see us using the release-branch and I don't quite understand what `make prepare-release-branch` would need to do in the jobset repo.

kannon92:
/hold

Thank you to @mbobrovskyi and @mimowo for help in reviewing.

I would like your lgtm on this before we merge.

mimowo:
> Without EXTRA_TAG you also need to add `-t $(IMAGE_REPO):$(RELEASE_BRANCH) \`:
> 
> https://github.com/kubernetes-sigs/jobset/blob/dcce7502f16e56b136ca910defa230f840def346/Makefile#L178-L184
> 
> Like we have on kueue:
> 
> https://github.com/kubernetes-sigs/kueue/blob/06cb5118a7a155c96cf358358d676fdccc7de19a/Makefile#L274-L276

I think @mbobrovskyi is right that you need to somehow ensure the images for branches are pushed, as they are currently: 

image

To achieve that you don't really need `RELEASE_BRANCH`  as a variable. I would just do:

`-t $(IMAGE_REPO):$(BRANCH_NAME)` where `BRANCH_NAME=git branch --show-current`

`RELEASE_BRANCH` is a confusing part here, because for `main` it is not really "release" branch (we could consider simplifying it in Kueue too)

EDIT: we had this issue in Kueue already https://github.com/kubernetes-sigs/kueue/issues/5535 but decided not to cleanup yet. I'm on the fence about RELEASE_BRANCH still. In any case, I think you can just use  `BRANCH_NAME=git branch --show-current`

kannon92:
@mimowo @mbobrovskyi Added a fix for branch name.

Makefile
```diff
@@ -176,7 +170,7 @@ image-local-push: image-local-build
 
 .PHONY: image-build
 image-build:
-	$(IMAGE_BUILD_CMD) -t $(IMAGE_TAG) \
+	$(IMAGE_BUILD_CMD) -t $(IMAGE_REPO):$(BRANCH_NAME) \
```
mimowo: Wait, but now it will only push the branch name tag. I think you want to push both, no?

kannon92: Oh I thought it was just one. I can add two tags.

kannon92: done.

mimowo: > Oh I thought it was just one. 

Well I check here: http://us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset

And it seems you are already pushing two tags: branch name, and git tag.

image

So, I think otherwise it would regress

mimowo:
LGTM, thanks 👍

andreyvelich (commented):
/lgtm
/hold
@kannon92 Feel free to unhlod it.

kannon92:
:cry: 

This did not work: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/post-jobset-push-images/1963631960895000576

Looks like `BRANCH_NAME ?= $(shell git branch --show-current)` is resolving to nill in the cloud build...

```
/buildx-entrypoint build -t us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset:v20250904-v0.10.0-devel-10-g32bb401 -t us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset: \
	--platform=linux/amd64,linux/arm64,linux/s390x \
	--build-arg BASE_IMAGE=gcr.io/distroless/static:nonroot \
	--build-arg BUILDER_IMAGE=golang:1.24.0 \
	--push \
	 ./
```

mbobrovskyi:
In Kueue, we have a static constant for RELEASE_BRANCH at https://github.com/kubernetes-sigs/kueue/blob/ffd0f09af9b132fe560479be6a3289a3898d8c3a/Makefile#L89-L90, and we use it as an additional tag.

kannon92:
> In Kueue, we have a static constant for RELEASE_BRANCH at https://github.com/kubernetes-sigs/kueue/blob/ffd0f09af9b132fe560479be6a3289a3898d8c3a/Makefile#L89-L90, and we use it as an additional tag.

Yea, I ended up opening https://github.com/kubernetes-sigs/jobset/pull/1008.

BenTheElder:
> Looks like BRANCH_NAME ?= $(shell git branch --show-current) is resolving to nill in the cloud build...



You have to opt into uploading .git, but the git tag is supplied via `_GIT_TAG` when using our tool to kick off these builds from prow:

https://github.com/kubernetes/test-infra/tree/master/images/builder#image-builder

You can opt-into uploading .git instead `--with-git-dir`.

hack/helm-chart-package.sh
```diff
@@ -20,27 +20,28 @@ set -o pipefail
...
+# going to be promoted to registry.k8s.io.
+readonly promoted_version_regex='^v([0-9]+)(\.[0-9]+){1,2}$'
 
-image_repository=${IMAGE_REPO}
-chart_version="${GIT_TAG/#v/}"
-if [[ ${EXTRA_TAG} =~ ${semver_regex} ]]
+if [[ ${GIT_TAG} =~ ${promoted_version_regex} ]]
 then
-	image_repository=${k8s_registry}/jobset
+	IMAGE_REGISTRY=${k8s_registry}
```
toVersus, line 39: @kannon92 
This change breaks the default image name in the JobSet Helm chart version 0.9.0. It points to `registry.k8s.io/jobset:v0.9.0` instead of the correct `registry.k8s.io/jobset/jobset:v0.9.0`, causing an ImagePullBackOff and preventing the controller from starting.

```console
❯ helm template oci://registry.k8s.io/jobset/charts/jobset --version 0.9.0 | grep "image:"
Pulled: registry.k8s.io/jobset/charts/jobset:0.9.0
Digest: sha256:6f435c55e6cf757d1f55c4f232617bc8a381c63e2bfe80f90b5fc620499de05e
        image: "registry.k8s.io/jobset:v0.9.0"

❯ crane ls registry.k8s.io/jobset | grep v0.9.0

❯ crane ls registry.k8s.io/jobset/jobset | grep v0.9.0
v0.9.0
```

Shouldn't it be like this?

```diff
-	IMAGE_REGISTRY=${k8s_registry}
+	IMAGE_REGISTRY=${k8s_registry}/jobset
```

andreyvelich: Thanks for reporting this @toVersus!
I think there is a bug, and we should use update it as in Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/hack/helm-chart-package.sh#L63

I will submit a hot fix.

PR #1010, opened 2025-09-05
match kueue and jobset cloud build

kannon92:
https://github.com/kubernetes-sigs/kueue/blob/main/cloudbuild.yaml

Aiming to match jobset/kueue cloud builds so we can have similar tagging.

### DIFF
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -10,8 +10,6 @@ steps:
       - image-push
       - helm-chart-push
     env:
-    - IMAGE_REGISTRY=us-central1-docker.pkg.dev/k8s-staging-images/jobset
-    - GIT_TAG=$_GIT_TAG
     - DOCKER_BUILDX_CMD=/buildx-entrypoint
     - GOTOOLCHAIN=auto
 substitutions:


kannon92:
One difference between jobset/kueue is that Kueue uses --with-git-dir in its image-builder postsubmit.

https://github.com/kubernetes/test-infra/pull/35478

I think once we merge that PR, we can use git commands to get tags rather than relying on cloudbuild.

mimowo:
Yeah. we decided not to use variables injected by cloud build because it was very hard to debug when the variables have different values locally and on ci

mimowo:
> Do we know if that is going to work for us since we use GIT_TAG here ?

I thinks so, this looks correct. As long you build env has git so that you can compute it .

kannon92:
/hold

We need https://github.com/kubernetes/test-infra/pull/35478 before this can merge I think.

kannon92:
/hold cancel

@andreyvelich @mimowo this is ready for merge now.

andreyvelich (commented):
/lgtm

Issue #1016, opened 2025-09-09
v0.9.0 helm installation - registry.k8s.io/jobset:v0.9.0: not found


eicca:
Hi team!

First, thank you for all the great work on JobSet!

## Issue Description

I was installing JobSet v0.9.0 using the Helm chart and encountered an issue with the image repository path that prevented the controller from starting.
It seems that the image path specified in the Helm chart is incorrect.

## What I observed

When installing the Helm chart, the deployment tries to pull the image from:
```
registry.k8s.io/jobset:v0.9.0
```

This results in an ImagePullBackOff error:
```
Failed to pull image "registry.k8s.io/jobset:v0.9.0": 
rpc error: code = NotFound desc = failed to pull and unpack image "registry.k8s.io/jobset:v0.9.0": 
failed to resolve reference "registry.k8s.io/jobset:v0.9.0": registry.k8s.io/jobset:v0.9.0: not found
```

## The working image path

After some investigation, I found that the correct image path is:
```
registry.k8s.io/jobset/jobset:v0.9.0
```

## Workaround

I was able to successfully deploy JobSet by overriding the image repository in the Helm values:
```yaml
image:
  repository: registry.k8s.io/jobset/jobset
```

With this override, everything works perfectly!

## Steps to reproduce

1. Install JobSet using the Helm chart:
   ```bash
   helm install jobset oci://registry.k8s.io/jobset/charts/jobset --version 0.9.0 --create-namespace --namespace=jobset-system
   ```
2. Check the pod status:
   ```bash
   kubectl get pods -n jobset-system
   ```

## Environment
- Helm v3
- JobSet chart version: 0.9.0
- Kubernetes 1.28

I hope this helps! Let me know if you need any additional information.

andreyvelich:
Thanks for reporting this @eicca!
Yes, it looks like a bug as @toVersus  mentioned here: https://github.com/kubernetes-sigs/jobset/pull/1005#discussion_r2332254879 

/assign @andreyvelich

PR #1018, opened 2025-09-09
fix(charts): Fix the JobSet image name

andreyvelich:
#### What type of PR is this?

/kind bug

We should use the correct JobSet image registry name in the Helm charts: `registry.k8s.io/jobset/jobset:v0.9.0`
I make it consistent with Kueue scripts: https://github.com/kubernetes-sigs/kueue/blob/main/hack/helm-chart-package.sh#L63

/assign @kannon92 @GiuseppeTT
/cc @toVersus @mbobrovskyi @tenzen-y @mimowo

Fixes https://github.com/kubernetes-sigs/jobset/issues/1016

```release-note

```

### DIFF
--- a/hack/helm-chart-package.sh
+++ b/hack/helm-chart-package.sh
@@ -39,14 +39,13 @@ then
 	IMAGE_REGISTRY=${k8s_registry}
 fi
 
-image_repository=${IMAGE_REGISTRY}
 chart_version="${GIT_TAG/#v/}"
 
 default_image_repo=$(${YQ} ".image.repository" charts/jobset/values.yaml)
 readonly default_image_repo
 
 # Update the image repo, tag and policy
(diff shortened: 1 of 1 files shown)


GiuseppeTT:
/lgtm

Thanks for fixing this bug!

For future reference. I tested by running `make helm-chart-package` and extracting `image.repository` from the resulting `values.yaml`. I got `us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset` so it's working

kannon92:
/cherry-pick release-0.9

I think we would need to release a new chart, no?

andreyvelich:
>I think we would need to release a new chart, no?

It's not really the blocker, since users have workaround: https://github.com/kubernetes-sigs/jobset/issues/1016
Maybe we can release v0.9.1 in the 1 month, in case we find more bugs to cherry-pick ?

PR #1022, opened 2025-09-15
Filter out terminating leader Pods

### LINKED ISSUE #1021: JobSet controller doesn't check if the leader Pod is stale, leading to misplacing follower Pods
**What happened**:
Using Kueue I preempted JobSet and then right after I admitted it. The preemption process send requests to delete all Pods but it doesn't wait until. After Kueue once again admitted (unsuspended) it there was a terminating leader Pod with stale value of the `ExclusiveKey` annotation. As a result, part of follower Pods were assigned stale `ExclusiveKey` value that led to JobSet misplacement 

I'd expect JobSet to filter out terminating leader Pods and wait until the new leader Pod with valid data is created

**Environment**:
- Install tools:
- Others:

PBundyra:
#### What type of PR is this?
/kind bug

- Filter outs terminating leader Pods
- Add unit tests for `leaderPodForFollower` function
- Small refactor of indexes so they can be used in unit tests

Fixes #1021

```release-note
JobSet pod webhook validates if leader Pods are terminating before copying metadate from them  
```

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -75,34 +75,37 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
 func SetupPodIndexes(ctx context.Context, indexer client.FieldIndexer) error {
 	// Build index where key is the hash of the namespaced job name of the job that owns this pod,
 	// and value is the pod itself.
-	if err := indexer.IndexField(ctx, &corev1.Pod{}, podJobKey, func(obj client.Object) []string {
-		pod := obj.(*corev1.Pod)
-		// Make sure the pod is part of a JobSet using exclusive placement.
-		if _, exists := pod.Annotations[jobset.ExclusiveKey]; !exists {
-			return nil
-		}
-		jobKey, exists := pod.Labels[jobset.JobKey]
-		if !exists {
-			return nil
-		}
-		return []string{jobKey}
-	}); err != nil {
+	if err := indexer.IndexField(ctx, &corev1.Pod{}, podJobKey, IndexPodJob); err != nil {
 		return err
 	}
-
 	// Build index where the key is the pod name (without the random suffix), and the value is the pod itself.
-	return indexer.IndexField(ctx, &corev1.Pod{}, PodNameKey, func(obj client.Object) []string {
-		pod := obj.(*corev1.Pod)
-		// Make sure the pod is part of a JobSet using exclusive placement.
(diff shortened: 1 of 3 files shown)


pkg/controllers/pod_controller.go
```diff
@@ -75,34 +75,37 @@ func (r *PodReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
-		if _, exists := pod.Annotations[jobset.ExclusiveKey]; !exists {
-			return nil
-		}
-		jobKey, exists := pod.Labels[jobset.JobKey]
-		if !exists {
-			return nil
-		}
-		return []string{jobKey}
-	}); err != nil {
+	if err := indexer.IndexField(ctx, &corev1.Pod{}, podJobKey, IndexPodJob); err != nil {
```
ahg-g, line 78: IIUC, the changes in this file are just a refactor; if true, I recommend avoiding pass-by changes, lets focus PR on fixing the bug and have this change sent in a separate "cleanup" PR if you like.

ahg-g (commented):
I think this is a reasonable thing to do irrespective of Kueue behavior; left one nit, but will leave the full review to Giuseppe

pkg/controllers/pod_controller.go
GiuseppeTT: > IIUC, the changes in this file are just a refactor

IIUC the refactor was done for the new tests added to `pkg/webhooks/pod_admission_webhook_test.go`. See https://github.com/kubernetes-sigs/jobset/pull/1022/files#diff-5308e6bd886282c7740a169b50d80ddd9dd01f49e41e41733683c6481c27d31bR205

GiuseppeTT:
> but will leave the full review to Giuseppe

I reviewed the code : )

I'm fine with the code. It's basically adding a [check that only consider leader Pods that are not marked for deletion](https://github.com/kubernetes-sigs/jobset/pull/1022/files#diff-4414aff8efe7c4c3e19085eeeca447bfc61219bde7bf458f124010c3a34ff4ccR106) when admitting a follower pod. The rest is tests (which are welcomed!) and a small refactor for the tests.

If @PBundyra was able to create a repro, I'd like to see a test (can be a manual test) to make sure the fix works. Otherwise we run the risk of advertising a fix that doesn't work.

That being said, I can't think of a case that follower Pods should be admitted when the leader Pod is marked for deletion so the PR is low risk. I'll LGTM it and leave the final call to @ahg-g

PR #1047, opened 2025-09-28
copy crd to helm chart on manifest

kannon92:
#### What type of PR is this?
/kind bug

When we generated the CRD we did not update the CRD for the helm chart.

Fixes #

This is why we created #820. 

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -102,6 +102,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
 		webhook output:webhook:artifacts:config=config/components/webhook\
 		paths="./pkg/..."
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -102,6 +102,7 @@ manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and Cust
 		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
 		webhook output:webhook:artifacts:config=config/components/webhook\
 		paths="./pkg/..."
+	cp ./config/components/crd/bases/jobset.x-k8s.io_jobsets.yaml charts/jobset/crds/.
```
andreyvelich: I think, we should do force copy: https://github.com/kubeflow/trainer/blob/master/Makefile#L122C2-L123C1

```suggestion
	cp -f ./config/components/crd/bases/jobset.x-k8s.io_jobsets.yaml ./charts/jobset/crds/
```

kannon92: updated.

PR #1053, opened 2025-10-01
KEP-572: VolumeClaimPolicies API for Stateful JobSet

andreyvelich:
#### What type of PR is this?

/kind feature
/kind api-change

This is KEP for stateful JobSet with the `VolumeClaimPolicies` API. It should allow to orchestrate shared volumes claims for JobSet or volume claims per ReplicatedJob.

I was trying to follow [the StatefulSets API](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#persistentvolumeclaim-retention).

<details>
<summary>Click to expand simple YAML example</summary>

```yaml
kind: JobSet
metadata:
  name: distributed-trainjob
spec:
  volumeClaimPolicies:
    - targetReplicatedJobs: ["node"]
      templates:
        - metadata:
            name: checkpoint-storage
          spec:
            accessModes: ["ReadWriteOnce"]
            resources:
              requests:
                storage: 100Gi
      retentionPolicy:
        whenComplete: Retain
        whenFailed: Retain
        whenDeleted: Retain
  replicatedJobs:
    - name: node
      replicas: 4
      template:
        spec:
          parallelism: 2
          completions: 2
          template:
            spec:
              containers:
                - name: node
                  image: ghcr.io/kubeflow/trainer/deepspeed-runtime
                  volumeMounts:
                    - name: checkpoint-storage
                      mountPath: /checkpoints
```

</details>

Part of: https://github.com/kubernetes-sigs/jobset/issues/572, https://github.com/kubernetes/kubernetes/issues/115066

/cc @kannon92 @GiuseppeTT @ahg-g @imreddy13 @carreter @tenzen-y @astefanutti @Electronic-Waste @vsoch @xigang @akshaychitneni @shravan-achar @bigsur0 

```release-note
Add the VolumeClaimPolicies API to support stateful JobSet.
```


keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+- [Alternatives](#alternatives)
+  - [Alternative 1: Job-level VolumeClaimTemplates](#alternative-1-job-level-volumeclaimtemplates)
+  - [Alternative 2: External Volume Management](#alternative-2-external-volume-management)
+  - [Alternative 3: Pre-created PVC References](#alternative-3-pre-created-pvc-references)
+
+<!-- /toc -->
+
+## Summary
+
+This KEP proposes adding VolumePolicies API to support stateful JobSets, enabling automated creation
```
vsoch:
```suggestion
This KEP proposes adding VolumeClaimPolicies API to support stateful JobSets, enabling automated creation
```
I would also add "a" or "the"

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+and management of PersistentVolumeClaims (PVCs) for stateful batch workloads. The feature addresses two
+primary use cases: per-job persistent storage (similar to StatefulSet VolumeClaimTemplates) and
+shared storage across multiple ReplicatedJobs within a JobSet. This enhancement bridges the gap
+between stateless batch processing and stateful storage requirements for modern distributed
+computing workloads, particularly in AI training and HPC domains.
+
+## Motivation
+
+Current JobSet implementation provides no native volume management capabilities, requiring users to
```
vsoch:
```suggestion
The current JobSet implementation provides no native volume management capabilities, requiring users to
```

keps/572-stateful-jobset/README.md
vsoch: Unless `VolumePolicies` is intended as a higher level umbrella that includes `VolumeClalimPolicies`. If that is the case, a brief reference would be helpful for the reader.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+manually create and manage PVCs for stateful workloads. This limitation creates significant
+operational overhead and prevents efficient resource utilization for distributed workloads that
+require persistent storage for checkpoints, datasets, models, or intermediate results.
+
+### Goals
+
+- Automate PVC creation similar to StatefulSet capability.
+- Support per ReplicatedJob and shared volume patterns withing a single JobSet.
```
vsoch:
```suggestion
- Supports each ReplicatedJob and shared volume pattern withing a single JobSet.
```
You are saying the policy supports a paired ReplicatedJob and shared volume? Does the volume have to be owned by the JobSet somehow (e.g,. within)?

keps/572-stateful-jobset/README.md
vsoch: Or it is this? https://github.com/kubernetes-sigs/jobset/issues/572

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+- Design configurable retention policy for created PVCs.
+
+### Non-Goals
+
+- Storage performance tuning and optimization features are not covered.
+- Support per Job index volume claim templates.
+- Cross-JobSet volume sharing. Sharing volumes between different JobSet instances is out of scope.
```
vsoch, line 68: What prevents a user from creating a volume type separately and then mounting to different jobsets?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+- Orchestration for PersistentVolumes. JobSet users should rely on StorageClasses to provision
```
vsoch, line 69: This is a huge area/challenge and indeed JobSet shouldn't take it all on! I used to have the Flux Operator create volumes and realized what a bad idea that was. :upside_down_face:

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+  and delete PVs.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Distributed ML Training with Per-Worker Checkpoints
+
+As AI practitioner, I want to run distributed training workloads where each worker maintains its
```
vsoch:
```suggestion
As an AI practitioner, I want to run distributed training workloads where each worker maintains its
```

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+own checkpoint storage, allowing for fault tolerance and efficient restart capabilities.
```
vsoch: Does this mean one ssd per node, each worker == one node? Or workers writing their own namespaced files (e.g., worker-0 ... worker-N in a shared location, single volume?)

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+                  image: ghcr.io/kubeflow/trainer/deepspeed-runtime
+                  volumeMounts:
+                    - name: checkpoint-storage
+                      mountPath: /checkpoints
+                  env:
+                    - name: CHECKPOINT_PATH
+                      value: /checkpoints
+```
+
+**Expected Behavior**: Creates 8 individual PVCs named `checkpoint-storage-distributed-trainjob-node-{0-3}-{0-1}`,
```
vsoch, line 124: Different underlying storage devices?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+each mounted to the corresponding worker pod for independent checkpoint storage.
+
+#### Story 2: Shared model and dataset across workers.
+
+As an AI practitioner, I want to initialize model and dataset before distributed fine-tuning,
```
vsoch:
```suggestion
As an AI practitioner, I want to initialize a model and dataset before distributed fine-tuning,
```

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+spec:
+  volumeClaimPolicies:
+    - targetReplicatedJobs: []
+      templates:
+        - metadata:
+            name: initializer
+            labels:
+              content-type: model
+          spec:
+            accessModes: ["ReadWriteMany"]
```
vsoch, line 146: How is it possible to support more than one access mode? Can you give an example? Does the most or least permissive take precedence?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+            resources:
+              requests:
+                storage: 50Gi
+            storageClassName: nfs-storage
+      retentionPolicy:
```
vsoch, line 151: To clarify defaults, if this entire block is left out, it would be akin to setting them all to Retain.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+Two shared PVC with names `dataset-cache-hybrid` and `results-storage-hybrid`.
+
+Worker scratch PVCs:
+
+- 4 PVCs with names `worker-scratch-data-loader-{0-3}-{0}` for data-loader ReplicatedJob
+- 12 PVCs with names `worker-scratch-trainer-{0-2}-{0-5}` for trainer ReplicatedJob
+
+#### Story 4: HPC Workloads with Node-Local Storage
+
+As an HPC researcher, I want to utilize node-local storage for high-performance simulation
```
vsoch, line 320: Many HPC simulation / apps that use MPI and assume a traditional system are going to want a fully shared filesystem, in addition to NVME.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+**Expected Behavior**: Creates 16 individual PVCs named `simulation-data-hpc-simulation-compute-node-{0-15}-{0}`
+with node-local storage for high-performance simulation data, persisting across pod failures and restarts.
+
+### Risks and Mitigations
+
+#### Storage Resource Exhaustion
+
+If automatically created PVCs won't get proper cleanup, it might lead to quota exhaustion and
+increased costs. We mitigate this by defaulting retention policies to safe values
+(`Delete` as a default) while providing clear documentation on cleanup strategies and cost
```
vsoch, line 376: ah ok, so I think this is the opposite of StatefulSet? So what is the purpose to define many of the blocks above with `Delete` if it is the default? Maybe a comment earlier about this would clear that up for the reader.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+implications.
+
+#### PVC Naming Conflicts
+
+Naming conflicts could occur with deterministic naming patterns, but we mitigate this through
```
vsoch, line 381: So I understand - when a pod fails, and if we retain the volume and it is recreated, we _want_ the naming conflict to occur (the pod to be created with its predecessor checkpoint). We likely don't want the naming conflict if the entire JobSet is deleted (but not volumes) and then recreated with the same names, and data received in all different states without the context of some lead worker that expects it.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+    // VolumeClaimPolicies defines volume claim templates and lifecycle management policies.
+    // Each policy targets specific ReplicatedJobs and defines templates and retention rules.
+    VolumeClaimPolicies []VolumeClaimPolicy `json:"volumeClaimPolicies,omitempty"`
+}
+
+// VolumeClaimPolicy defines volume claim templates and lifecycle management for targeted ReplicatedJobs.
+type VolumeClaimPolicy struct {
+    // TargetReplicatedJobs specifies which ReplicatedJobs this policy applies to.
+    // If empty, the policy applies to all ReplicatedJobs in the JobSet.
+    TargetReplicatedJobs []string `json:"targetReplicatedJobs,omitempty"`
```
vsoch, line 413: In the future it could be as simple as a selector, e.g., to target specific pods within the jobs.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+
+    // Templates is a list of claims that the targeted ReplicatedJobs are allowed to reference.
+    // The JobSet controller is responsible for mapping volume identities to claims in a way that
+    // maintains the identity of the jobs. Every claim in this list must have a matching (by name)
+    // volumeMount in one container in the job template. ReplicatedJob template must not have
+    // volumes with the same name as defined in this template.
+    //
+    // Generated PVC naming convention depends on TargetReplicatedJobs:
+    // - If targeting specific ReplicatedJobs: <claim-name>-<jobset-name>-<replicated-job-name>-<replica-index>-<pod-index>
```
vsoch, line 422: Depending on the length, this could get long. I think the max is 253 (this is 76). Something to keep in mind because I ran into this issue with the DNS names for JobSet.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+    // - "model-cache-training" (volume across all ReplicatedJobs)
+    Templates []corev1.PersistentVolumeClaim `json:"templates,omitempty"`
+
+    // RetentionPolicy defines the retention policy for PVCs created from this policy's templates.
+    RetentionPolicy *VolumeRetentionPolicy `json:"retentionPolicy,omitempty"`
+}
+
+// VolumeRetentionPolicy defines the retention policy for PVCs created by the JobSet.
+type VolumeRetentionPolicy struct {
+    // WhenComplete specifies what happens to PVCs when jobs complete successfully.
```
vsoch, line 436: Is there nuanced different to complete successfully vs. just complete? is the implication that if I want a generic `WhenComplete` (success or failure) I just set both of these?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+```go
+func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+    js := &jobset.JobSet{}
+    if err := r.Get(ctx, req.NamespacedName, js); err != nil {
+        return ctrl.Result{}, client.IgnoreNotFound(err)
+    }
+
+    // Reconcile volume claims before jobs
+    if len(js.Spec.VolumeClaimPolicies) > 0 {
+        if err := r.reconcileVolumeClaimPolicies(ctx, js); err != nil {
```
vsoch, line 482: So this is mutable?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+                    // Create PVC for each replica and pod index combination
+                    for replicaIdx := 0; replicaIdx < int(rjob.Replicas); replicaIdx++ {
+                        parallelism := 1
+                        if rjob.Template.Spec.Parallelism != nil {
+                            parallelism = int(*rjob.Template.Spec.Parallelism)
+                        }
+
+                        for podIdx := 0; podIdx < parallelism; podIdx++ {
+                            pvcName := generatePerJobPVCName(template.Name, js.Name, targetJobName, replicaIdx, podIdx)
+                            if err := r.ensurePVCExists(ctx, js, pvcName, &template, &policy); err != nil {
```
vsoch, line 537: If there is one error for a pod, is the probability high for an entire set? I'm imagining a group of like, 128 pods, and getting some error 128x.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+        }
+    }
+
+    if err := r.Create(ctx, pvc); err != nil {
+        r.Record.Eventf(js, corev1.EventTypeWarning, "FailedCreatePVC",
+            "Failed to create PVC %s: %v", pvcName, err)
+        return err
+    }
+
+    r.Record.Eventf(js, corev1.EventTypeNormal, "SuccessfulCreatePVC",
```
vsoch, line 601: Is there some concept of a group abstraction to reduce the verbosity?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+            }
+
+            // Determine PVC name based on targeting
+            var pvcName string
+            if len(policy.TargetReplicatedJobs) == 0 {
+                // Shared volume across all ReplicatedJobs
+                pvcName = generateSharedPVCName(template.Name, js.Name)
+            } else {
+                // Per-job volume for specific ReplicatedJob
+                pvcName = generatePerJobPVCName(template.Name, js.Name, rjob.Name, replicaIdx, podIdx)
```
vsoch, line 642: Should we check length somewhere or is that handled better by Kubernetes?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+
+2. **Completion/Failure Policies (`whenComplete`/`whenFailed`)**: Handled explicitly by the controller when JobSet reaches terminal states.
+
+```go
+func (r *JobSetReconciler) handleVolumeClaimPolicies(ctx context.Context, js *jobset.JobSet) error {
+    var allErrors []error
+
+    // Process each volume claim policy
+    for _, policy := range js.Spec.VolumeClaimPolicies {
+        var shouldDelete bool
```
vsoch, line 725: If the default of `WhenComplete` is `Delete` should this variable not be set to true, and then we set to false under specific conditions?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+                }
+
+                for replicaIdx := 0; replicaIdx < int(rjob.Replicas); replicaIdx++ {
+                    parallelism := 1
+                    if rjob.Template.Spec.Parallelism != nil {
+                        parallelism = int(*rjob.Template.Spec.Parallelism)
+                    }
+
+                    for podIdx := 0; podIdx < parallelism; podIdx++ {
+                        pvcName := generatePerJobPVCName(template.Name, js.Name, targetJobName, replicaIdx, podIdx)
```
vsoch, line 772: What about cases when an indexed job scales up and down and recreates a pod (to replace a previous one) but does not reuse the index? We would likely miss the new pods here beyond the index, and hit a few that were since deleted.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+        }
+    }
+
+    return errors.Join(allErrors...)
+}
+```
+
+### Defaulting/Validation
+
+- `VolumeClaimPolicies` API is immutable.
```
vsoch, line 788: Ah ok. If this is the case, how do we separate concern between the reconciler redundantly trying to check/create abstractions (the ensure* functions) vs. creating for new / updated pods in the set? Is there no way aside from always checking everything?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+- Each PVC template must have a corresponding `volumeMount` in at least one container.
+- Template namespace must not be set.
+- Template names must be unique within each policy
+- Template names must be valid DNS-1123 subdomain names
+- Generated PVC names must not exceed Kubernetes name length limits
```
vsoch, line 793: ah ok, here this is! :+1:

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+- [Deprecation policy][deprecation-policy]
+
+[feature gate]: https://git.k8s.io/community/contributors/devel/sig-architecture/feature-gates.md
+[maturity-levels]: https://git.k8s.io/community/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions
+[deprecation-policy]: https://kubernetes.io/docs/reference/using-api/deprecation-policy/
+-->
+
+## Implementation History
+
+- Draft KEP: October 1st 2025
```
vsoch, line 950: omg you are living in tomorrow, and tomorrow is October!! :sob: 

Oh wait, Halloween! :smile:

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+
+## Drawbacks
+
+### Complexity of Volume Management
+
+Adding volume management increases JobSet controller complexity and surface area for bugs. It
```
vsoch, line 956: Agree.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+adds additional overhead to troubleshoot volume-related issues.
+
+### Storage Dependencies
+
+Stateful JobSet becomes dependant on underlying storage infrastructure and provisioners
```
vsoch, line 961: This might be challenging if a cloud can provision some kind of storage but it isn't publicly available for inspection.

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,996 @@
...
+Additional resource tracking and quota management complexity.
+
+## Alternatives
+
+### Alternative 1: Job-level VolumeClaimTemplates
+
+Instead of managing volumes at the JobSet level, extend the Kubernetes core Job API to support
+VolumeClaimTemplates, similar to how StatefulSets work.
+
+This is a solid long-term approach, but it requires changes to Kubernetes core APIs, which implies
```
vsoch, line 974: You could demonstrate the need here, then come back with evidence for upstream.

keps/572-stateful-jobset/README.md
vsoch: I'm guessing this is the Kubeflow use case? Does it make sense to add a comment about that?

keps/572-stateful-jobset/README.md
andreyvelich: It should be `the VolumeClaimPolicies API`.

keps/572-stateful-jobset/README.md
andreyvelich: Make it clearer, how does it sound @vsoch ?
> Support volumes per ReplicatedJob and shared volume for all ReplicatedJobs within a single JobSet.

keps/572-stateful-jobset/README.md
andreyvelich: Yeah, this is possible, if users want to do that manually. This is just out of scope for the `VolumeClaimPolicies` API.

keps/572-stateful-jobset/README.md
andreyvelich: one ssh per node, let me clarify here to rename workers to nodes.

keps/572-stateful-jobset/README.md
andreyvelich: This is not necessary, you might use shared NFS-like filesystem that mounted to all Pods.

keps/572-stateful-jobset/README.md
andreyvelich: IIUC, In Kubernetes, a PVC can request multiple access modes, such as ReadWriteOnce, ReadOnlyMany, and ReadWriteMany. However, when a PVC is created with multiple access modes, Kubernetes will attempt to bind it to a PersistentVolume (PV) that supports at least one of the requested modes. The actual access mode used will be determined by the capabilities of the bound PV.

keps/572-stateful-jobset/README.md
andreyvelich: Good point, let me clarify this in defaults.

keps/572-stateful-jobset/README.md
andreyvelich: Oh, actually, it is a mistake here. As I said in the Defaulting/Validation section the default value is `Retain` similar to StatefulSets.
@kannon92 @ahg-g @GiuseppeTT Any thoughts on this ? We can change defaults to `Delete` if we think that this is more applicable for batch workloads.

keps/572-stateful-jobset/README.md
andreyvelich: Hmm, it is a good point. How do we want to manage labels for PVCs that we Retain after JobSet gets deleted?
Currently, in the KEP we just apply label to the PVC with the first JobSet that created this claim:
```
jobset.sigs.k8s.io/jobset-name=jobset-name
```

keps/572-stateful-jobset/README.md
andreyvelich: I think here we should wait for JobSet Complete status. It would be tricky to implement other conditions given the FailurePolicy and SuccessPolicy APIs.

keps/572-stateful-jobset/README.md
andreyvelich: @vsoch Do you mean whether PVCs are mutable after creation?

keps/572-stateful-jobset/README.md
andreyvelich: good point, shall we directly return error if one of the creation fails ?

keps/572-stateful-jobset/README.md
andreyvelich: Not that I am aware of. Maybe we should reduce number of events, since currently it will generate x number of events for every PVC we create (e.g. 100 replicas for ReplicatedJob = 100 events)

keps/572-stateful-jobset/README.md
andreyvelich: Check it here: https://github.com/kubernetes-sigs/jobset/blob/8bea072441ef00ce57cab72ed9316acf18228175/keps/572-stateful-jobset/README.md?plain=1#L879-L882

keps/572-stateful-jobset/README.md
andreyvelich: Yeah, let's clarify the default values here: https://github.com/kubernetes-sigs/jobset/pull/1053#discussion_r2395891754

keps/572-stateful-jobset/README.md
andreyvelich: good point, do we trigger JobSet reconcile loop today when we run elastic indexed jobs ?
IIUC, currently JobSet doesn't support elastic jobs: https://github.com/kubernetes-sigs/jobset/issues/463
cc @kannon92 @tenzen-y @ahg-g

keps/572-stateful-jobset/README.md
vsoch: ah interesting. So the "which one" decision gets forwarded along.

keps/572-stateful-jobset/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+replaces:
+  - "NA"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.10.0"
```
kannon92:
```suggestion
latest-milestone: "v0.11.0"
```

keps/572-stateful-jobset/kep.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.11.0"
```
kannon92, line 29: I'm not entirely sure what alpha means for JobSet.

Are we going to introduce a feature gate or just let people use the API as is? So its beta effectively?

keps/572-stateful-jobset/README.md
```diff
@@ -0,0 +1,998 @@
...
+- Support per Job index volume claim templates.
+- Cross-JobSet volume sharing. Sharing volumes between different JobSet instances is out of scope.
+- Orchestration for PersistentVolumes. JobSet users should rely on StorageClasses to provision
+  and delete PVs.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Distributed ML Training with Per-Worker Checkpoints
```
kannon92, line 76: Do you have a use case where each replicated job can have a different PVC template?

Thinking of the case where targetReplicatedJobs has more than 1 entry.

Your examples are either the first replicated job has the PVC or all of them.

kannon92: Ah I see #story-3 actually.

keps/572-stateful-jobset/kep.yaml
andreyvelich: Currently, we don't have feature gate support in JobSet. I marked this as alpha since we still use `v1alpha2` API for JobSet.

kannon92: I don't agree with that conclusion.

We have GA features in v1alpha2. In theory that would mean everything in JobSet is alpha?

I don't think that reflects the reality for our users.

I'm personally fine with releasing this feature without a feature gate as its an opt-in API.

But we also could consider introducing a feature gate.

andreyvelich:
After today's discussion in Batch WG, we decided to:
- Remove WhenComplete and WhenFailed from retention policy since we should not tie PVC to JobSet statuses.
- Make `Delete` as a default retention policy in opposite to StatefulSet, where the default retention policy is `Retain`.

cc @tenzen-y @mimowo @mwielgus @kannon92

andreyvelich:
For some reason this PR is out of sync with the branch: https://github.com/andreyvelich/jobset/commits/kep-572-stateful-jobset/
Let me try to re-open it.

andreyvelich:
I have to recreate it since GitHub doesn't allow to re-open the PR: https://github.com/kubernetes-sigs/jobset/pull/1062

PR #1063, opened 2025-10-09
restrict controller-manager Secrets access to jobset install namespace

### LINKED ISSUE #1061: Why do we need cluster‑wide Secret permissions?
Currently, the JobSet controller only uses Secrets when updating self‑signed certificates.
A single shared Secret object is mounted onto all JobSet controller pods and used as the webhook server certificate.
In this scenario, the Secret and the controller are deployed in the same namespace ({{ .Release.Namespace }}), so namespace‑scoped Secret update permissions should be sufficient.

Why then is cluster‑scoped Secret permission declared(charts/jobset/templates/controller/cluster_role.yaml)?
```yaml
apiVersion: rbac.authorization.k8s.io/v1
kind: ClusterRole
metadata:
  name: {{ include "jobset.controller.clusterRole.name" . }}
  labels:
    {{- include "jobset.controller.labels" . | nindent 4 }}
rules:
- apiGroups:
  - ""
  resources:
  - secrets
  verbs:
  - get
  - list
  - update
  - watch
```

In addition to restricting Secret permissions to the namespace level, another possible approach is to declare the specific Secret name to be authorized directly in the ClusterRole. Reference: charts/jobset/templates/controller/deployment.yaml
```yaml
- name: webhook-cert
  secret:
    secretName: {{ include "jobset.webhook.secret.name" . }}
    defaultMode: 420
```
We would only need to add the following (**resourceNames**) to charts/jobset/templates/controller/cluster_role.yaml:

```yaml
- apiGroups:
  - ""
  resources:
  - secrets
  resourceNames:
  - {{ include "jobset.webhook.secret.name" . }}
  verbs:
  - get
  - list
  - update
  - watch
```
This would scope the Secret access in the ClusterRole to the specific webhook-cert Secret, rather than all Secrets.

If I am missing any background information, please let me know.
Thank you!

bcfre:
#### What type of PR is this?
/kind feature

Currently, the JobSet controller only uses Secrets when updating self‑signed certificates.
A single shared Secret object is mounted onto all JobSet controller pods and used as the webhook server certificate.
In this scenario, the Secret and the controller are deployed in the same namespace ({{ .Release.Namespace }}), so namespace‑scoped Secret update permissions should be sufficient.

A similar approach is taken to the existing **leader-election-role** Role and RoleBinding. Specifically:
- Remove Secrets RBAC marker from pkg/util/cert/cert.go, in turn removing Secrets permissions from the generated ClusterRole.
- Add Role manager-secrets-role and RoleBinding manager-secrets-rolebinding to config/components/rbac.
- Update Helm chart templates for the above.

Fixes #1061

Reference related PR: https://github.com/kubernetes-sigs/kueue/pull/7188

```release-note
NONE
```

### DIFF
--- a/charts/jobset/templates/controller/cluster_role.yaml
+++ b/charts/jobset/templates/controller/cluster_role.yaml
@@ -53,15 +53,6 @@ rules:
 - apiGroups:
   - ""
   resources:
-  - secrets
-  verbs:
-  - get
-  - list
-  - update
-  - watch
-- apiGroups:
-  - ""
-  resources:
   - services
   verbs:
   - create
--- a/charts/jobset/templates/controller/role.yaml
+++ b/charts/jobset/templates/controller/role.yaml
@@ -34,6 +34,15 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
+  - update
+  - watch
 {{- if .Values.controller.leaderElection.enable }}
 - apiGroups:
   - coordination.k8s.io
--- a/config/components/rbac/kustomization.yaml
+++ b/config/components/rbac/kustomization.yaml
@@ -9,6 +9,8 @@ resources:
 - role_binding.yaml
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
+- manager_secrets_role.yaml
+- manager_secrets_role_binding.yaml
 # Comment the following 4 lines if you want to disable
 # protecting your /metrics endpoint.
 - auth_proxy_role.yaml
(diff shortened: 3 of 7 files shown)


GiuseppeTT:
/lgtm

@kannon92 I reviewed the PR and it makes sense. It correctly updates the Helm and Kustomize files to move the Secret permissions from the `ClusterRole` to a `Role`. It also updates `pkg/util/cert/cert.go` to ensure the overly broad `ClusterRole` rule is not regenerated.

That being said, I'm not familiar with how JobSet uses its Secret permissions, so I'm leaving the final confirmation on that aspect to you : )

kannon92:
Agree with you!

PR looks good.

/approve

PR #1065, opened 2025-10-14
Use explicit bash image path instead of ambiguous shortname

ardaguclu:
#### What type of PR is this?
/kind cleanup

`bash:latest` image is used in e2e tests heavily. However, these tests do not reliably work on the environments that do not allow ambiguous image registries (i.e. due to the enforced policy). For example, if `bash:latest` is hosted in multiple registries, node simply errors out without pulling the image. 

In order to fix this issue and make e2e tests working on all environment, this PR uses explicit image path.

```release-note
None
```


kannon92:
Code looks good.

failure is related to https://github.com/kubernetes-sigs/lws/issues/677.

we either need to upgrade to Golang 1.25 or pin envtest as main branch switched to Golang 1.25

PR #1069, opened 2025-10-17
enable nobools for future api calls. ignore current violations though.

kannon92:
#### What type of PR is this?
/kind cleanup

Add more kube-api-linter fixes

Fixes #

For API promotion we should probably apply this rule but this would be a breaking change for v1alpha2.

```release-note
NONE
```


GiuseppeTT:
> For API promotion we should probably apply this rule but this would be a breaking change for v1alpha2.

Agreed

PR #1074, opened 2025-10-24
Remove pod create permission from controller

### LINKED ISSUE #1073: Why do we request create permission for Pods
In the current JobSet design, the actual Pod creation is delegated to the Job operator.
We only modify or delete Pods in two scenarios:

- When handling Pods in topology domain exclusive mode.
- When performing mutating/validating logic in the webhook.

Therefore, in [pkg/controllers/pod_controller.go](https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/pod_controller.go#L112) we don’t really need create permission for Pods.
```
// +kubebuilder:rbac:groups=core,resources=pods,verbs=get;list;watch;create;update;patch;delete
```
Should we consider narrowing down the current permission scope? Personally, I believe we should follow the principle of least privilege, and by removing create we can better reflect the relationship between JobSet and Job.

If I am missing any background information, please let me know.
Thank you!

bcfre:
#### What type of PR is this?
/kind cleanup

In the current JobSet design, the actual Pod creation is delegated to the Job operator.
We only modify or delete Pods in two scenarios:

- When handling Pods in topology domain exclusive mode.
- When performing mutating/validating logic in the webhook.

We no need for pod create permissions.

Fixes #1073

NONE

```release-note
NONE
```

### DIFF
--- a/charts/jobset/templates/controller/cluster_role.yaml
+++ b/charts/jobset/templates/controller/cluster_role.yaml
@@ -43,7 +43,6 @@ rules:
   resources:
   - pods
   verbs:
-  - create
   - delete
   - get
   - list
--- a/config/components/rbac/role.yaml
+++ b/config/components/rbac/role.yaml
@@ -25,6 +25,16 @@ rules:
   - ""
   resources:
   - pods
+  verbs:
+  - delete
+  - get
+  - list
+  - patch
+  - update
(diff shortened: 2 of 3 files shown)


ahg-g:
/lgtm
/approve

thanks

PR #1076, opened 2025-10-24
jobset controller: Emit event on create job failed

tchap:
#### What type of PR is this?

/kind feature

Emit a warning event when job creation fails.

N/A

Let me know if this is desired and I will add some tests if needed.

```release-note
Added a warning event emitted when job creation fails
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -537,6 +537,7 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
 		// Create jobs as necessary.
 		if err := r.createJobs(ctx, js, jobs); err != nil {
 			log.Error(err, "creating jobs")
+			r.Record.Eventf(js, corev1.EventTypeWarning, constants.JobCreationFailedReason, err.Error())
 			return err
 		}
 
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -30,6 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -2495,6 +2496,74 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 					return false
 				}
 				return !fresh1.DeletionTimestamp.IsZero() && fresh2.DeletionTimestamp.IsZero()
+			}, timeout, interval).Should(gomega.BeTrue())
+		})
+	})
+
+	ginkgo.When("a JobSet is created with an invalid Job template", func() {
+		ginkgo.It("should emit a warning event", func() {
+			ns := &corev1.Namespace{
+				ObjectMeta: metav1.ObjectMeta{
+					GenerateName: "jobset-ns-",
+				},
+			}
+
+			ginkgo.By("creating namespace")
+			gomega.Expect(k8sClient.Create(ctx, ns)).To(gomega.Succeed())
+
+			defer func() {
+				gomega.Expect(testutil.DeleteNamespace(ctx, k8sClient, ns)).To(gomega.Succeed())
+			}()
+
+			ginkgo.By("creating jobset with an invalid job template")
+			podSpec := testing.TestPodSpec
+			podSpec.Containers[0].Image = ""
+
+			js := testing.MakeJobSet("invalid-jobset", ns.Name).
+				ReplicatedJob(testing.MakeReplicatedJob("replicated-job-a").
+					Job(testing.MakeJobTemplate("test-job-A", ns.Name).PodSpec(podSpec).Obj()).
+					Replicas(1).
+					Obj()).
+				Obj()
+
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
+
+			ginkgo.By("checking that a warning event is emitted")
+			gomega.Eventually(func() (bool, error) {
+				var events corev1.EventList
+				if err := k8sClient.List(ctx, &events, client.InNamespace(ns.Name), client.MatchingFieldsSelector{
+					Selector: fields.AndSelectors(
+						fields.OneTermEqualSelector("involvedObject.kind", "JobSet"),
+						fields.OneTermEqualSelector("involvedObject.name", js.Name),
+						fields.OneTermEqualSelector("reason", "JobCreationFailed"),
+					),
+				}); err != nil {
+					return false, err
+				}
+
+				if len(events.Items) < 1 {
+					return false, nil
+				}
+
+				event := events.Items[0]
+				gomega.Expect(event.Type).To(gomega.Equal(corev1.EventTypeWarning))
+				return true, nil
+			}, timeout, interval).Should(gomega.BeTrue())
+
+			ginkgo.By("checking that the events are accumulated")
+			gomega.Eventually(func() (bool, error) {
+				var events corev1.EventList
+				if err := k8sClient.List(ctx, &events, client.InNamespace(ns.Name), client.MatchingFieldsSelector{
+					Selector: fields.AndSelectors(
+						fields.OneTermEqualSelector("involvedObject.kind", "JobSet"),
+						fields.OneTermEqualSelector("involvedObject.name", js.Name),
+						fields.OneTermEqualSelector("reason", "JobCreationFailed"),
+					),
+				}); err != nil {
+					return false, err
+				}
+
+				return len(events.Items) == 1 && events.Items[0].Count > 1, nil
 			}, timeout, interval).Should(gomega.BeTrue())
 		})
 	})


tchap:
Hmm, it seems `constants.JobCreationFailedReason` was in use before, wondering why it's unused now. Also perhaps it might be better to batch errors somehow and not emit an event for each job...

pkg/controllers/jobset_controller.go
```diff
@@ -577,7 +577,9 @@ func (r *JobSetReconciler) createJobs(ctx context.Context, js *jobset.JobSet, jo
 		if err := r.Create(ctx, job); err != nil {
 			lock.Lock()
 			defer lock.Unlock()
-			finalErrs = append(finalErrs, fmt.Errorf("job %q creation failed with error: %v", job.Name, err))
+			err = fmt.Errorf("job %q creation failed with error: %w", job.Name, err)
+			finalErrs = append(finalErrs, err)
+			r.Record.Event(js, corev1.EventTypeWarning, constants.JobCreationFailedReason, err.Error())
```
ahg-g: emit the event outside the parallel loop pls, the errors are already aggregated.

tchap: Moved the event next to the log statement outside the `createJobs` call actually.

kannon92:
This change is needed. Thanks for adding this. We are getting quite a few reports that finding that job failing is not intuitive and requires viewing the controller logs.

I also wonder if we should consider a condition since events can be ephemeral/brittle in a chatty cluster. 

Maybe a condition can be a follow up.

tchap:
> This change is needed. Thanks for adding this. We are getting quite a few reports that finding that job failing is not intuitive and requires viewing the controller logs.
> 
> I also wonder if we should consider a condition since events can be ephemeral/brittle in a chatty cluster.
> 
> Maybe a condition can be a follow up.

Yeah, perhaps a followup. I am not sure conditions map well to issues with particular resources, though. Seems to me issues with particular jobs are on the other hand too ephemeral for a condition 🤷🏻‍♂️

tchap:
I am wondering whether we shouldn't throttle the events somehow. There will be too many since this error is not gonna get fixed on its own and an event will be emitted on every reconciliation retry...

kannon92:
> I am wondering whether we shouldn't throttle the events somehow. There will be too many since this error is not gonna get fixed on its own and an event will be emitted on every reconciliation retry...

Good point.

/hold

JobSet does have a way to batch events if the status update needs to change. Maybe you can look into something like that so we only emit a single event on job failed.

tchap:
> Good point.
> 
> /hold
> 
> JobSet does have a way to batch events if the status update needs to change. Maybe you can look into something like that so we only emit a single event on job failed.

Yeah, that can help and I will use that. But for the rest I mean something like, only post this event every N minutes.

pkg/controllers/jobset_controller.go
```diff
@@ -652,7 +658,12 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js
 
 		// Create headless service.
 		if err := r.Create(ctx, &headlessSvc); err != nil {
-			r.Record.Eventf(js, corev1.EventTypeWarning, constants.HeadlessServiceCreationFailedReason, err.Error())
+			enqueueEvent(updateStatusOpts, &eventParams{
```
tchap: Migrates this to use the helper as well.

pkg/controllers/jobset_controller.go
```diff
@@ -556,6 +569,56 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
 	return nil
 }
 
+func (r *JobSetReconciler) recordJobCreationFailed(js *jobset.JobSet, err error) {
+	key := types.NamespacedName{Namespace: js.GetNamespace(), Name: js.GetName()}.String()
+	r.recordThrottledEvent(key, js, js.Spec.ReplicatedJobs, corev1.EventTypeWarning, constants.JobCreationFailedReason, err.Error())
```
tchap: Passing `js.Spec.ReplicatedJobs` as `hashObject` basically means the event is throttled as long as the object remains the same. So changing the spec causes the event to be emitted immediately.

pkg/controllers/jobset_controller.go
```diff
@@ -556,6 +569,56 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
...
+	terminatedCh := make(chan struct{})
+	go func() {
+		defer close(terminatedCh)
+		ticker := time.NewTicker(eventTimestampCacheResetPeriod)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ticker.C:
+				r.eventTimestampLock.Lock()
+				r.eventTimestamps.Clear()
```
tchap: Not possible to implement a more clever mechanism with the LRU cache, but just resetting it periodically works just fine, I think. In the worst case you get 2 events very close to each other, who cares.

pkg/controllers/jobset_controller.go
```diff
@@ -556,6 +569,51 @@ func (r *JobSetReconciler) reconcileReplicatedJobs(ctx context.Context, js *jobs
...
+}
+
+func (r *JobSetReconciler) RunCleaner(ctx context.Context) {
+	ticker := time.NewTicker(eventTimestampCacheResetPeriod)
+	defer ticker.Stop()
+	for {
+		select {
+		case <-ticker.C:
+			r.eventTimestampLock.Lock()
+			r.eventTimestamps.Clear()
```
tchap: Not possible to implement a more clever mechanism with the LRU cache, but just resetting it periodically works just fine, I think. In the worst case you get 2 events very close to each other, who cares.

tchap:
I pushed some event throttling PoC, let me know what you think. This is definitely not finished. Like, I defined the time intervals as more or less random constants for now, also there are no tests.

pkg/controllers/jobset_controller.go
```diff
@@ -48,6 +51,12 @@ import (
 	"sigs.k8s.io/jobset/pkg/util/placement"
 )
 
+const (
+	eventThrottlePeriod            = 1 * time.Minute
```
kannon92: I would move this to our component config and set these as defaults. That way if there are any issues someone can tune these.

tchap: Done. I also actually pulled the throttling logic into a separate package. Tests still largely missing.

pkg/controllers/jobset_controller.go
```diff
@@ -89,8 +93,14 @@ type eventParams struct {
...
 }
 
-func NewJobSetReconciler(client client.Client, scheme *runtime.Scheme, record record.EventRecorder) *JobSetReconciler {
-	return &JobSetReconciler{Client: client, Scheme: scheme, Record: record, clock: clock.RealClock{}}
+func NewJobSetReconciler(client client.Client, scheme *runtime.Scheme, record record.EventRecorder, throttledRecord ThrottledEventRecorder) *JobSetReconciler {
+	return &JobSetReconciler{
+		Client:          client,
+		Scheme:          scheme,
+		Record:          record,
+		throttledRecord: throttledRecord,
```
tchap: Not entirely sure why Record is public, made throttledRecord private...

api/config/v1alpha1/configuration_types.go
```diff
@@ -54,6 +54,20 @@ type ControllerManager struct {
...
+	// Events contains the controller event emitting configuration.
+	Events ControllerEvents `json:"events,omitempty"`
+}
+
+type ControllerEvents struct {
+	// ThrottlingWindowSize specified the time window during which a particular event is emitted at most once.
+	ThrottlingWindowSize *metav1.Duration `json:"throttlingWindowSize,omitempty"`
+
+	// ThrottlingCacheSize specified the number of unique events being tracked in the throttling cache.
+	ThrottlingCacheSize *int `json:"throttlingCacheSize,omitempty"`
```
kannon92:
```suggestion
	ThrottlingCacheSize *int32 `json:"throttlingCacheSize,omitempty"`
```

k8s conventions ask to not use general integers.

kannon92:
@GiuseppeTT can you do a first path for review?

api/config/v1alpha1/configuration_types.go
tchap: Amended.

GiuseppeTT:
>  @GiuseppeTT can you do a first path for review?

I'm wondering if we need event throttling.

The Kubernetes API convention mentions [event aggregation for frequent events](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#events). Looking into the controller manager code, the [EventBroadcaster](https://github.com/kubernetes-sigs/controller-runtime/blob/7a1b16d039d204e7e4c5c8df57124d4e060a55f0/pkg/manager/manager.go#L512) is set up, which should take care of event aggregation automatically.

Before making a custom event throttling, I'd first test what happens if you just record the events. Will it generate any noise? Or will it correctly aggregate?

tchap:
> I'm wondering if we need event throttling.

Yeah, well, it seems we don't. Weird, I thought I checked the code before 😞 Anyway, I will test that there really is just a single event.

Thanks for catching this.

tchap:
@GiuseppeTT Rolled back throttling and added a test to check for event accumulation.

tchap:
Throttling is not needed, it's implemented in the broadcaster.

/unhold

test/integration/controller/jobset_controller_test.go
```diff
@@ -2498,6 +2499,73 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+					return false, nil
+				}
+
+				event := events.Items[0]
+				gomega.Expect(event.Type).To(gomega.Equal(corev1.EventTypeWarning))
+				gomega.Expect(event.Reason).To(gomega.Equal("JobCreationFailed"))
+				return true, nil
+			}, timeout, interval).Should(gomega.BeTrue())
+
+			ginkgo.By("checking that the events are accumulated")
```
tchap, line 2553: Strictly speaking the following test doesn't check there is just a single API request to add the events, but the correlation broadcaster setup can be seen here: https://github.com/kubernetes/client-go/blob/v0.34.1/tools/record/event.go#L276

tchap: Ensuring this properly would require mocking the `EventSink` and creating an `EvenBroadcaster` manually:
https://pkg.go.dev/k8s.io/client-go/tools/record#EventSink

GiuseppeTT:
Great work! I'll review the new changes shortly.

About checking API requests, one option is doing an experiment (a manual test) to confirm things are working as expected. For instance:

1. Create a k8s cluster (a kind cluster may be enough)
2. Create a JobSet that triggers your event repetitively 
3. Check audit logs for API requests from the controller Pod

IMO the experiment is not a blocker for the PR since your test confirms the events are grouped. Worst case scenario we add your event throttler in a new PR : )

test/integration/controller/jobset_controller_test.go
```diff
@@ -2498,6 +2499,73 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+					),
+				}); err != nil {
+					return false, err
+				}
+
+				if len(events.Items) < 1 {
+					return false, nil
+				}
+
+				event := events.Items[0]
```
GiuseppeTT, line 2548: This is brittle. What happens if another event (e.g., `HeadlessServiceCreationFailed`) is emitted before the event `JobCreationFailed`?

tchap: Added reason to the selector.

PR #1079, opened 2025-10-27
Validate coordinator label value

### LINKED ISSUE #1056: Improve validation webhook to block JobSets that will lead to invalid labels

Improve validation webhook to block JobSets that will lead to invalid label values.

One useful rule is blocking JobSets such that [this function](https://github.com/kubernetes-sigs/jobset/blob/193ee03ba68ed2f60c4017a3d637bfb3af2ccc05/pkg/controllers/jobset_controller.go#L1086-L1090) will return a string larger than 63 characters.

**Why is this needed**:

Improve user experience.

For instance, the label `jobset.sigs.k8s.io/coordinator` is set to `{jobset name}-{replicated job name}-{job index}-{pod index}.{subdomain}` which can easily exceed the limit of 63 characters. This error is only detected in the controller logs when the creation of a child Job fails due to the invalid label value. A better approach is to check if the JobSet config will lead to a invalid workload during the validation webhook check.

GiuseppeTT:
#### What type of PR is this?

/bug

Add validation to coordinator label value at JobSet creation.

Fixes #1056

```release-note
Add validation to coordinator label value at JobSet creation. Now JobSets with a spec that will lead to a invalid label value for label `jobset.sigs.k8s.io/coordinator` will be blocked from being created by the JobSet webhook.
```

### DIFF
--- a/test/integration/webhook/jobset_webhook_test.go
+++ b/test/integration/webhook/jobset_webhook_test.go
@@ -539,7 +539,7 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 			},
 			jobSetCreationShouldFail: false,
 		}),
-		ginkgo.Entry("spec that will lead to invalid coordinator label value (in this case too long) should be rejected", &testCase{
+		ginkgo.Entry("spec that will lead to invalid coordinator label value (in this case too long due to long JobSet name) should be rejected", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
 				return testing.MakeJobSet(strings.Repeat("a", 60), ns.Name).
 					Coordinator(&jobset.Coordinator{


GiuseppeTT:
I also did a manual test to confirm it fixes the bug.

# Test

Try to create a JobSet with a long name.

```bash
kubectl apply -f - <<EOF
kind: JobSet
metadata:
  name: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
spec:
  coordinator:
    replicatedJob: replicatedjob-a
    jobIndex: 0
    podIndex: 0
  replicatedJobs:
  - name: replicatedjob-a
    replicas: 1
    template:
      spec:
        completions: 1
        parallelism: 1
        template:
          spec:
            containers:
            - name: worker
              image: busybox
              command: ['/bin/sh', '-c']
              args:
              - |
                echo "Start"
                sleep infinity
EOF
```

Error

```
Error from server (Forbidden): error when creating "STDIN": admission webhook "vjobset.kb.io" denied the request: spec.replicatedJobs[0].name: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-replicatedjob-a-0": JobSet name is too long, job names generated for this JobSet will exceed 63 characters
spec.replicatedJobs[0].name: Invalid value: "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-replicatedjob-a-0": JobSet name is too long, pod names generated for this JobSet will exceed 63 characters
spec will lead to invalid label value "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-replicatedjob-a-0-0.aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa" for coordinator label "jobset.sigs.k8s.io/coordinator" (long JobSet / ReplicatedJob / SubDomain name?): must be no more than 63 characters
```

pkg/controllers/jobset_controller.go
```diff
@@ -1083,9 +1083,9 @@ func exclusiveConditions(cond1, cond2 metav1.Condition) bool {
 	return inProgressAndCompleted || completedAndInProgress
 }
 
-// coordinatorEndpoint returns the stable network endpoint where the coordinator pod can be reached.
+// CoordinatorEndpoint returns the stable network endpoint where the coordinator pod can be reached.
 // This function assumes the caller has validated that jobset.Spec.Coordinator != nil.
-func coordinatorEndpoint(js *jobset.JobSet) string {
+func CoordinatorEndpoint(js *jobset.JobSet) string {
```
GiuseppeTT, line 1088: I had to export this function to use in the `webhooks` package

pkg/controllers/jobset_controller.go
```diff
@@ -837,7 +837,7 @@ func GetSubdomain(js *jobset.JobSet) string {
 	// This must be done in the controller rather than in the request-time defaulting, since if a JobSet
 	// uses generateName rather than setting the name explicitly, the JobSet name will still be an empty
 	// string at that time.
-	if js.Spec.Network.Subdomain != "" {
+	if js.Spec.Network != nil && js.Spec.Network.Subdomain != "" {
```
GiuseppeTT, line 840: I added this because the test was failing. In either case checking a subfield of a possibly nil struct is not right.

kannon92:
Can you add an integration test [here](https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/webhook/jobset_webhook_test.go)?

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1583,6 +1583,37 @@ func TestValidateCreate(t *testing.T) {
 				fmt.Errorf("coordinator pod index 2 is invalid for replicatedJob replicatedjob-a job index 0"),
 			),
 		},
+		{
+			name: "coordinator label is too long",
```
kannon92: Can you add a test case where you specify Network also in the spec?

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -538,5 +539,21 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 			},
 			jobSetCreationShouldFail: false,
 		}),
+		ginkgo.Entry("spec that will lead to invalid coordinator label value (in this case too long due to long JobSet name) should be rejected", &testCase{
```
kannon92, line 542: Its worth adding more test cases here.

kannon92:
/hold

My comment is a nit (unit seems to cover this code well enough) but there is gap in coverage in integration.

/approve
/lgtm

test/integration/webhook/jobset_webhook_test.go
GiuseppeTT: This was a conscious decision. Since the unit tests are covering all cases, I added just one case to the integration test.  The reasoning is that integration tests are more reliable but lose on execution speed, cost and maintainability.

I'm happy to add the missing cases if you think the tradeoff is worth it : )

PR #1081, opened 2025-10-29
Refactor kustomize manifests to keep the same structure as Helm chart

ChenYi015:
#### What type of PR is this?

/kind cleanup

- Reorganize the Kustomize manifests to keep the same structure as Helm chart.

It will make PR #824 easier to review after this refactoring.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -92,11 +92,9 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) \
-		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
 		crd:generateEmbeddedObjectMeta=true output:crd:artifacts:config=config/components/crd/bases\
 		paths="./api/..."
 	$(CONTROLLER_GEN) \
-		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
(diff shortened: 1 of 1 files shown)


GiuseppeTT:
Running `make generate` shows a missing file `config/components/webhook/manifests.yaml`.

GiuseppeTT:
Can you provide more details? Like, if you are only moving yaml content around, be explicit about where content "X" was and where it will be and how it makes kustomize consistent with the helm template structure. If you are also adding or removing yaml content, be explicit about it.

Also, how is this PR going to help with #824? Is it a subset of #824?

Makefile
```diff
@@ -92,11 +92,9 @@ help: ## Display this help.
 .PHONY: manifests
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) \
-		rbac:roleName=manager-role output:rbac:artifacts:config=config/components/rbac\
```
GiuseppeTT, line 95: Why are you removing it? What happens if we modify the rbacs in `./api/...` and `./pkg/...`?

GiuseppeTT:
I thought more about this. IMO a better approach is just getting rid of kustomize and use only the helm chart. We can do that if we use the helm chart to generate the release manifest. See https://github.com/kubernetes-sigs/jobset/issues/820#issuecomment-3465328087

kannon92:
/close

We ended up bringing in yaml-processor and this PR is no longer needed.

Thank you for the discussion this created!

Issue #976, opened 2025-08-05
Finer control over maxRestarts for RestartJobSet vs RecreateJob


carreter:
**What would you like to be added**:
Currently, the `RecreateJob` and `RestartJobSet` failure policies both increment the same `.status.restartsCountTowardsMax` whenever they are triggered. When said counter reaches the value of `.spec.failurePolicy.maxRestarts`, triggering either the `RecreateJob` or `RestartJobSet` will mark the entire JobSet as failed.

It'd be desirable to allow separate counters for `RecreateJob` and `RestartJobSet`. For example, imagine a JobSet with three ReplicatedJobs: `workers-heavy`, `workers-light`, and `controllers`. We want the following behaviours:
- failures in `workers-heavy` trigger `RecreateJob` up to 3 times for each child `Job`, as these are expensive to recreate
- failures in `workers-light` trigger `RecreateJob` up to 10 times for each child `Job`, as these are easy and fast to recreate
- failures in `controllers` trigger `RestartJobSet` up to 3 times total

The configuration could look something like this
```yaml
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata:
  name: test
spec:
  failurePolicy:
    maxRestarts: 3 # How many times can a full JobSet restart happen?
    rules:
    - action: RecreateJob
      maxRecreatesPerIndex: 4 # How many times can a child job in workers-heavy be recreated before failing the JobSet?
      targetReplicatedJobs:
      - workers-heavy
    - action: RecreateJob
      maxRecreatesPerIndex: 10 # How many times can a child job in workers-light be recreated before failing the JobSet?
      targetReplicatedJobs:
      - workers-light
```

**Why is this needed**:
Not all ReplicatedJob failures are the same :) In the previous example, `workers-heavy` take a long time to recreate and as such should fail in fewer attempts than `workers-light`, which can be recreated quickly more times without incurring too heavy a cost. 

This enhancement requires the following artifacts:

- [ ] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

GiuseppeTT:
I wonder if it makes more sense to support a counter per `replicatedJob` instead of per `failureActionRule`. Would that be enough for your use case?

carreter:
[...]

PR #1083, opened 2025-11-04
KEP 467: Fast failure recovery with in place restarts

### LINKED ISSUE #467: Investigate restarting pods in-place as a mechanism for faster failure recovery
A faster restart/failure recovery mechanism that doesn't involve recreating all the child jobs, rescheduling all of the pods, etc.

**Why is this needed**:
For faster failure recovery, reducing downtime for batch/training workloads which experience some kind of error.

GiuseppeTT:
#### What type of PR is this?

/kind documentation

Add the KEP for in place restarts.

Fixes #467

Refer to the [design doc](https://docs.google.com/document/d/16zexVooHKPc80F4dVtUjDYK9DOpkVPRNfSv0zRtfFpk/edit?usp=sharing) for more details.

cc @ahg-g 

```release-note
NONE
```

### DIFF
--- a/keps/467-InPlaceRestart/README.md
+++ b/keps/467-InPlaceRestart/README.md
@@ -579,7 +579,7 @@ The image of the agent sidecar will be buildable from code in the JobSet repo (e
 
 ### Implementation - Permissions for the agent sidecar
 
-The agent sidecar requires permissions to watch its parent JobSet. It also requires permissions to update its Pod in-place restart attempt, which can be integrated with a validating admission policy to make sure only the in-place restart attempt annotation can be modified. The following manifest shows an example of how a service account can be set up for the worker Pod. It will be part of the documentation of in-place restart since it's up to the user to integrate these extra permissions to any existing service account.
+The agent sidecar requires permissions to watch its parent JobSet. It also requires permissions to update its Pod in-place restart attempt, which can be integrated with a validating admission policy to make sure only the in-place restart attempt annotation can be modified. The following manifest shows an example of how a service account can be set up for the worker Pod. It will be part of the documentation of in-place restart since it's up to the user to integrate these extra permissions to any existing service account. A helm chart will also be created to make it easier to add the extra permissions.
 
 ```yaml
 # Service account
@@ -827,6 +827,7 @@ milestones with these graduation criteria:
 - Address reviews and bug reports from alpha users
 - Add a way to automatically build and release the agent sidecar image
 - Add a way to add the agent sidecar automatically
+- Add a helm chart for installing the extra permissions required by the agent sidecar
 
 #### GA
 


keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,775 @@
...
+    maxRestarts: ... # Maximum number of full restarts
+    restartStrategy: InPlaceRestart # Enable in place restart
+  replicatedJobs:
+  - name: workers
+    replicas: ... # Number of Node pools
+    template:
+      spec:
+        completions: ... # Number of Nodes per Node pool
+        parallelism: ... # Number of Nodes per Node pool
+        backoffLimit: 2147483647 # MaxInt32. Avoid full recreation by not failing Jobs due to failed Pods and containers
```
kannon92, line 215: Is there anything different if someone specifies 2147483646 (-1)?

Is it just the point that this needs to be a really high backOffLimit or are we keying off MaxInit to change behavior?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,775 @@
...
+  //
+  // When a Job has failed, fall back to action `Recreate` and execute the matching failure policy rule.
+  InPlaceRestart JobSetRestartStrategy = "InPlaceRestart"
+)
+```
+
+Changes to the JobSet status.
+
+```go
+// README: Add new fields `deprecatedEpoch` and `syncedEpoch`.
```
kannon92: I really don't like the idea of epochs in the JobSet API.

This reminds me of the same issues I have with https://github.com/kubeflow/trainer/pull/2905.

For kubeflow I don't think its as big as deal as the project is designed for training workloads. JobSet was not designed to be only meant for AI training.

With adding epocs, we are forcing JobSet to be a API designed for tracking neural networking training as far as I understand.

How would epocs work with a generic MPI workload that does not do training?

keps/467-InPlaceRestart/README.md
GiuseppeTT: Thanks for the feedback : )

`MaxInt32` (2147483647) is just a big number. `MaxInt32 - 1` (2147483646) could be used but arguably the maximum value should be picked.

For context, I based my choice on the Job API. When `backoffLimitPerIndex` is specified, `backoffLimit` defaults to `MaxInt32` (see [here](https://kubernetes.io/docs/concepts/workloads/controllers/job/#pod-backoff-failure-policy)). There is nothing in the Job documentation nor in the Job controller code that treats `MaxInt32` differently from `MaxInt32 - 1`.

keps/467-InPlaceRestart/README.md
GiuseppeTT: I think there is a misunderstanding : )

What I'm calling "epoch" in the proposal is not the same "epoch" from deep learning, which means "a complete pass through the entire training dataset". Instead, I'm referring to the generic term "epoch" from distributed systems, which means something like "a logical timestamp that processes can use to determine if they are in sync". It has a similar meaning to `jobset.sigs.k8s.io/restart-attempt` from JobSet.

Here is an example of the term "epoch" being used in the book "Distributed Systems" from Maarten van Steen and Andrew S. Tanenbaum.

image

To give more context. I originally used the term "generation" in the design doc, but [it was suggested to change the term](https://docs.google.com/document/d/16zexVooHKPc80F4dVtUjDYK9DOpkVPRNfSv0zRtfFpk/edit?disco=AAABrc6hZ1I) to another name because "generation" is already used in the Pod API. So, I changed the term "generation" to "epoch".

> How would epocs work with a generic MPI workload that does not do training?

As long as the workload depends on the pattern "if one worker process fails, all worker processes must be restarted", it should work. It doesn't matter what the worker process is doing.

kannon92: Ah alright. That sounds good to me then.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,789 @@
...
+
+  // The most recent deprecated epoch of the JobSet.
+  //
+  // Healthy Pods that have an epoch smaller than or equal to this value should be restarted in place.
+  //
+  // This is written by the JobSet controller and read by the agent sidecars.
+  //
+  // +optional
+  // +kubebuilder:default=0
+  DeprecatedEpoch int32 `json:"deprecatedEpoch,omitempty"`
```
kannon92:
```suggestion
  DeprecatedEpoch *int32 `json:"deprecatedEpoch,omitempty"`
```

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,789 @@
...
+
+  // The most recent synced epoch of the JobSet.
+  //
+  // Pods that have an epoch equal to this value should lift their barrier to allow the worker containers to start running.
+  //
+  // This is written by the JobSet controller and read by the agent sidecars.
+  //
+  // +optional
+  // +kubebuilder:default=0
+  SyncedEpoch int32 `json:"syncedEpoch,omitempty"`
```
kannon92:
```suggestion
  SyncedEpoch *int32 `json:"syncedEpoch,omitempty"`
```

keps/467-InPlaceRestart/README.md
GiuseppeTT: Any reason for using a pointer? I was using the `Restarts` and `RestartsCountTowardsMax` status fields as reference, which are `int32`

kannon92: With the linter I think we may require it.

Typically pointers are good to use to distinguish between 0 and nil case.

kannon92: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#optional-vs-required

GiuseppeTT: Makes sense. Thanks!

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,790 @@
...
+* If worker container exits `Z`: fail the Pod without failing the parent Job (worker failure that is recoverable with Pod recreation but not Pod in place restart)
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: jobset-example-in-place-restart-failure-policy
+  annotations:
+    # Force 1 Job per Node pool
+    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool
```
kannon92, line 299: Does this feature dependent on exclusive-topology?

I'm trying to understand if one could enable this feature on-prem or aws?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,790 @@
...
+
+* `jobSet.spec.replicatedJobs[].template.spec.backOffLimit` should be set to `MaxInt32` (i.e., `2147483647`) to avoid unnecessarily failing Jobs if a child Pod fails individually or a Pod is restarted in place  
+* `jobSet.spec.replicatedJobs[].template.spec.podReplacementPolicy` should be set to `Failed` to make sure the replacement Pod is created only after the original Pod has fully failed  
+* `jobSet.spec.replicatedJobs[].template.spec.template.spec.initContainers` should contain the agent sidecar to handle in place restart operations at the worker Pod level for the JobSet controller. This is partially verified by checking for the existence of a sidecar that contains the `RestartPod` rule
+
+### Implementation - Agent sidecar
+
+The only way to trigger the `RestartPod` action on demand is by making one container exit with a specified exit code. Therefore, we propose to add a new container image that will be buildable from new code in the JobSet repo. The container should be added by the user to the JobSet manifest as the “agent sidecar”. The agent sidecar is responsible for handling in place restart operations at the worker Pod level for the JobSet controller.
```
kannon92: There should be some thought on how to streamline this for users.

Asking users to add an "agent sidecar" seems prone to bugs/challenges.

Maybe we consider a parameter in a helm chart to inject the agent sidecar.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,790 @@
...
+when drafting this test plan.
+
+[testing-guidelines]: https://git.k8s.io/community/contributors/devel/sig-testing/testing.md
+-->
+
+[X] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+##### Prerequisite testing updates
```
kannon92, line 746: Have you given any thought on how you will be able to test this e2e?

And will it be possible to run a test case like this in our CI?

There is a lot of things going on here so we should make sure to try and cover the testing of this feature in upstream as much as possible.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,790 @@
...
+-->
+
+Make sure the following cases are covered:
+
+- All Pods in sync: JobSet controller should not update the JobSet status
+- All Pods just became in sync: JobSet controller should update `syncedEpoch`
+- All Pods were in sync but one restarted into a new epoch: JobSet controller should update `deprecatedEpoch`
+- Pods just lost sync and a second Pod restarted into the new epoch but not all yet: JobSet controller should not update the JobSet status
+
+### Graduation Criteria
```
kannon92, line 799: Since you are adding a feature gate let's put some information here.

kannon92: What criteria do you want to promote this to beta?

keps/467-InPlaceRestart/README.md
GiuseppeTT: > Does this feature dependent on exclusive-topology?

In place restart can be used without exclusive-topology.

I included exclusive-topology along with other things in the example to make it more realistic, but I can prioritize simplicity and remove everything that is not necessary from the YAML. What do you think?

> I'm trying to understand if one could enable this feature on-prem or aws?

It should be possible. There is nothing GKE specific in the design.

keps/467-InPlaceRestart/README.md
GiuseppeTT: This is addressed [here](https://github.com/kubernetes-sigs/jobset/pull/1083/files#diff-a482883fe99622fdaa4c090337a4565774b5b8ba6c32fa92b3672583d740eb7aR776).

I agree that adding the agent sidecar manually is cumbersome and error-prone. However, I think it makes sense to defer automatic injection to beta. Keeping it manual for alpha simplifies the review process and offers more flexibility for testing different configurations and gathering initial user feedback. I will update the graduation criteria.

keps/467-InPlaceRestart/README.md
GiuseppeTT: > Have you given any thought on how you will be able to test this e2e?

The framework we have for e2e tests is very flexible. It should be possible to create a JobSet that has the agent sidecar and a simplified worker container that exits non-zero on demand (e.g., by checking for the existence of a file, k8s resource or an HTTP request). In place restart can be tested by making one of the worker containers fail and then checking the Pod and JobSet manifests to make sure the group restart succeeded.

> And will it be possible to run a test case like this in our CI?

I don't see any reason why a small test (2 Pods) cannot be run to test the core logic.

keps/467-InPlaceRestart/README.md
GiuseppeTT: I couldn't find a graduation criteria in the other JobSet KEPs to use as reference. What do you think about the following?

```md
### Graduation Criteria

#### Alpha

- Feature implemented behind a feature flag
- Initial unit and e2e tests are implemented

#### Beta

- Address reviews and bug reports from alpha users
- Add a way to add the agent sidecar automatically

#### GA

- Address reviews and bug reports from beta users
- Feature flag enabled by default
```

kannon92: I think future work could be beta in some cases.

I think we should inject the side car before we graduate this feature.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,790 @@
...
+```
+
+The highlights are:
+
+- Calculate the Pod epoch at start up as `jobSet.status.syncedEpoch + 1`. This makes sure the worker container will start running only when the JobSet controller updates `jobSet.status.syncedEpoch`. This is done only when all worker Pods are at the same epoch
+- Restart the Pod in place if its epoch has been deprecated by `checking epoch <= jobSet.status.deprecatedEpoch`. This is done only when a group restart is necessary
+- Lift the Pod barrier if its epoch has been marked as synced by checking `epoch == mostRecentSyncedEpoch`. This is done only when all worker Pods are at the same epoch
+
+When `RestartPod` is used, the barrier lift can be done by setting up a startup probe for the agent sidecar and making sure the agent sidecar creates the specified endpoint when the barrier must be lifted. If `restartPolicy = OnFailure` is used instead, the barrier lift can be done starting the worker process command when the barrier must be lifted.
+
```
kannon92, line 577: Will we build and publish a side car image?

keps/467-InPlaceRestart/README.md
kannon92: I agree. Can you update the e2e section with how you would implemen this?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,790 @@
...
+### Make the barrier optional
+
+This is technically possible, but it adds little overhead while allowing the design to make less assumptions over the worker containers.
+
+### Restart only specified Pods instead of all of them
+
+Currently JobSet only supports restarting the whole workload. Partial restart like restarting only a single Job or a group of ReplicatedJobs is desirable but it is still being [discussed](https://github.com/kubernetes-sigs/jobset/issues/918). 
+
+## Future work
```
kannon92, line 886: What features will you need to implement as part of this KEP?

keps/467-InPlaceRestart/README.md
GiuseppeTT: > I think future work could be beta in some cases.

> I think we should inject the side car before we graduate this feature.

Sounds good. This seems in line with what I wrote. I will update the KEP.

keps/467-InPlaceRestart/README.md
GiuseppeTT: That's the plan : )

keps/467-InPlaceRestart/README.md
GiuseppeTT: I am not sure I follow 🤔 . The only feature that needs to be implemented is the in place restart feature.

That being said, I have talked to Anu to make sure the new design for recreate-single-job is compatible with in place restart. But the recreate-single-job feature is an independent feature that is outside the scope of this KEP.

kannon92: There is sidecar and restartContainers.

Are you going to implement these in this KEP as part of graduating the feature?

Or will there be a separate KEP for this?

GiuseppeTT: What is `restartContainers`?

To me, everything is part of this KEP:

- Add the agent code and image to the JobSet repo to support Pods being restarted in place on demand
- Change the JobSet controller to orchestrate in place restarts
- Add extra validation to the JobSet webhook to validate JobSet manifests that have in place restart enabled
- Add tests for in place restart
- Add documentation for in place restart

kannon92: RestartPolicyRules

GiuseppeTT: Oh, I see what you mean. The new `RestartPod` action in the `RestartPolicyRules` API is upstream work. It's being done independently by @yuanwang04 (see [this](https://kubernetes.io/blog/2025/08/29/kubernetes-v1-34-per-container-restart-policy/#roadmap) and [this](https://docs.google.com/document/d/1UmJHJzdmMA1hWwkoP1f3rG9nS0oZ2cRcOx8rO8MsExA/edit?usp=sharing&resourcekey=0-OuKspBji_1KJlj2JbnZkgQ)). In other words, `RestartPod` is not part of this KEP.

keps/467-InPlaceRestart/README.md
GiuseppeTT: See more info in https://github.com/kubernetes-sigs/jobset/pull/1083/files#diff-a482883fe99622fdaa4c090337a4565774b5b8ba6c32fa92b3672583d740eb7aR577

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,823 @@
...
+kind: JobSet
+metadata:
+  name: jobset-example-recreate-all-pods
+  annotations:
+    # Force 1 Job per Node pool
+    alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool
+spec:
+  failurePolicy:
+    maxRestarts: ... # Maximum number of full restarts
+    restartStrategy: Recreate # (Default) Restart the JobSet by recreating the Jobs
```
andreyvelich, line 110: Would it be tough for us to rename restartStrategy to restartPolicy to be consistent with Pod API ?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,823 @@
...
+                  path: /barrier-is-lifted
+                  port: 8080
+                failureThreshold: ...
+                periodSeconds: 1
+              env:
+                # Required env variables for agent sidecar
+                - name: NAMESPACE
+                  valueFrom:
+                    fieldRef:
+                      fieldPath: metadata.namespace
```
andreyvelich, line 248: you don't need this since info about namespace can be taken from:
```
/var/run/secrets/kubernetes.io/serviceaccount/namespace
```

keps/467-InPlaceRestart/README.md
andreyvelich: I would agree with @kannon92. Although, epoch is a general term I would prefer to find a better name to represent this. Otherwise, users might miss-interpreted this. Maybe @vsoch has some thoughts from the HPC world.

What is the problem to introduce generation status to the JobSet, similar to Pods ?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,823 @@
...
+  // The most recent synced epoch of the JobSet.
+  //
+  // Pods that have an epoch equal to this value should lift their barrier to allow the worker containers to start running.
+  //
+  // This is written by the JobSet controller and read by the agent sidecars.
+  //
+  // +optional
+  // +kubebuilder:default=0
+  SyncedEpoch *int32 `json:"syncedEpoch,omitempty"`
+}
```
andreyvelich, line 504: I guess, these status will be introduced only when in-place restarts are enable ? 
Could we discuss some alternative approaches ? For example, storing generations in the Pod file-system which can be read from the control plane ? For example, by SSH-ing to the worker pods from the JobSet controller ?

cc @astefanutti @tenzen-y

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,823 @@
...
+```python
+# Initialize
+parentJobSet = getJobSet(env.namespace, env.jobSetName)
+mostRecentSyncedEpoch = jobset.status.syncedEpoch
+epoch = mostRecentSyncedEpoch + 1
+patch = {"metadata" : {"annotations" : {"jobset.sigs.k8s.io/epoch" : epoch}}}
+patchPod(env.namespace, env.podName, patch)
+
+# Watch
+for event in watchJobSet(env.namespace, env.jobSetName):
```
andreyvelich, line 555: Would you require to add RBAC to the sidecar to watch for the JobSet updates ?

andreyvelich (commented):
Thank you @GiuseppeTT!

keps/467-InPlaceRestart/README.md
GiuseppeTT: As much as I like making things consistent, renaming a field is a breaking change : /

keps/467-InPlaceRestart/README.md
GiuseppeTT: That is a valid approach. If the namespace were the only piece of data required, I would definitely use that path. 

But, since we already need the Downward API for the Pod name and JobSet name, it keeps the implementation simpler to use the same mechanism for the namespace.

The automatic injection of the agent sidecar (a requirement for the Beta promotion) will handle this configuration, so the user won't have to manage it manually.

keps/467-InPlaceRestart/README.md
GiuseppeTT: > I would agree with @kannon92. Although, epoch is a general term I would prefer to find a better name to represent this. Otherwise, users might miss-interpreted this. Maybe @vsoch has some thoughts from the HPC world.

I am happy to consider other names : )

> What is the problem to introduce generation status to the JobSet, similar to Pods ?

Just as "epoch" is an overloaded term, "generation" is too. See [this comment thread](https://docs.google.com/document/d/16zexVooHKPc80F4dVtUjDYK9DOpkVPRNfSv0zRtfFpk/edit?disco=AAABrc6hZ1I) in the design doc.

keps/467-InPlaceRestart/README.md
GiuseppeTT: The main feedback on the original proposal was the hackiness of not relying on the API server. The [new proposal](https://docs.google.com/document/d/16zexVooHKPc80F4dVtUjDYK9DOpkVPRNfSv0zRtfFpk/edit?tab=t.0#heading=h.8duhxd21z8jh) consciously embraces [Kubernetes API conventions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status), adopting a level-based and declarative approach.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Yes! Any necessary RBAC rules will be in the documentation of in place restart.

The agent sidecar also requires specific RBAC rules to update its Pod epoch (and only its Pod epoch) as [discussed in the design doc](https://docs.google.com/document/d/16zexVooHKPc80F4dVtUjDYK9DOpkVPRNfSv0zRtfFpk/edit?tab=t.0#heading=h.ti7m3xqxahqv).

keps/467-InPlaceRestart/README.md
vsoch: if it was a life service and the timepoint was a check, I'd call it a heartbeat. If it was backing something up, then a checkpoint, snapshot, or image. You could also just call it a timepoint. I'll check with others and report back if anyone has other ideas.

keps/467-InPlaceRestart/README.md
kannon92: @GiuseppeTT KEPs should contain all the information.

I tend to prefer writing the KEP document because I don't expect the google doc to reflect the actual implementation once you draft the KEP.

I realize your design doc is much more detailed but if there are gaps in this KEP document I would prefer that the KEP contains this and not just the google doc.

For upstream Kubernetes we discourage linking google docs in KEPs as our goal is for the KEP to be the design record not the google doc.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Vanessa, I appreciate the suggestions, but I don't think they make sense in this context : / . Checkpoint could be one, but it's also overloaded.

Everybody. What do you think about renaming "epoch" to "inplace restart attempt"? I think it makes sense since the proposed epoch counter is similar to `jobset.sigs.k8s.io/restart-attempt` from JobSet.

keps/467-InPlaceRestart/README.md
GiuseppeTT: [Done](https://github.com/kubernetes-sigs/jobset/pull/1083/files#diff-a482883fe99622fdaa4c090337a4565774b5b8ba6c32fa92b3672583d740eb7aR579).

keps/467-InPlaceRestart/README.md
vsoch: Another https://en.wikipedia.org/wiki/Lamport_timestamp

andreyvelich: Yeah, I like this to be specific to in place restart feature.
@kannon92 thoughts ?

keps/467-InPlaceRestart/README.md
andreyvelich: @GiuseppeTT Can we track this as part of JobSet v1 ?
Like we do for Complete status rename: https://github.com/kubernetes-sigs/jobset/issues/723

keps/467-InPlaceRestart/README.md
andreyvelich: @GiuseppeTT Am my understanding correct that based on our findings earlier, it will be worse to store this information in JobSet/Pods annotations instead of status?

keps/467-InPlaceRestart/README.md
kannon92: I like that suggestion.

keps/467-InPlaceRestart/README.md
GiuseppeTT: > @GiuseppeTT Am my understanding correct that based on our findings earlier, it will be worse to store this information in JobSet/Pods annotations instead of status?

I'm not sure I understand your question : / can you clarify?

The epoch of each worker Pod is stored in its annotation `jobset.sigs.k8s.io/epoch`. The most recent synced epoch and the most recently deprecated epoch are stored in the JobSet status `jobset.status.syncedEpoch` + `jobset.status.deprecatedEpoch`.  From my scales test, this choice did not impact speed of in place restart. Storing all numbers (proportional to the number of worker Pods) in the same object (a worker Pod or JobSet) offers no benefit.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Great! I will update the KEP.

keps/467-InPlaceRestart/README.md
GiuseppeTT: > @GiuseppeTT Can we track this as part of JobSet v1 ?

Feel free to propose it : )

I don't have strong opinions about the names.

kannon92 (commented):
This is looking really good for alpha imo!

I'm going to ask @ahg-g to review before I get my LGTM though.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done!

keps/467-InPlaceRestart/README.md
andreyvelich: @GiuseppeTT I am trying to understand if we store all of this information in Pod and JobSet annotations, why do we need to make changes to the JobSet status API ?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+// README: No change. Only here for reference.
+type FailurePolicy struct {
+  // Limit on the number of JobSet restarts.
+  MaxRestarts int32 `json:"maxRestarts,omitempty"`
+
+  // Strategy to use when restarting the JobSet.
+  //
+  // +optional
+  // +kubebuilder:default=Recreate
+  RestartStrategy JobSetRestartStrategy `json:"restartStrategy,omitempty"`
```
andreyvelich, line 441: I also was thinking about what if users want to apply in-place restarts only for the particular ReplicatedJob's pods ?
With this API design, we don't give such option to users in the future.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+
+  // The most recent deprecated in place restart attempt of the JobSet.
+  //
+  // Healthy Pods that have an in place restart attempt smaller than or equal to this value should be restarted in place.
+  //
+  // This is written by the JobSet controller and read by the agent sidecars.
+  //
+  // +optional
+  // +kubebuilder:default=0
+  DeprecatedInPlaceRestartAttempt *int32 `json:"deprecatedInPlaceRestartAttempt,omitempty"`
```
andreyvelich: Shall we use unsyncedInPlaceRestartAttempt instead of deprecated ?
Deprecated might sound confusing, imo.

keps/467-InPlaceRestart/README.md
GiuseppeTT: In the proposal, no information is stored in the JobSet annotations. We could do that, but it would be an anti pattern given that we have control over the JobSet status IMO.

The design proposes an annotation to Pods to store the in place restart attempt because we don't have control over the Pod status.

keps/467-InPlaceRestart/README.md
GiuseppeTT: > With this API design, we don't give such option to users in the future.

Can you clarify? 

JobSet currently doesn't support restarts of only specified `ReplicatedJobs` and there is no proposal to do so (the closest we have is #918 which is about *failed* Jobs). If this is proposed, in place restart would face the same challenges of normal restarts (Job recreations) IMO.

keps/467-InPlaceRestart/README.md
GiuseppeTT: I don't think "unsynced" is a good choice.

The reasoning behind "deprecated" is that the IPRA (in place restart attempt) was valid (all worker Pods had the same value for the IPRA), then a failure happened, a new IPRA was created and the old IPRA one must be deprecated to make their Pods move to the new IPRA. In a sense, the new IPRA is also unsynced until all Pods reach it.

But I understand the disconfort with "deprecated". What do you think about "retired"? : )

keps/467-InPlaceRestart/README.md
andreyvelich: @kannon92 @GiuseppeTT @ahg-g Do we have any concerns that we are going to populate this status only when in-place restart is enabled ?

keps/467-InPlaceRestart/README.md
andreyvelich: IIRC, @carreter wanted to introduce `backOffLimit` API on the replicatedJob level which allows to restart specific job: https://github.com/kubernetes-sigs/jobset/issues/976#issuecomment-3207208253.
So if we go ahead with this approach, users might want to enable in-place restarts only for individual rJobs.

keps/467-InPlaceRestart/README.md
kannon92: I thought I wrote something up here.

I asked this question to my api-review team (thank you @everettraven!) but there isn't really a concern to have status fields populate on optional features.

The main feedback I got back was to make sure that if the feature is not enabled you should make sure to not display these status fields (so I think pointer here would suffice).

We should add integration testing to make sure that the status fields only show up if people are using this feature.

keps/467-InPlaceRestart/README.md
andreyvelich: What do we think about active / inactive or current / previous?

kannon92: I am happy with currentInPlaceRestartAttempt and previousInPlaceRestartAttempt.

keps/467-InPlaceRestart/README.md
GiuseppeTT: AFAIK #976 is concerned more with restarting only failed Jobs rather than restarting a group of Jobs if one fails.

Regardless, the main priority should be ensuring that in place restart can be extended to support this behavior, similar to how the current JobSet failure policy is extensible. I don't see anything in the current in place restart design that would block this. It would likely just require a counter per Job, as Willow originally proposed for standard Job recreation.

Once restart-only-failed Job is implemented, I'm happy to add support to it to JobSet : )

keps/467-InPlaceRestart/README.md
GiuseppeTT: > I am happy with currentInPlaceRestartAttempt and previousInPlaceRestartAttempt.

Sounds good! I'll update the KEP.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Agreed

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
+# KEP-467: Fast failure recovery with in place restarts
```
ahg-g: nit: in-place, here and all other places

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+```
+
+The highlights are:
+
+- Calculate the Pod in place restart attempt at start up as `jobSet.status.currentInPlaceRestartAttempt + 1`. This makes sure the worker container will start running only when the JobSet controller updates `jobSet.status.currentInPlaceRestartAttempt`. This is done only when all worker Pods are at the same in place restart attempt
+- Restart the Pod in place if its in place restart attempt is outdated by checking `inPlaceRestartAttempt <= jobSet.status.previousInPlaceRestartAttempt`. This is done only when a group restart is necessary
+- Lift the Pod barrier if its in place restart attempt has been marked as current by checking `inPlaceRestartAttempt == currentInPlaceRestartAttempt`. This is done only when all worker Pods are at the same in place restart attempt
+
+When `RestartPod` is used, the barrier lift can be done by setting up a startup probe for the agent sidecar and making sure the agent sidecar creates the specified endpoint when the barrier must be lifted. If `restartPolicy = OnFailure` is used instead, the barrier lift can be done starting the worker process command when the barrier must be lifted.
```
ahg-g: s/"the barrier lift"/"lifting the barrier"?

ahg-g: I am a little confused by the sentence "If `restartPolicy = OnFailure` is used instead", instead of what?

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+kind: ServiceAccount
+metadata:
+  name: in-place-restart-sa
+---
+# JobSet watcher role
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: jobset-watcher
+  namespace: default
```
ahg-g: so we need to create this in every namespace the user wants to create jobsets in? if so, how will we do that?

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done

GiuseppeTT: tl;dr: "If `restartPolicy = OnFailure` is used instead" = "If restartPolicy = OnFailure is used instead of `RestartPod`"

This paragraph is explaining how to set up the barrier using the agent sidecar.

One option is to use `action = RestartPod` + a start up probe like

```yaml
apiVersion: v1
kind: Pod
metadata:
    name: RestartPod-example
spec:
  initContainers:
  - name: agent
    restartPolicy: Always
    restartPolicyRules:
    - action: RestartPod
      exitCodes:
        operator: In
        values: [X]
    # Barrier
    # Only allow the worker container to start when the agent sidecar creates this endpoint
    startupProbe:
      httpGet:
        path: /barrier-is-lifted
        port: 8080
      failureThreshold: ...
      periodSeconds: 1
  containers:
  - name: worker
    ...
```

Another option is `restartPolicy = OnFailure` + start the worker process as a subprcess of the agent only when the barrier is lifted. This is useful while `action = RestartPod` is not fully released.

```yaml
apiVersion: v1
kind: Pod
metadata:
  name: OnFailure-example
spec:
  restartPolicy: OnFailure
  containers:
  # Start the worker process (as a subprocess) only when the barrier is lifted
  - name: agent
  ...
```

keps/467-InPlaceRestart/README.md
GiuseppeTT: > so we need to create this in every namespace the user wants to create jobsets in?

Yep. AFAIK since service accounts are namespaced, there is no way to grant the extra permissions required by the agent sidecar without setting up stuff for each namespace the user wants to create jobsets in.

It is technically possible to use a ClusterRoleBinding that binds to all service accounts, but I assume anyone using namespaces for isolation would reject that security risk.

> if so, how will we do that?

Since the effort scales with the number of namespaces, I assume this is manageable for the infra team (since they already manage the namespaces).

That said, I can create a Helm chart to make installing these permissions easier.

What do you think? : )

andreyvelich: How would you configure VAP for every namespace that this service account is created ? Currently, it checks only `default` ns:
```
request.userInfo.username != 'system:serviceaccount:default:in-place-restart-sa'
```

Does it support wildcards ?
```
system:serviceaccount:*:jobset-in-place-restart
```

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+### Implementation - Permissions for the agent sidecar
+
+The agent sidecar requires permissions to watch its parent JobSet. It also requires permissions to update its Pod in place restart attempt, which can be integrated with a validating admission policy to make sure only the in place restart attempt annotation can be modified. The following manifest shows an example of how a service account can be set up for the worker Pod. It will be part of the documentation of in place restart since it's up to the user to integrate these extra permissions to any existing service account.
+
+```yaml
+# Service account
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: in-place-restart-sa
```
andreyvelich: Can we make those SA, JobSet specific ?
```suggestion
  name: jobset-in-place-restart
```

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+---
+# JobSet watcher role
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: jobset-watcher
```
andreyvelich:
```suggestion
  name: jobset-in-place-restart
```

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+  namespace: default
+rules:
+- apiGroups: ["jobset.x-k8s.io"]
+  resources: ["jobsets"]
+  verbs: ["get", "list", "watch"]
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: bind-in-place-restart-sa-to-jobset-watcher
```
andreyvelich:
```suggestion
  name: jobset-in-place-restart
```

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+  namespace: default
+subjects:
+- kind: ServiceAccount
+  name: in-place-restart-sa
+  namespace: default
+roleRef:
+  kind: Role
+  name: jobset-watcher
```
andreyvelich:
```suggestion
  name: jobset-in-place-restart
  namespace: default
roleRef:
  kind: Role
  name: jobset-in-place-restart
```

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,918 @@
...
+  apiGroup: rbac.authorization.k8s.io
+---
+# Pod patcher role
+apiVersion: rbac.authorization.k8s.io/v1
```
andreyvelich, line 593: Why do you need two separate Roles to watch for JobSet and patch pods ?
Can you combine RBAC to the one Role: `jobset-in-place-restart`

keps/467-InPlaceRestart/README.md
ahg-g: > That said, I can create a Helm chart to make installing these permissions easier.

yes, otherwise this will be hard to use

kannon92: I don't quite understand how you would use a helm chart here. We would install the helm chart into a particular namespace for jobset installation. 

I guess it could be a separate helm chart from jobset. You have templates for these roles and you could do something like `helm install inplaceRestart SOME_CHART_REGISTRY --namespace user-a`.

I just don't want someone to accidently install jobset in a different namespace as they are using the same helm chart to create the roles in that namespace.

Probably worth adding how you would do this in the KEP.

GiuseppeTT: > How would you configure VAP for every namespace that this service account is created ?

It works if I use 

```yaml
  matchConditions:
  - name: "is-target-service-account"
    expression: "request.userInfo.username.matches('^system:serviceaccount:[^:]+:in-place-restart-sa$')"
```

See https://gist.github.com/GiuseppeTT/6fc2b4cfa8727869bc52b3b500b89e62

keps/467-InPlaceRestart/README.md
GiuseppeTT: I don't mind the name change, but keep in mind that this is the service account used in the worker Pod. The most probable scenario is that the user already has a service account for their worker Pods. So they will adapt their service account with the extra permissions and Validating Admission Policy.

keps/467-InPlaceRestart/README.md
GiuseppeTT: I did it to make it more clear the permissions being added, but I agree, the manifest is better with a single role.

I updated the KEP.

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done

keps/467-InPlaceRestart/README.md
GiuseppeTT: Done.

keps/467-InPlaceRestart/README.md
GiuseppeTT: > yes, otherwise this will be hard to use

Sounds good.

I updated the KEP to mention this helm chart and added it to the graduation criteria.

ahg-g:
/lgtm

will leave the approval to @andreyvelich and @kannon92 

Thanks @GiuseppeTT , this is one of the most exciting and differentiating features in JobSet

kannon92 (approved):
/approve

I think this is good for alpha.

There will be some UX improvements needed to make this GA and that is in scope for KEP.

keps/467-InPlaceRestart/README.md
andreyvelich: Awesome!

andreyvelich:
Great! Excited to see this moving forward!

/lgtm

GiuseppeTT:
Thank you all!

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,902 @@
...
+
+Both in this section and below, follow the guidelines of the [documentation
+style guide]. In particular, wrap lines to a reasonable length, to make it
+easier for reviewers to cite specific portions, and to minimize diff churn on
+updates.
+
+[documentation style guide]: https://github.com/kubernetes/community/blob/master/contributors/guide/style-guide.md
+-->
+
+This KEP proposes the new restart strategy `InPlaceRestart` for JobSet. The objective is to speed up the restart time of distributed ML model training, which is traditionally done by recreating all Pods. This is especially important for large scales, where frequent failures that take longer to recover can cost millions. The proposal leverages the incoming `RestartPod` action to enable in-place restart of Pods for JobSet workloads, which avoids the overhead of recreating Pods and significantly reduces recovery time. The solution involves adding an agent sidecar to each worker Pod to allow Pods to be restarted in-place on demand and updating the JobSet controller to orchestrate group restarts by restarting healthy Pods in-place while allowing the Job controller to recreate failed Pods.
```
samos123, line 82: Has this approach been validated with a Jax workload? In McJax all pods will fail if 1 of the pods fail so I wonder if it would still help.

> restarting healthy Pods in-place while allowing the Job controller to recreate failed Pods.

I think for Jax all pods will be considered failed pods when there is a failure.

CC @GiuseppeTT @imreddy13

GiuseppeTT: > Has this approach been validated with a Jax workload? In McJax all pods will fail if 1 of the pods fail so I wonder if it would still help.

The approach hasn't been explicitly tested with a Jax workload, but I did desgined for and test for its behavior.

That is, the in place restart design is extremely flexible. When it notices that a worker restarted (either because its Pod was recreated or restarted in place), it makes sure that all other workers restart. The other workers might be forcibly restarted by restarting their Pod in place. Another supported case is the other workers restarting on their own. The important fact is that the workers restart, on way or another.

Importantly:

- The proposed JobSet spec allows for Pods to be restarted in place if their worker containers exit with a special exit code. That is, if the worker container failed due to a recoverable failure, its Pod can be automatically restarted in place without the agent sidecar acting
- When a worker Pod restarts, it doesn't start the worker process immediately. Instead, the Pod waits at the barrier and only starts the worker processes when all worker Pods are ready

Also, from my experience, workers in the same TPU slice fail pretty fast when one of them fail, but the workers in other TPU slices take a longer time to fail.

Does this answer your question? : )

samos123: I think what will happen in-practice when using McJax is that all pods will get recreated even with this new feature. This is because all processes will fail when one of the pods gets killed or dies due to failure. Maybe that's handled by `if their worker containers exit with a special exit code`. I haven't checked what the exit code is when Jax propogates sigterm when one of the processes terminate, so that indeed may be a way to make this work well for a McJax workload.

keps/467-InPlaceRestart/README.md
```diff
@@ -0,0 +1,902 @@
...
+  namespace: default
+roleRef:
+  kind: Role
+  name: jobset-in-place-restart
+  apiGroup: rbac.authorization.k8s.io
+---
+# Validation admission policy
+# Allow only the pod annotation "jobset.sigs.k8s.io/in-place-restart-attempt" to be updated
+apiVersion: admissionregistration.k8s.io/v1
+kind: ValidatingAdmissionPolicy
```
andreyvelich, line 625: We briefly talked about this KEP with @astefanutti on the last Training WG call.
@GiuseppeTT @kannon92 @ahg-g What do you think about creating these resources in the JobSet controller once user creates JobSet with in-place restart capability?
If we want to be more restricted, we can create dedicated RBAC + VAP per JobSet, and remove them once JobSet is complete.

GiuseppeTT: Setting this automatically is nice but I wonder what users think about creating security resources automatically.

IMO this change is a good candidate for beta, once we gather more feedback from users. WDYT?

andreyvelich: We can always explain this in the docs that using this feature will require JobSet controller to create these security resources. I just feel that if we create them automatically, that will reduce number of errors we get from users who want to try in-place restarts.

WDYT @kannon92 ?

astefanutti: I tend to agree with @andreyvelich, those security resources are complex enough so it's error-prone for users and seems more secured in practice to have them created by the JobSet controller. It would also create a lot of friction for users to adopt the feature.
The Role could be restricted to the JobSet name and those resources could be owned by the parent JobSet to make sure they are properly cleaned-up.

GiuseppeTT: Sounds good! I'll add this to the beta

Implementation wise, I'm considering the following:

- Scan the Pod template in the JobSet spec for the service account name
- Use the JobSet controller to create the security resources ([example from the in-place restart PR](https://github.com/kubernetes-sigs/jobset/blob/4f93865291cc69468a18b6ffb8c986b37a6a8af0/cmd/in-place-restart-agent/service-account-example.yaml)) for the given service account

What do you think?

Adding the JobSet name to the field `role.rules[].resourceNames[]` is a great addition!

kannon92: VAP greatly exceeds the scope of jobset controller. I agree that they are complicated but I worry that the job set controller is now required to create cluster wide resources. Usually we lead the creation of cluster wide resources to operators/helm/kustomize on deployment of an operand.

I think this would mean we would need to give admin privileges to the job set controller to provision these resources.

GiuseppeTT: Moving the discussion from here to https://github.com/kubernetes-sigs/jobset/pull/1099#discussion_r2616675430

PR #1089, opened 2025-11-14
Make Pod admission webhook fail if the Node of the leader Pod is not found

### LINKED ISSUE #1088: Node selector of follower Pod is set to empty value when the leader Pod exists but its Node does not exist anymore
**What happened**:

Creating a JobSet with the annotation `alpha.jobset.sigs.k8s.io/exclusive-topology: cloud.google.com/gke-nodepool` makes sure child Jobs are mapped 1:1 to GKE node pools. For a given child Job, this is achieved by the JobSet controller and its webhooks by first only allowing the leader Pod to be created. Once the leader Pod is created and scheduled, the follower Pods are also allowed to be created and they will contain the node selector `cloud.google.com/gke-nodepool="<Node Pool of Node that is running the Leader Pod>"` to make sure they follow the leader Pod to the right Node pool.

The bug I experienced was a follower Pod being created with the node selector containing `cloud.google.com/gke-nodepool=""` (notice the empty value). This happened when the leader Pod still existed, but its Node did not exist anymore.

I expected the follower Pod creation request to fail instead of creating a follower Pod with an invalid node selector.

No easy to reproduce.

The solution is easy. [This line](https://github.com/kubernetes-sigs/jobset/blob/8ba41d3a44ae529da07792b77a799fad665cddef/pkg/webhooks/pod_mutating_webhook.go#L202) should be `return "", err` instead of `return "", client.IgnoreNotFound(err)`.

[This other line](https://github.com/kubernetes-sigs/jobset/blob/8ba41d3a44ae529da07792b77a799fad665cddef/pkg/controllers/pod_controller.go#L257) is not related to the bug I experienced but might also be a bug.

GiuseppeTT:
#### What type of PR is this?

/kind bug

Fixes a bug in the mutation Pod webhook. When modifying a follower Pod, the webhook should fail if the Node of the leader Pod does not exist. The current behavior is setting the node selector in the follower Pod to `<topology key>=""`  (e.g., `cloud.google.com/gke-nodepool=""`).

Fixes #1088

```release-note
NONE
```

### DIFF
--- a/pkg/webhooks/pod_mutating_webhook_test.go
+++ b/pkg/webhooks/pod_mutating_webhook_test.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
 	"sigs.k8s.io/jobset/pkg/controllers"
 )
 


GiuseppeTT:
/hold

I'll check this tomorrow with a clearer head

GiuseppeTT:
/assign @kannon92

Can you review this one? : )

[This comment](https://github.com/kubernetes-sigs/jobset/issues/1088#issuecomment-3534042960) explains step by step how the bug manifests in the webhook code

pkg/webhooks/pod_mutating_webhook.go
```diff
@@ -196,10 +196,8 @@ func (p *podWebhook) topologyFromPod(ctx context.Context, pod *corev1.Pod, topol
 	// Get node the leader pod is running on.
 	var node corev1.Node
 	if err := p.client.Get(ctx, types.NamespacedName{Name: nodeName}, &node); err != nil {
-		// We'll ignore not-found errors, since there is nothing we can do here.
```
kannon92, line 199: So is this comment wrong?

It seems that someone did think of this case.

If a node temporarily doesn't exist or "other scenario" are you okay with always failing?

kannon92: blame seems to point to @danielvegamyhre.

kannon92: Maybe @ahg-g has more context here?

GiuseppeTT: > So is this comment wrong?

I suspect so. It might have been copied from [here](https://github.com/kubernetes-sigs/jobset/blob/8ba41d3a44ae529da07792b77a799fad665cddef/pkg/controllers/jobset_controller.go#L109-L112). Ignoring objects not found is a common pattern in controllers, but it is not the right approach in this case.

> If a node temporarily doesn't exist or "other scenario" are you okay with always failing?

I am. If the Pod mutation fails, the Job controller will just retry creating the Pod (we already rely on this). Eventually the leader Pod is garbage collected (because its Node does not exist anymore), a new leader Pod is created (and scheduled) and the follower Pods can be created.

This is not ideal, but it is better than creating follower Pods with invalid node selectors.

ahg-g: good catch, did a customer hit this?

kannon92:
If this is a customer related bug, id be fine to cherry pick.

pkg/webhooks/pod_mutating_webhook.go
GiuseppeTT: > good catch, did a customer hit this?

Yep. A big user of JobSet in fact.

PR #1096, opened 2025-12-02
Add `InPlaceRestart` feature gate as alpha

GiuseppeTT:
#### What type of PR is this?

/kind feature

- Add the `InPlaceRestart` feature gate as alpha, which is a prerequisite for [KEP-467](https://github.com/kubernetes-sigs/jobset/blob/main/keps/467-InPlaceRestart/README.md)
- Add support for configuring feature gates via the Configuration API, which enables easier management of feature flags. This is a standard feature from feature gates which was not available in JobSet. I based my changes on [Kueue](https://github.com/kubernetes-sigs/kueue/blob/81cae0608cac3d457b8ecf8a0480e7994c2151c1/cmd/kueue/main.go#L150-L160)

Part of #467

Since in-place restart is a large feature, I am breaking it into multiple PRs, starting with the feature gate to ensure all subsequent changes are protected.

The development plan is as follows:

- Add feature gate
- Add API changes with basic webhook validation to only allow the new fields to be set if the feature gate is enabled
- Add controller changes with unit tests, behind the feature gate
- Add agent sidecar changes with unit tests, behind the feature gate
- Add extra webhook validations with validation tests, behind the feature gate
- Add integration and e2e tests, behind the feature gate
- Add documentation

I manually tested the feature gate. It works with either editing the deployment to use `--feature-gates="InPlaceRestart=true"` or editing the JobSet configuration to use `featureGates: InPlaceRestart: true`

```release-note
Add InPlaceRestart feature gate as alpha. Add support for configuring feature gates via the Configuration API.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -71,7 +71,6 @@ func main() {
 	var probeAddr string
 	var qps float64
 	var burst int
-	var featureGates string
 	var configFile string
 	flag.StringVar(&metricsAddr, "metrics-bind-address", ":8443", "The address the metric endpoint binds to.")
 	flag.StringVar(&probeAddr, "health-probe-bind-address", ":8081", "The address the probe endpoint binds to.")
@@ -80,7 +79,6 @@ func main() {
 			"Enabling this will ensure there is only one active controller manager.")
 	flag.Float64Var(&qps, "kube-api-qps", 500, "Maximum QPS to use while talking with Kubernetes API")
 	flag.IntVar(&burst, "kube-api-burst", 500, "Maximum burst for throttle while talking with Kubernetes API")
-	flag.StringVar(&featureGates, "feature-gates", "", "A set of key=value pairs that describe feature gates for alpha/experimental features.")
 	flag.StringVar(&configFile, "config", "",
 		"The controller will load its initial configuration from this file. "+
 			"Command-line flags will override any configurations set in this file. "+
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -118,6 +113,19 @@ func main() {
 		setupLog.Error(err, "Unable to load the configuration")
 		os.Exit(1)
 	}
+
+	if featureGates != "" {
```
kannon92: I think this would take feature gates from the cli and not from the config.

```suggestion
	if cfg.featureGates != "" {
```

I would like to deprecate feature gate CLI in place of the configuration option. Since we don't have any feature gates I think we could just remove the CLI option.

WDYT @andreyvelich @ahg-g ?

GiuseppeTT: > I think this would take feature gates from the cli and not from the config.

I followed the same approach from [Kueue](https://github.com/kubernetes-sigs/kueue/blob/81cae0608cac3d457b8ecf8a0480e7994c2151c1/cmd/kueue/main.go#L150-L160). If the CLI is provided (i.e., `featureGates != ""`), use it. Otherwise, use the configuration configmap

kannon92: Ah. Yea we added config later in Kueue and decided to go forward with recommending config feature gates rather than CLI.

kannon92: I don't see a reason why we need to carry both if we don't have any feature gates yet.

GiuseppeTT: > I would like to deprecate feature gate CLI in place of the configuration option. Since we don't have any feature gates I think we could just remove the CLI option.

IMO we should favor simplicity and support only one method. Kueue is good example of the success of the configuration method. We can add the CLI method later if someone requires it.

andreyvelich: I would also support to introduce Feature Gates only in JobSet Config, not in CLI.
Ideally, we should manage all system parameters via JobSet configMap

PR #1098, opened 2025-12-06
KEP-572: Implement VolumeClaimPolicies API for Stateful JobSet

andreyvelich:
#### What type of PR is this?

/kind api-change

This PR implemented VolumeClaimPolicies API to manage volumes for JobSet. Working example:
```yaml
kind: JobSet
metadata:
  name: volume-claim-trainjob
spec:
  volumeClaimPolicies:
    - templates:
        - metadata:
            name: initializer
          spec:
            accessModes: ["ReadWriteOnce"]
            resources:
              requests:
                storage: 1Gi
    - templates:
        - metadata:
            name: checkpoints
          spec:
            accessModes: ["ReadWriteOnce"]
            resources:
              requests:
                storage: 1Gi
      retentionPolicy:
        whenDeleted: Retain
  replicatedJobs:
    - name: initializer
      template:
        spec:
          template:
            spec:
              containers:
                - name: initializer
                  image: busybox
                  command:
                    - /bin/sh
                    - -c
                    - |
                      echo "Download pre-trained model into /workspace/model/qwen3.txt"
                      echo 'Qwen3-30b' > /workspace/model/qwen3.txt
                  volumeMounts:
                    - mountPath: /workspace/model
                      name: initializer
    - name: node
      dependsOn:
        - name: initializer
          status: Complete
      template:
        spec:
          parallelism: 3
          completions: 3
          template:
            spec:
              containers:
                - name: node
                  image: busybox
                  command:
                    - /bin/sh
                    - -c
                    - |
                      echo "Read pre-trained model"
                      cat /workspace/model/qwen3.txt
                      echo "Write model checkpoint to /workspace/checkpoints/model_node_${JOB_COMPLETION_INDEX}"
                      echo "Checkpoint: model_${JOB_COMPLETION_INDEX}" > /workspace/checkpoints/model_${JOB_COMPLETION_INDEX}.txt
                  volumeMounts:
                    - mountPath: /workspace/model
                      name: initializer
                    - mountPath: /workspace/checkpoints
                      name: checkpoints
```

TODO items:

- [x] Add unit tests for VolumeClaimPolicies controller changes
- [x] Add integration tests for VolumeClaimPolicies API
- [x] Add E2E tests for VolumeClaimPolicies API

/hold for review

Fixes: https://github.com/kubernetes-sigs/jobset/issues/572

/cc @kannon92 @GiuseppeTT @ahg-g @imreddy13 @carreter @tenzen-y @astefanutti @Electronic-Waste @vsoch @xigang @akshaychitneni @shravan-achar @bigsur0 

```release-note
Add the VolumeClaimPolicies API to support stateful JobSet.
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -100,7 +100,7 @@ func NewJobSetReconciler(client client.Client, scheme *runtime.Scheme, record re
 //+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=batch,resources=jobs/status,verbs=get;patch;update
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
-//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=create
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create
 
 // Reconcile is part of the main kubernetes reconciliation loop which aims to
 // move the current state of the cluster closer to the desired state.
--- a/charts/jobset/templates/controller/cluster_role.yaml
+++ b/charts/jobset/templates/controller/cluster_role.yaml
@@ -44,6 +44,9 @@ rules:
   - persistentvolumeclaims
   verbs:
   - create
+  - get
+  - list
+  - watch
 - apiGroups:
   - ""
   resources:
--- a/config/components/rbac/role.yaml
+++ b/config/components/rbac/role.yaml
@@ -27,6 +27,9 @@ rules:
   - persistentvolumeclaims
   verbs:
   - create
+  - get
+  - list
+  - watch
 - apiGroups:
   - ""
   resources:


pkg/controllers/jobset_controller.go
```diff
@@ -203,6 +204,14 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 		return ctrl.Result{}, err
 	}
 
+	// Reconcile VolumeClaimPolicies if it is set.
+	if len(js.Spec.VolumeClaimPolicies) > 0 {
```
kannon92, line 208: Does it make sense to reconcile volume claim polcies when Jobset is suspended?

kannon92: I guess they aren't any pods so they wouldn't bind but something to think through.

andreyvelich: Hm, good point, I think we should create PVCs even if JobSet is created with suspended status, since this API object only consumes etcd memory. It looks like LWS controller doesn't check suspended state (e.g. Replicas=0) before creating PVCs: https://github.com/kubernetes-sigs/lws/blob/main/pkg/controllers/leaderworkerset_controller.go#L778-L780

Is my understand correct @ahg-g @Edwinhr716 @andyzhangx ?

Maybe @soltysh has some suggestions from Job controller perspective on this?

kannon92: I think since we are only creating a global PVC for the entire workload this would be fine.

The main problem will be if we/when we start having PVC per ReplicatedJob. In that case, I think creating n replicas PVC could be a potential issue.

kannon92:
Want to test out CoPilot reviews.

kannon92:
I'm fine if you do this as a follow up but please don't forget to update our website with this feature.

kannon92:
Copilot gets a F for this one.

andreyvelich:
> I'm fine if you do this as a follow up but please don't forget to update our website with this feature.

Sure, let me track it in this issue: https://github.com/kubernetes-sigs/jobset/issues/1101

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -159,6 +160,16 @@ type JobSetSpec struct {
 	// +kubebuilder:validation:Minimum=0
 	// +optional
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
+
+	// volumeClaimPolicies is a list of policies for persistent volume claims that pods are allowed
+	// to reference. JobSet controller automatically adds the required volume claims to the
+	// pod template. Every claim in this list must have at least one matching (by name)
+	// volumeMount in one container in the template.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
```
kannon92, line 168: So I guess its good that this is immutable.

In some cases I'd like to be able to change JobSet objects while they are suspended. But in this case, I guess we are saying that this should never be modified even if the workload is suspended.

And the fact that we create even if suspended I guess this is fine.

Just calling this out.

andreyvelich: Good point, I couldn't imagine a use-case when volumes might be mutable during workload suspension.

Theoretically, users might want to mount a new PVC after JobSet preemption, however that will require to make Pod spec `.volumes` mutable, which is impossible today, right?

pkg/controllers/volume_claim_policy.go
```diff
@@ -0,0 +1,122 @@
...
+			}
+		} else {
+			log.V(2).Info("successfully created PVC", "pvc", klog.KObj(&pvc))
+		}
+	}
+
+	return nil
+}
+
+// generateSharedPVCName creates a name for the PVC in this format: <pvc-template-name>-<jobset-name>
```
kannon92:
```suggestion
// generatePVCName creates a name for the PVC in this format: <pvc-template-name>-<jobset-name>
```

pkg/controllers/volume_claim_policy.go
```diff
@@ -0,0 +1,122 @@
...
+
+		// Set PVC owner reference if retention policy is Delete.
+		if volumeClaimPolicy.RetentionPolicy != nil && volumeClaimPolicy.RetentionPolicy.WhenDeleted == jobset.RetentionPolicyDelete {
+			if err := ctrl.SetControllerReference(js, &pvc, r.Scheme); err != nil {
+				return err
+			}
+		}
+
+		// Create PVC if it doesn't exist.
+		// TODO (andreyvelich): Shall we consider to use workqueue.ParallelizeUntil, similar to rJobs creation?
```
kannon92: Rather than TODO, could we create an issue to document follow ups?

test/integration/controller/jobset_controller_test.go
```diff
@@ -1257,6 +1258,66 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("jobset with VolumeClaimPolicies should create PVCs", &testCase{
```
kannon92: We should test retention policies.

kannon92: Let's have a test for suspended PVC also.

pkg/controllers/jobset_controller.go
```diff
@@ -726,6 +735,11 @@ func constructJob(js *jobset.JobSet, rjob *jobset.ReplicatedJob, jobIdx int) *ba
 	jobsetSuspended := jobSetSuspended(js)
 	job.Spec.Suspend = ptr.To(jobsetSuspended)
 
+	// If VolumeClaimPolicies are set, update Job spec to set volumes and volumeMounts.
```
kannon92, line 738: One issue we can have is if VolumeClaimTempaltes have syntax errors or not not fully create (create goes back with no error but other issues occur)

We would then add volumes to a job that actually does not have an existing volumeClaimTemplate.

api/jobset/v1alpha2/jobset_types.go
kannon92, line 168: Can you add a test case confirming immutability in integrations

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -458,6 +469,46 @@ type Coordinator struct {
...
+type VolumeClaimPolicy struct {
+	// templates is a list of shared PVC claims that ReplicatedJobs are allowed to reference.
+	// The JobSet controller is responsible for creating shared PVCs that can be mounted by
+	// multiple ReplicatedJobs. Every claim in this list must have a matching (by name)
+	// volumeMount in one container or initContainer in at least one ReplicatedJob template.
+	// ReplicatedJob template must not have volumes with the same name as defined in this template.
+	// Generated PVC naming convention: <claim-name>-<jobset-name>
+	// Example: "model-cache-trainjob" (shared volume across all ReplicatedJobs).
+	// +kubebuilder:validation:MaxItems=50
+	// +listType=atomic
```
kannon92, line 485: I'm trying to fix the linter but I think it would complain here once I enabled Option or required settings.

Each field should have an optional tag.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -458,6 +469,46 @@ type Coordinator struct {
...
+	// By default, all persistent volume claims are deleted once JobSet is deleted.
+	// +optional
+	RetentionPolicy *VolumeRetentionPolicy `json:"retentionPolicy,omitempty"`
+}
+
+// volumeRetentionPolicy defines the retention policy used for PVCs created from the JobSet VolumeClaimPolicies.
+type VolumeRetentionPolicy struct {
+	// whenDeleted specifies what happens to PVCs when JobSet is deleted.
+	// +kubebuilder:validation:Enum=Delete;Retain
+	// +kubebuilder:default=Delete
```
kannon92, line 498: I think this should have +required

pkg/controllers/volume_claim_policy.go
```diff
@@ -0,0 +1,122 @@
...
+		if err := r.Create(ctx, &pvc); err != nil {
+			if !apierrors.IsAlreadyExists(err) {
```
kannon92, line 62: If retentionPolicy is Retain and someone creates a JobSet with a PVC.
And then they delete the JobSet with updated PVC templates will this cause any error?

pkg/webhooks/jobset_webhook.go
```diff
@@ -409,6 +425,115 @@ func validateCoordinatorLabelValue(js *jobset.JobSet) error {
...
+	var allErrs []error
+	claimNames := sets.New[string]()
+
+	// Collect all claim names from templates.
+	for policyIdx, policy := range volumeClaimPolicies {
+		fieldPath := field.NewPath("spec", "volumeClaimPolicies").Index(policyIdx)
+		for templateIdx, template := range policy.Templates {
+			templateFieldPath := fieldPath.Child("template").Index(templateIdx)
+
+			// Validate template namespace.
```
kannon92: Can we try and validate this with CEL on the API?

If not, we should also document at the API level the restrictions we put on VolumeClaimPolicies.

pkg/webhooks/jobset_webhook.go
```diff
@@ -409,6 +425,115 @@ func validateCoordinatorLabelValue(js *jobset.JobSet) error {
...
+				))
+			}
+
+			// Validate DNS-1123 subdomain name
+			for _, err := range validation.IsDNS1123Subdomain(template.Name) {
+				allErrs = append(allErrs, field.Invalid(templateFieldPath.Child("name"), template.Name, err))
+			}
+
+			// Validate PVC name length limits
+			estimatedNameLength := len(template.Name) + len(js.Name) + 10 // Buffer for indices and separators
```
kannon92: Not really obvious to me where 10 comes from.

What indices and separators are you talking about?

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -2032,11 +2101,453 @@ func TestValidateCreate(t *testing.T) {
...
+											},
+										},
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			want: errors.Join(fmt.Errorf("namespace cannot be set for the VolumeClaimPolicies template")),
```
kannon92: Could we make the error message more descriptive on what template is failing (maybe name plus namespace?)

test/integration/controller/jobset_controller_test.go
kannon92: Can we add a test for retention polciy Retain but include a delete and recreate of a JobSet with update volume claims?

pkg/controllers/volume_claim_policy.go
kannon92: I guess this is a potential bug as PVC will be the old PVC.

test/e2e/e2e_test.go
```diff
@@ -581,6 +582,38 @@ var _ = ginkgo.Describe("JobSet", func() {
...
+
+			// Create JobSet with VolumeClaimPolicies.
+			ginkgo.By("creating jobset with VolumeClaimPolicies")
+			js := volumeClaimTestJobSet(ns).Obj()
+
+			// Verify jobset created successfully.
+			ginkgo.By("checking that jobset creation succeeds")
+			gomega.Expect(k8sClient.Create(ctx, js)).Should(gomega.Succeed())
+
+			// Verify PVCs are created.
```
kannon92, line 602: There could be a potential flake issue here where JobSet does not created.

May be worth verifying that jobset is created before checking for PVC.

test/e2e/e2e_test.go
```diff
@@ -581,6 +582,38 @@ var _ = ginkgo.Describe("JobSet", func() {
 			gomega.Expect(k8sClient.Delete(ctx, jobSet)).To(gomega.Succeed())
 		})
 	})
+	// This test runs JobSet with the VolumeClaimPolicies API.
+	ginkgo.When("VolumeClaimPolicies is enabled on JobSet", func() {
```
kannon92, line 586: Do you think adding a test case with multiple replicas would be useful??

Curious on how this work handles that all the jobs will write to the same volume claim.

kannon92 (commented):
Really nice work.

I have a few comments on test cases and some potential edge cases.
But this is looking really good!

pkg/controllers/volume_claim_policy.go
andreyvelich: Sure, I am just unsure whether we want to parallelize PVC creation in the future.
@kannon92 Do you think that might be useful?

kannon92: It feels a bit like YAGNI right now tbh.

pkg/controllers/jobset_controller.go
andreyvelich: I believe, for the syntax error we verify the DNS-1123 name: https://github.com/kubernetes-sigs/jobset/blob/07ced4f814f1cc66beb604d50df0f82293770e30/pkg/webhooks/jobset_webhook.go#L459

For other errors, we expose PVCCreationFailedReason to the JobSet events. Should that be sufficient?

pkg/controllers/volume_claim_policy.go
andreyvelich: How would they be able to modify the PVC template of running/suspended JobSet if API is immutable?

pkg/webhooks/jobset_webhook.go
andreyvelich: Maybe we should decrease it to 1?
Since we construct name as PVC template name + JobSet name.

pkg/webhooks/jobset_webhook_test.go
andreyvelich: This is the full error example:
```
            	Error:      	"spec.volumeClaimPolicies[0].template[0].namespace: Invalid value: \"invalid-namespace\": 
namespace cannot be set for the VolumeClaimPolicies template" does not contain "namespace cannot be set for
 the VolumeClaimPoliciesa template"
```

Since we bypass the index into `field.Invalid()`, the error will contain the index of the invalid slice.

@kannon92 Do you think that is not sufficient?

test/integration/controller/jobset_controller_test.go
andreyvelich: In this test do you want to modify the JobSet's pod volume mount path before JobSet is recreated?

test/e2e/e2e_test.go
andreyvelich: Sure, let me create simple tests for multiple Replicas, maybe we can also verify Retain policy in E2Es.

pkg/controllers/volume_claim_policy.go
andreyvelich: Let me remove it for now, we can come back to it later.

pkg/webhooks/jobset_webhook_test.go
kannon92: Good point. This is probably fine.

test/integration/controller/jobset_controller_test.go
kannon92: I mean suspended JobSet

pkg/controllers/jobset_controller.go
kannon92: We still would addVolumes even if we got this error though?

The events will emit it but I don't think we check for success on adding Volumes.

api/jobset/v1alpha2/jobset_types.go
kannon92: Rebase should pick these up now.

pkg/controllers/volume_claim_policy.go
kannon92: In that case you are correct.

I am thinking Retain means that even if you remove the JobSet the PVC will still stay around.

So I'm thinking of the following scenario:

1) Create JobSet With PVC Retain.
2) Delete JobSet.
3) PVC should still be left around.
4) Create a new JobSet with same name with new PVC parameters.
5) Will the PVC get updated with new parameters or will it be left with parameters in 1?

pkg/webhooks/jobset_webhook.go
kannon92: could we just export the function that computes this and use that?

I don't get why we have to estimate this. We should know exactly what the length would be.

pkg/controllers/jobset_controller.go
andreyvelich: Yes, we still try to add Volumes to the Pod spec, and Pods should stuck in the Pending phase.
But that should be fine, what do you think?

pkg/controllers/volume_claim_policy.go
andreyvelich: oh, that is good point.
It looks like StatefulSet controller also doesn't validate it: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/statefulset/stateful_pod_control.go#L354

I am wondering if we should introduce any additional validation for such use-cases?

pkg/controllers/jobset_controller.go
kannon92: I'm not sure they would be stuck in pending. Pending means that the scheduler wasn't able to find a node. I think it would go to ContainerCreateFailed or PodReadyToStartContainers=False.


I think they would end up in a failed state since the PVC wasn't created.

pkg/controllers/volume_claim_policy.go
kannon92: Yea maybe we can check that PVC matches at least.

pkg/webhooks/jobset_webhook.go
andreyvelich: Yes, CEL looks to be working fine for namespace.
Verifying the name uniqueness is not possible in CEL, since the complexity is O(n^2) which won't work in CEL.

pkg/controllers/volume_claim_policy.go
andreyvelich: @kannon92 Added validation of existing PVCs here: https://github.com/kubernetes-sigs/jobset/pull/1098/commits/500b7ab7338be4286387fa184458ca9517df4df9

kannon92:
I'll review this on Friday.

pkg/controllers/jobset_controller.go
andreyvelich: Just verified it, and I can see that Pods will be in the Pending phase::
```yaml
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2025-12-29T19:10:26Z"
    message: '0/1 nodes are available: persistentvolumeclaim "test-volume" not
      found. preemption: 0/1 nodes are available: 1 Preemption is not helpful for
      scheduling.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
  qosClass: BestEffort
```

andreyvelich:
@kannon92 Thanks for the review! I believe, I addressed all the suggestions, appreciate if you could take a look again!

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -159,6 +160,16 @@ type JobSetSpec struct {
...
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
+
+	// volumeClaimPolicies is a list of policies for persistent volume claims that pods are allowed
+	// to reference. JobSet controller automatically adds the required volume claims to the
+	// pod template. Every claim in this list must have at least one matching (by name)
+	// volumeMount in one container in the template.
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
+	// +optional
+	// +listType=atomic
+	// +kubebuilder:validation:MaxItems=50
```
kannon92, line 171: can we have a integratoin test for webhook for this validation also?

pkg/controllers/jobset_controller.go
kannon92: sounds good then.

pkg/controllers/jobset_controller.go
```diff
@@ -100,6 +100,7 @@ func NewJobSetReconciler(client client.Client, scheme *runtime.Scheme, record re
 //+kubebuilder:rbac:groups=batch,resources=jobs,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=batch,resources=jobs/status,verbs=get;patch;update
 //+kubebuilder:rbac:groups=core,resources=services,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=core,resources=persistentvolumeclaims,verbs=get;list;watch;create;update;patch;delete
```
kannon92: Are all these rbac roles really required?

test/integration/controller/jobset_controller_test.go
kannon92: This test seems to be testing both retain and Deleted in the same test case. Could you update the title or consider separating these as two separate tests.

kannon92 (approved):
/lgtm

I have a few nits but I think this is good to go.

pkg/controllers/volume_claim_policy.go
kannon92: Looks good!

test/e2e/e2e_test.go
kannon92: I missed this when I was reviewing. Was this done?

andreyvelich: Yes, I used 2 Replicas for the E2E: https://github.com/kubernetes-sigs/jobset/blob/c2d28e68563afe253385c0c2a20739fcc9ef22a0/test/e2e/e2e_test.go#L895

pkg/controllers/jobset_controller.go
andreyvelich: Good point, actually we probably need only `create`.

andreyvelich:
> Should we update this function to include PersistentVolumeClaims?
> 
> https://github.com/kubernetes-sigs/jobset/blob/main/pkg/controllers/jobset_controller.go#L227-L235

Not really, since client uses cache for [this call](https://github.com/kubernetes-sigs/jobset/blob/133a94c457f7992616177f315d46e1023d836b7a/pkg/webhooks/jobset_webhook.go#L472), we just need to add `list` and `watch` RBAC.

Also, we don't trigger JobSet reconciler for Service or PVCs updates, so we might want to watch only for Jobs:
WDYT @kannon92 ?

```go
// SetupWithManager sets up the controller with the Manager.
func (r *JobSetReconciler) SetupWithManager(mgr ctrl.Manager) error {
	return ctrl.NewControllerManagedBy(mgr).
		For(&jobset.JobSet{}).
		Owns(&batchv1.Job{}).
		Complete(r)
}
```

andreyvelich:
@kannon92 Please let me know if you have more suggestion to this PR!

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -555,5 +555,203 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+										},
+									},
+								},
+							}).
+							CompletionMode(batchv1.IndexedCompletion).Obj()).
+						Obj())
+			},
+			jobSetCreationShouldFail: true,
+		}),
+		ginkgo.Entry("VolumeClaimPolicies exceeding MaxItems=50 should be rejected", &testCase{
```
kannon92, line 676: I'm having a hard time telling the difference between these two test cases.

kannon92: Ah okay. I read the comments.

this seems fine.

andreyvelich:
Thanks for your review and help @kannon92!

/hold cancel

PR #1110, opened 2025-12-22
Add GangScheduling for JobSet

### LINKED ISSUE #969: Gang Scheduling of JobSets

Following [LWS gang scheduling](https://github.com/kubernetes-sigs/lws/tree/main/keps/407-gang-scheduling#startuppolicy-handling), I think we should have a way to support gang scheduling via PodGroups.

**Why is this needed**:

JobSets usually require tight grouping and I think gang scheduling would be useful to enforce scheduling at one time.

Projects like Kubeflow Traininer and KAI use PodGroups to enforce gang scheduling but our controller does not create PodGroups.

This means that these projects will create their own PodGroups which is not ideal.

**Why not Kueue**:

In previous asks for this, we used to say that we would recommend the use of Kueue for gang scheduling. I still think this is valid but there are other options. Also, it is heavyweight to recommend Kueue for gang scheduling if someone only wants a JobSet to be scheduled together.

So I think having this repo create the PodGroups would make sense.

- [x] Design doc
- [x] API change
- [x] Docs update

kannon92:
#### What type of PR is this?
/kind feature

Fixes #969 

```release-note
Add gang scheduling support for JobSet
```


kannon92:
Sounds like @imreddy13 is interested in taking this on.

Going to close this out for now.

/close

PR #1068, opened 2025-10-15
KEP for GangConfig (to support Gang Scheduling)

imreddy13:
#### What type of PR is this?
/kind documentation

#### What this PR does / why we need it:

Proposal to add a new field `GangConfig` to ReplicatedJob API  to support Gang Scheduling (KEP: https://github.com/kubernetes/enhancements/tree/master/keps/sig-scheduling/4671-gang-scheduling)

#### Which issue(s) this PR fixes:
Fixes #969

#### Does this PR introduce a user-facing change?
```release-note
     NONE
```

PR #1111, opened 2025-12-22
Possible KEP for gang scheduling

kannon92:
#### What type of PR is this?
/kind documentation

KEP to demonstrate how we could use workload API

Fixes #

```release-note
NONE
```

### DIFF
--- a/keps/969-GangScheduling/README.md
+++ b/keps/969-GangScheduling/README.md
@@ -9,7 +9,7 @@
   - [User Stories](#user-stories)
     - [Story 1: All-or-Nothing Scheduling for Distributed Training](#story-1-all-or-nothing-scheduling-for-distributed-training)
     - [Story 2: Per-ReplicatedJob Gang Scheduling](#story-2-per-replicatedjob-gang-scheduling)
-    - [Story 3: Advanced Workload Template Control](#story-3-advanced-workload-template-control)
+    - [Story 3: Custom Pod Groupings](#story-3-custom-pod-groupings)
   - [Risks and Mitigations](#risks-and-mitigations)
 - [Design Details](#design-details)
   - [API Proposal](#api-proposal)
@@ -25,6 +25,7 @@
 - [Implementation History](#implementation-history)
 - [Drawbacks](#drawbacks)
 - [Alternatives](#alternatives)
+  - [Using an Enum with Preset Gang Policies](#using-an-enum-with-preset-gang-policies)
   - [Using Annotations Instead of API Field](#using-annotations-instead-of-api-field)
   - [Hard-coding Gang Policy Behavior](#hard-coding-gang-policy-behavior)
 <!-- /toc -->
@@ -36,10 +37,8 @@ Gang scheduling ensures that groups of pods are scheduled atomically - either al
 together, or none are scheduled. This is critical for distributed AI/ML training workloads and HPC applications
 where partial scheduling leads to resource deadlock and wasted cluster capacity.
 
-The proposal introduces three gang scheduling modes:
-- **JobSetAsGang**: The entire JobSet is treated as a single gang (all-or-nothing scheduling)
-- **JobSetGangPerReplicatedJob**: Each ReplicatedJob within the JobSet is treated as a separate gang
-- **JobSetWorkloadTemplate**: Advanced mode allowing users to provide custom Workload specifications for fine-grained control
+The proposal uses a `WorkloadTemplate` approach that allows users to provide a custom Workload specification
+for fine-grained control over pod groupings and gang scheduling behavior.
 
 ## Motivation
 
@@ -85,7 +84,7 @@ but currently lacks first-class integration with JobSet. This KEP bridges that g
 As a ML engineer running distributed PyTorch training with 64 GPUs across 8 nodes, I want all 64 pods to be
 scheduled together or not at all. Partial scheduling wastes expensive GPU resources and blocks other jobs.
 
-With this KEP, I can specify:
+With this KEP, I can specify a WorkloadTemplate that defines a single PodSet containing all pods:
 
 ```yaml
 apiVersion: jobset.x-k8s.io/v1alpha2
@@ -94,7 +93,11 @@ metadata:
   name: pytorch-training
 spec:
   gangPolicy:
-    gangPolicyOption: JobSetAsGang
+    workloadTemplate:
+      spec:
+        podSets:
+        - name: all-pods
+          count: 8  # Total: 1 master + 7 workers
   replicatedJobs:
   - name: master
     replicas: 1
@@ -134,6 +137,8 @@ As a HPC user running a multi-tier application with separate parameter servers a
 to be gang-scheduled independently. Parameter servers can start as soon as their gang is ready, while workers
 wait for their separate gang to be satisfied.
 
+With the WorkloadTemplate approach, I define multiple PodSets - one for each ReplicatedJob:
+
 ```yaml
 apiVersion: jobset.x-k8s.io/v1alpha2
 kind: JobSet
@@ -141,7 +146,13 @@ metadata:
   name: multi-tier-training
 spec:
   gangPolicy:
-    gangPolicyOption: JobSetGangPerReplicatedJob
+    workloadTemplate:
+      spec:
+        podSets:
+        - name: ps
+          count: 4
+        - name: worker
+          count: 16
   replicatedJobs:
   - name: ps
     replicas: 4
@@ -169,14 +180,13 @@ spec:
 
 Parameter server pods (4) will be gang-scheduled together, and worker pods (16) will be gang-scheduled as a separate group.
 
-#### Story 3: Advanced Workload Template Control
+#### Story 3: Custom Pod Groupings
 
 As a ML engineer running distributed training, I have an initialization phase that must complete before training starts,
-but I want the leader and worker pods to be gang-scheduled together as a single unit (not separately). The simple gang
-policies don't support this pattern where I need custom grouping across ReplicatedJobs.
+but I want the leader and worker pods to be gang-scheduled together as a single unit (not separately).
 
-With the WorkloadTemplate option, I can define custom pod sets that group the leader and worker together while keeping
-the initializer separate:
+With the WorkloadTemplate approach, I can define custom pod groupings that combine pods from different ReplicatedJobs
+into a single gang:
 
 ```yaml
 apiVersion: jobset.x-k8s.io/v1alpha2
@@ -185,8 +195,7 @@ metadata:
   name: multi-phase-training
 spec:
   gangPolicy:
-    gangPolicyOption: JobSetWorkloadTemplate
-    workload:
+    workloadTemplate:
       metadata:
         labels:
           workload-type: multi-phase-training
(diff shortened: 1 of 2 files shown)


kannon92:
cc @andreyvelich

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+
+**Webhook Validation**:
+```go
+// Validating webhook checks:
+// 1. Workload field is only set when using JobSetWorkloadTemplate mode
+// 2. PodSet counts in Workload template match ReplicatedJob specs
+// 3. Workload API is available in the cluster
+```
+
+### User Experience
```
kannon92, line 384: Will remove this.

Workload UX should be used not added into JobSet.

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+### Graduation Criteria
+
+**Alpha** (v0.11):
+- API field added with validation
+- Basic implementation for JobSetAsGang and JobSetGangPerReplicatedJob modes
+- Integration with Kueue scheduler
+- Unit and integration test coverage
+- Documentation for basic usage
+
+**Beta** (v0.7.0):
```
kannon92: Beta should follow feature gate for Workload in tree.

The rest of this is not useful.

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+
+This KEP introduces gang scheduling support for JobSets through a new `gangPolicy` field in the JobSet API.
+Gang scheduling ensures that groups of pods are scheduled atomically - either all pods in the gang are scheduled
+together, or none are scheduled. This is critical for distributed AI/ML training workloads and HPC applications
+where partial scheduling leads to resource deadlock and wasted cluster capacity.
+
+The proposal introduces three gang scheduling modes:
+- **JobSetAsGang**: The entire JobSet is treated as a single gang (all-or-nothing scheduling)
+- **JobSetGangPerReplicatedJob**: Each ReplicatedJob within the JobSet is treated as a separate gang
+- **JobSetWorkloadTemplate**: Advanced mode allowing users to provide custom Workload specifications for fine-grained control
```
andreyvelich: If we plan to introduce Workload template, why do we really need to have other policies? Is it just to simplify UX?
I am wondering if documentation which explains how to configure Workload API is sufficient?

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+With the WorkloadTemplate option, I can define custom pod sets that group the leader and worker together while keeping
+the initializer separate:
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: multi-phase-training
+spec:
+  gangPolicy:
```
andreyvelich, line 197: Given that in the future Workload API will be integrated with TAS and DRA, how are we planning to integrate it into JobSet API?
- https://github.com/kubernetes/enhancements/pull/5711
- https://github.com/kubernetes/enhancements/pull/5733

If we want to support templates via JobSet API, do we really need to introduce `gangPolicy` API?

One of the idea could be to directly introduce the `spec.workloadTemplate` API similarly to `spec.volumeClaimTemplates` in StatefulSet.

With that, we can gradually support more Workload features in JobSet controller.

@ahg-g @Edwinhr716 Do we have any updates from the LWS API on the Workload API will be supported?

cc @macsko  @wojtek-t @erictune @44past4 @johnbelamaric

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+1. **Workload Creation**:
+   - When a JobSet with `gangPolicy` is created, the controller creates a corresponding Workload object
+   - For `JobSetAsGang`: Create a single Workload with one PodSet containing all pods
+   - For `JobSetGangPerReplicatedJob`: Create a single Workload with multiple PodSets (one per ReplicatedJob)
+   - For `JobSetWorkloadTemplate`: Create the user-provided Workload template
+
+2. **Ownership and Lifecycle**:
+   - The Workload is owned by the JobSet (via OwnerReference) and shares its lifecycle
+   - Deleting the JobSet cascades deletion to the Workload
+   - The Workload name follows the pattern: `<jobset-name>-workload`
```
andreyvelich: I would suggest to just use JobSet name, similar to how we create PVCs.
```suggestion
   - The Workload name follows the pattern: `<jobset-name>`
```

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+
+3. **Job Creation Coordination**:
+   - Jobs are created with appropriate labels/annotations to link them to the Workload's PodSets
+   - The scheduler (e.g., Kueue) uses these labels to enforce gang scheduling constraints
+
+4. **Status Synchronization**:
+   - JobSet status is updated based on both Job status and Workload admission status
+   - New condition: `WorkloadAdmitted` indicates when the gang has been admitted for scheduling
```
andreyvelich: How would we check that gang has been admitted?

andreyvelich (commented):
Thanks for this @kannon92!
Shall we close this KEP: https://github.com/kubernetes-sigs/jobset/pull/1068 to continue discussion in this PR?

cc @astefanutti @tenzen-y @GiuseppeTT @imreddy13

keps/969-GangScheduling/README.md
kannon92: I need to update this actually.

Right now there is no status field on Workloads and I think the only way to tell if a gang has been admitted is if the workload is scheduled.

I believe there is a KEP to add more status fields to Workload but it is not yet here. I don't see conditions or events. So the way may be (for now) to read the kubescheduler logs.

keps/969-GangScheduling/README.md
kannon92: That was sorta my thought on Workload Template for more advanced features.

Same for supporting DependsOn with GangScheduling.

keps/969-GangScheduling/README.md
kannon92: yea its simpler to have this handled for someone.

For Job, I don't necessarly see a reason to have WorkloadTemplate (plus workload api is v1alpha1 and the Job API is v1).

For JobSet, I figured we can use this API for more advanced features but maybe the quick options are easier to support for simple use cases (all Gang or JobSetGangPerReplicatedJob).

kannon92:
> Thanks for this @kannon92! Shall we close this KEP: #1068 to continue discussion in this PR?
> 
> cc @astefanutti @tenzen-y @GiuseppeTT @imreddy13

I'll wait for the author to get back to me before I close.

I created this as there was some discussion on JobSet in a google doc and I wanted to see what options we could have.

I also have a parallel KEP for Job.

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+	// and scheduling logic for a JobSet.
+	JobSetWorkloadTemplate GangPolicyOptions = "JobSetWorkloadTemplate"
+)
+
+type GangPolicy struct {
+	// gangPolicyOption determines the gang scheduling policy for JobSet.
+	// Defaults to nil (no gang scheduling).
+	// +kubebuilder:validation:Enum=JobSetAsGang;JobSetGangPerReplicatedJob;JobSetWorkloadTemplate
+	// +optional
+	GangPolicyOption *GangPolicyOptions `json:"gangPolicyOption,omitempty"`
```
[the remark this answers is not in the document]
kannon92: This is wrong.

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,589 @@
...
+   - **Failure Policy**: Gang scheduling is respected during failure recovery and restarts
+   - **StartupPolicy**: Gang admission happens before startup sequencing begins
+
+### Integration with Kubernetes Workload API
+
+The Workload API (k8s.io/api/scheduling/v1alpha1) provides the abstraction for gang scheduling:
+
+```go
+// Example generated Workload for JobSetAsGang mode
+apiVersion: scheduling.x-k8s.io/v1alpha1
```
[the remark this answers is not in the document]
kannon92: This isn't necessary.

keps/969-GangScheduling/README.md
Edwinhr716: Initial plan on the LWS side is to have an API field when the user wants the controller to create the workload template. Though as new features are added, it does make more sense to add the workload template itself to the spec.

cc @helayoty

keps/969-GangScheduling/README.md
kannon92: updated.

keps/969-GangScheduling/README.md
kannon92: updated. removed this number actually.

keps/969-GangScheduling/README.md
andreyvelich: I would be nice to keep API consistent between JobSet and LWS, and introducing WorkloadTemplate makes sense to me.

kannon92: Hmm okay.

For Job I don't think we can have Workload Template so I was going to keep the enum structure there.

kannon92: Updated the KEP with this.

imreddy13:
I haven't read this in detail yet (back from vacation) but I had a KEP for this a few months ago: https://github.com/kubernetes-sigs/jobset/pull/1068

We were actually planning to add it for 1.36 when Workload API will be beta after discussing with Wojtekt/Eric that 1.35 might be too early.

@ahg-g is separately designing for LWS too

kannon92:
> I haven't read this in detail yet (back from vacation) but I had a KEP for this a few months ago: #1068
> 
> We were actually planning to add it for 1.36 when Workload API will be beta after discussing with Wojtekt/Eric that 1.35 might be too early.
> 
> @ahg-g is separately designing for LWS too

Sounds good.

I have the [Job](https://github.com/kubernetes/enhancements/pull/5760) opened up.

I'll let you review this but we can close this out. 

I wanted to draft a potential API just to see what gaps we could have with current implementation.

keps/969-GangScheduling/README.md
```diff
@@ -0,0 +1,569 @@
...
+### API Proposal
+
+```go
+type GangPolicy struct {
+	// workloadTemplate provides a Workload template to create on JobSet creation.
+	// When specified, the JobSet controller will create this Workload object
+	// and link it to the JobSet for gang scheduling.
+	// The Workload's podSets define how pods should be grouped for gang scheduling.
+	// +optional
+	WorkloadTemplate *schedulingv1alpha1.WorkloadSpec `json:"workloadTemplate,omitempty"`
```
andreyvelich, line 290: I think, if we are going ahead with `workloadTemplate` approach, we should move it directly under JobSet spec, and we should allow users to set Workload object metadata like labels and annotations.
The API might look as follows:

```go
type JobSetSpec struct {
  WorkloadTemplate *WorkloadTemplate`json:"workloadTemplate,omitempty"`
}

type WorkloadTemplate struct {
    metav1.ObjectMeta `json:"metadata,omitempty"`

    spec *schedulingv1alpha1.WorkloadSpec `json:"spec,omitempty"`
}
```

PR #1119, opened 2026-01-13
Switch controller events to events.k8s.io/v1

erain:
#### What type of PR is this?

/kind cleanup

- Switch controller event emission to `events.k8s.io/v1` using the client-go events broadcaster.
- Fail fast on startup if the events.k8s.io/v1 API is not available.
- Map action to reason consistently for v1 events.
- Keep leader-election events on core/v1 (no forked behavior).
- Update RBAC/Helm roles to allow events.k8s.io writes.
- Update unit/integration tests to assert v1 event shape and series behavior.
- Add a benchmark for v1 event recording throughput.

Fixes # (n/a)

- The new event broadcaster is wired for controllers only; leader-election remains unchanged.
- Tests cover v1 event emission and series aggregation via `events.EventRecorder`.

#### Performance results:
```
# Baseline (upstream/main + local temp benchmark using core/v1 recorder)
$ go test -run=^$ -bench=BenchmarkEventRecorderCoreV1 ./pkg/controllers
BenchmarkEventRecorderCoreV1-12  578979  4312 ns/op  1347 B/op  16 allocs/op

# This PR (events.k8s.io/v1)
$ go test -run=^$ -bench=BenchmarkEventRecorderV1 ./pkg/controllers
BenchmarkEventRecorderV1-12  458974  2531 ns/op  1626 B/op  10 allocs/op
```

```release-note
JobSet controllers now emit Kubernetes events using the `events.k8s.io/v1` API.
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -17,9 +17,11 @@ limitations under the License.
 package main
 
 import (
+	"context"
 	"crypto/tls"
 	"errors"
 	"flag"
+	"fmt"
 	"net/http"
 	"os"
 
@@ -27,10 +29,14 @@ import (
 	// to ensure that exec-entrypoint and run can make use of them.
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
 
+	eventsv1 "k8s.io/api/events/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
+	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/rest"
+	k8sevents "k8s.io/client-go/tools/events"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/healthz"
 	"sigs.k8s.io/controller-runtime/pkg/log/zap"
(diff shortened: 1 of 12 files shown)


GiuseppeTT:
/retest

I'm rerunning the tests, but I suspect this is not a flake.

kannon92:
Trying out copilot reviews right now.

kannon92:
@GiuseppeTT or @andreyvelich could one of you take a look?

pkg/controllers/jobset_controller.go
```diff
@@ -91,11 +91,11 @@ type eventParams struct {
 	eventMessage string
 }
 
-func NewJobSetReconciler(client client.Client, scheme *runtime.Scheme, record record.EventRecorder) *JobSetReconciler {
+func NewJobSetReconciler(client client.Client, scheme *runtime.Scheme, record events.EventRecorder) *JobSetReconciler {
 	return &JobSetReconciler{Client: client, Scheme: scheme, Record: record, clock: clock.RealClock{}}
 }
 
-//+kubebuilder:rbac:groups="",resources=events,verbs=create;watch;update;patch
+//+kubebuilder:rbac:groups=events.k8s.io,resources=events,verbs=create;watch;update;patch
```
andreyvelich, line 98: We should keep RBAC for the old events as well, otherwise controller won't be able to read previously generated events and return errors: https://github.com/kubeflow/trainer/pull/3127

pkg/controllers/volume_claim_policy.go
```diff
@@ -60,7 +60,7 @@ func (r *JobSetReconciler) createPVCsIfNecessary(ctx context.Context, js *jobset
 		// Create PVC if it doesn't exist.
 		if err := r.Create(ctx, &pvc); err != nil {
 			if !apierrors.IsAlreadyExists(err) {
-				r.Record.Eventf(js, corev1.EventTypeWarning, constants.PVCCreationFailedReason, err.Error())
+				r.Record.Eventf(js, nil, corev1.EventTypeWarning, constants.PVCCreationFailedReason, constants.PVCCreationFailedReason, err.Error())
```
andreyvelich, line 63: Similar to this comment, we can use Reconciling action for such event: https://github.com/kubeflow/trainer/pull/3127#discussion_r2732535498

```suggestion
				r.Record.Eventf(js, nil, corev1.EventTypeWarning, constants.PVCCreationFailedReason, "Reconciling", err.Error())
```

main.go
```diff
@@ -187,6 +193,18 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "unable to initialize events broadcaster")
+		os.Exit(1)
+	}
+	go func() {
+		<-ctx.Done()
+		eventBroadcaster.Shutdown()
+	}()
+	jobSetRecorder := eventBroadcaster.NewRecorder(mgr.GetScheme(), "jobset")
+	podRecorder := eventBroadcaster.NewRecorder(mgr.GetScheme(), "pod")
```
andreyvelich, line 206: Do we really need this event broadcaster ?

andreyvelich (commented):
Thank you @erain!

main.go
GiuseppeTT: > Do we really need this event broadcaster ?

+1

IMO we should have the simplest PR that migrates JobSet to the new events API.

Issue #1084, opened 2025-11-08
Failed to deploy JobSet and LeaderWorkerSet controllers in the same namespace


andreyvelich:
**What happened**:

As part of testing the Kubeflow Distributed Data Cache, @akshaychitneni discovered that the LWS and JobSet controllers cannot be deployed within the same namespace. 
Currently, as a workaround, we need to manually patch the deployments to use different label selectors for the `metrics-service` and `webhook-service`: https://github.com/kubeflow/trainer/pull/2923

The issue arises because both JobSet and LWS use the same `controller-manager` label selector to bind their services to the deployment:

- JobSet: https://github.com/kubernetes-sigs/jobset/blob/main/config/components/manager/manager.yaml#L28-L30
- LWS: https://github.com/kubernetes-sigs/lws/blob/main/config/manager/manager.yaml#L28-L30

We should add the correct prefix for selectors to distinguish them.

cc @kannon92 @ahg-g @kerthcet @ardaguclu @GiuseppeTT @astefanutti @tenzen-y 

**What you expected to happen**:

JobSet and LWS control plane should be deployed in `kubeflow-system` namespace.

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): v1.31
- JobSet version (use `git describe --tags --dirty --always`): v0.10.1
- Cloud provider or hardware configuration:
- Install tools:
- Others:

kannon92:
@rphillps proposed https://github.com/kubernetes-sigs/kueue/pull/7371 in Kueue.

I think we can follow a similar pattern.

GiuseppeTT:
The helm chart of JobSet uses the following labels in the selector. It should fix the problem.

```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/component: controller
    app.kubernetes.io/instance: jobset
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: jobset
    app.kubernetes.io/version: 0.10.1
    helm.sh/chart: jobset-0.10.1
  name: jobset-controller
  namespace: jobset-system
spec:
  replicas: 1
  selector:
    matchLabels:
      app.kubernetes.io/component: controller
      app.kubernetes.io/instance: jobset
      app.kubernetes.io/name: jobset
  template:
    metadata:
      annotations:
[...]

PR #1120, opened 2026-01-14
Add jobset labels when deployed with kustomize

akshaychitneni:
#### What type of PR is this?
/kind bug

This PR adds labels to deployment and service objects to avoid conflicts when deployed using kustomize

Fixes # https://github.com/kubernetes-sigs/jobset/issues/1084

jobset upgrades should also work as expected

```release-note

```

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -8,9 +8,11 @@ namespace: jobset-system
 # field above.
 namePrefix: jobset-
 
-# Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+# Common labels to add to all resources and selectors (matches Helm behavior).
+commonLabels:
+  app.kubernetes.io/name: jobset
+  app.kubernetes.io/instance: jobset
+  app.kubernetes.io/managed-by: kustomize
 
 resources:
 - ../components/crd


kannon92:
Please sign the CLA.

kannon92:
cc @andreyvelich

config/default/kustomization.yaml
```diff
@@ -8,9 +8,12 @@ namespace: jobset-system
...
 
-# Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+# Labels to add to all resources (metadata only, does not affect selectors/matchLabels).
+labels:
+  - pairs:
+      app.kubernetes.io/name: jobset
+      app.kubernetes.io/instance: jobset
+    includeTemplates: true
```
andreyvelich: We only change selectors for the Services, not in Deployment right?
In that case, do we need to make any changes to LWS to ensure that their service won't get attached to the JobSet deployment?
cc @ahg-g @GiuseppeTT @Edwinhr716

akshaychitneni: labels also gets applied to deployment pod template and not matchlabels which should be fine. I believe lws already adds labels?

andreyvelich: Are you sure, I can see that webhook service has only one selector: https://github.com/kubernetes-sigs/lws/blob/main/config/webhook/service.yaml#L18-L19

GiuseppeTT: Andrey is right. The selector from LWS is too broad.

This is well worth testing. @akshaychitneni, can you deploy JobSet and LWS to the same namespace? This will help us:

- Get the real manifests that are applied in production to have something concrete to look at
- Check the interaction between the selectors of JobSet and LWS

GiuseppeTT: Hopefully kustomize injects more labels to the selector for the final manifest on top `control-plane: controller-manager`

GiuseppeTT: Running `make artifacts` in the JobSet repo I get only 

```yaml
apiVersion: v1
kind: Service
metadata:
  labels:
    app.kubernetes.io/component: webhook
    app.kubernetes.io/created-by: jobset
    app.kubernetes.io/instance: webhook-service
    app.kubernetes.io/managed-by: kustomize
    app.kubernetes.io/name: service
    app.kubernetes.io/part-of: jobset
  name: jobset-webhook-service
  namespace: jobset-system
spec:
  ports:
  - port: 443
    protocol: TCP
    targetPort: 9443
  selector:
    control-plane: controller-manager
```

Something similar will probably happen to LWS. So the selector is likely too broad.

akshaychitneni: Ah yes. Helm configures fine but lws kustomize still doesnt add labels as we found with [cache deployment](https://github.com/kubeflow/trainer/tree/master/manifests/third-party/leaderworkerset/patches). Let me create a PR with lws.

akshaychitneni: > Running `make artifacts` in the JobSet repo I get only

This PR should fix this with jobset

kannon92: Do we still want controller-manager as a label here?

I know for Kueue we found this to be not great as multiple services conflicted with each other so we ended up adding a more descriptive name for each controller to make sure services bind to the right deployment.

We found cases where Prometheus was matching on both controller and visibility since they used the same control plane label.

akshaychitneni: I think controller-manager label cant be removed completely from pod template labels as it is being used in deployment matchLabels which is immutable? This might require recreation to migrate?

akshaychitneni: Also created https://github.com/kubernetes-sigs/lws/issues/734 with lws

andreyvelich: Yes, removing the `controller-manager` label might introduce breaking change for existing JobSet users who are using Kustomize.
How do we feel about it @GiuseppeTT @kannon92?

GiuseppeTT: > How do we feel about it @GiuseppeTT @kannon92?

Can we just add more labels? I am not aware of a specific use case that would break, but extending (instead of replacing) the selector would solve the problem (assuming a sister PR on LWS). So I don't see a reason to risk removing `controller-manager`. Like, it's better to let the user fix an already broken setup with Prometheus than trying to fix it in JobSet while risking breaking another user.

GiuseppeTT: > I think controller-manager label cant be removed completely from pod template labels as it is being used in deployment matchLabels which is immutable? This might require recreation to migrate?

AFAIK the selector is immutable. So even if we just add new labels, the user will need recreation

andreyvelich: So you suggest to just use `commonLabels` in Kustomize overlay since it should apply labels to the Deployment selectors as well?

GiuseppeTT: > So you suggest to just use commonLabels in Kustomize overlay since it should apply labels to the Deployment selectors as well?

That is one option.

My point is that we should move from 

```yaml
selector:
    control-plane: controller-manager
```

to

```yaml
selector:
    control-plane: controller-manager
    label-a: value-a
    label-b: value-b
```

instead of 

```yaml
selector:
    label-a: value-a
    label-b: value-b
```

in JobSet and do the same in LWS. This way JobSet and LWS can be set up both in the same namespace.

Keeping `control-plane: controller-manager` is annoying, but I don't see a good reason to remove it and risk breaking people.

Upgrade from JobSet with `control-plane: controller-manager` to JobSet with `control-plane: controller-manager` + other labels will also be annoying for users (e.g., requires recreating the deployment) but I don't see an alternative.

GiuseppeTT: And ideally extend the selector with the labels from helm to decrease the gap between kustomize and helm.

https://github.com/kubernetes-sigs/jobset/issues/1084#issuecomment-3513519805

kannon92: My main point was due to Kueue making more descriptive names for the selectors: https://github.com/kubernetes-sigs/kueue/pull/7371

For Kueue we did see a real problem where multiple services are picked by control-plane:controller-manager and that was breaking the prometheus setup.

GiuseppeTT: I see. I agree that descriptive names are better. I'm just leaning toward extending the old selector instead of replacing it. Both fix #1084 (assuming sister PR in LWS for the extension solution) but extending seems safer.

kannon92: It's probably fine without my suggestion.

I'd love to get confirmation that this works for both JobSet/LWS before merging. Maybe just a ascidemo to confirm would be fine?

akshaychitneni: I see. If we just want to extend selector with additional labels, it would still need recreation. Should we update docs to let users recreate instead of upgrade? I think adding labels to commonLabels would just work for this case.

akshaychitneni: However, if we are thinking of recreation, why not just use the labels same as with helm which already has needed labels?

andreyvelich: I think using the same labels as of Helm is fine.
@akshaychitneni Do you need to use `commonLabels` in that case?
IIUC, it should update deployment selectors too.

akshaychitneni: > I think using the same labels as of Helm is fine.
> 
> @akshaychitneni Do you need to use `commonLabels` in that case?
> 
> IIUC, it should update deployment selectors too.
> 
> 

Yes using commonLabels should work in this case. Will update

akshaychitneni: Updated. Verified with simialr change in lws.

Deploy jobset+lws in same namespace using - 
```
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization
namespace: jobset-lws-system

patches:
  # Delete Namespace resources from both manifests to avoid conflicts
  - patch: |-
      apiVersion: v1
      kind: Namespace
      metadata:
        name: not-used
      $patch: delete
    target:
      kind: Namespace

resources:
  # Local JobSet with new commonLabels
  - ../config/default
  # Local LWS
  - ../../lws/config/default
```

verify labels for both lws and jobset
```
kubectl get services -n jobset-lws-system -o custom-columns=NAME:.metadata.name,SELECTOR:.spec.selector
NAME                                        SELECTOR
jobset-controller-manager-metrics-service   map[app.kubernetes.io/instance:jobset app.kubernetes.io/managed-by:kustomize app.kubernetes.io/name:jobset control-plane:controller-manager]
jobset-webhook-service                      map[app.kubernetes.io/instance:jobset app.kubernetes.io/managed-by:kustomize app.kubernetes.io/name:jobset control-plane:controller-manager]
lws-controller-manager-metrics-service      map[app.kubernetes.io/instance:lws app.kubernetes.io/managed-by:kustomize app.kubernetes.io/name:lws control-plane:controller-manager]
lws-webhook-service                         map[app.kubernetes.io/instance:lws app.kubernetes.io/managed-by:kustomize app.kubernetes.io/name:lws control-plane:controller-manager]
```
endpoints
```
kubectl -n jobset-lws-system get endpoints   
Warning: v1 Endpoints is deprecated in v1.33+; use discovery.k8s.io/v1 EndpointSlice
NAME                                        ENDPOINTS         AGE
jobset-controller-manager-metrics-service   10.244.0.7:8443   4m10s
jobset-webhook-service                      10.244.0.7:9443   4m10s
lws-controller-manager-metrics-service      10.244.0.8:8443   4m10s
lws-webhook-service                         10.244.0.8:9443   4m10s
```

andreyvelich: Great thank you @akshaychitneni!
/lgtm

PR #1139, opened 2026-01-26
separate pvc retain/delete into a separate test case to control timeline better.

### LINKED ISSUE #1133: [Flaking Test]  [It] JobSet controller jobset is created and its jobs go through a series of updates jobset with VolumeClaimPolicies should create two PVC with Delete and Retain policies
https://testgrid.k8s.io/sig-apps#periodic-jobset-test-integration-main

The above test is flaking ever so often.

Failure:

```
{Timed out after 10.000s.
Expected
    <bool>: false
to equal
    <bool>: true failed [FAILED] Timed out after 10.000s.
Expected
    <bool>: false
to equal
    <bool>: true
In [It] at: /home/prow/go/src/sigs.k8s.io/jobset/test/util/util.go:101 @ 01/22/26 03:56:49.381
```

kannon92:
#### What type of PR is this?
/kind flake

Separate integration test with PVC to a separate test case as there is a lot of orchestration needed.

Moving to a separate test case makes it clear what we need to run in order and hopefully deflakes the test.

Fixes #1133 

```release-note
NONE
```

### DIFF
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -1258,206 +1258,6 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
-		ginkgo.Entry("jobset with VolumeClaimPolicies should create two PVC with Delete and Retain policies", &testCase{
-			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
-				return testing.MakeJobSet("volume-test", ns.Name).
-					SuccessPolicy(&jobset.SuccessPolicy{Operator: jobset.OperatorAll}).
-					EnableDNSHostnames(true).
-					VolumeClaimPolicies([]jobset.VolumeClaimPolicy{
-						{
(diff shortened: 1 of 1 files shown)


test/integration/controller/jobset_controller_test.go
```diff
@@ -1258,206 +1258,6 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
-								Namespace: namespace,
-							}, &js2)
-						}, timeout, interval).Should(gomega.Succeed())
-
-						testutil.JobSetCompleted(ctx, k8sClient, &js2, timeout)
-					},
-				},
-			},
-		}),
 		ginkgo.Entry("suspended jobset with VolumeClaimPolicies should create PVCs", &testCase{
```
andreyvelich, line 1261: Do we need to move suspend tests as well?

kannon92: I'll check today but I didn't see any flakes on that job.

kannon92:
/hold

Looks to still be flaky.

PR #1140, opened 2026-01-27
Add sidecar container mode to the in-place restart agent

GiuseppeTT:
#### What type of PR is this?

/kind feature

**tl;dr:** Allow the in-place restart agent to run as a sidecar (sidecar container mode) instead of the main container (main container mode).

Up until now, the in-place restart agent could only be used in the "main container container" mode (see `site/static/examples/in-place-restart/jobset-main-container-mode.yaml`) in which the worker process runs in the container of the agent. In-place restarts are done by restarting only the agent-worker container. The barrier is done by making the agent start the worker process only when the barrier is lifted.

This PR adds the "sidecar" mode (see `site/static/examples/in-place-restart/jobset-sidecar-container-mode.yaml`) in which the agent runs as a sidecar while the worker runs as the main container. In-place restarts are done using the `RestartAllContainers` feature from upstream (see https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#restart-all-containers). This feature was introduced in 1.35 and its feature gate will be enabled by default in 1.36. The barrier is done by making the agent succeed a start up probe when the barrier is lifted, which allows the worker container to start running. The main advantage of the sidecar container mode is that it is easier to set up (add a new sidecar vs figure out how to build an image that contains both the agent and the worker).

API-wise, the main container mode is used when the env variable `WORKER_COMMAND` is used. Otherwise, the agent runs in sidecar container mode. The start up probe has defaults but can be configured by the env vars `STARTUP_PROBE_PATH` and `STARTUP_PROBE_PORT`.

The agent image (for the sidecar container mode) can be built with `IN_PLACE_RESTART_AGENT_IMAGE_REGISTRY=<CHANGE ME> make in-place-restart-agent-image-push` which uses the Dockerfile `cmd/in-place-restart-agent/Dockerfile`. I plan to add the required pipeline to automatically build and push the agent image to the JobSet registry in a follow up PR.

I also made sure to update the examples in `site/static/examples/in-place-restart/`. Documentation will be added in a follow up PR.

Part of #467

None

```release-note
Add the sidecar container mode to the in-place restart agent.
```


GiuseppeTT:
cc @ahg-g @kannon92 @andreyvelich

pkg/controllers/in_place_restart.go
```diff
@@ -134,6 +134,11 @@ func getMaxContainerRestartCount(associatedPods *corev1.PodList) int32 {
 func getPodInPlaceRestartAttempts(associatedPods *corev1.PodList) ([]int32, error) {
 	podInPlaceRestartAttempts := []int32{}
 	for _, pod := range associatedPods.Items {
+		// Skip it if the pod is failed
+		// Failed Pods might persist while their new copy already exists
+		if pod.Status.Phase == corev1.PodFailed {
```
ahg-g, line 139: was this a bug?

cmd/in-place-restart-agent/main.go
```diff
@@ -166,43 +174,75 @@ type env struct {
...
 	}
-	workerCommand := getEnvOrDie(EnvWorkerCommand)
+
+	if os.Getenv(EnvWorkerCommand) != "" && (os.Getenv(EnvStartupProbePath) != "" || os.Getenv(EnvStartupProbePort) != "") {
+		setupLog.Error(nil, "invalid env var configuration: WORKER_COMMAND cannot be set with STARTUP_PROBE_PATH or STARTUP_PROBE_PORT")
+		os.Exit(1)
+	}
+	workerCommand := os.Getenv(EnvWorkerCommand)
+	if workerCommand != "" {
+		setupLog.Info("env var WORKER_COMMAND is set. In-place restart agent will run in big container mode")
```
ahg-g: "big container" is not a known terminology, I suggest to call them `main` and `sidecar` modes

pkg/controllers/in_place_restart.go
GiuseppeTT: Yes. I am not sure how I didn't catch this in earlier tests 😅 .

# Example of how the bug might manifest

Starting case. Workload has only 2 Pods.

- `pod-0` : Running
- `pod-1` : Running

`pod-0` fails.

- `pod-0` : Failed
- `pod-1` : Running

Job controller creates the replacement `pod-0-new` because `backoffLimit > 0`.

- `pod-0` : Failed
- `pod-0-new` : Waiting at the barrier
- `pod-1` : Running

`pod-0` is fully terminated but the object might persist for a long time. The JobSet controller should ignore  `pod-0` and consider only `pod-0-new` and `pod-1` for in-place restart. This change makes sure this is true.

cmd/in-place-restart-agent/main.go
GiuseppeTT: Sounds good. I'll change that.

GiuseppeTT: Renamed "big container mode" to "main container mode" and "sidecar mode" to "sidecar container mode".

Issue #1108, opened 2025-12-20
Mutable JobSet spec with opportunistic updates


GiuseppeTT:
cc @imreddy13 @kannon92 @andreyvelich

# Introduction

Currently, the JobSet spec (specifically the Pod / Job template) is immutable. If a user needs to change a configuration in a running JobSet, they must delete and recreate the object. This interrupts running workloads and causes the loss of progress.

I would love to gather feedback from the community on making JobSets mutable:

- Have you heard similar requests from users where mutation of JobSets is required?
- What are your thoughts on allowing JobSets to become mutable in general?
- What do you think about an "opportunistic" update strategy (applying changes only during natural restarts)?

# Real use case

Here is a real use case that I got from a big user:

1. **Setup:** The user has multiple JobSets running simultaneously and many more scheduled for when capacity becomes available
2. **Trigger:** The user develops a new Pod template (e.g., to fix a bug in the worker container or apply an optimization)
3. **Current limitation:** Currently, the user can only apply this new template to **newly** submitted JobSets. To fix existing ones, they have to recreate them, killing healthy running JobSets, which loses progress
4. **Goal:** The user wants to update the template for the currently running JobSets as well
5. **Constraint:** The user **does not** want the interrupt the running Jobs
6. **Desired behavior:** Instead, the user wants to use natural JobSet restarts as an opportunity to change the Pod template. If a Job fails and the JobSet controller recreates it, it should come back up with the new Pod template

# Potential solution

One thing that I thought is the introduction of an `updatePolicy` for the Pod / Job template, similar to the idea of the existing `failurePolicy`, `startupPolicy`, and `successPolicy`. It could potentially include `updateStrategies` such as:

- `Never`: (Current behavior) The validating webhook blocks updates to the template
- `Opportunistic`: The webhook allows updates to the spec. The JobSet controller applies the updated Job template only when recreating the child Jobs during a restart. It does not force the deletion of running Jobs

kannon92:
We do have a limitation on what we can update on job/pod templates.

https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/5440-mutable-job-pod-resource-updates/README.md

In this KEP I wanted to relax immutability while Jobs are suspended. Running jobs will reject any changes other than parallelism/completions.

The job immutability may limit updates to be delete and recreate.

On suspended Jobset and Job my hope would be that we can change requests/limits.

kannon92:
[...]

PR #1147, opened 2026-02-04
Add KEP for Elastic Jobsets

### LINKED ISSUE #463: Support Elastic JobSets

aniket2405:
#### What type of PR is this?

/kind documentation

Add the KEP for Elastic Jobsets

Fixes #463 

NONE

```release-note
NONE

```

### DIFF
--- a/keps/463-ElasticJobsets/kep.yaml
+++ b/keps/463-ElasticJobsets/kep.yaml
@@ -2,7 +2,7 @@ title: Support Elastic JobSets
 kep-number: 463
 authors:
   - "@aniket2405"
-status: provisional
+status: implementable
 creation-date: 2026-02-04
 reviewers:
   - "@GiuseppeTT"
@@ -20,3 +20,7 @@ replaces:
 
 # The target maturity stage in the current dev cycle for this KEP.
 stage: alpha
+
+# List the feature gate name and the components for which it must be enabled
+feature-gates:
+  - name: ElasticJobsets


aniket2405:
@andreyvelich @ahg-g 
Please review the KEP for Elastic Jobsets.
Thanks!

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,223 @@
...
+
+type ReplicatedJob struct {
+	// Name of the ReplicatedJob.
+	Name string `json:"name"`
+
+	// Replicas is the number of desired Jobs for this ReplicatedJob.
+	// This field is changed from immutable to mutable.
+	Replicas int32 `json:"replicas"`
+
+	// ... (Template and other fields remain immutable)
```
kannon92: This design wouldn't satisfy story 1. You still can't change parallelism/completions on ReplicatedJobTemplates.

kannon92:
/ok-to-test

Please sign the CLA.

keps/463-ElasticJobsets/README.md
aniket2405: RIght. The JobSet webhook currently enforces immutability on the entire template. I've updated the KEP to explicitly relax the validation for `parallelism` and `completions` in the `ReplicatedJob` template, allowing us to propagate those patches to the underlying Jobs.

Please review and let me know if this is acceptable.

aniket2405:
Hi @kannon92 @andreyvelich @tenzen-y Would like to have your feedback here.

kannon92:
Can you fix the verify job?

kannon92:
I'll leave the first round of reviews to @andreyvelich (or @tenzen-y if he has time).
I'm not entirely clear what the Kubeflow community would like here.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This KEP outlines the proposal to support "elasticity" within JobSets by making the `replicas` field of a `ReplicatedJob` mutable. This allows external controllers or users to scale the number of Job replicas in a JobSet up or down dynamically during its execution.
```
andreyvelich: What about Job's `parallelism` and `completions`? Shall we also allow to modify that field?
Those fields are immutable now in JobSet: https://github.com/kubernetes-sigs/jobset/blob/main/pkg/webhooks/jobset_webhook.go#L350C24-L350C46

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Motivation
+
+Currently, most fields in a JobSet are immutable after creation. This poses a challenge for distributed training workloads that are designed to be elastic. For example, PyTorch Elastic (Torchrun) can handle a varying number of nodes. If a JobSet starts with 4 replicas and one node fails or is preempted, the JobSet should ideally scale down to 3 replicas and allow the training to continue, rather than failing the entire workload.
+
+### Goals
+
```
andreyvelich, line 64: One of the goal can be to support Elastic IndexedJob in JobSet: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/3715-elastic-indexed-job

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+- Supporting mutability of the `JobTemplate` or container images mid-run.
+- Automated scaling within the JobSet controller (scaling should be triggered by external actors like HPA or Kueue).
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+As a Kubeflow Training Operator v2 user, I want PyTorch elastic training that scales worker count during failures or when GPUs become available.
```
andreyvelich:
```suggestion
As a Kubeflow Trainer v2 user, I want PyTorch elastic training that scales worker count during failures or when GPUs become available.
```

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+      spec:
+        completionMode: Indexed
+        parallelism: 2      # Can scale: 2→8
+        completions: 2
+        template:
+          spec:
+            containers:
+            - name: trainer
+              image: pytorch/torchserve
+              command: ["torchrun", "--nproc_per_node=1", "train.py"]
```
andreyvelich: Can you add example how `torchrun` allows to set min and max value for nnodes for elastic training: https://docs.pytorch.org/docs/stable/elastic/run.html#elastic-min-1-max-4-tolerates-up-to-3-membership-changes-or-failures

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+              resources:
+                limits:
+                  nvidia.com/gpu: 1
+# kubectl patch jobset pytorch-elastic -p '{"spec":{"replicatedJobs":[{"replicas":1,"template":{"spec":{"parallelism":8,"completions":8}}}]}}'
+```
+
+#### Story 2
+As a Google Cloud TPU user, I want my training to continue when one TPU slice fails by scaling down replicas.
```
andreyvelich: Not sure if JAX support elasticity. IIRC, we had some discussions about it before.
cc @GiuseppeTT @imreddy13

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Motivation
+
+Currently, most fields in a JobSet are immutable after creation. This poses a challenge for distributed training workloads that are designed to be elastic. For example, PyTorch Elastic (Torchrun) can handle a varying number of nodes. If a JobSet starts with 4 replicas and one node fails or is preempted, the JobSet should ideally scale down to 3 replicas and allow the training to continue, rather than failing the entire workload.
+
+### Goals
+
+- Support scaling the number of replicas in a `ReplicatedJob` after the JobSet has been created.
+- Ensure the JobSet controller correctly manages the lifecycle (creation/deletion) of child Jobs based on the updated replica count.
+- Maintain the stability of existing JobSet features like `StartupPolicy` and `ExclusivePlacement` during scaling events.
```
andreyvelich:
```suggestion
- Maintain the stability of existing JobSet features like `DependsOn` and `ExclusivePlacement` during scaling events.
```

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+As a cluster admin, I want JobSets to integrate with Kueue elastic quotas for dynamic GPU scaling.
+
+Kueue cannot scale JobSets today. With elastic support, Kueue scales based on cluster utilization:
+
+```
+Cluster 100 GPUs @ 80% → Kueue scales: replicas 20→25
+Cluster 100 GPUs @ 95% → Kueue scales: replicas 25→20
+```
+
+### Risks and Mitigations
```
andreyvelich, line 120: Would be good to review elastic Index Job risks as well: https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/3715-elastic-indexed-job#risks-and-mitigations
cc @ahg-g

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+## Design Details
+
+### API Details
+
+The validation webhook for `jobset.x-k8s.io/v1alpha2` will be modified to allow updates to:
+
+1. spec.replicatedJobs[i].replicas (Horizontal Job scaling)
+2. spec.replicatedJobs[i].template.spec.parallelism (Internal Pod scaling)
+3. spec.replicatedJobs[i].template.spec.completions (Internal Pod scaling)
```
andreyvelich: You need to update goal + overview to say that we also allow to modify parallelism/completions.

keps/463-ElasticJobsets/README.md
andreyvelich, line 120: Any risks with FailurePolicies or SuccessPolicies API?

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+#### Index Sorting:
+The controller fetches all child Jobs owned by the JobSet and sorts them numerically by their index suffix (e.g., workers-0, workers-1).
+
+#### Horizontal Scaling (Job-level):
+
+Scale Up: If spec.replicas > actual Job count, the controller creates new Jobs for the missing indices sequentially (e.g., if scaling from 2 to 4, it creates indices 2 and 3).
+
+Scale Down (LIFO): If actual Job count > spec.replicas, the controller deletes Jobs starting from the highest index downward. This preserves lower-indexed Jobs, which often host "leader" or "master" processes.
+
+#### Internal Scaling (Pod-level):
```
andreyvelich: Shall we just name this:
- Horizontal Job Scaling (ReplicatedJob level)
- Horizontal Pod Scaling (Pod-Level)

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+Template Propagation: When the controller detects an update to spec.replicatedJobs[i].template.spec.parallelism or completions, it will iterate through all existing healthy child Jobs.
+
+Opportunistic Patching: The controller will issue a Patch request to each existing child Job to synchronize its spec.parallelism and spec.completions with the new values in the JobSet template.
```
andreyvelich: How would that work in practice? Is there a risk that some jobs will be patched before others?

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+Elastic Integration: By patching the child Jobs, the JobSet controller leverages the native Kubernetes Elastic Indexed Job functionality to scale the pod count within each job replica dynamically.
+
+#### Failure Handling Priority
+To prevent "scaling into a fire," the controller will prioritize stabili  ty over elasticity:
```
andreyvelich:
```suggestion
To prevent "scaling into a fire," the controller will prioritize stability over elasticity:
```

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+#### Graceful Termination:
+During any scale-down event (either removing a Job replica or reducing internal parallelism), the controller relies on the terminationGracePeriodSeconds of the Jobs and Pods. This allows applications like PyTorch Elastic to perform final checkpoints or clean shutdowns.
+
+### Status Management
+
+The JobSet.Status.ReplicatedJobStatus must reflect the scaling operation.
+
+Ready/Active Counts: During a scale-down, the active count must be decremented as soon as the Job deletion is initiated.
+
+Conditions: A new condition Scaling may be added to the JobSet status to indicate a scale-up or scale-down is in progress.
```
andreyvelich, line 187: Do we know if elastic Job has any condition for scaling up/down?

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+Template Propagation: When the controller detects an update to spec.replicatedJobs[i].template.spec.parallelism or completions, it will iterate through all existing healthy child Jobs.
+
+Opportunistic Patching: The controller will issue a Patch request to each existing child Job to synchronize its spec.parallelism and spec.completions with the new values in the JobSet template.
+
+Elastic Integration: By patching the child Jobs, the JobSet controller leverages the native Kubernetes Elastic Indexed Job functionality to scale the pod count within each job replica dynamically.
+
+#### Failure Handling Priority
+To prevent "scaling into a fire," the controller will prioritize stabili  ty over elasticity:
+
+If any Job in the JobSet is in a Failed state, the controller will pause both horizontal and internal scaling operations until the FailurePolicy has reconciled the failed resource (either by restarting the set or hitting the backoffLimit).
```
andreyvelich: Does it mean that scaling up and down will be executed after FailurePolicy rule is applied?
What about SuccessPolicy?

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+
+### Status Management
+
+The JobSet.Status.ReplicatedJobStatus must reflect the scaling operation.
+
+Ready/Active Counts: During a scale-down, the active count must be decremented as soon as the Job deletion is initiated.
+
+Conditions: A new condition Scaling may be added to the JobSet status to indicate a scale-up or scale-down is in progress.
+
+### Defaulting/Validation
```
andreyvelich, line 189: Might need to also validate that values stay `>= 1`

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,245 @@
...
+jobset_controller_test.go: Verify the logic that calculates which Job indices to delete during a scale-down.
+
+#### Integration Tests
+Verify that a JobSet correctly scales up from 1 to 3 replicas.
+
+Verify that a JobSet correctly scales down from 3 to 1 replica, ensuring the Jobs with index 1 and 2 are the ones deleted.
+
+Verify that updating parallelism on the JobSet propagates to existing child Jobs.
+
+Test scaling in a JobSet that uses ExclusivePlacement to ensure pods are still placed correctly on new nodes.
```
andreyvelich: We might need to add tests for FailurePolicies as well.

andreyvelich (commented):
Thank you for driving this @aniket2405!
I left a few comments, overall looks great!
/cc @tenzen-y @astefanutti

keps/463-ElasticJobsets/README.md
andreyvelich: Found it later in the doc, so we just need to update Summary.

aniket2405:
Thanks for the review @andreyvelich 
I'll try to address them by tomorrow.

keps/463-ElasticJobsets/README.md
kannon92: No. There is no condition on scaling.

keps/463-ElasticJobsets/README.md
aniket2405: Right. Added that

keps/463-ElasticJobsets/README.md
aniket2405: Done

keps/463-ElasticJobsets/README.md
aniket2405: Done

keps/463-ElasticJobsets/README.md
aniket2405: Ack.

keps/463-ElasticJobsets/README.md
aniket2405: My understanding is that while pure JAX historically didn't support dynamic scaling out of the box, modern orchestration layers (like Pathways) have introduced mechanisms to handle fault tolerance and dynamic resizing for JAX workloads when a TPU slice fails. That being said, I'll wait for @GiuseppeTT or @imreddy13 to confirm the current state of JAX elasticity.

keps/463-ElasticJobsets/README.md
aniket2405: Done

keps/463-ElasticJobsets/README.md
aniket2405: Added some risks based on elastic Index Job risks

keps/463-ElasticJobsets/README.md
aniket2405: Done

keps/463-ElasticJobsets/README.md
aniket2405: You're right, there are a few risks with FailurePolicies or SuccessPolicies API. Have added them in `JobSet Policy & State Management` section of `Risks and Mitigation`

keps/463-ElasticJobsets/README.md
aniket2405: Done.

keps/463-ElasticJobsets/README.md
aniket2405: Yes, the Jobs will be patched sequentially, but I don't see that as a risk. You'll be able to help out better here, but what I believe is:

In practice, this can be handled through eventual consistency at the controller level and rendezvous barriers at the application level:

Controller Level: The JobSet controller issues PATCH requests sequentially. If the controller crashes or is interrupted mid-loop (e.g., patching 2 out of 4 Jobs), it relies on its standard reconciliation to catch the remaining unpatched Jobs on the next sync.

Application Level: This asynchronous scaling is safe because frameworks designed for elasticity expect it. For example, in PyTorch Elastic, when the first Jobs scale up and their new pods start, they don't immediately resume training. They enter a rendezvous phase and wait at the barrier until the remaining nodes from the other Jobs scale up and join the network, satisfying the nnodes requirement.

I have updated the Horizontal Pod Scaling (Pod-level) section in the KEP to explicitly document this eventual consistency behaviour.

keps/463-ElasticJobsets/README.md
aniket2405: Thanks for confirming, @kannon92 

Does elastic Job rely on eventual consistency of the status fields instead?

If so, we can probably remove the condition for scaling and use the former instead.

keps/463-ElasticJobsets/README.md
aniket2405: Yes. Success and Failure policy evaluations take priority over applying scaling mutations.

For `FailurePolicy`: If a Job fails, the controller evaluates the FailurePolicy first. If the policy triggers a restart, the JobSet executes the restart. Once the JobSet is back in a healthy/recovering state, any pending scale operations are applied. If the FailurePolicy dictates failing the entire JobSet, the JobSet enters a terminal Failed state, and the scale operation is ignored.

For `SuccessPolicy`: If the JobSet meets its SuccessPolicy criteria (e.g., a specific target Job finishes), it transitions to a terminal Completed state. Once terminal, all pending or future scaling operations are halted.

keps/463-ElasticJobsets/README.md
aniket2405: Right. Good catch.

keps/463-ElasticJobsets/README.md
aniket2405: Agreed. Added tests for them.

aniket2405:
@andreyvelich 
Thanks for the comprehensive review. 
Addressed most of them (left comments unresolved for comments that merit further discussion).
Please have a look.

GiuseppeTT:
First of all, thank you for this. I like this feature and I think it makes sense.

Secondly, I want to double check. Are there clear real-world use cases for this feature? I ask this because I believe any feature added to JobSet should have a clear real-world use case behind it and any design should solve those specific requirements while keeping options open for the future. This is the reason this feature wasn't done before.

For instance:

- Use case 1: I assume this is true based on [this comment](https://github.com/kubernetes-sigs/jobset/issues/463#issuecomment-3761359034). @andreyvelich, can you confirm? If so, can you also summarize what the requirements are for JobSet?
- Use case 2: I don't think this is the case or at least I haven't heard from a user who is currently blocked by this. From my experience, users are either not worried about this or they want to handle failures by recreating Jobs individually (see my [KEP here](https://github.com/kubernetes-sigs/jobset/pull/1152)) while the other Jobs keep running. I am double checking this is the case
- Use case 3: I don't have context here. I can see the value of making JobSet and Kueue support changes in the size of workloads together, but this is tricky and requires changes to Kueue (see [this thread](https://github.com/kubernetes-sigs/jobset/issues/1108)). @mimowo, can you comment on this?

keps/463-ElasticJobsets/README.md
andreyvelich: Yeah, that makes sense, it looks like additional workers will wait for barrier to join the rendezvous: https://docs.pytorch.org/docs/stable/elastic/rendezvous.html
We should verify how it works in practice, if the JobSet controller fails during scale up, and only subset of Pods will join the training group.

andreyvelich:
>Use case 1: I assume this is true based on https://github.com/kubernetes-sigs/jobset/issues/463#issuecomment-3761359034. @andreyvelich, can you confirm? If so, can you also summarize what the requirements are for JobSet?

Yes, we have the real-world use-case to implement elastic PyTorch training with JobSet. Similarly to how we support elastic PyTorchJob in Training Operator v1: https://github.com/kubeflow/trainer/pull/1453
As we discussed in the thread, `torchrun` allows to dynamically add more nodes to the running training cluster.
Additionally, frameworks like Horovod also allows elasticity with Rendezvous server, which we can integrate into TrainJob since we support node discovery via MPI's host files: https://horovod.readthedocs.io/en/latest/elastic_include.html

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+          spec:
+            nodeSelector:
+              cloud.google.com/gke-tpu-topology: 4x4
+# TPU slice 2 fails → scale: replicas: 4→3 (controller deletes workers-3)
+```
+
+#### Story 3
+As a cluster admin, I want JobSets to integrate with Kueue elastic quotas for dynamic GPU scaling.
+
+Kueue cannot scale JobSets today. With elastic support, Kueue scales based on cluster utilization:
```
mimowo: The support for Kueue sounds great, I opened an issue in Kueue: https://github.com/kubernetes-sigs/kueue/issues/9411 to make sure we move the effort on both sides.

Does that entail autoscaling decisions by Kueue about scale up/down? This is currently not supported in Kueue.

What I currently consider the scope for the integration is to react to the size changes in JobSet based on user changes or 3rd party controller changes (eg. in case of the built-in RayCluster autoscaler). This is what we typically understand by elastic Jobs support in Kueue.

aniket2405: @mimowo

Thanks for opening the Kueue issue, and for catching that nuance. I didn't mean to imply that Kueue would make the active autoscaling decisions. As you mentioned, the scope here is purely that JobSet allows the API mutability, an external autoscaler (or user) triggers the size change, and Kueue can react to that change without restarting the JobSet.

I will update Story 3 in the KEP to ensure it accurately reflects Kueue's role in reacting to quota changes rather than acting as the autoscaler itself.

aniket2405:
@andreyvelich @mimowo @GiuseppeTT 
Updated the KEP to sync with the latest discussion.
Do we have any other considerations here?

kannon92:
can you fix the verify job?

keps/463-ElasticJobsets/README.md
GiuseppeTT: I would just remove this use case until we get a request for it from a user.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+### Non-Goals
+
+- Supporting mutability of the `JobTemplate` or container images mid-run.
+- Automated scaling within the JobSet controller (scaling should be triggered by external actors like HPA or Kueue).
+
+## Proposal
+
+### User Stories
+
+#### Story 1
```
GiuseppeTT, line 79: @andreyvelich does this use case also require mutability of `jobSet.spec.replicatedJobs[].replicas`?

In general, what are the requirements for the JobSet API? Is it allowing mutability of `jobSet.spec.replicatedJobs[].spec.parallelism` and `jobSet.spec.replicatedJobs[].spec.completions` and making the JobSet controller update the corresponding child Jobs?

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+        parallelism: 16    # 16 TPU v4 cores per slice
+        completions: 16
+        template:
+          spec:
+            nodeSelector:
+              cloud.google.com/gke-tpu-topology: 4x4
+# TPU slice 2 fails → scale: replicas: 4→3 (controller deletes workers-3)
+```
+
+#### Story 3
```
GiuseppeTT: My understanding of this use case is that Kueue needs to be able to react to changes to `jobSet.spec.replicatedJobs[].replicas`. But is there a use case to make `jobSet.spec.replicatedJobs[].replicas` mutable? Perhaps from PyTorch elastic training + Kubeflow trainer?

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+```
+Cluster 100 GPUs @ 80% → Kueue scales: replicas 20→25
+Cluster 100 GPUs @ 95% → Kueue scales: replicas 25→20
+```
+
+### Risks and Mitigations
+
+#### Resource & Index Management
+- **Resource Leakage:** Deleting a replica might leave orphaned pods if the child Job is not cleaned up properly.
+  - *Mitigation:* The controller will rely on OwnerReferences and the standard Job deletion lifecycle to ensure all pods are terminated.
```
GiuseppeTT: Make it foreground deletion. https://github.com/kubernetes-sigs/jobset/blob/8b7deca870a970ed0e6d06bad7b8ab7b38fca348/pkg/controllers/jobset_controller.go#L671

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+- **Race Conditions on Job Completion:** Changing completion semantics dynamically could conflict with the Job controller declaring a child Job finished.
+  - *Mitigation:* This is handled upstream by the native Job controller, which uses `Update` when setting the finished status (triggering object modified errors on concurrent spec changes). Additionally, `spec.completions` is strictly immutable once a Job reaches a terminal condition, preventing the JobSet controller from improperly modifying finished Jobs.
+
+#### JobSet Policy & State Management
+- **Scale-down Deletion Triggering FailurePolicy:** When horizontal scaling reduces `replicas`, the JobSet controller must delete the highest-indexed child Jobs. The reconciliation loop might misinterpret this deletion as an unexpected Job failure.
+  - *Mitigation:* The controller logic will explicitly differentiate between "Jobs deleted due to a scale-down event" and "Jobs that failed." Controlled deletions triggered by a `spec.replicas` update will bypass `FailurePolicy` evaluation.
```
GiuseppeTT: I like that this is said explicitly in the KEP so please keep it, but the implementation might not need to worry about this.

tl;dr: deleting healthy Jobs typically do not fail them. Currently, the JobSet controller only restarts when a Job fails, if it is "healthy deleted", the JobSet controller just recreates it.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+type ReplicatedJob struct {
+	// Name of the ReplicatedJob.
+	Name string `json:"name"`
+
+	// Replicas is the number of desired Jobs for this ReplicatedJob.
+	// This field is changed from immutable to mutable.
+	Replicas int32 `json:"replicas"`
+
+	// Template is now partially mutable.
+  // Specifically, parallelism and completions can be updated
```
GiuseppeTT, line 152: nit: inconsistent indentation. I also get this all the time when mixing spaces and tabs 😅

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+Ready/Active Counts: During a scale-down, the active count must be decremented as soon as the Job deletion is initiated.
+
+Conditions: A new condition Scaling may be added to the JobSet status to indicate a scale-up or scale-down is in progress.
+
+### Defaulting/Validation
+
+- **Mutable Fields:** `replicas`, `template.spec.parallelism`, and `template.spec.completions`.
+- **Minimum Value Constraints:** All three mutable scaling fields (`replicas`, `parallelism`, and `completions`) must be strictly `>= 1`. Setting `parallelism` to `0` natively suspends a Job, which conflicts with JobSet's explicit top-level `suspend` field. Elasticity cannot be used as a suspension mechanism.
+- **Elastic Indexed Job Constraints:** Horizontal pod-level scaling inherits the strict validation rules of upstream Elastic Indexed Jobs:
```
GiuseppeTT, line 193: Keep in mind that JobSet currently doesn't fully validate the Job template. So it is currently possible to create a JobSet with an invalid JobTemplate and get it stuck because it can't create Jobs.

The implementation should validate the changes manually.

keps/463-ElasticJobsets/README.md
kannon92: I'm not sure it really even looks at status fields.

The controller mainly allows one to change these values.

And then it reacts as it would normally if a pod failed. ie it checks if the expected pods matches desired and if it doesn't it creates them.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+### Drawbacks
+Application Awareness: This feature is only useful if the application running inside the JobSet can handle its membership changing dynamically.
+
+Controller Overhead: Frequent scaling updates could increase the load on the JobSet controller and the Kubernetes API server.
+
+### Alternatives
+HorizontalPodAutoscaler (HPA) on Jobs: HPA can scale the parallelism of a standard Job, but it cannot manage the multi-job coordination that JobSet provides (e.g., DNS, multi-interface networking).
+
+Manual Re-creation: Deleting the JobSet and creating a new one with the desired size. This causes a full restart of all workers, which is inefficient for large-scale training.
```
GiuseppeTT: Suggestion: Add a "Future Work" subsection with possible improvements. For instance, as the use of elastic JobSet expands it will probably make sense to add a `updatePolicy` API to handle behavior in different cases.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,273 @@
...
+
+	// Template is now partially mutable.
+  // Specifically, parallelism and completions can be updated
+  // to support Elastic Indexed Jobs.
+  Template batchv1.JobTemplateSpec `json:"template"`
+}
+
+```
+
+### Implementation
```
GiuseppeTT, line 159: I think it makes sense to use a feature gate.

WDYT? @kannon92 @andreyvelich

aniket2405:
> can you fix the verify job?

@kannon92 Could you help with why the verify test fails for a PR intending to add a KEP alone? And how can we fix that? Apologies if it has already been mentioned somewhere. It will be great if you can link a resource for this. Thanks!

GiuseppeTT:
> can you fix the verify job?

@aniket2405 for context. One test is failing because you need to update the table of content https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_jobset/1147/pull-jobset-verify-main/2025876741414719488

The other is failing because of a flaky test 😅 

/retest

kannon92:
https://github.com/kubernetes-sigs/jobset/pull/622

Was my first attempt at this.

aniket2405:
Thanks, @GiuseppeTT and @kannon92 for the help. Will fix this.
Also, lots of great observations mentioned here, will address them at the earliest.

keps/463-ElasticJobsets/README.md
andreyvelich: Makes sense, we can have a feature flag for some time.

keps/463-ElasticJobsets/README.md
andreyvelich: Ideally, we want to allow both for the long term. Replicas are useful for large-scale training jobs requiring over 700-1000 GPUs, since it will reduce reconciliation load on a single Job controller.
However, we didn't get a chance to integrate it with TrainJob yet: https://github.com/kubeflow/trainer/issues/2318 (cc @tenzen-y)

If we want, we can relax immutability only for `parallelism` and `completions` in the first iteration.
WDYT @aniket2405 @kannon92 @GiuseppeTT ?

kannon92: I'm fine to support Elastic Indexed Job use case first (scaling parallelism/completions on Jobs).

I like the idea of solving specific usecases first.

keps/463-ElasticJobsets/README.md
GiuseppeTT: Sounds good. 

A good plan is:

- Start as alpha
- Promote to beta once it's been integrated to Kubeflow / Kueue for one release. Also enable by default
- Promote to GA once it's been beta one release and nothing broke

keps/463-ElasticJobsets/README.md
GiuseppeTT: Let's start with `parallelism` and `completions`. We can add `replicas` once it's available in TrainJob and the use case / api is clear : )

keps/463-ElasticJobsets/README.md
kannon92: @GiuseppeTT 
Are you thinking of a feature gate  to iterate on alpha and add phases?

Or to control the behavior for users who don't want this behavior?

I'm fine with a feature gate but just want to clarify your ask.

keps/463-ElasticJobsets/README.md
GiuseppeTT: That being said. I think we should add `replicas` mutability to a "Future work" section

andreyvelich: Sounds good, @aniket2405 when you get a chance please update the proposal to start with elastic indexed job first.

keps/463-ElasticJobsets/README.md
GiuseppeTT: > Are you thinking of a feature gate to iterate on alpha and add phases?

This one.

I think we should do more feature gates. Despite JobSet being alpha, it is treated as stable. The graduation cycles should be fast though. Like, 1 release with clear usage + no complains before graduating to beta + enabled by default.

kannon92: Yea I think feature gates are a good process to follow.

keps/463-ElasticJobsets/README.md
aniket2405: Tackling specific use cases first definitely makes more sense, let's go with that! 
However, updating this will significantly change many aspects of the KEP, since the complexity will now reduce. Do we want me to push the trimmed down version directly to this PR for another round of review, or would you prefer I open a fresh PR since the scope is narrowing so much?

cc: @GiuseppeTT @andreyvelich @kannon92

andreyvelich: >However, updating this will significantly change many aspects of the KEP, since the complexity will now reduce.

Can we just add this design ideas to the Future Work for now, so we keep them in the proposal?
E.g. Horizontal Job Scaling (ReplicatedJob level)

Once we have initial implementation for Horizontal Pod Scaling, we can update this KEP to introduce `replicas` mutability.

keps/463-ElasticJobsets/README.md
aniket2405: Yes, PyTorch Elastic + Kubeflow Training Operator is a use case, I believe.
If replicas is immutable, handling a failed TPU slice requires deleting and recreating the entire JobSet. That forces a complete job restart. Making the field mutable allows the controller to scale the JobSet down without dropping the healthy pods.

keps/463-ElasticJobsets/README.md
aniket2405: Good call. I've updated the mitigation text to mention foreground deletion explicitly.

keps/463-ElasticJobsets/README.md
aniket2405: Acknowledged. I made this clearer in the mitigation.

keps/463-ElasticJobsets/README.md
aniket2405: Ah, thanks!
Opening this in the code editor does not display the inconsistency unless we explicitly try to change it from spaces to tabs.

keps/463-ElasticJobsets/README.md
aniket2405: I see. Great point.
I've updated the KEP to explicitly state that we will implement synchronous validation in the JobSet Validating Admission Webhook to manually verify these mutable fields.

keps/463-ElasticJobsets/README.md
aniket2405: I've added a 'Future Work' section to capture this.

keps/463-ElasticJobsets/README.md
aniket2405: Sounds good.
I will add the feature gate criteria to the KEP.

keps/463-ElasticJobsets/README.md
aniket2405: Makes sense. I've done that

aniket2405:
Thanks @GiuseppeTT, @andreyvelich and @kannon92 for the great discussion and guidance!

I have updated the KEP after addressing the comments and added a future scope for mutable `replicas` field.

Let me know what you think of this.

keps/463-ElasticJobsets/README.md
GiuseppeTT: > > However, updating this will significantly change many aspects of the KEP, since the complexity will now reduce.
> 
> Can we just add this design ideas to the Future Work for now, so we keep them in the proposal? E.g. Horizontal Job Scaling (ReplicatedJob level)
> 
> Once we have initial implementation for Horizontal Pod Scaling, we can update this KEP to introduce `replicas` mutability.

+1

The KEP (design) and implementation should take into account that `replicas` mutability will likely be added later.

keps/463-ElasticJobsets/kep.yaml
```diff
@@ -0,0 +1,26 @@
+title: Support Elastic JobSets
+kep-number: 463
+authors:
+  - "@aniket2405"
+status: provisional
+creation-date: 2024-03-21
```
GiuseppeTT: Update this and other fields. See the other KEPs in the JobSet repo for reference.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,284 @@
...
+Manual Re-creation: Deleting the JobSet and creating a new one with the desired size. This causes a full restart of all workers, which is inefficient for large-scale training.
+
+## Future Work
+
+While this initial implementation focuses strictly on Pod-level scaling (`parallelism` and `completions`), the ultimate goal is to support full elasticity, including Job-level scaling. The following items capture the design ideas, user stories, and identified risks for these future iterations.
+
+### 1. Horizontal Job Scaling (`replicas` mutability)
+Future iterations will introduce mutability to `jobSet.spec.replicatedJobs[].replicas`. This will allow the JobSet controller to natively handle scaling down (e.g., in response to a hardware failure) and scaling up (when new nodes become available) without requiring a full teardown.
+
+#### Motivating User Story: TPU Slice Failures
```
GiuseppeTT: Please remove this until we get a request for it from a user.

keps/463-ElasticJobsets/kep.yaml
aniket2405: Done

keps/463-ElasticJobsets/README.md
aniket2405: Done

aniket2405:
@GiuseppeTT Updated the KEP.
Let me know if we need to add any other field in the YAML file.

cc: @andreyvelich @kannon92

GiuseppeTT:
> @GiuseppeTT Updated the KEP.

LGTM.

I'll let the official LGTM to @andreyvelich since he owns the use case.

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,258 @@
...
+}
+
+```
+
+### Implementation
+
+The JobSet controller will mplement reconciliation logic to support internal elasticity. This ensures the JobSet state remains synchronized with the template.spec parameters.
+
+### Graduation Criteria
+To ensure stability for existing workloads, this feature will be introduced incrementally using a standard feature gate.
```
andreyvelich: Let's add the feature gate name here (e.g. `ElasticJobSet`)

keps/463-ElasticJobsets/README.md
```diff
@@ -0,0 +1,258 @@
...
+  // Specifically, parallelism and completions can be updated
+  // to support Elastic Indexed Jobs.
+  Template batchv1.JobTemplateSpec `json:"template"`
+}
+
+```
+
+### Implementation
+
+The JobSet controller will mplement reconciliation logic to support internal elasticity. This ensures the JobSet state remains synchronized with the template.spec parameters.
```
andreyvelich:
```suggestion
The JobSet controller will implement reconciliation logic to support internal elasticity. This ensures the JobSet state remains synchronized with the template.spec parameters.
```

andreyvelich (commented):
Thanks for the update @aniket2405!
Looks good to me, just a few nits.

keps/463-ElasticJobsets/README.md
aniket2405: Oops, missed this.

keps/463-ElasticJobsets/README.md
aniket2405: Done

aniket2405:
Updated the CR after addressing the comments.
cc: @andreyvelich

aniket2405:
Sure, @andreyvelich!
It has been exciting, and hope we're able to close on this soon.

To implement this, are we going to create a new `jobset` issue or will we be tracking the implementation through the same issue?

andreyvelich:
>To implement this, are we going to create a new jobset issue or will we be tracking the implementation through the same issue?

We can track the progress in: https://github.com/kubernetes-sigs/jobset/issues/463

keps/463-ElasticJobsets/kep.yaml
```diff
@@ -0,0 +1,22 @@
+title: Support Elastic JobSets
+kep-number: 463
+authors:
+  - "@aniket2405"
+status: provisional
```
kannon92:
```suggestion
status: implementable
```

keps/463-ElasticJobsets/kep.yaml
```diff
@@ -0,0 +1,22 @@
...
+  - "@GiuseppeTT"
+  - "@andreyvelich"
+  - "@kannon92"
+
+see-also:
+  - "NA"
+replaces:
+  - "NA"
+
+# The target maturity stage in the current dev cycle for this KEP.
```
kannon92, line 21: I think you need a section on the feature gate here.

See https://github.com/kubernetes-sigs/jobset/blob/main/keps/467-InPlaceRestart/kep.yaml#L29

keps/463-ElasticJobsets/kep.yaml
aniket2405: Done

keps/463-ElasticJobsets/kep.yaml
aniket2405: Done.
Do we also need a field for milestone, as in: https://github.com/kubernetes-sigs/jobset/blob/963f92a0e0233f9b35ef838cc64246bc24ab78fa/keps/467-InPlaceRestart/kep.yaml#L26

aniket2405:
cc: @kannon92

kannon92:
One thing that may be good to explore after you implement this @aniket2405 is kubeflow/katib#1150 

It should not be done in the same KEP as the feature gate dependencies are not the same but just want to see if you are interested.

aniket2405:
kubeflow/katib#1150 sounds interesting, @kannon92 
I'd be happy to pick it up after this implementation.
However, I have one other ongoing CR as well - https://github.com/kubeflow/trainer/issues/3562
If the timeline for kubeflow/katib#1150 is not strict, it should be interesting to work on!

andreyvelich:
Also, kubeflow/katib#1150 might be required for TrainJob to support intelligent GPU assignment as we discussed with @VassilisVassiliadis here: https://github.com/kubeflow/trainer/issues/2599#issuecomment-3908875212
Since we want to update TrainJob's container resource by our classifier controller, when the TrainJob is suspended.

So once we have elastic JobSet ready, it would be great to investigate it!

cc @robert-bell @astefanutti @kaisoz

Issue #1129, opened 2026-01-21
Add single Job recreation to JobSet restarts


GiuseppeTT:
# Introduction

Currently, JobSet allows workloads to recover from failures by performing a JobSet restart in which all child Jobs are recreated. The recreate-all strategy is a powerful method that supports many types of workloads. But, there are now use cases that could be more optimally recovered from failures by recreating only the failed child Job.

A solution to this problem was designed in https://github.com/kubernetes-sigs/jobset/pull/925 and implemented in https://github.com/kubernetes-sigs/jobset/pull/920 and https://github.com/kubernetes-sigs/jobset/pull/958. But, the feature was later reverted in https://github.com/kubernetes-sigs/jobset/pull/982 before being released. This was due to new requirements that required breaking changes.

The objective of this issues is to get consensus on the use cases, requirements and the API for this feature request. Once we have that, I will update the KEP of failure policy.

# User cases

## Case 1: Recreate only the failed Job

As a user, I have a JobSet workload with one replicateJob that contains multiple replicas. When a failure happens, my workload needs only the failed Job to be recreated. Currently, JobSet only supports recreating everything, which increases mean time to recovery and the load on the control plane. Therefore, I need a way to configure my JobSet to recreate only the failed child Job instead of all of them.

```yaml
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata:
  name: js
spec:
  replicatedJobs:
  # Recreate only the failed child Job
  - name: rj
    replicas: ...
    template:
      spec:
        parallelism: ...
        completions: ...
        backoffLimit: 0 # Job fails if any Pod fails
        template:
         ...
```

## Case 2: Different replicatedJobs with different needs

As a user, I have a JobSet workload with two replicatedJobs: one for controlling and another for working. When a failure affects the control Jobs, I need to recreate all Jobs. Whereas, when a failure affects the worker Jobs, I need to recreate only the failed worker Job. Therefore, I need a way to configure my JobSet to recreate either all Jobs or only the failed Job depending on what happened.

On top of that, failures in control Jobs are rare while failures in worker Jobs are frequent. Therefore, I need a way to configure my JobSet to limit global restarts to a small value (e.g., 3) and single Job restarts to a large value (e.g., 100).

```yaml
apiVersion: jobset.x-k8s.io/v1alpha2
kind: JobSet
metadata:
  name: js
spec:
  replicatedJobs:
[...]

PR #1152, opened 2026-02-13
KEP: Add restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts`

GiuseppeTT:
#### What type of PR is this?

/kind documentation

Update the KEP of the failure policy API.

Add restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts`. These actions allow the JobSet to be restarted by recreating only the failed Job instead of all Jobs. This is very useful for workloads that can recover by restarting only the failed Job.

Part of #1129

API Changes are in the KEP.

Current restart logic for `RestartJobSet` case:

- Reconciliation loop 1
    - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
    - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (let's say `restartJobSetActionApplier` from `RestartJobSet`)
    - In `restartJobSetActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts `). Otherwise, it bumps `js.Status.Restarts += 1` and `js.Status.RestartsCountTowardsMax += 1`
- Reconciliation loop 2
    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts`) and group them to `ownedJobs.previous`
    - Later in the `reconcile` function, the Jobs in `ownedJobs.previous` are deleted
- Reconciliation loop 3...N
    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs`

Future restart logic for `RestartJobSet` case:

- Reconciliation loop 1
    - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
    - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (let's say `restartJobSetActionApplier` from `RestartJobSet`)
    - In `restartJobSetActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts ` or `js.Status.JobStatus[jobName].RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.Restarts += 1` and `js.Status.RestartsCountTowardsMax += 1` and `js.Status.JobStatus[jobName].Restarts += 1` and `js.Status.JobStatus[jobName].RestartsCountTowardsMax += 1`
- Reconciliation loop 2
    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts` or `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.JobStatus[jobName].Restarts`) and group them to `ownedJobs.previous`
    - Later in the `reconcile` function, the Jobs in `ownedJobs.previous` are deleted
- Reconciliation loop 3...N
    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs`

Future restart logic for `RestartJob` case:

- Reconciliation loop 1
    - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
    - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (let's say `restartJobActionApplier` from `RestartJob`)
    - In `restartJobActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts ` or `js.Status.JobStatus[jobName].RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.JobStatus[jobName].Restarts += 1` and `js.Status.JobStatus[jobName].RestartsCountTowardsMax += 1`
- Reconciliation loop 2
    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts` or `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.JobStatus[jobName].Restarts`) and group them to `ownedJobs.previous`
    - Later in the `reconcile` function, the Jobs in `ownedJobs.previous` are deleted
- Reconciliation loop 3...N
    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs`

Observability:

- `jobSet.status.restarts`, `jobSet.status.restartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/restart-attempt']` track only recreate-all-jobs restarts
- `jobSet.status.jobsStatus[].restarts`, `jobSet.status.jobsStatus[].restartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` track recreate-single-job and recreate-all-jobs for a given job. If the user wants to know only the number of recreate-single-job, they can subtract the fields such as `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` - `job.labels['jobset.sigs.k8s.io/restart-attempt']`

Future changes:

- Once we agree on how to implement multiple restart limits for the same JobSet, I believe that this KEP should be easy to adapt. This is because we are tracking the number of restarts for each Job. So, it should be easy to compare these values to the JobSet spec in any way to decide if the JobSet should be failed

Relation to in-place restart:

- When used with in-place restart, these changes allow the failed Job to be recreated while all the other Pods are restarted in-place. This is a desired behavior. For instance, when using `alpha.jobset.sigs.k8s.io/exclusive-topology: rack`, we want to recreate all Pods from a failed rack, while restarting the Pods from other racks in-place.

```release-note
NONE
```

### DIFF
--- a/keps/262-ConfigurableFailurePolicy/README.md
+++ b/keps/262-ConfigurableFailurePolicy/README.md
@@ -662,17 +662,17 @@ type ReplicatedJobStatus struct {
 	// suspended is the number of child Jobs which are in a suspended state.
 	Suspended int32 `json:"suspended"`
 
-  // restarts tracks the number of times each Job has restarted (i.e. recreated in case of RestartJob action).
+  // jobRestarts tracks the number of times each Job has restarted (i.e. recreated in case of RestartJob action).
   // The first value corresponds to the job with index 0, the second value corresponds to the job with index 1, etc.
   // Example: "0,1,0" implies that job 1 has restarted once, while jobs 0 and 2 haven't restarted.
   // +optional
-  Restarts *string `json:"restarts,omitempty"`
+  JobRestarts *string `json:"jobRestarts,omitempty"`
 
-  // restartsCountTowardsMax tracks the number of times each Job has restarted that counts towards the maximum allowed number of restarts.
+  // jobRestartsCountTowardsMax tracks the number of times each Job has restarted that counts towards the maximum allowed number of restarts.
   // The first value corresponds to the job with index 0, the second value corresponds to the job with index 1, etc.
   // Example: "0,1,0" implies that job 1 has restarted once, while jobs 0 and 2 haven't restarted.
   // +optional
-  RestartsCountTowardsMax *string `json:"restartsCountTowardsMax,omitempty"`
+  JobRestartsCountTowardsMax *string `json:"jobRestartsCountTowardsMax,omitempty"`
 }
 
 const (
@@ -730,29 +730,29 @@ Future restart logic for `action: RestartJobSet`:
 - Reconciliation loop 1
     - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
     - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (`restartJobSetActionApplier` from `RestartJobSet` in this example)
-    - In `restartJobSetActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts` or `js.Status.ReplicatedJobsStatus[rjName].RestartsCountTowardsMax[jobIndex] >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.Restarts += 1` and `js.Status.RestartsCountTowardsMax += 1` and `js.Status.ReplicatedJobsStatus[*].Restarts[*] += 1` and `js.Status.ReplicatedJobsStatus[*].RestartsCountTowardsMax[*] += 1`
+    - In `restartJobSetActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts` or `js.Status.ReplicatedJobsStatus[rjName].JobRestartsCountTowardsMax[jobIndex] >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.Restarts += 1` and `js.Status.RestartsCountTowardsMax += 1` and `js.Status.ReplicatedJobsStatus[*].JobRestarts[*] += 1` and `js.Status.ReplicatedJobsStatus[*].JobRestartsCountTowardsMax[*] += 1`
 - Reconciliation loop 2
-    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts` or `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.ReplicatedJobsStatus[rjName].Restarts[jobIndex]`) and group them to `ownedJobs.previous`
+    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts` or `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.ReplicatedJobsStatus[rjName].JobRestarts[jobIndex]`) and group them to `ownedJobs.previous`
     - Later in the `reconcile` function, the Jobs in `ownedJobs.previous` are deleted
 - Reconciliation loop 3...N
-    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs` with the updated labels `job.labels['jobset.sigs.k8s.io/restart-attempt'] = js.Status.Restarts` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] = js.Status.ReplicatedJobsStatus[rjName].Restarts[jobIndex]`
+    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs` with the updated labels `job.labels['jobset.sigs.k8s.io/restart-attempt'] = js.Status.Restarts` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] = js.Status.ReplicatedJobsStatus[rjName].JobRestarts[jobIndex]`
 
 Future restart logic for `action: RestartJob`:
 
 - Reconciliation loop 1
     - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
     - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (`restartJobActionApplier` from `RestartJob` in this example)
-    - In `restartJobActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts ` or `js.Status.ReplicatedJobsStatus[rjName].RestartsCountTowardsMax[jobIndex] >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.ReplicatedJobsStatus[rjName].Restarts[jobIndex] += 1` and `js.Status.ReplicatedJobsStatus[rjName].RestartsCountTowardsMax[jobIndex] += 1`
+    - In `restartJobActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts ` or `js.Status.ReplicatedJobsStatus[rjName].JobRestartsCountTowardsMax[jobIndex] >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.ReplicatedJobsStatus[rjName].JobRestarts[jobIndex] += 1` and `js.Status.ReplicatedJobsStatus[rjName].JobRestartsCountTowardsMax[jobIndex] += 1`
 - Reconciliation loop 2
-    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts` or `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.ReplicatedJobsStatus[rjName].Restarts[jobIndex]`) and group them to `ownedJobs.previous`
+    - In `getChildJobs`, the JobSet controller detects outdated Jobs (i.e., `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts` or `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.ReplicatedJobsStatus[rjName].JobRestarts[jobIndex]`) and group them to `ownedJobs.previous`
     - Later in the `reconcile` function, the Jobs in `ownedJobs.previous` are deleted
 - Reconciliation loop 3...N
-    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs` with the updated labels `job.labels['jobset.sigs.k8s.io/restart-attempt'] = js.Status.Restarts` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] = js.Status.ReplicatedJobsStatus[rjName].Restarts[jobIndex]`
+    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs` with the updated labels `job.labels['jobset.sigs.k8s.io/restart-attempt'] = js.Status.Restarts` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] = js.Status.ReplicatedJobsStatus[rjName].JobRestarts[jobIndex]`
 
 Observability:
 
 - `jobSet.status.restarts`, `jobSet.status.restartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/restart-attempt']` track only recreate-all-jobs restarts
-- `jobSet.status.replicatedJobsStatus[].restarts`, `jobSet.status.replicatedJobsStatus[].restartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` track recreate-single-job and recreate-all-jobs for a given job. If the user wants to know only the number of recreate-single-job, they can subtract the fields such as `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` - `job.labels['jobset.sigs.k8s.io/restart-attempt']`
+- `jobSet.status.replicatedJobsStatus[].jobRestarts`, `jobSet.status.replicatedJobsStatus[].jobRestartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` track recreate-single-job and recreate-all-jobs for a given job. If the user wants to know only the number of recreate-single-job, they can subtract the fields such as `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` - `job.labels['jobset.sigs.k8s.io/restart-attempt']`
 
 Future changes:
 


keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,39 @@ type FailurePolicy struct {
   Rules []FailurePolicyRule `json:"rules,omitempty"`
 }
 
+
+// JobSetStatus holds the status of a JobSet
+type JobSetStatus struct {
+  // JobsStatus holds the status of Jobs
+  // +optional
+	// +listType=map
```
kannon92: nit: formatting seems off here and not consistent.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,39 @@ type FailurePolicy struct {
...
+}
+
+// JobStatus holds the status of a Job
+type JobStatus struct {
+	// Name of the Job
+	Name string `json:"name"`
+
+	// Restarts tracks the number of times the Job has restarted (i.e. recreated in case of RestartJob action)
+	// +optional
+	Restarts int32 `json:"restarts"`
```
kannon92: You may want to try adding this API in the code and see what kube-api-linter says about this.

We should consider the behavior if this policy is not set and whether or not we expect these fields to be populated?

kannon92:
/hold

You wrote some detailed implementation details in this issue but the KEP doesn't have this.

I would prefer that we put more implementation details of these policies in the KEP.

I did not read the issue as I can't really comment on the design in a github issue description.

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: Good catch. It seems that:

- (1) There is a mixture of tabs and spaces for indentation
- (2) My IDE shows tabs as 2 spaces whereas GitHub uses 4 spaces

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: > You may want to try adding this API in the code and see what kube-api-linter says about this.

Good idea. Done!

> We should consider the behavior if this policy is not set and whether or not we expect these fields to be populated?

Makes sense. I am fine with keeping `jobSet.status.jobsStatus = nil` if there is no restart-single-job action in the JobSet object.

When implementing, I'll check if it is possible to only add entries to `jobSet.status.jobsStatus` for Jobs that restarted individually.

GiuseppeTT:
> I would prefer that we put more implementation details of these policies in the KEP.

Done! See https://github.com/kubernetes-sigs/jobset/pull/1152/changes#diff-95ff3f969c4dd0a07a5d8082df42bcb6f8f985e07bbe517b306635e397b4fca2R688

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -568,6 +685,57 @@ will add an annotation `jobset.sigs.k8s.io/restart` to mark Jobs which need to b
...
+- Reconciliation loop 3...N
+    - In `shouldCreateJob`, the JobSet controller blocks recreation for outdated Jobs that still exist. Jobs that do not exist are later created in `r.createJobs` with the updated labels `job.labels['jobset.sigs.k8s.io/restart-attempt'] = js.Status.Restarts` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] = js.Status.JobStatus[jobName].Restarts`
+
+Future restart logic for `action: RestartJob`:
+
+- Reconciliation loop 1
+    - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
+    - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (`restartJobActionApplier` from `RestartJob` in this example)
+    - In `restartJobActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts ` or `js.Status.JobStatus[jobName].RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.JobStatus[jobName].Restarts += 1` and `js.Status.JobStatus[jobName].RestartsCountTowardsMax += 1`
```
imreddy13: `js.Status.JobStatus[jobName].RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts`

> so we don't want to support a max # restarts per job before restarting the whole jobset? We only allow setting max restarts at the jobset level?

GiuseppeTT: Multiple restart limits for the same JobSet object did not get consensus on https://github.com/kubernetes-sigs/jobset/issues/1129, so I am focusing this PR on adding the restart action `RestartJob`  and `RestartJobAndIgnoreMaxRestarts`  which is the most important part of the feature request.

In this KEP, there is a single global restart limit. Any failure that is not of type "AndIgnoreMaxRestarts" will count towards it and the JobSet will fail if it is exceeded.

I am being careful on the API and implementation to keep the options open for a future KEP to add multiple restart limits.

GiuseppeTT:
@kannon92 @andreyvelich the KEP is ready for review : )

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -568,6 +685,57 @@ will add an annotation `jobset.sigs.k8s.io/restart` to mark Jobs which need to b
...
+- `jobSet.status.jobsStatus[].restarts`, `jobSet.status.jobsStatus[].restartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` track recreate-single-job and recreate-all-jobs for a given job. If the user wants to know only the number of recreate-single-job, they can subtract the fields such as `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` - `job.labels['jobset.sigs.k8s.io/restart-attempt']`
+
+Future changes:
+
+- Once we agree on how to implement multiple restart limits for the same JobSet, I believe that this KEP should be easy to adapt. This is because we are tracking the number of restarts for each Job. So, it should be easy to compare these values to the JobSet spec in any way to decide if the JobSet should be failed
+
+Relation to in-place restart:
+
+- When used with in-place restart, these changes allow the failed Job to be recreated while all the other Pods are restarted in-place. This is a desired behavior. For instance, when using `alpha.jobset.sigs.k8s.io/exclusive-topology: rack`, we want to recreate all Pods from a failed rack, while restarting the Pods from other racks in-place.
+
```
andreyvelich, line 764: Could you update Implementation History please ?
https://github.com/GiuseppeTT/jobset/blob/da596e58579de65f44e1afdd68e0a13858af9c32/keps/262-ConfigurableFailurePolicy/README.md#implementation-history

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -450,6 +450,81 @@ spec:
...
+metadata:
+  name: js
+spec:
+  failurePolicy:
+    maxRestarts: ...
+    rules:
+    # If any Job fails, restart the JobSet by recreating only the failed Job
+    # Count the restart towards `jobSet.spec.failurePolicy.maxRestarts`
+    - name: r
+      action: RestartJob
```
andreyvelich, line 471: What if user doesn't want to count the restart of `rj` ReplicatedJob towards `jobSet.spec.failurePolicy.maxRestarts`?
Any thoughts on this comment: https://github.com/kubernetes-sigs/jobset/issues/1129#issuecomment-3819882500 ?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,42 @@ type FailurePolicy struct {
...
 }
 
+// JobSetStatus defines the observed state of JobSet
+type JobSetStatus struct {
+  // jobsStatus holds the status of Jobs
+  // +optional
+  // +listType=map
+  // +listMapKey=name
+  // +kubebuilder:validation:MaxItems=1024
+  JobsStatus []JobStatus `json:"jobsStatus,omitempty"`
```
andreyvelich: Shall this be ReplicatedJobsStatus?
https://github.com/GiuseppeTT/jobset/blob/da596e58579de65f44e1afdd68e0a13858af9c32/api/jobset/v1alpha2/jobset_types.go#L221C25-L221C44

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,42 @@ type FailurePolicy struct {
...
+  // +kubebuilder:validation:MaxLength=253
+  Name string `json:"name,omitempty"`
+
+  // restarts tracks the number of times the Job has restarted (i.e. recreated in case of RestartJob action)
+  // +optional
+  Restarts *int32 `json:"restarts,omitempty"`
+
+  // restartsCountTowardsMax tracks the number of times the Job has restarted that counts towards the maximum allowed number of restarts
+  // +optional
+  RestartsCountTowardsMax *int32 `json:"restartsCountTowardsMax,omitempty"`
```
andreyvelich: Why do we need this if user can just simple check `replicatedJobsStatus.restarts` and `failurePolicy` configuration ?

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: They can use the new restart action `RestartJobAndIgnoreMaxRestarts`

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: I don't have a strong opinion here.

@kannon92 any thoughts?

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: I don't think `replicatedJobsStatus.restarts` exists https://github.com/kubernetes-sigs/jobset/blob/e803fa2436520dfb2f1f5a99eed1348d22ae6a1c/api/jobset/v1alpha2/jobset_types.go#L244-L265

The rationale of the KEP is that `jobSet.status.restarts` and `jobSet.status.restartsCountTowardsMax` are currently being used for global restarts, so we keep their semantics instead of making them count all types of restarts. 

With the KEP, Jobs will be able to be recreated individually. Consequently, we need individual counters (at least one for each Job replica that has been recreated individually). That's why I introduced `jobSet.status.jobsStatus[].restarts` and `jobSet.status.jobsStatus[].restartsCountTowardsMax`. 

For instance, the JobSet controller needs a way to detect if a Job should be deleted to be recreated individually. This is done with `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.JobStatus[jobName].Restarts`.

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: I'll do it : )

keps/262-ConfigurableFailurePolicy/README.md
andreyvelich: I guess, adding this should be sufficient, isn't ? That will indicate how many restarts each ReplicatedJob has.
```
.status.replicatedJobsStatus[x].restarts
```

keps/262-ConfigurableFailurePolicy/README.md
kannon92: I think this make sense to me. 

Are you suggesting to add this info to the exsting field?

andreyvelich: Yes, I don't think we should introduce a new `jobsStatus` field.

GiuseppeTT: I'll do it 🤝

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: I see 3 problems with this approach.

First. The JobSet controller needs a way to determine if a Job is in the right "generation". For instance, for recreate-all-jobs restarts, the JobSet controller does that by comparing `job.labels['jobset.sigs.k8s.io/restart-attempt'] < js.Status.Restarts`. What I am proposing is to also check `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < js.Status.JobStatus[jobName].Restarts` to handle recreate-single-job restarts.

Second. IMO we should reserve the field `.status.replicatedJobsStatus[x].restarts` in case someone requests the restart action `RestartReplicatedJob` in the future. That is, recreate all Jobs of a replicatedJob if one of them fails without touching the Jobs from other replicatedJobs.

Third. `js.Status.JobStatus[jobName].Restarts` is better for observability IMO. Since Jobs are being recreated individually and they have identity (they are indexed), we can observe how many times a specific Job was individually recreated and identify problems better.

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: Done

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: Done.

GiuseppeTT:
@kannon92 @andreyvelich 

I have address the feedback points. Are there any other concerns? Otherwise, I would like to proceed with the implementation : )

keps/262-ConfigurableFailurePolicy/README.md
kannon92: Hmm I was thinking we wouldn't need this field at all.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -450,6 +450,81 @@ spec:
               values: [42]
 ```
 
+#### Story 6: RestartJob and RestartJobAndIgnoreMaxRestarts
+
+As a user, I am running a ML workload using JobSet. My workload is able to recover from failures by restarting only the worker processes in the affected Job. The traditional restart logic `RestartJobSet` is too wasteful because it restarts all worker processes. Therefore, JobSet should provide a way to restart by recreating only the failed Job.
```
kannon92, line 455: In all these examples I don't know how we need to track the individual jobs in the ReplicatedJob.

It seems to say if any of the jobs fail within a replicated job trigger a restart.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,80 @@ type FailurePolicy struct {
   Rules []FailurePolicyRule `json:"rules,omitempty"`
 }
 
+// JobSetStatus defines the observed state of JobSet
```
kannon92, line 625: Its not really clear what you are adding to the API.

I think all you are adding is JobsStatus?

Everything else exists?

keps/262-ConfigurableFailurePolicy/README.md
kannon92: So this granularity is odd to me.

Why can't we just let the Job controller decide the failed and restart the failed jobs.

Our API only targets the replciatedJob so I don't understand why we are tracking the individual jobs restart counts. 

We don't provide the ability to control restarts based on the individual jobs so I don't know why we need to store this data in JobSet.

keps/262-ConfigurableFailurePolicy/README.md
kannon92: In our use cases I don't really know what the point of this field is. Our API says to RestartJob and targets the ReplicatedJob.

We don't provide the ability to control the number of restarts on the Job itself. In fact I think this may be difficult to do.

The other complicatation is that sometimes folks want to control the backoffLimit per index.

https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/3850-backoff-limits-per-index-for-indexed-jobs#goals

It still feels like to me we are overloading JobSet to control things that the Job itself should control.

GiuseppeTT: I think there is a confusion here. I am not proposing for JobSet to control the `backoffLimit`. In fact, all the use cases I have in mind will use `backoffLimit = 0`.

What I am proposing is a new restart action `RestartJob` (just like `RestartJobSet` is a restart action). In `RestartJobSet`, if a Job fails, the JobSet controller recreates **all** Jobs. In `RestartJob`, if a Job fails, the JobSet controller recreates **only the failed** Job.

The way JobSet is implemented, for the restart action `RestartJobSet`, it uses the JobSet field `jobSet.status.restarts` and the Job label `job.labels['jobset.sigs.k8s.io/restart-attempt']`. This works perfect for the recreate-all-jobs behavior but falls short for the recreate-only-failed-jobs behavior.

Similarly, to support the new restart action`RestartJob`, I want to use the new field `jobSet.status.jobsStatus[jobName].restarts` and the new Job label `job.labels['jobset.sigs.k8s.io/job-restart-attempt']`.

This is an approach very similar to what Willow proposed the first this feature was attempted.

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: @kannon92 I believe there is a miss understanding here. See my [other comment](https://github.com/kubernetes-sigs/jobset/pull/1152#discussion_r2866201285).

keps/262-ConfigurableFailurePolicy/README.md
kannon92: >  I think there is a confusion here. I am not proposing for JobSet to control the backoffLimit. In fact, all the use cases I have in mind will use backoffLimit = 0.

yea I get that. But I just think there are usecases where people want to control the retry ability at the job level.

My point is that our API doesn't really provide the granularity needed at the Job level. We are targeting a replicated job but there can be a bunch of replicas. Right now we have a general API that says RecreateJobs at the ReplicatedJobLevel. What happens if replica-1 of a Job needs to have a different restart level compared to replica-2 of a job?


It feels like what we really want is a Job policy like RecreateOnFailure. Essentially we want the job controller to trigger a recreation/rescheduling if it fails.

GiuseppeTT:
Synced offline. The main concern was the scalability of the field `jobsStatus`. So I replaced `jobsStatus` (which was an array of objects) with two new string fields: `jobRestarts` and `jobRestartsCountTowardsMax`. This is similar to the approach of `FailedIndexes` from the Job API.

@kannon92 @andreyvelich let me know if it is ready for implementation.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -568,6 +711,57 @@ will add an annotation `jobset.sigs.k8s.io/restart` to mark Jobs which need to b
...
+
+Future restart logic for `action: RestartJob`:
+
+- Reconciliation loop 1
+    - In `getChildJobs`, the JobSet controller detects failed Jobs (i.e., contain condition type `"Failed"`) and group them to `ownedJobs.failed`
+    - If `len(ownedJobs.failed) > 0`, `executeFailurePolicy` is executed and the right restart action is used (`restartJobActionApplier` from `RestartJob` in this example)
+    - In `restartJobActionApplier`, the JobSet controller fails the JobSet if it failed too much (i.e., `js.Status.RestartsCountTowardsMax >= js.Spec.FailurePolicy.MaxRestarts ` or `js.Status.ReplicatedJobsStatus[rjName].JobRestartsCountTowardsMax[jobIndex] >= js.Spec.FailurePolicy.MaxRestarts`). Otherwise, it bumps `js.Status.ReplicatedJobsStatus[rjName].JobRestarts[jobIndex] += 1` and `js.Status.ReplicatedJobsStatus[rjName].JobRestartsCountTowardsMax[jobIndex] += 1`
```
kannon92: JobRestartsCountTowardsMax is not an array anymore right?

kannon92 (commented):
LGTM!

/assign @andreyvelich

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: Exactly. It is a string now

https://github.com/kubernetes-sigs/jobset/pull/1152/changes#diff-95ff3f969c4dd0a07a5d8082df42bcb6f8f985e07bbe517b306635e397b4fca2R675

kannon92: But `js.Status.ReplicatedJobsStatus[rjName].JobRestartsCountTowardsMax[jobIndex] += 1` is saying it is an array right?

Saying this needs to be updated to reflect the new design.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -493,6 +567,13 @@ const (
 
   // Don't count the failure against maxRestarts.
   RestartJobSetAndIgnoreMaxRestarts FailurePolicyAction = "RestartJobSetAndIgnoreMaxRestarts"
+
+  // Restart only the failed Job if the number of restart attempts is less than MaxRestarts.
+  // Otherwise, fail the JobSet.
+  RestartJob FailurePolicyAction = "RestartJob"
+
+  // Same as RestartJob but do not count the failure against maxRestarts.
+  RestartJobAndIgnoreMaxRestarts FailurePolicyAction = "RestartJobAndIgnoreMaxRestarts"
```
andreyvelich, line 576: @GiuseppeTT @kannon92 I am wondering shall we call this: `RestartReplicatedJob` to be more specific to the ReplicatedJob resource?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,68 @@ type FailurePolicy struct {
...
 
+// JobSetStatus defines the observed state of JobSet
+type JobSetStatus struct {
+  // restarts tracks the number of times the JobSet has restarted (i.e. recreated in case of RecreateAll policy).
+  // +optional
+  Restarts int32 `json:"restarts"`
+
+  // restartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
+  // +optional
+  RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
```
andreyvelich, line 633: Why this is omitempty?
Shall this always be `==.restarts` unless we use `RestartJobSetAndIgnoreMaxRestarts`?

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,68 @@ type FailurePolicy struct {
...
+	Active int32 `json:"active"`
+
+	// suspended is the number of child Jobs which are in a suspended state.
+	Suspended int32 `json:"suspended"`
+
+  // jobRestarts tracks the number of times each Job has restarted (i.e. recreated in case of RestartJob action).
+  // The first value corresponds to the job with index 0, the second value corresponds to the job with index 1, etc.
+  // Example: "0,1,0" implies that job 1 has restarted once, while jobs 0 and 2 haven't restarted.
+  // +optional
+  JobRestarts *string `json:"jobRestarts,omitempty"`
```
andreyvelich: Shall this be just named `restarts` since it is already under `ReplicatedJobStatus`?
```suggestion
 Restarts *string `json:"jobRestarts,omitempty"`
```

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,68 @@ type FailurePolicy struct {
...
+	// active is the number of child Jobs with at least 1 pod in a running or pending state
+	// which are not marked for deletion.
+	Active int32 `json:"active"`
+
+	// suspended is the number of child Jobs which are in a suspended state.
+	Suspended int32 `json:"suspended"`
+
+  // jobRestarts tracks the number of times each Job has restarted (i.e. recreated in case of RestartJob action).
+  // The first value corresponds to the job with index 0, the second value corresponds to the job with index 1, etc.
+  // Example: "0,1,0" implies that job 1 has restarted once, while jobs 0 and 2 haven't restarted.
```
andreyvelich, line 667: @GiuseppeTT @kannon92 Instead of having logic to use `string` would it be more declarative to just define struct here ? 

For example:
```yaml
jobRestarts:
  - index: 0
     restarts: 0
     restartsCountTowardsMax: 0
  - index: 1
     restarts: 1
     restartsCountTowardsMax: 1
  - index: 2
     restarts: 0
     restartsCountTowardsMax: 0

```
Using string might introduce several downsides since it is hard to extend and doesn't have schema validation.

cc @tenzen-y @astefanutti

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -541,6 +622,68 @@ type FailurePolicy struct {
...
+  // +optional
+  JobRestartsCountTowardsMax *string `json:"jobRestartsCountTowardsMax,omitempty"`
+}
+
+const (
+  // RestartsKey is an annotation and label key which defines the restart attempt number
+  // the JobSet is currently on.
+  RestartsKey = "jobset.sigs.k8s.io/restart-attempt"
```
andreyvelich, line 681: I still don't get it why do we need it?
Users can read info from `.status` and `.failurePolicy` to understand the restart attempt number, isn't?

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: `restartsCountTowardsMax` is already part of the API of JobSet. Let's keep out of scope discussions outside the review of this KEP : )

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: > @GiuseppeTT @kannon92 Instead of having logic to use string would it be more declarative to just define struct here ?

This has been [discussed](https://github.com/kubernetes-sigs/jobset/pull/1152#issuecomment-3976542588). We went with a string due to scalability concerns (i.e., the JobSet object might get too large due to uncapped number of replicas). This is the same approach of the field `FailedIndexes` from the Job API.

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: `jobset.sigs.k8s.io/restart-attempt` is already part of the API of JobSet. Let's keep out of scope discussions outside the review of this KEP : )

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: When I read `RestartJob` it's clear to me that a Job is being restarted (i.e., recreated).

When I read `RestartReplicatedJob` I understand that a ReplicatedJob is being restarted, that is, all Jobs in a ReplicatedJob are being recreated. This is even said in the HEAD version of this file.

https://github.com/kubernetes-sigs/jobset/blob/963f92a0e0233f9b35ef838cc64246bc24ab78fa/keps/262-ConfigurableFailurePolicy/README.md?plain=1#L667

ahg-g: this is not restarting the whole replicatedJob though, this is restarting an individual job only

keps/262-ConfigurableFailurePolicy/README.md
ahg-g: not really, this is specific to individual job restarts, and so jobRestarts seems appropriate to me.

keps/262-ConfigurableFailurePolicy/README.md
andreyvelich: Oh, I see.

keps/262-ConfigurableFailurePolicy/README.md
andreyvelich: I see, the struct will grow significantly if user requests > 1000 Jobs.
Shall we call it `jobIndexesRestarts` and `jobIndexesRestartsCountTowardsMax` to be more explicit?

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: Done

keps/262-ConfigurableFailurePolicy/README.md
andreyvelich: I think, if we add `indexes` prefix that will make things clear. WDYT @GiuseppeTT @ahg-g @kannon92 ?
```yaml
status:
  replicatedJobsStatus:
    - name: launcher
      indexesRestarts: 0,2,0
      indexesRestartsCountTowardsMax: 0,2,0
```

ahg-g: I don't think it will add too much information, it might actually be confusing here since the values are not indexes, they are restart counts. I would keep it as is

GiuseppeTT: > I don't think it will add too much information, it might actually be confusing here since the values are not indexes, they are restart counts. I would keep it as is

+1

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: Reverted.

GiuseppeTT:
@andreyvelich are there any further concerns? : )

ahg-g:
/approve
/lgtm

/hold 
In case Andrey has any other followups

andreyvelich:
Sure, that looks good to me guys, just a small nit as I asked in Slack, do we want to keep API name as `restartsPerIndex` or `restarts` ?

GiuseppeTT:
> Sure, that looks good to me guys, just a small nit as I asked in Slack, do we want to keep API name as `restartsPerIndex` or `restarts` ?

Let's keep it as `jobRestarts`.

See https://github.com/kubernetes-sigs/jobset/pull/1152#discussion_r2873960262

andreyvelich:
sounds good.
/hold cancel

PR #1167, opened 2026-02-25
List child Jobs by JobSet UID instead of JobSet name + namespace

### LINKED ISSUE #1165: The JobSet controller should list child Jobs by JobSet UID
**Summary**:

The JobSet controller lists child Jobs by JobSet **name** but it should be by JobSet **UID**. Otherwise, the JobSet controller can malfunction by reconciling Jobs from different JobSets that share the same name.

**What happened**:

User recreated a JobSet with the same name but different replicatedJob names and got `panic: assignment to entry in nil map` in [this line](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L373).

No panic.

It is not deterministic but the steps are the following:

1. Create JobSet `js` with replicatedJob `rj-v-0` with 2 Jobs `rj-v-0-0` and `rj-v-0-1`
2. Delete JobSet `js`
3. Recreate JobSet `js` with replicatedJob `rj-v-1` with 2 Jobs `rj-v-1-0` and `rj-v-1-1`

To make this work, Job `rj-v-0-0` (or `rj-v-0-1`) should fail as close as possible to step 3 and take as long as possible to be garbage collected.

The culprit is [this line](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L373).

```go
for _, job := range jobs.failed {
    replicatedJobsReady[job.Labels[jobset.ReplicatedJobNameKey]]["failed"]++
}
```

That is, the JobSet controller is trying to access the key `"failed"` in the map `replicatedJobsReady[replicatedJobName]` that is `nil` because it was not initialized. Good golang code should check for a key instead of just panicking, but this is not the only bug. The real bug is that the JobSet controller lists child Jobs by JobSet name (`job.metadata.ownerReferences.name`) (see [1](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L280) and [2](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L248)). It should be by JobSet UID (`job.metadata.ownerReferences.uid`).

So, what likely happened is:

1. User creates JobSet `js` with replicatedJob `rj-v-0`
2. JobSet controller creates child Jobs `js-rj-v-0-0` and `js-rj-v-0-1`
3. User deletes JobSet `js`
4. Garbage controller starts deleting child Jobs `js-rj-v-0-0` and `js-rj-v-0-1`. `js-rj-v-0-0` is deleted fast, but `js-rj-v-0-1` takes longer. `js-rj-v-0-1` also fails
5. In parallel to the garbage collection, user recreates JobSet `js` with replicatedJob `rj-v-1`
6. JobSet controller creates child Jobs `js-rj-v-1-0` and `js-rj-v-1-1`
7. The JobSet controller executes [this line](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L280). The child Jobs include `js-rj-v-1-0` and `js-rj-v-1-1` but also include `js-rj-v-0-1` which hasn't been deleted yet
8. The JobSet controller executes [this line](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L335). The map `replicatedJobsReady` only contains the key `rj-v-1`
9. The JobSet controller executes [this line](https://github.com/kubernetes-sigs/jobset/blob/19960d097cdbdb8a2435ae095181832e5476e44e/pkg/controllers/jobset_controller.go#L373). It works for `js-rj-v-1-0` and `js-rj-v-1-1` but fails for `js-rj-v-0-1` because `replicatedJobsReady[job.Labels[jobset.ReplicatedJobNameKey]]["failed"]++ = replicatedJobsReady["rj-v-0"]["failed"]++ = nil["failed"]++` which panics

**Environment**:

- Kubernetes version (use `kubectl version`): N/A
- JobSet version (use `git describe --tags --dirty --always`): 0.9.1
- Cloud provider or hardware configuration: GCP, but it's irrelevant
- Install tools: N/A
- Others: N/A

GiuseppeTT:
#### What type of PR is this?

/kind bug

Make the JobSet controller list child Jobs by JobSet UID instead of JobSet name. Otherwise the JobSet controller can malfunction due to reconciling not owned Jobs. See example of bug in #1165.

Fixes #1165

The main fix is in the lines [1](https://github.com/kubernetes-sigs/jobset/compare/main...GiuseppeTT:list-jobs-by-uid?expand=1#diff-68677f8530f00fbbb399ac36268f928ee584948dd09091644a8fad81a8658153R299) and [2](https://github.com/kubernetes-sigs/jobset/compare/main...GiuseppeTT:list-jobs-by-uid?expand=1#diff-68677f8530f00fbbb399ac36268f928ee584948dd09091644a8fad81a8658153R349). The rest is just improving the logging and adding tests.

```release-note
JobSet controller reconciles owned Jobs by JobSet UID instead of JobSet name + namespace.
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -138,7 +138,7 @@ func (r *JobSetReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 
 // reconcile is the internal method containing the core JobSet reconciliation logic.
 func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, updateStatusOpts *statusUpdateOpts) (ctrl.Result, error) {
-	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js))
+	log := ctrl.LoggerFrom(ctx).WithValues("jobset", klog.KObj(js), "jobSetUID", js.UID)
 	ctx = ctrl.LoggerInto(ctx, log)
 
 	// Check the controller configured for the JobSet.


kannon92:
/lgtm

Feel free to add a cherry pick label if you think this should go into 0.10.

Issue #380, opened 2024-01-18
Graduate the API to v1beta1


ahg-g:
Graduate JobSet API to v1beta1. We need to keep v1apha1 for a few more releases to make it easier to customers to migrate.

Ref: https://book.kubebuilder.io/multiversion-tutorial/api-changes

danielvegamyhre:
@ahg-g and I were discussing skipping v1beta1 and graduating directly to v1, since several companies have been using for their production training workloads for a few months now, so it effectively GA already.

@kannon92 @vsoch what do you think?

vsoch:
It depends on if you think it's really free of errors and potential issues, or not. I don't see any harm in doing v1beta1 and then having that wiggle room, but it's up to you!

kannon92:
I think the main thing about v1 promotion is that means we should not break any existing user. I think we have been pretty careful on not breaking the API anymore so I think its fine to promote to GA.

Could you ask those companies (or even you!) to create an adopters page? It'd be nice to convey that people are actually using this project for something.

ahg-g:
@vsoch it is not about free of errors, this is more about what commitment we are making to the API stability. Since multiple users are already dependent on it, this is practically becoming GA because everything we do moving forward must be backward compatible, so might as well just make that commitment official in the API.

> Could you ask those companies (or even you!) to create an adopters page? It'd be nice to convey that people are actually using this project for something.

we can mention that Google Cloud is using it (we are not at liberty to list the customers); @vsoch if you feel comfortable perhaps we can list Lawrence Livermore National Laboratory?

vsoch:
I will ask! To be clear - "using it" meaning for development and prototyping or in production? We do not have a production Kubernetes cluster. That's what we are working towards.

kannon92:
@ahg-g @danielvegamyhre from the KubeFlow discussions, do we want to table this?

ahg-g:
Yes, I think so.

andreyvelich:
[...]

PR #1177, opened 2026-03-04
Add restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts`

### LINKED ISSUE #1129: Add single Job recreation to JobSet restarts

GiuseppeTT:
#### What type of PR is this?

/feature

Add restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts` from KEP [262](https://github.com/kubernetes-sigs/jobset/tree/main/keps/262-ConfigurableFailurePolicy).

I also made a small change to the KEP:

- `jobSet.status.replicatedJobs[].jobRestarts`, `jobSet.status.replicatedJobs[].jobRestartsCountTowardsMax` and `job.labels['jobset.sigs.k8s.io/job-restart-attempt']` now count the number of individual Job restarts instead of all the times the job has restarted (global restarts and individual restarts). This simplifies the API, implementation and improve backward compatibility because the individual job restart will default to `0` if not present
- Added fields `jobSet.status.totalRestarts` and `jobSet.status.totalRestartsCountTowardsMax` to improve observability. `jobSet.status.restarts` only counts global restarts (recreate all jobs), so these new field help to count all restarts. This was a direct request from a user

Fixes #1129

Main changes for helping to review files:

- `api/jobset/v1alpha2/jobset_types.go`
    - Add the new API fields
- `pkg/constants/constants.go`
    - Add the new Job label / annotation `jobset.sigs.k8s.io/job-restart-attempt`
- `pkg/controllers/jobset_controller.go`
    - Create Jobs with `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] = jobSet.status.replicatedJobs[replicatedJobName].jobRestarts[jobIndex]`
    - Delete old jobs that have `job.labels['jobset.sigs.k8s.io/job-restart-attempt'] < jobSet.status.replicatedJobs[replicatedJobName].jobRestarts[jobIndex]`
- `pkg/controllers/failure_policy.go`
    - Implement `RestartJob` and `RestartJobAndIgnoreMaxRestarts`
    - If they are triggered, bump `jobSet.status.replicatedJobs[replicatedJobName].jobRestarts[jobIndex]` and `jobSet.status.replicatedJobs[replicatedJobName].jobRestartsCountTowardsMax[jobIndex]`
- `pkg/controllers/jobset_controller_test.go` and `pkg/controllers/failure_policy.go`
    - Tests
- All the other files
    - Generated files

```release-note
Add restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts` that allow JobSet to restart by recreating only the failed Job instead of all Jobs.
```

### DIFF
--- a/pkg/controllers/failure_policy.go
+++ b/pkg/controllers/failure_policy.go
@@ -307,7 +307,7 @@ func failurePolicyRecreateJob(ctx context.Context, js *jobset.JobSet, matchingFa
 // restartJobActionApplier applies the RestartJob FailurePolicyAction
 var restartJobActionApplier failurePolicyActionApplier = func(ctx context.Context, js *jobset.JobSet, matchingFailedJob *batchv1.Job, updateStatusOpts *statusUpdateOpts) error {
 	if !features.Enabled(features.RestartJob) {
-		return fmt.Errorf("RestartJob failure policy action cannot be used when the feature gate is disabled")
+		return fmt.Errorf("RestartJob failure policy action cannot be used when the feature gate RestartJob is disabled")
 	}
 
 	if totalRestartsCountTowardsMax(js) >= js.Spec.FailurePolicy.MaxRestarts {
@@ -334,7 +334,7 @@ var restartJobActionApplier failurePolicyActionApplier = func(ctx context.Contex
 // restartJobAndIgnoreMaxRestartsActionApplier applies the RestartJobAndIgnoreMaxRestarts FailurePolicyAction
 var restartJobAndIgnoreMaxRestartsActionApplier failurePolicyActionApplier = func(ctx context.Context, js *jobset.JobSet, matchingFailedJob *batchv1.Job, updateStatusOpts *statusUpdateOpts) error {
 	if !features.Enabled(features.RestartJob) {
-		return fmt.Errorf("RestartJobAndIgnoreMaxRestarts failure policy action cannot be used when the feature gate is disabled")
+		return fmt.Errorf("RestartJobAndIgnoreMaxRestarts failure policy action cannot be used when the feature gate RestartJob is disabled")
 	}
 
 	baseMessage := constants.RestartJobAndIgnoreMaxRestartsActionMessage


api/jobset/v1alpha2/jobset_types.go
```diff
@@ -201,14 +201,24 @@ type JobSetStatus struct {
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 
-	// restarts tracks the number of times the JobSet has restarted (i.e. recreated in case of RecreateAll policy).
+	// restarts tracks the number of times the JobSet has globally restarted (i.e. recreated all Jobs due to a restart action such as RestartJobSet).
```
kannon92, line 204: Idk what you mean by globally.

This tracks the number of restarts a JobSet does.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -262,6 +272,18 @@ type ReplicatedJobStatus struct {
 
 	// suspended is the number of child Jobs which are in a suspended state.
 	Suspended int32 `json:"suspended"`
+
+	// jobRestarts tracks the number of times the Job has individually restarted for each job index.
+	// It is encoded as `<restarts of job 0>,...,<restarts of job replicas - 1>`
```
kannon92: Can you add an API comment on the MaxLength?

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -201,14 +201,24 @@ type JobSetStatus struct {
...
+	// restarts tracks the number of times the JobSet has globally restarted (i.e. recreated all Jobs due to a restart action such as RestartJobSet).
 	// +optional
 	Restarts int32 `json:"restarts"`
 
-	// restartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
+	// restartsCountTowardsMax tracks the number of times the JobSet has globally restarted that counts towards the maximum allowed number of restarts (i.e. recreated all Jobs due to a restart action such as RestartJobSet).
 	// +optional
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// totalRestarts tracks the number of times the JobSet has restarted in any way (e.g., this also counts restart actions such as RestartJob).
```
kannon92: I am struggling to understand why we need both Restarts and TotalRestarts.

How is an end user supposed to understand the difference of these two values

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: Global restarts mean recreate-all-jobs restarts which are triggered by the restart actions `RestartJobSet` and `RestartJobSetAndIgnoreMaxRestarts`.

I can improve the comments.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: That's a valid question : )

Before this KEP, the field `jobSet.status.restarts` had two functions:

- (1) Count how many times the JobSet has restarted. That is, how many times the JobSet has triggered the restart actions `RestartJobSet` and `RestartJobSetAndIgnoreMaxRestarts`
- (2) Reconcile child Jobs. A child Job should always satisfy `job.labels['jobset.sigs.k8s.io/restart-attempt'] = jobSet.status.restarts`, so JobSet restarts are achieved by bumping `jobSet.status.restarts += 1`, which makes the JobSet controller delete Jobs with `job.labels['jobset.sigs.k8s.io/restart-attempt'] < jobSet.status.restarts` and later recreate them with `job.labels['jobset.sigs.k8s.io/restart-attempt'] = jobSet.status.restarts`

This KEP is adding the restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts`. Having a clear way to count the total number of times any restart action has been triggered is important (I got this directly from a user). At first, I considered adapting (1) from `jobSet.status.restarts`  to also include `RestartJob` and `RestartJobAndIgnoreMaxRestarts`. But this is not possible because `jobSet.status.restarts` is already deeply used to reconcile Jobs in recreate-all-jobs restarts in (2).

So, to not break backward compatibility, I kept the function (2) of `jobSet.status.restarts` and added the new field `jobSet.status.totalRestarts` for (1). In the future, if we ever implement the action `RestartReplicatedJob`, we can keep `jobSet.status.totalRestarts` because it is not used to reconcile Jobs.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: Done.

This limit was required by the API linter. We can bump it in the future if the use is needed and if the JobSet controller can handle it.

api/jobset/v1alpha2/jobset_types.go
kannon92: So is jobSet.Status.Restarts a reliable field to use for users if this feature is enabled?

I guess technically you never want to restart the full JobSet so this would always be 0 if RestartJob or RestartJobAndIgnoreMaxRestarts?

api/jobset/v1alpha2/jobset_types.go
imreddy13: Can it be a map of {job, # restarts}?

api/jobset/v1alpha2/jobset_types.go
imreddy13: @GiuseppeTT why not have separate fields for each level of restart:

1) current fields Restarts and RestartsCountTowardsMax => JobSet restarts
2) new fields JobRestarts and JobRestartsCountTowardsMax => Job restarts
3) (in the future) new fields ReplicatedJobRestarts and ReplicatedJobRestartsCountTowardsMax => ReplicatedJob restarts

when would we need total restarts?

pkg/controllers/jobset_controller.go
```diff
@@ -350,25 +350,59 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
 
-	// Categorize each job into a bucket: active, successful, failed, or delete.
+	// Categorize each job into a bucket: active, successful, failed or previous.
 	ownedJobs := childJobs{}
 	for i, job := range childJobList.Items {
-		// Jobs with jobset.sigs.k8s.io/restart-attempt < restarts are marked for deletion.
-
-		jobRestarts, err := strconv.Atoi(job.Labels[constants.RestartsKey])
+		// A job is marked for deletion if its global restart attempt is less than the JobSet's global restarts.
+		globalRestarts, err := strconv.Atoi(job.Labels[constants.RestartsKey])
```
imreddy13: Why do we need to touch this code? I know you're just renaming vars but I prefer to leave it alone if there's no feature gate. It's critical jobset recreation code

pkg/controllers/jobset_controller.go
```diff
@@ -350,25 +350,59 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		targetGlobalRestarts := js.Status.Restarts
+		if int32(globalRestarts) < targetGlobalRestarts {
+			log.V(2).Info("child Job marked for recreation as value of global restarts label is less than target", "name", job.Name, constants.RestartsKey, globalRestarts, "targetGlobalRestarts", js.Status.Restarts)
 			ownedJobs.previous = append(ownedJobs.previous, &childJobList.Items[i])
 			continue
 		}
 
-		// Jobs with jobset.sigs.k8s.io/restart-attempt == jobset.status.restarts are part of
-		// the current JobSet run, and marked either active, successful, or failed.
+		// A job is marked for deletion if its individual restart attempt is less than the target individual restart attempt.
```
imreddy13: Can you make a helper method for this logic like shouldMarkJobForDeletion()?

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: > So is jobSet.Status.Restarts a reliable field to use for users if this feature is enabled?

Yes.

With and without this feature enabled, `jobSet.status.restarts` will count how many times the restart actions `RestartJobSet` and `RestartJobSetAndIgnoreMaxRestarts` have been triggered

With and without this feature enabled, `jobSet.status.totalRestarts` will count how many times the restart actions `RestartJobSet`, `RestartJobSetAndIgnoreMaxRestarts`, `RestartJob` and `RestartJobAndIgnoreMaxRestarts` have been triggered

GiuseppeTT: > I guess technically you never want to restart the full JobSet so this would always be 0 if RestartJob or RestartJobAndIgnoreMaxRestarts?

Story 6, example 2 shows a JobSet with `RestartJobSet` and `RestartJob`

https://github.com/kubernetes-sigs/jobset/tree/main/keps/262-ConfigurableFailurePolicy#story-6-restartjob-and-restartjobandignoremaxrestarts

To give you an example of a few different scenarios:

If `RestartJobSet` is triggered:

- Bump `jobSet.status.restarts += 1`
- Bump `jobSet.status. restartsCountTowardsMax  += 1`
- Bump `jobSet.status.totalRestarts += 1`
- Bump `jobSet.status.totalRestartsCountTowardsMax  += 1`

If `RestartJob` is triggered:

- Bump `jobSet.status.replicatedJob[replicatedJobName].jobRestarts[jobIndex] += 1`
- Bump `jobSet.status.replicatedJob[replicatedJobName]. jobRestartsCountTowardsMax[jobIndex] += 1`
- Bump `jobSet.status.totalRestarts += 1`
- Bump `jobSet.status.totalRestartsCountTowardsMax  += 1`

GiuseppeTT: > @GiuseppeTT why not have separate fields for each level of restart:

That is an option. Let me check with the owner of the use case if they have a preference.

api/jobset/v1alpha2/jobset_types.go
imreddy13: @kannon92 do you have a strong preference on this not being a map? I doubt we will exceed 1k replicas & this string makes debugging harder. Plus we already have a map for replicated jobs.

Anyway I won't block on this since we already agreed on the string previously

pkg/controllers/jobset_controller.go
GiuseppeTT: Done.

pkg/controllers/jobset_controller.go
GiuseppeTT: I don't think that helps given how the loop is structured. Depending on intermediary values the function should either error or just continue to the next iteration.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -27,12 +27,14 @@ tags, and then generate with `hack/update-toc.sh`.
     - [Story 3: RestartJobSetAndIgnoreMaxRestarts](#story-3-restartjobsetandignoremaxrestarts)
     - [Story 4: Different failure policies for different replicated jobs](#story-4-different-failure-policies-for-different-replicated-jobs)
     - [Story 5: Distinguishing retriable from non retriable Pod failure policies](#story-5-distinguishing-retriable-from-non-retriable-pod-failure-policies)
+    - [Story 6: RestartJob and RestartJobAndIgnoreMaxRestarts](#story-6-restartjob-and-restartjobandignoremaxrestarts)
```
kannon92, line 30: Can you fix the toc as a separate PR?

Its failing a few PRs and we should get our CI green.

GiuseppeTT: Done.

https://github.com/kubernetes-sigs/jobset/pull/1180

kannon92:
/assign

I'm going to try and make time for this. I'll be at Kubecon for last week of month so I'll try to give a review before then.

I mostly looked at API so far but haven't dove into the implementation.

api/jobset/v1alpha2/jobset_types.go
kannon92: I like this idea.

We can't do this for restarts and restartsCountTowardsMax but it may be worth considering this for v1 promotion.

kannon92 (changes requested):
Please add some integration tests for this feature.

I would like to see a e2e test also.

pkg/constants/constants.go
```diff
@@ -99,7 +103,17 @@ const (
 	RestartJobSetAndIgnoreMaxRestartsActionReason  = "RestartJobSetAndIgnoreMaxRestartsFailurePolicyAction"
 	RestartJobSetAndIgnoreMaxRestartsActionMessage = "applying RestartJobSetAndIgnoreMaxRestarts failure policy action"
 
+	// Event reason used when a PVC creation fails.
+	// The event uses the error(s) as the message.
 	// Event reason used when a PVC creation fails.
 	// The event uses the error(s) as the message.
```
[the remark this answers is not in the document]
GiuseppeTT: Done.

pkg/controllers/jobset_controller.go
```diff
@@ -367,8 +367,42 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		replicatedJobName, ok := job.Labels[jobset.ReplicatedJobNameKey]
+		if !ok {
+			log.Error(err, fmt.Sprintf("missing label %s", jobset.ReplicatedJobNameKey))
+			ownedJobs.previous = append(ownedJobs.previous, &childJobList.Items[i])
+			return nil, err
+		}
```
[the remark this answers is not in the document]
GiuseppeTT: Done.

pkg/controllers/jobset_controller.go
```diff
@@ -367,8 +367,42 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		if err != nil {
+			log.Error(err, fmt.Sprintf("invalid value for label %s, must be integer", jobset.JobIndexKey))
+			ownedJobs.previous = append(ownedJobs.previous, &childJobList.Items[i])
+			return nil, err
+		}
+		individualRestartsStr, ok := job.Labels[constants.JobRestartAttemptKey]
+		// If the label is missing, just log instead of deleting the Job for backward compatibility
+		// If this Job is ever recreated, it will have the label
+		if !ok {
+			log.Error(err, fmt.Sprintf("missing label %s", constants.JobRestartAttemptKey))
```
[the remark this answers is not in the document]
GiuseppeTT: Done.

pkg/controllers/failure_policy.go
```diff
@@ -186,14 +190,27 @@ func failurePolicyRecreateAll(ctx context.Context, js *jobset.JobSet, shouldCoun
...
+		js.Status.TotalRestarts = ptr.To(js.Status.Restarts)
+	}
+	*js.Status.TotalRestarts += 1
+	if shouldCountTowardsMax {
+		if js.Status.TotalRestartsCountTowardsMax == nil {
+			// Default to `js.Status.RestartsCountTowardsMax` for backward compatibility
+			js.Status.TotalRestartsCountTowardsMax = ptr.To(js.Status.RestartsCountTowardsMax)
+		}
+		*js.Status.TotalRestartsCountTowardsMax += 1
+	}
```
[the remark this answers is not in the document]
GiuseppeTT: Done.

pkg/controllers/failure_policy.go
```diff
@@ -256,6 +274,94 @@ var restartJobSetAndIgnoreMaxRestartsActionApplier failurePolicyActionApplier =
...
+	if err != nil {
+		return fmt.Errorf("failed job %s has missing or invalid %s label: %s", matchingFailedJob.Name, jobset.JobIndexKey, matchingFailedJob.Labels[jobset.JobIndexKey])
+	}
+
+	// Increment the individual restart count for the specific job
+	jobRestarts := getJobRestarts(js, replicatedJobName)
+	jobRestarts[jobIndex] += 1
+	setJobRestarts(js, replicatedJobName, jobRestarts)
+	if shouldCountTowardsMax {
+		jobRestartsCountTowardsMax := getJobRestartsCountTowardsMax(js, replicatedJobName)
```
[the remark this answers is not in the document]
GiuseppeTT: Done.

pkg/controllers/jobset_controller.go
```diff
@@ -367,8 +367,42 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+		} else {
+			individualRestarts, err := strconv.Atoi(individualRestartsStr)
+			if err != nil {
+				log.Error(err, fmt.Sprintf("invalid value for label %s, must be integer", constants.JobRestartAttemptKey))
+				ownedJobs.previous = append(ownedJobs.previous, &childJobList.Items[i])
+				return nil, err
+			}
+			targetIndividualRestarts := getJobRestarts(js, replicatedJobName)[jobIndex]
+			if int32(individualRestarts) < targetIndividualRestarts {
```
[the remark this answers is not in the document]
GiuseppeTT: Done

GiuseppeTT:
> Please add some integration tests for this feature.
> I would like to see a e2e test also.

Done.

kannon92:
Can you add a doc for this feature also? We could probably add an example in FailurePolicy

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -262,6 +272,20 @@ type ReplicatedJobStatus struct {
...
+	// It is encoded as `<restarts of job 0>,...,<restarts of job replicas - 1>`
+	// Max length is set to 32KB (32768 bytes). This is enough to handle 2 978 replicas per replicatedJob
+	// +kubebuilder:validation:MaxLength=32768
+	// +optional
+	JobRestarts *string `json:"jobRestarts,omitempty"`
+
+	// jobRestartsCountTowardsMax tracks the number of times the Job has individually restarted that counts towards the maximum allowed number of restarts for each job index.
+	// It is encoded as `<restarts of job 0>,...,<restarts of job replicas - 1>`
+	// Max length is set to 32KB (32768 bytes). This is enough to handle 2 978 replicas per replicatedJob
+	// +kubebuilder:validation:MaxLength=32768
```
kannon92: What is going to happen if the replicas causes this to exceed?

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: > > @GiuseppeTT why not have separate fields for each level of restart:
> 
> That is an option. Let me check with the owner of the use case if they have a preference.

I talked to the user. They prefer `jobset.status.totalRestarts` (count `RestartJobSet` + `RestartJob` triggers) over something like `jobset.status.individualJobRestarts` (count `RestartJob` triggers)

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: AFAIK the control plane will reject any jobSet object that fails this validation and this will cause the controller to malfunction.

In my last commit, I made the webhook block creation of JobSets with `any(jobset.spec.replicatedJob[].replicas > 2978)` if `RestartJob` is used in the JobSet spec. Let me know what you think

This is not a problem for existing JobSets because this field is nil if `RestartJob` is not used. See the e2e test for context.

GiuseppeTT:
> Can you add a doc for this feature also? We could probably add an example in FailurePolicy

Done!

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -393,6 +417,13 @@ const (
 
 	// Do not count the failure against maxRestarts.
 	RestartJobSetAndIgnoreMaxRestarts FailurePolicyAction = "RestartJobSetAndIgnoreMaxRestarts"
+
+	// Restart (i.e., recreate) the failed Job individually if the number of restart attempts is less than MaxRestarts.
+	// Otherwise, fail the JobSet.
+	RestartJob FailurePolicyAction = "RestartJob"
+
+	// Restart (i.e., recreate) the failed Job individually and do not count the failure against maxRestarts.
```
kannon92, line 418: Why do we need both options honestly?

Couldn't you just set MaxRestarts very high?

pkg/controllers/failure_policy.go
```diff
@@ -256,6 +276,100 @@ var restartJobSetAndIgnoreMaxRestartsActionApplier failurePolicyActionApplier =
...
+		updateStatusOpts = &statusUpdateOpts{}
+	}
+
+	replicatedJobName, ok := matchingFailedJob.Labels[jobset.ReplicatedJobNameKey]
+	if !ok {
+		return fmt.Errorf("failed job %s missing %s label", matchingFailedJob.Name, jobset.ReplicatedJobNameKey)
+	}
+	jobIndex, err := strconv.Atoi(matchingFailedJob.Labels[jobset.JobIndexKey])
+	if err != nil {
+		return fmt.Errorf("failed job %s has missing or invalid %s label: %s", matchingFailedJob.Name, jobset.JobIndexKey, matchingFailedJob.Labels[jobset.JobIndexKey])
```
kannon92: This could potentially panic if the label doesn't exist, right?

Can you check if label exists and then do Atoi?

Also you are swallowing the error from ATOI here.

pkg/controllers/failure_policy.go
```diff
@@ -256,6 +276,100 @@ var restartJobSetAndIgnoreMaxRestartsActionApplier failurePolicyActionApplier =
...
+	}
+
+	// Increment the individual restart count for the specific job
+	jobRestarts := getJobRestarts(js, replicatedJobName)
+	if jobIndex >= len(jobRestarts) {
+		return fmt.Errorf("failed job %s has invalid %s label: %s", matchingFailedJob.Name, jobset.JobIndexKey, matchingFailedJob.Labels[jobset.JobIndexKey])
```
kannon92: This error message does not seem to match line 298?

Is this really an invalid label?

pkg/controllers/failure_policy.go
```diff
@@ -256,6 +276,100 @@ var restartJobSetAndIgnoreMaxRestartsActionApplier failurePolicyActionApplier =
...
+	}
+	jobRestarts[jobIndex] += 1
+	setJobRestarts(js, replicatedJobName, jobRestarts)
+	if shouldCountTowardsMax {
+		jobRestartsCountTowardsMax := getJobRestartsCountTowardsMax(js, replicatedJobName)
+		if jobIndex >= len(jobRestartsCountTowardsMax) {
+			return fmt.Errorf("failed job %s has invalid %s label: %s", matchingFailedJob.Name, jobset.JobIndexKey, matchingFailedJob.Labels[jobset.JobIndexKey])
```
kannon92: Same here. https://github.com/kubernetes-sigs/jobset/pull/1177/changes#r2897662717

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1752,6 +1752,40 @@ func TestValidateCreate(t *testing.T) {
...
+									Completions:    ptr.To(int32(1)),
+									Parallelism:    ptr.To(int32(1)),
+								},
+							},
+						},
+					},
+					FailurePolicy: &jobset.FailurePolicy{
+						Rules: []jobset.FailurePolicyRule{
+							{
+								Action: jobset.RestartJob,
```
kannon92, line 1779: What about the other action you added? Seems this would fail.

pkg/controllers/jobset_controller.go
```diff
@@ -367,8 +367,50 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
 			continue
 		}
 
-		// Jobs with jobset.sigs.k8s.io/restart-attempt == jobset.status.restarts are part of
-		// the current JobSet run, and marked either active, successful, or failed.
+		// A job is marked for deletion if its individual restart attempt is less than the target individual restart attempt.
+		replicatedJobName, ok := job.Labels[jobset.ReplicatedJobNameKey]
+		if !ok {
+			err := fmt.Errorf("missing label %s", jobset.ReplicatedJobNameKey)
```
kannon92: This whole block seems odd.

You could just feed line 373 into log.Error(err, "").

And why would we add a value into an array if there is an error?

pkg/controllers/jobset_controller.go
```diff
@@ -367,8 +367,50 @@ func (r *JobSetReconciler) getChildJobs(ctx context.Context, js *jobset.JobSet)
...
+			log.Error(err, "")
+			ownedJobs.previous = append(ownedJobs.previous, &childJobList.Items[i])
+			return nil, err
+		}
+		jobIndex, err := strconv.Atoi(job.Labels[jobset.JobIndexKey])
```
kannon92: This would panic if label doesn't exist and you are not really returning a reliable error message here.

I would check for label exists (return err if does not) and then do Atoi.

test/integration/controller/jobset_controller_test.go
```diff
@@ -743,6 +743,46 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
 						matchJobSetRestartsCountTowardsMax(js, 1)
+						matchJobSetTotalRestarts(js, int32(1))
+						matchJobSetTotalRestartsCountTowardsMax(js, int32(1))
+						matchJobRestarts(js, "replicated-job-a", nil)
+						matchJobRestartsCountTowardsMax(js, "replicated-job-a", nil)
+					},
+				},
+			},
+		}),
+		ginkgo.Entry("[failure policy] jobset restarts with RestartJob failure policy action.", &testCase{
```
kannon92, line 788: You are missing another test case here.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: For the same reason we need `RestartJobSet` and `RestartJobSetAndIgnoreMaxRestarts`. We have users that use `RestartJobSetAndIgnoreMaxRestarts` to handle expected failures such as infra failures and use `RestartJobSet` along with a small `maxRestarts` to retry unexpected failures without wasting time in errors that will never be fixed with a restart.

We have the same expectations and use case for `RestartJob`.

pkg/controllers/failure_policy.go
GiuseppeTT: It's the same pattern used by the JobSet controller.

https://github.com/kubernetes-sigs/jobset/blob/9f40c7cdb53c3e6107d82962b06bb4967458bad7/pkg/controllers/jobset_controller.go#L358

It doesn't panic.

image

I'll improve it, thoughl

GiuseppeTT: Done.

pkg/controllers/jobset_controller.go
GiuseppeTT: Done. See https://github.com/kubernetes-sigs/jobset/pull/1177#discussion_r2898018686

pkg/controllers/failure_policy.go
GiuseppeTT: Done. I improved the error message.

What I meant is that jobIndex should be checked for out of bound.

pkg/controllers/failure_policy.go
GiuseppeTT: Done. See https://github.com/kubernetes-sigs/jobset/pull/1177#discussion_r2898046198

pkg/webhooks/jobset_webhook_test.go
GiuseppeTT: Ok. This was a very silly mistake from me 🤦

GiuseppeTT: Done.

pkg/controllers/jobset_controller.go
GiuseppeTT: > You could just feed line 373 into log.Error(err, "").

The `err` variable is also returned 3 lines later.

> And why would we add a value into an array if there is an error?

Any job in `ownedJobs.previous` gets later deleted by the JobSet controller. This is part of the self healing process because the Job is recreated with the right labels in function `labelAndAnnotateObject`. This is important in the case that someone accidentally deletes one of these labels.

test/integration/controller/jobset_controller_test.go
GiuseppeTT: Done.

GiuseppeTT:
@kannon92 finished addressing the comments. Let me know if there is anything else.

kannon92:
@GiuseppeTT 
Thinking more on this I think this change is large enough to warrant a feature gate. 
Can you please introduce a feature gate here?

I know we haven't done it for FailurePolicy before but I want us to encourage feature gates with graduation criteria to best protect production users.

And I'm not sure where the ask was but I thought I saw @imreddy13 asking if I was really set on strings as API.

I reached out to some api-review folks to see if they can check my assumptions. I'll let you know what they say.

I also would like to see an additional review from @andreyvelich or @imreddy13 before I approve.

imreddy13:
> And I'm not sure where the ask was but I thought I saw @imreddy13 asking if I was really set on strings as API.
> I reached out to some api-review folks to see if they can check my assumptions. I'll let you know what they say.
> Thanks! Let us know. I don't have a strong opinion here.

> I also would like to see an additional review from @andreyvelich or @imreddy13 before I approve.
LGTM

GiuseppeTT:
> Can you please introduce a feature gate here?

Synced with Anu offline. I am adding it.

GiuseppeTT:
> Can you please introduce a feature gate here?

Done.

> I also would like to see an additional review from andreyvelich or imreddy13 before I approve.

Done.

@kannon92

GiuseppeTT:
@kannon92 are there any pending feedback? : )

kannon92:
Hi. @everettraven has agreed to take a look from api perspective.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -201,14 +201,24 @@ type JobSetStatus struct {
...
 	// +optional
 	Restarts int32 `json:"restarts"`
 
-	// restartsCountTowardsMax tracks the number of times the JobSet has restarted that counts towards the maximum allowed number of restarts.
+	// restartsCountTowardsMax tracks the number of times the JobSet has globally restarted that counts towards the maximum allowed number of restarts (i.e. recreated all Jobs due to a restart action such as RestartJobSet).
 	// +optional
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// totalRestarts tracks the number of times the JobSet has restarted in any way (e.g., this also counts restart actions such as RestartJob).
+	// Nil should be treated as `jobSet.status.restarts`
```
kannon92: This seems a bit informal for API docs.


```suggestion
	// This will be set to Restarts if not specified.
```

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -262,6 +272,20 @@ type ReplicatedJobStatus struct {
...
 	Suspended int32 `json:"suspended"`
+
+	// jobRestarts tracks the number of times the Job has individually restarted for each job index.
+	// It is encoded as `<restarts of job 0>,...,<restarts of job replicas - 1>`
+	// Max length is set to 32KB (32768 bytes). This is enough to handle 2 978 replicas per replicatedJob
+	// +kubebuilder:validation:MaxLength=32768
+	// +optional
+	JobRestarts *string `json:"jobRestarts,omitempty"`
+
+	// jobRestartsCountTowardsMax tracks the number of times the Job has individually restarted that counts towards the maximum allowed number of restarts for each job index.
```
kannon92, line 276: I still don't really know what this is actually trying to say.

And your comments make it sound like we are storing the same data.

ahg-g:
> Hi. @everettraven has agreed to take a look from api perspective.

Thanks, but what is the sticking point? Didn't we discuss the API on the KEP?

kannon92:
@imreddy13 wanted to know if we could continue to use a map. https://github.com/kubernetes-sigs/jobset/pull/1177/changes#r2893080474

api/jobset/v1alpha2/jobset_types.go
ahg-g: A string is the more efficient representation here, this is what we did for job and I think is the right approach here as well.

api/jobset/v1alpha2/jobset_types.go
andreyvelich: @GiuseppeTT @kannon92 Can we have this as a tracking issue for JobSet v1 graduation: https://github.com/kubernetes-sigs/jobset/pull/1177/changes#r2893243169?
Ref: https://github.com/kubernetes-sigs/jobset/issues/380

api/jobset/v1alpha2/jobset_types.go
andreyvelich: I agree.

api/jobset/v1alpha2/jobset_types.go
andreyvelich: @GiuseppeTT One question about: `TotalRestarts`.
Why users can't calculate this value as sum of all jobRestarts from each:
```
jobSet.status.replicatedJob[replicatedJobName].jobRestarts[jobIndex]
```

pkg/webhooks/jobset_webhook.go
```diff
@@ -342,12 +347,33 @@ const (
...
+	hasRestartJob := false
+	for _, rule := range failurePolicy.Rules {
+		if rule.Action == jobset.RestartJob || rule.Action == jobset.RestartJobAndIgnoreMaxRestarts {
+			hasRestartJob = true
+			break
+		}
+	}
+	if hasRestartJob {
+		if !features.Enabled(features.RestartJob) {
+			allErrs = append(allErrs, fmt.Errorf("RestartJob and RestartJobAndIgnoreMaxRestarts failure policy actions are not allowed when JobLevelRestart feature gate is disabled"))
```
kannon92: nit: should this just return?

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1752,6 +1753,111 @@ func TestValidateCreate(t *testing.T) {
...
+					},
+					SuccessPolicy: &jobset.SuccessPolicy{},
+				},
+			},
+			want: errors.Join(
+				fmt.Errorf("JobSet cannot have a failure policy rule with RestartJob or RestartJobAndIgnoreMaxRestarts action and a replicated job with replicas > 2978"),
+			),
+		},
+		{
+			name:             "jobset failure policy rule has RestartJob action and JobLevelRestart feature gate is disabled",
```
kannon92: Can you add RestartJobAndIgnoreMaxRestarts test case also?

test/e2e/e2e_test.go
```diff
@@ -629,6 +630,32 @@ var _ = ginkgo.Describe("JobSet", func() {
 
 		})
 	})
+
+	ginkgo.When("JobSet uses RestartJobSet failure policy action", func() {
```
kannon92, line 634: I'd like to try and cover the user stories in e2e tests as much as possible.

Can we add test cases covering https://github.com/kubernetes-sigs/jobset/tree/main/keps/262-ConfigurableFailurePolicy#story-6-restartjob-and-restartjobandignoremaxrestarts?

We should have RestartJ

kannon92, line 634: RestartJobSet isn't the policy you added in this code though. Can you also add test cases for those that you added?

pkg/controllers/failure_policy.go
```diff
@@ -337,3 +466,97 @@ func findFirstFailedJob(failedJobs []*batchv1.Job) *batchv1.Job {
...
+// decodeJobRestarts parses the given comma-separated string pointer into a slice of integers.
+// If the string is nil or empty, it returns a slice of 0s of the specified length.
+func decodeJobRestarts(restartsStr *string, length int32) []int32 {
+	restarts := make([]int32, length)
+	if restartsStr == nil || *restartsStr == "" {
+		return restarts
+	}
+
+	parts := strings.Split(*restartsStr, ",")
+	for i, p := range parts {
```
andreyvelich: We don't really need another var here.
```suggestion
	for i, p := range strings.Split(*restartsStr, ",") {
```

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -262,6 +272,20 @@ type ReplicatedJobStatus struct {
 
 	// suspended is the number of child Jobs which are in a suspended state.
 	Suspended int32 `json:"suspended"`
+
+	// jobRestarts tracks the number of times the Job has individually restarted for each job index.
+	// It is encoded as `<restarts of job 0>,...,<restarts of job replicas - 1>`
+	// Max length is set to 32KB (32768 bytes). This is enough to handle 2 978 replicas per replicatedJob
+	// +kubebuilder:validation:MaxLength=32768
```
andreyvelich: Shall we have minLength=1?

pkg/controllers/failure_policy.go
```diff
@@ -337,3 +466,97 @@ func findFirstFailedJob(failedJobs []*batchv1.Job) *batchv1.Job {
...
+func getJobRestartsCountTowardsMax(js *jobset.JobSet, replicatedJobName string) []int32 {
+	for _, rjs := range js.Status.ReplicatedJobsStatus {
+		if rjs.Name == replicatedJobName {
+			jobRestarts := rjs.JobRestartsCountTowardsMax
+			replicas := getReplicatedJobReplicas(js, replicatedJobName)
+			return decodeJobRestarts(jobRestarts, replicas)
+		}
+	}
+	replicas := getReplicatedJobReplicas(js, replicatedJobName)
+	return decodeJobRestarts(nil, replicas)
```
andreyvelich: Passing the nil to this helper function seems strange, can we just return empty map here?

test/e2e/e2e_test.go
GiuseppeTT: I had to remove the test case because it was not compatible with the feature flag (it is not as easy as `features.SetFeatureGateDuringTest(t, features.RestartJob, true)` used in the unit tests and integration tests). Notice that **none** of our e2e tests use feature flags. Unless there is a clear way to do that, can we skip this given the timeline?

GiuseppeTT: @kannon92 see https://github.com/kubernetes-sigs/jobset/pull/1177#discussion_r2912864438

pkg/controllers/failure_policy.go
GiuseppeTT: Using `nil` here is the correct approach. Notice that the first argument of `decodeJobRestarts` is a pointer to a string, not a map. This case is only done if `js.Status.ReplicatedJobsStatus` hasn't been populated yet, which means that the job restarts should be an array of zeroes.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: I mean, empty string is also a valid value and is treated as the starting case (array of zero restarts for all Jobs).

pkg/webhooks/jobset_webhook.go
GiuseppeTT: I mean, isn't the point of `allErrs` in `validateFailurePolicy` to gather all violations instead of returning on the first violation found?

test/e2e/e2e_test.go
kannon92: I think e2e testing of alpha features is important.

We could consider setting this feature gate in the e2e for our e2e.

ahg-g: In k8s, at least in the past, we mandated e2e on beta graduation when the feature is enabled by default, integration tests are sufficient for alpha

kannon92: True. I guess that makes sense.

I would like us to have alpha e2e testing though. We are pushing back on people writing code in k8s without e2e testing. Even if the test is not in periodic, at lease for sig-node, we ask that they have e2e tests and a presubmit that they can verify their change.

We see a lot of PRs merge that get tested for the first time in the CI and then we revert and ask that they actually verify e2e.

I'm fine to let this one go but Kueue has some patterns to follow where we can create a dedicated lane for running custom configs to verify features e2e.

pkg/controllers/failure_policy.go
andreyvelich: I see, thanks for the clarification.

pkg/controllers/failure_policy.go
GiuseppeTT: Done.

pkg/webhooks/jobset_webhook_test.go
GiuseppeTT: My bad. Done

test/e2e/e2e_test.go
ahg-g: Thanks, yes, we can block beta graduation on it.

pkg/webhooks/jobset_webhook.go
kannon92: At least for k8s if feature gate is disabled we would unset values. 

I think its fine but the true error is that the feature is off. The other stuff would no longer be used.

But this is not blocking so I'll leave it up to you to address.

test/e2e/e2e_test.go
GiuseppeTT: I created an issue to track the support and addition of alpha e2e tests to JobSet. https://github.com/kubernetes-sigs/jobset/issues/1184

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: I improved the comments. It now explains the relationship of these fields to the new restart actions `RestartJob` and `RestartJobAndIgnoreMaxRestarts`

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: Done.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: > Why users can't calculate this value as sum of all jobRestarts from each:

I think this is technically possible. That is, `totalRestarts = restarts + sum(jobRestarts)`. So this is a matter of redundancy vs have an easy field to get the total number of restarts. What do you say @kannon92 ?

andreyvelich: I would suggest avoiding adding additional API fields if that is not absolutely necessary.

pkg/webhooks/jobset_webhook.go
GiuseppeTT: I updated it to early return. Done.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: I see. I am not completely against removing these fields, but I do see a value in observability. That is, an easy way to get how many times a restart action has been triggered. This is similar to the fields `Succeeded` and `CompletedIndexes` from the Job API.

pkg/webhooks/jobset_webhook.go
kannon92: I don't see this here.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -634,6 +634,16 @@ type JobSetStatus struct {
   // +optional
   RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// totalRestarts tracks the number of times the JobSet has restarted in any way (e.g., this also counts restart actions such as RestartJob).
+	// Nil should be treated as `jobSet.status.restarts`
```
kannon92: The API here doesn't match the implementation. Can be done as a follow up.

andreyvelich:
Overall, looks good, my only suggestion is about removing `totalRestarts` for now: https://github.com/kubernetes-sigs/jobset/pull/1177#discussion_r2913417394
If we get strong user signal that we really need to have this API for observability, we can add it later.
WDYT @GiuseppeTT @kannon92 ?

pkg/webhooks/jobset_webhook.go
GiuseppeTT: My bad. I forgot to push the commit 🤦

GiuseppeTT:
> Overall, looks good, my only suggestion is about removing `totalRestarts` for now: [#1177 (comment)](https://github.com/kubernetes-sigs/jobset/pull/1177#discussion_r2913417394) If we get strong user signal that we really need to have this API for observability, we can add it later. WDYT @GiuseppeTT @kannon92 ?

Done. I removed the total restart fields. Let's wait for more user feedback before adding it again.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: I removed them.

keps/262-ConfigurableFailurePolicy/README.md
GiuseppeTT: I removed the total restarts fields.

GiuseppeTT:
@kannon92 @andreyvelich let me know what you think!

andreyvelich:
Thanks @GiuseppeTT!
/lgtm
/assign @kannon92 @imreddy13 for final approval.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -262,6 +264,23 @@ type ReplicatedJobStatus struct {
 
 	// suspended is the number of child Jobs which are in a suspended state.
 	Suspended int32 `json:"suspended"`
+
+	// jobRestarts tracks the number of times the Job has individually restarted for each job index.
+	// That is, for each job index, it tracks how many times the restart actions RestartJob and RestartJobAndIgnoreMaxRestarts
+	// were executed for that job index.
+	// It is encoded as `<restarts of job 0>,...,<restarts of job replicas - 1>`
+	// Max length is set to 32KB (32768 bytes). This is enough to handle 2 978 replicas per replicatedJob
+	// +kubebuilder:validation:MaxLength=32768
```
everettraven: Just a note, because this API is a CRD the schema is an OpenAPI v3 schema where `maxLength` is character-based, not byte based.

This means that without enforcing a specific format that constrains the character set of allowed characters, someone could specify unicode characters that are up to 4 bytes.

If you truly want this to be closer to an explicit byte length, make sure the allowed input is only single byte characters (ASCII).

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -406,7 +432,7 @@ type FailurePolicyRule struct {
 	// The name must match the regular expression "^[A-Za-z]([A-Za-z0-9_,:]*[A-Za-z0-9_])?$".
 	Name string `json:"name"`
 	// action to take if the rule is matched.
-	// +kubebuilder:validation:Enum:=FailJobSet;RestartJobSet;RestartJobSetAndIgnoreMaxRestarts
+	// +kubebuilder:validation:Enum:=FailJobSet;RestartJobSet;RestartJobSetAndIgnoreMaxRestarts;RestartJob;RestartJobAndIgnoreMaxRestarts
```
everettraven, line 433: Just a note, users using something like `kubectl explain` to explore your API surface never see the kubebuilder markers.

This means the only information they end up seeing for this field is:
> action to take if the rule is matched.

I normally encourage being more verbose so that generated documentation for API fields that may be read by end-users gives them a better understanding of what values are valid for a field and why they may want to set those values.

For example, updating this description to read along the lines of:
> action is a required field that specifies which action should be taken when this failure rule is matched.
> Allowed values are FailJobSet, RestartJobSet, RestartJobSetAndIgnoreMaxRestarts, RestartJob, and RestartJobAndIgnoreMaxRestarts.
> When set to FailJobSet, {more details on what this means here}
> When set to ....

Would, IMO, be a good improvement to the documentation for end-users to better understand the purpose of this field, when they should set it, what values can be set, and why they may want to select a specific value (based on understanding of the behaviors for each value).

api/jobset/v1alpha2/jobset_types.go
everettraven: FWIW, I agree that the map approach is a significantly better representation from a UX standpoint.

If there are concerns around the number of jobs you need to scale to being unbounded - why do you need to support a virtually unbounded number of jobs being defined in the jobset? Would it make sense to update the validations in place such that a reasonable limit is enforced in the number of jobs that can be specified in a single jobset resource?

Would a limit on the number of jobs that can be specified for a single jobset effectively resolve concerns around the scale problem of making this a map?

Another alternative to a pure string - have you considered using a `[]int32` here? Instead of something like:
```yaml
jobRestarts: "123,456,789,..."
```
you would end up with:
```yaml
jobRestarts:
  - 123
  - 456
  - 789
```

You increase the size in bytes by a factor of ~1.57 by taking the `[]int32` approach (string based example is ~26 bytes, the int slice string representation is ~41 bytes), but it is significantly easier to read and automate against because as an end-user I don't have to do any string manipulation to parse out a value I might care about.

GiuseppeTT: Thanks for the review Bryce.

I like the idea of using `[]int32`. Let's start with a limit of 1024 Jobs that is only enforced if the JobSet contains `RestartJob` (I will create a new commit with this solution and scale test it). Since the feature will be released as alpha, it will give us time to collect more user feedback on the size of JobSets.

What do you think @kannon92 @andreyvelich @ahg-g @ahg-g ?

andreyvelich: Yeah, I like the idea of array instead of string. However, it makes API inconsistent with the `CompletedIndexes`  Job API. @tenzen-y @ahg-g @kannon92 Do you know why did we use `string` instead of `[]int32` for Job in the first place?

kannon92: That was before my time.

IndexedJobs support up to 10000 parallism so I think that was probably why it was a string. But I wasn't around for the review of that implementation. Maybe @soltysh has an idea here?

GiuseppeTT: My guess is that:

- (1) The Job API has a bigger and more diverse user base than JobSet
- (2) The Job API handles Pods which are expected to scale more with the infra size than Jobs 

So that the Job API is interested in extracting every single byte out of the API

GiuseppeTT: I finished the scale tests. The feature worked with 1024 Jobs. Moreover, the new fields `jobRestarts` and `jobRestartsCountTowardsMax` only show up in `jobSet.status.replicatedJobStatus[]` if the replicatedJob went through a `RestartJob` restart.

<details>

<summary>JobSet yaml for reference</summary>

```yaml
kind: JobSet
metadata:
  name: js
spec:
  failurePolicy:
    maxRestarts: 2
    rules:
    # If any Job from replicatedJob `rj-leader` fails, restart the JobSet by recreating ALL Jobs
    # Count the restart towards `jobSet.spec.failurePolicy.maxRestarts`
    - name: r_leader
      action: RestartJobSet
      targetReplicatedJobs:
      - rj-leader
    # If any Job from replicatedJob `rj-worker` fails, restart the JobSet by recreating only the failed Job
    # Count the restart towards `jobSet.spec.failurePolicy.maxRestarts`
    - name: r_worker
      action: RestartJob
      targetReplicatedJobs:
      - rj-worker
  replicatedJobs:
  - name: rj-leader
    replicas: 1
    template:
      spec:
        parallelism: 2
        completions: 2
        backoffLimit: 0 # Job fails if any Pod fails
        template:
          spec:
            restartPolicy: Never
            terminationGracePeriodSeconds: 0
            containers:
            - name: leader
              image: busybox:1.37
              command: ["/bin/sh", "-c"]
              args:
              - |
                echo "Start"
                while true; do
                  if [ -f /tmp/fail ]; then
                    echo "Exiting 1"
                    exit 1
                  elif [ -f /tmp/succeed ]; then
                    echo "Exiting 0"
                    exit 0
                  fi
                  sleep 1
                done
  - name: rj-worker
    replicas: 1024
    template:
      spec:
        parallelism: 2
        completions: 2
        backoffLimit: 0 # Job fails if any Pod fails
        template:
          spec:
            restartPolicy: Never
            terminationGracePeriodSeconds: 0
            containers:
            - name: worker
              image: busybox:1.37
              command: ["/bin/sh", "-c"]
              args:
              - |
                echo "Start"
                while true; do
                  if [ -f /tmp/fail ]; then
                    echo "Exiting 1"
                    exit 1
                  elif [ -f /tmp/succeed ]; then
                    echo "Exiting 0"
                    exit 0
                  fi
                  sleep 1
                done
```

</details>

GiuseppeTT:
@kannon92 @andreyvelich 

I synced with @alculquicondor who created the field `CompletedIndexes` in https://github.com/kubernetes/enhancements/pull/2245

He explained that he used a string because he could use the range trick (e.g., `0,1,2,5` = `0-2,5`) to reduce the size as much as possible.

I don't think it is applicable to our case in JobSet because:

- (1) The range trick does not work for us. We are counting restarts for each Job (so each Job has an int32, which requires an array-like representation). He was counting Pods that completed (so each Pod is completed or not, which can use a set representation with the range trick)
- (2) The range trick does not solve the scalability problem, it only mitigates the problem. That is because worst case scenario you can have alternating indexes like `0,2,4,6,...` which cannot be reduced

Having that in mind, are there any remaining concerns for this PR?

api/jobset/v1alpha2/jobset_types.go
alculquicondor: I'm not here to provide an answer, but just context:

`completedIndexes` can collapse as index terminate. So, it can go from `1,3,5` to `1-5` once jobs 2 and 4 complete. I don't think the same can be said for this.

But having an array of mostly zeros sounds like a waste. Could the information be stored in a job annotation (as a job is recreated, the next job gets a prev_job+1 in the annotation)? But if precision is necessary and you can't deal with the possible race conditions, then I suppose an array is an ok option, if the number of jobs is generally a ~small number

pkg/webhooks/jobset_webhook.go
```diff
@@ -342,12 +347,34 @@ const (
...
```
kannon92:
```suggestion
			allErrs = append(allErrs, fmt.Errorf("RestartJob and RestartJobAndIgnoreMaxRestarts failure policy actions are not allowed when RestartJob feature gate is disabled"))
```

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1752,6 +1753,146 @@ func TestValidateCreate(t *testing.T) {
...
+					},
+					SuccessPolicy: &jobset.SuccessPolicy{},
+				},
+			},
+			want: errors.Join(
+				fmt.Errorf("JobSet cannot have a failure policy rule with RestartJob or RestartJobAndIgnoreMaxRestarts action and a replicated job with replicas > 1024"),
+			),
+		},
+		{
+			name:             "jobset failure policy rule has RestartJob action and JobLevelRestart feature gate is disabled",
```
kannon92:
```suggestion
			name:             "jobset failure policy rule has RestartJob action and RestartJob feature gate is disabled",
```

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -1752,6 +1753,146 @@ func TestValidateCreate(t *testing.T) {
...
+					},
+					SuccessPolicy: &jobset.SuccessPolicy{},
+				},
+			},
+			want: errors.Join(
+				fmt.Errorf("RestartJob and RestartJobAndIgnoreMaxRestarts failure policy actions are not allowed when JobLevelRestart feature gate is disabled"),
+			),
+		},
+		{
+			name:             "jobset failure policy rule has RestartJobAndIgnoreMaxRestarts action and JobLevelRestart feature gate is disabled",
```
kannon92: Seems like its misnamed everywhere in this file. Going to stop comment on them but please fix that.

api/jobset/v1alpha2/jobset_types.go
kannon92: The API changes for this still need this.

pkg/controllers/failure_policy.go
```diff
@@ -256,6 +259,97 @@ var restartJobSetAndIgnoreMaxRestartsActionApplier failurePolicyActionApplier =
...
+	// Emit event for each Job restarts for observability and debugability.
+	enqueueEvent(updateStatusOpts, event)
+	log.V(2).Info("attempting restart for job", "job", matchingFailedJob.Name)
+	return nil
+}
+
+// restartJobActionApplier applies the RestartJob FailurePolicyAction
+var restartJobActionApplier failurePolicyActionApplier = func(ctx context.Context, js *jobset.JobSet, matchingFailedJob *batchv1.Job, updateStatusOpts *statusUpdateOpts) error {
+	if !features.Enabled(features.RestartJob) {
+		return fmt.Errorf("RestartJob failure policy action cannot be used when the feature gate is disabled")
```
kannon92:
```suggestion
		return fmt.Errorf("RestartJob failure policy action cannot be used when the feature gate RestartJob is disabled")
```

pkg/webhooks/jobset_webhook.go
GiuseppeTT: My bad. I renamed the feature gate and forgot to update the messages. 🤦 

I fixed it.

pkg/controllers/failure_policy.go
GiuseppeTT: Done.

api/jobset/v1alpha2/jobset_types.go
kannon92: ty! this reads better.

keps/262-ConfigurableFailurePolicy/README.md
```diff
@@ -664,27 +664,29 @@ type ReplicatedJobStatus struct {
...
-  // Example: "0,1,0" implies that job 1 has restarted once, while jobs 0 and 2 haven't restarted.
-  // +optional
-  JobRestarts *string `json:"jobRestarts,omitempty"`
-
-  // jobRestartsCountTowardsMax tracks the number of times each Job has restarted that counts towards the maximum allowed number of restarts.
-  // The first value corresponds to the job with index 0, the second value corresponds to the job with index 1, etc.
-  // Example: "0,1,0" implies that job 1 has restarted once, while jobs 0 and 2 haven't restarted.
-  // +optional
-  JobRestartsCountTowardsMax *string `json:"jobRestartsCountTowardsMax,omitempty"`
+	// jobRestarts tracks the number of times the Job has individually restarted for each job index.
```
kannon92, line 667: not blocking.

After we merge let's sync this with the actual implementation.

kannon92: Or honestly we can remove this update and do it after the merge. I'm not sure how out dated the KEP is with our latest updates.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -2947,6 +3088,7 @@ func TestValidateCreate(t *testing.T) {
 				t.Fatalf("error creating jobset webhook: %v", err)
 			}
 			features.SetFeatureGateDuringTest(t, features.InPlaceRestart, tc.enableInPlaceRestart)
```
kannon92, line 3090: This does remind me that we should have some testing / scenarios for InPlaceRestart and make sure these features all work together.

Not blocking for this though.

pkg/controllers/jobset_controller_test.go
```diff
@@ -897,10 +898,51 @@ func TestConstructJobsFromTemplate(t *testing.T) {
...
+					jobIdx:            1,
+					jobRestartAttempt: 0}).
+					Suspend(false).Obj(),
+			},
+		},
 	}
 
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
+			features.SetFeatureGateDuringTest(t, features.RestartJob, true)
```
kannon92, line 945: Can we have a test case where this feature is disabled?

pkg/features/features.go
```diff
@@ -38,6 +38,14 @@ const (
 	//
 	// Enables TLSOptions for TLSMinVersion and CipherSuites for JobSet servers.
 	TLSOptions featuregate.Feature = "TLSOptions"
+
+	// owner: @giuseppett
```
kannon92, line 42:
```suggestion
	// owner: @GiuseppeTT
```

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -262,6 +264,21 @@ type ReplicatedJobStatus struct {
 
 	// suspended is the number of child Jobs which are in a suspended state.
 	Suspended int32 `json:"suspended"`
+
+	// jobRestarts tracks the number of times the Job has individually restarted for each job index.
+	// That is, for each job index, it tracks how many times the restart actions RestartJob and RestartJobAndIgnoreMaxRestarts
+	// were executed for that job index.
+	// +kubebuilder:validation:MaxItems=1024
```
kannon92, line 271: Given that this puts a limit on the Replicas this should be a user facing on the docs.

api/jobset/v1alpha2/jobset_types.go
GiuseppeTT: Thanks for the input!

> Could the information be stored in a job annotation (as a job is recreated, the next job gets a prev_job+1 in the annotation)?

I considered this but there are drawbacks:

- The old Job and the replacement Job have to exist at the same time since the JobSet cannot hold the information in the JobSet object. This could be a problem. Even if we suspend the Jobs, it could be a problem since other components like Kueue rely on suspended Jobs
- If a Job is deleted by something else, the JobSet controller loses the information on how many times it has restarted. This is a problem because JobSet should compare this restart count with `jobSet.spec.failurePolicy.maxRestarts` to decide to fail if it has restarted too many times

andreyvelich:
Thank you for this @GiuseppeTT! Excited to see this moving forward.
/lgtm
/approve

GiuseppeTT:
I am lifting the hold since the PR has been LGTMed by [Kevin](https://github.com/kubernetes-sigs/jobset/pull/1177#pullrequestreview-3932499868) and [Andrey](https://github.com/kubernetes-sigs/jobset/pull/1177#issuecomment-4035005192).

I'll create a follow up PR to address remaining feedbacks.

Thank you all for the review!

/hold cancel

PR #1178, opened 2026-03-05
Add KEP to support granular suspend/resume for individual Jobs within a JobSet

### LINKED ISSUE #1172: Support granular suspend/resume for individual Jobs within a JobSet

Introduce fine-grained suspend and resume capabilities at the individual Job level within a JobSet. Specifically, we need the ability to suspend and resume specific child Jobs independently, without altering the execution state of other actively running Jobs within the same JobSet.

**Why is this needed**:

This feature is necessary to support advanced scheduling and placement optimizations in Kueue, particularly around topology awareness.

If the underlying topology of a specific Job's placement becomes fragmented or broken, Kueue needs the ability to independently relocate (recreate) that single Job. Currently, lacking granular control, adjusting the placement of one Job disrupts the others. By enabling individual Job suspension, Kueue will be able to suspend a specific Job, safely recreate it in a new topological location, and resume it—all without affecting the uninterrupted execution of the other sibling Jobs in the JobSet.

- [ ] Design doc
- [ ] API change
- [ ] Docs update

imreddy13:
#### What type of PR is this?
/kind feature

This KEP introduces fine-grained suspend and resume capabilities at the individual Job level within a JobSet. This allows specific child Jobs to be independently suspended and resumed without altering the execution state of other actively running Jobs within the same JobSet.

This feature is necessary to support advanced scheduling and placement optimizations in Kueue, particularly around topology awareness. 

If the underlying topology of a specific Job's placement becomes fragmented or broken, Kueue needs the ability to independently relocate (recreate) that single Job. Currently, lacking granular control, adjusting the placement of one Job disrupts the others in the JobSet. By enabling individual Job suspension, Kueue will be able to suspend a specific Job, safely recreate it in a new topological location, and resume it, all without affecting the uninterrupted execution of the other sibling Jobs in the JobSet.

Fixes https://github.com/kubernetes-sigs/jobset/issues/1172

Yes.
```release-note
Action required to use this feature: add the annotation `jobset.sigs.k8s.io/reconciliation-mode: Independent` to the JobSet spec.
```


GiuseppeTT:
@mwielgus @mwysokin @mimowo can you take a look for the Kueue side?

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This KEP introduces fine-grained suspend and resume capabilities at the individual Job level within a JobSet. This allows specific child Jobs to be independently suspended and resumed without altering the execution state of other actively running Jobs within the same JobSet.
+
+## Motivation
+
+This feature is necessary to support advanced scheduling and placement optimizations in Kueue, particularly around topology awareness.
```
GiuseppeTT: I would add that JobSet started with the assumption that if one Job is being changed (recreated, suspended, etc), then all Jobs should go through the same process. But now there are use cases that can handle and require only a single Job being changed without affecting the rest.

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+None currently identified.
+
+### Risks and Mitigations
+
+The primary risk is introducing complexity to the JobSet controller, as it now has to manage the suspension state of individual jobs rather than just propagating the JobSet's suspension down to all jobs uniformly. This will be mitigated by thorough unit and integration testing.
+
+## Design Details
+
+Initially we will support this via an annotation on the JobSet: `jobset.sigs.k8s.io/skip-suspend-reconciliation`. If this annotation is set, then the JobSet webhook enforces `jobSet.spec.suspend = nil and does not reconcile the field job.spec.suspend of child Jobs.
```
GiuseppeTT: nit: missing "`" in "jobSet.spec.suspend = nil"

keps/1172-JobSuspendResume/kep.yaml
```diff
@@ -0,0 +1,28 @@
+title: Support granular suspend/resume for individual Jobs within a JobSet
+kep-number: 1172
+authors:
+  - "@imreddy"
+status: implementable
+creation-date: 2026-03-05
+reviewers:
+  - - "@kannon92"
```
kannon92:
```suggestion
  - "@kannon92"
```

keps/1172-JobSuspendResume/kep.yaml
```diff
@@ -0,0 +1,28 @@
...
+approvers:
+  - - "@ahg"
```
kannon92:
```suggestion
  - "@ahg-g"
```

Probably worth reaching out to @ahg-g if he has cycles for this.

keps/1172-JobSuspendResume/kep.yaml
```diff
@@ -0,0 +1,28 @@
...
+replaces:
+  - "NA"
+
+# The target maturity stage in the current dev cycle for this KEP.
+stage: alpha
+
+# The most recent milestone for which work toward delivery of this KEP has been
+# done. This can be the current (upcoming) milestone, if it is being actively
+# worked on.
+latest-milestone: "v0.11.2"
```
kannon92:
```suggestion
latest-milestone: "v0.12.0"
```

keps/1172-JobSuspendResume/kep.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+# The milestone at which this feature was, or is targeted to be, at each stage.
+milestone:
+  alpha: "v0.11.2"
```
kannon92:
```suggestion
  alpha: "v0.12.0"
```

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+N/A
+
+## Implementation History
+
+- 2026-02-27: Initial issue opened requesting the feature.
+
+## Drawbacks
+
+Increases the API surface and complexity of the JobSet controller, Job suspension is now managed by an external controller (e.g. Kueue).
```
kannon92: Job suspension has always been managed by an external entity (user or Kueue)

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+- Changing the underlying Kubernetes Job behavior itself beyond what JobSet controls.
+
+## Proposal
+
+We propose adding a mechanism (e.g., via an API field or annotation in the JobSet) to individually target specific Jobs within a JobSet and control their suspension state without changing the overall JobSet suspension state.
+
+### User Stories (Optional)
+
+#### Story 1
+As a user utilizing Kueue for topology-aware scheduling, I want to relocate a single fragmented Job to a better topological location without killing or pausing the other jobs in my JobSet, so that large-scale parallel processing is more resilient to location constraints.
```
kannon92, line 55: So is Kueue setting a different NodeSelector in addition to resuming the individual job?

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+1. Create a JobSet with the `jobset.sigs.k8s.io/skip-suspend-reconciliation` annotation and verify the webhook enforces `jobSet.spec.suspend = nil`.
+2. Create a JobSet with the annotation, wait for child Jobs to be created, then manually toggle `suspend: true` and `suspend: false` on specific child Jobs via Kueue. Verify the JobSet controller does not revert these changes.
+3. (Backwards Compatibility) Create a standard JobSet without the annotation. Toggle the JobSet's `suspend` field and verify all child Jobs are correctly suspended or resumed, ensuring existing behavior is unbroken.
+
+### Graduation Criteria
+
+N/A
+
+## Implementation History
```
kannon92, line 179: IMO We should have an alternative.

Is it possible to support this via https://github.com/kubernetes-sigs/kueue/tree/main/pkg/controller/jobs/jobset or adding this into a proper Job interface for Kueue?

I'm suspect that this wouldn't be useful for LWS or RayJobs/RayClusters

GiuseppeTT:
The test failed because the toc of another KEP is not up to date. https://github.com/kubernetes-sigs/jobset/pull/1177 should fix it shortly.

keps/1172-JobSuspendResume/README.md
GiuseppeTT: I think what she meant is that the suspension management is currently done as "kueue -> jobset -> job" whereas this change will allow "kueue -> job"

keps/1172-JobSuspendResume/README.md
GiuseppeTT: Yes.

cc @mwielgus @mwysokin @mimowo

keps/1172-JobSuspendResume/README.md
GiuseppeTT: @mwielgus @mwysokin @mimowo

keps/1172-JobSuspendResume/README.md
kannon92: not quite.

Kueue would still handle flow: kueue -> jobset -> job.

But then somehow kueue will trigger suspends/resumes of jobs while the JobSet is resumed.

I think I get the meaning.

GiuseppeTT:
Please rebase to get this [fix](https://github.com/kubernetes-sigs/jobset/pull/1180)

keps/1172-JobSuspendResume/README.md
kannon92: Also cc @andreyvelich 

for TrainJob.

kannon92:
@imreddy13 This proposal was well done but I feel like I am missing the big picture on this.

We can add this feature into JobSet but I don't really understand the kueue work to force these restarts. Would that be done after we have support for this in our API?

And we would need to consider other workloads for supporting this as I know that relocating failing jobs would be useful for any multi-node workload.

GiuseppeTT:
> @imreddy13 This proposal was well done but I feel like I am missing the big picture on this.

@mwielgus @mwysokin @mimowo for context on Kueue.

keps/1172-JobSuspendResume/README.md
mwielgus: Most likely node selector defined at the job level won't change (unless a different Resource Flavor is selected). Topology Aware Scheduling in Kueue operates at the Pod level.

mwielgus:
> > @imreddy13 This proposal was well done but I feel like I am missing the big picture on this.
> 
> @mwielgus @mwysokin @mimowo for context on Kueue.

On the Kueue side, we will be creating a Workload per Jobset Job, not for the whole JobSet. As a result Jobset jobs will be admitted somewhat independently, with an option for synchronized start/admission. The appropriate KEPs on the Kueue side are in the works.

keps/1172-JobSuspendResume/README.md
ahg-g: This API (the annotation) is expressed in terms of the JobSet controller implementation details. The API should express the user intent of how they want a JobSet to behave, not how the controller should execute on the behavior.

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+If this annotation is set:
+1. The JobSet webhook will enforce `jobSet.spec.suspend = nil`.
+2. The JobSet controller will not reconcile the `job.spec.suspend` field of child Jobs.
+
+Based on usage & feedback, we will evaluate if we should modify the JobSet API to control this behavior.
+
+We cannot use `jobSet.spec.suspend = nil` as a signal to not reconcile the field `job.spec.suspend` of child Jobs becuase the JobSet controller currently treats `jobSet.spec.suspend = nil` as a synonym for `jobSet.spec.suspend = false`. If users are unsuspending JobSets by setting `jobSet.spec.suspend = nil` instead of `false`, this change would break their workflow. 
+
+This simple annotation-based solution gives the use case time to evolve before proposing a more complete API (such as an `updatePolicy` field in the future).
```
ahg-g, line 77: I am concerned about continuing to use annotations as temporary api, which in the exclusive placement case never graduated to a proper API on the spec.

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+
+#### Future API Evolution
+
+If the `jobset.sigs.k8s.io/skip-suspend-reconciliation` annotation proves successful, we may promote this configuration to a strongly-typed API field within the `JobSetSpec`. A dedicated policy field (e.g. `UpdatePolicy`) would cleanly move this behavior out of annotations, providing better validation, discoverability via OpenAPI schemas, and extensibility if other child Job fields need decoupled reconciliation in the future.
```
ahg-g: I am concerned about this gaining adoption and having to go through the process of migrating users to the new api.

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,145 @@
...
+    // +optional
+    UpdatePolicy *JobSetUpdatePolicy `json:"updatePolicy,omitempty"`
+}
+
+type JobSetUpdatePolicy struct {
+    // SuspendReconciliation controls whether the JobSet controller actively 
+    // reconciles the `suspend` field of child Jobs to match the JobSet.
+    // Valid values are "Sync" (default) and "Ignore".
+    // +optional
+    SuspendReconciliation *SuspendReconciliationMode `json:"suspendReconciliation,omitempty"`
```
ahg-g: ditto: https://github.com/kubernetes-sigs/jobset/pull/1178/changes#r2904939202

mwielgus:
We created a corresponding issue on the Kueue side (https://github.com/kubernetes-sigs/kueue/issues/9940). It should give more context why this is needed and how this will work in the bigger picture.

keps/1172-JobSuspendResume/README.md
kannon92: I agree. I would prefer feature gate with real API if we have real use cases and customer demand.

keps/1172-JobSuspendResume/README.md
imreddy13: @kannon92 I think an alternative is to leverage the RestartJob feature Giuseppe just built, but it doesn't work for some Kueue use cases.

@mimowo can comment on when they plan to introduce the partial admission functionality in Kueue via https://github.com/kubernetes-sigs/kueue/issues/9940

keps/1172-JobSuspendResume/README.md
imreddy13: I agree that an API is cleaner and customers would not need to be educated to move away from the annotation in the future. My main concern is that the Kueue design in the future may not be compatible with the API we introduce. So I preferred the annotation. Wdyt @ahg-g ? The Kueue issue for reference: https://github.com/kubernetes-sigs/kueue/issues/9940

keps/1172-JobSuspendResume/kep.yaml
imreddy13: Also added you!

keps/1172-JobSuspendResume/README.md
imreddy13: Added some clarification

keps/1172-JobSuspendResume/README.md
imreddy13: Does the new name (ReconcilationMode) work better? Also open to other suggestions

kannon92 (commented):
Can you comment on how this feature will work with depends on?

It may be good to also enumerate the behavior with our other policies (Failure, Success, PVC).

Some of these may be obvious but I think we should think through the interactions of this. Ideally we should have test cases too.

imreddy13:
> Can you comment on how this feature will work with depends on?
> 
> It may be good to also enumerate the behavior with our other policies (Failure, Success, PVC).
> 
> Some of these may be obvious but I think we should think through the interactions of this. Ideally we should have test cases too.

1. DependsOn
With `reconciliation-mode: Independent`, suspending an already running child Job should have no effect on the dependsOn logic. The "gate" for dependsOn is only checked at the time of Job creation or re-creation on failure. It should be up to the scheduler to choose what order to suspend/unsuspend Jobs. This feature will still respect dependsOn ordering at Job creation time.

2. FailurePolicy
With `reconciliation-mode: Independent`, suspending a Job will not trigger the failure policy action for the ReplicatedJob since the Job doesn't "fail". Jobs being suspended or unsuspended will not count towards maxRestarts.

JobSet failure policy applies to suspended jobs too i.e. if job A is suspended & job B fails, job A will also be restarted. If a Job fails and then is suspended by Kueue before the JobSet controller processes the failure, the controller will still respect the FailurePolicy.

As a follow up, we will work with @mwielgus to support failure policy actions with Kueue suspend/resume. We can also add a new field to track number of per job restarts that includes suspensions + failure policy restarts. 

3. SuccessPolicy
With `reconciliation-mode: Independent`, if the SuccessPolicy is set to All, the JobSet will never reach a Succeeded state as long as any Job is suspended.

If a SuccessPolicy targets a specific ReplicatedJob (e.g., "Succeed if the 'leader' Job completes"), suspending Jobs in other ReplicatedJobs will not prevent the JobSet from succeeding once the leader finishes.

Once a JobSet meets its SuccessPolicy, the controller deletes or cleans up all Jobs including suspended ones.

4. PVCs
With `reconciliation-mode: Independent`, when a Job is suspended, its Pods are terminated, but the Job object and its associated PVCs should persist. When the Job is resumed, the new Pods will mount the same PVCs.

Test Cases

1. Basic Functionality
- Independent Sync: Verify that when `reconciliation-mode: Independent` is set, changing `JobSet.spec.suspend` from false to true does not suspend Jobs that were already running, and vice versa.

- Granular Control: Manually suspend 1 of 3 Jobs in a ReplicatedJob and verify the other 2 continue running. Verify the JobSet status remains Active.

2. Policy Interactions with `reconciliation-mode: Independent`:
- DependsOn + Suspension: Create a JobSet where Job B depends on Job A. Suspend one replica of Job A. Verify Job B is never created. Resume the replic, and verify Job B is then created.

- FailurePolicy + Restart: Set a FailurePolicy with maxRestarts: 1. Suspend one Job, then force a failure in another Job. Verify the JobSet restarts and check if the previously suspended Job is recreated as "running" (demonstrating the state loss during restart).

- SuccessPolicy (All): Suspend one Job and let all others succeed. Verify the JobSet does not transition to Succeeded. Unsuspend the Job, let it finish, and verify the JobSet then succeeds.

- SuccessPolicy (Any): Define a policy where only the first ReplicatedJob needs to succeed. Suspend a Job in the second ReplicatedJob. Verify the JobSet still succeeds when the first finishes.

3. PVC Retention with `reconciliation-mode: Independent`
Suspend a Job that has a JobSet-managed PVC. Verify the PVC is not deleted. Resume the Job and verify it successfully mounts the same PVC.

4. JobSet Deletion with `reconciliation-mode: Independent`: Delete a JobSet while some child Jobs are suspended. Verify all Jobs (running and suspended) and their associated resources are cleaned up correctly.

@kannon92 PTAL, if these make sense to you I will update the KEP with them

kannon92:
Thank you for that detailed answer. Please go ahead and add that to the KEP.

imreddy13:
> Thank you for that detailed answer. Please go ahead and add that to the KEP.

Added! @kannon92 @GiuseppeTT  PTAL

GiuseppeTT:
> > Thank you for that detailed answer. Please go ahead and add that to the KEP.
> 
> Added! @kannon92 @GiuseppeTT PTAL

Looks good to me. I'll leave the approval for @kannon92. I have a tentative implementation for it in https://github.com/kubernetes-sigs/jobset/pull/1202.

/lgtm

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,189 @@
...
+N/A
+
+### Risks and Mitigations
+
+The primary risk is introducing complexity to the JobSet controller, as it now has to manage the suspension state of individual jobs rather than just propagating the JobSet's suspension down to all jobs uniformly. This will be mitigated by thorough unit and integration testing.
+
+## Design Details
+
+Initially we will support this via an annotation on the JobSet: `jobset.sigs.k8s.io/reconciliation-mode: Independent`. If this annotation is set, then the JobSet webhook enforces `jobSet.spec.suspend = nil` and does not reconcile the field job.spec.suspend of child Jobs.
+
```
andreyvelich, line 68: Instead of introducing annotation, why we cannot just add validation webhook that doesn't allow to set `suspend` API in the individual ReplicatedJobs when this feature is disabled?
If they goal is to allow users/Kueue suspend ReplicatedJobs why not just allow to update that field?

ahg-g: Because it is not just about update validation in the webhook, but also about signaling to the operator whether or not to reconcile.

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,189 @@
...
+N/A
+
+### Risks and Mitigations
+
+The primary risk is introducing complexity to the JobSet controller, as it now has to manage the suspension state of individual jobs rather than just propagating the JobSet's suspension down to all jobs uniformly. This will be mitigated by thorough unit and integration testing.
+
+## Design Details
+
+Initially we will support this via an annotation on the JobSet: `jobset.sigs.k8s.io/reconciliation-mode: Independent`. If this annotation is set, then the JobSet webhook enforces `jobSet.spec.suspend = nil` and does not reconcile the field job.spec.suspend of child Jobs.
```
ahg-g, line 67: I recommend to name it `jobset.sigs.k8s.io/suspend-reconciliation-mode` since this is specific to suspend reconciliation, not everything else.

keps/1172-JobSuspendResume/README.md
andreyvelich: Why we cannot just check that if JobSet reconciler is triggered by `job.spec.suspend` update, we end the reconciler loop?

keps/1172-JobSuspendResume/kep.yaml
```diff
@@ -0,0 +1,29 @@
+title: Support granular suspend/resume for individual Jobs within a JobSet
+kep-number: 1172
+authors:
+  - "@imreddy"
+status: implementable
+creation-date: 2026-03-05
+reviewers:
+  - "@kannon92"
+  - "@ahg"
```
kannon92, line 9:
```suggestion
  - "@ahg-g"
```

kannon92 (commented):
Since @ahg-g is a reviewer on this, I'd like a lgtm to make sure his comments are addressed.

Otherwise I think this is good to merge.

keps/1172-JobSuspendResume/README.md
kannon92: This is still unaddressed. 

I also don't love the overuse of annotations.

keps/1172-JobSuspendResume/README.md
```diff
@@ -0,0 +1,189 @@
...
+If a SuccessPolicy targets a specific ReplicatedJob (e.g., "Succeed if the 'leader' Job completes"), suspending Jobs in other ReplicatedJobs will not prevent the JobSet from succeeding once the leader finishes.
+
+Once a JobSet meets its SuccessPolicy, the controller deletes or cleans up all Jobs including suspended ones.
+
+4. PVCs
+With `reconciliation-mode: Independent`, when a Job is suspended, its Pods are terminated, but the Job object and its associated PVCs should persist. When the Job is resumed, the new Pods will mount the same PVCs.
+
+#### Future API Evolution
+
+If the `jobset.sigs.k8s.io/reconciliation-mode: Independent` annotation proves successful, we may promote this configuration to a strongly-typed API field within the `JobSetSpec`. A dedicated policy field (e.g. `UpdatePolicy`) would cleanly move this behavior out of annotations, providing better validation, discoverability via OpenAPI schemas, and extensibility if other child Job fields need decoupled reconciliation in the future.
```
kannon92, line 102: IMO I want to make proper API a requirement for this. 

Can we do something like alpha with feature gate can have annotation but to promote this feature to beta we should move to API?

GiuseppeTT:
/close

@jianqiaol will assume this work and create new PRs.

PR #1179, opened 2026-03-05
feat(controller,webhook): implement Elastic JobSets (Pod-level scaling)

### LINKED ISSUE #463: Support Elastic JobSets

aniket2405:
#### What type of PR is this?
/kind feature
/kind api-change

This PR implements the pod-level horizontal scaling capabilities outlined in KEP-#463 (Support Elastic JobSets)

Currently, large-scale elastic distributed training workloads require the ability to dynamically scale worker counts up and down in response to hardware failures or resource availability. Previously, JobSet rejected these changes as the entire template was strictly immutable. 

This PR enables elasticity by:

1. Webhook Updates: Relaxing the strict immutability constraint on `parallelism` and `completions`, allowing these specific fields to be mutated mid-run.

2. Constraints: Adding webhook validation to ensure scaling fields are strictly >= 1 and blocking any scaling mutations if the JobSet has already reached a terminal state (Completed or Failed).

3. Controller Synchronization: Through `syncJobScaling`, the controller now detects scaling drift and issues opportunistic in-place PATCH requests to synchronize the active child Jobs with the updated JobSet template.

Fixes #463 

1. As defined in the KEP, this PR strictly focuses on Pod-level scaling (Elastic Indexed Jobs via KEP-3715). Horizontal Job Scaling (mutating `replicas`) is intentionally deferred to future work.

2. Feature Gate: All new webhook immutability bypasses and controller patching logic are wrapped behind the new `ElasticJobSet` feature gate (Alpha, disabled by default).

```
Users can now dynamically scale the number of pods in a running JobSet by updating the `parallelism` and `completions` fields of a ReplicatedJob template. Updates are seamlessly propagated to active child Jobs in-place. JobSets in a terminal state (Completed or Failed) will reject scaling requests.
```

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -212,6 +212,12 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 		}
 	}
 
+	// Opportunistically patch existing active jobs for pod-level scaling
+	if err := r.syncJobScaling(ctx, js, ownedJobs.active); err != nil {
+		log.Error(err, "syncing job scaling")
+		return ctrl.Result{}, err
+	}
+
 	// If job has not failed or succeeded, reconcile the state of the replicatedJobs.
 	if err := r.reconcileReplicatedJobs(ctx, js, ownedJobs, rjobStatuses, updateStatusOpts); err != nil {
 		log.Error(err, "creating jobs")
@@ -818,6 +824,64 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js 
 		log.V(2).Info("successfully created headless service", "service", klog.KObj(&headlessSvc))
 	}
 	return nil
+}
+
+// syncJobScaling applies in-place updates to the parallelism and completions of existing active child jobs
+// to match the desired state in the JobSet template (Elastic Indexed Jobs).
+func (r *JobSetReconciler) syncJobScaling(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job) error {
+	log := ctrl.LoggerFrom(ctx)
+	var finalErrs []error
+
+	// Lookup map for the desired scaling configurations
+	type scalingConfig struct {
+		parallelism *int32
+		completions *int32
+	}
+	desiredScaling := make(map[string]scalingConfig)
+	for _, rjob := range js.Spec.ReplicatedJobs {
+		desiredScaling[rjob.Name] = scalingConfig{
+			parallelism: rjob.Template.Spec.Parallelism,
+			completions: rjob.Template.Spec.Completions,
+		}
+	}
+
+	// Iterate through active jobs and patch them if their scaling configuration is out of sync
+	for _, job := range activeJobs {
+		rjobName, ok := job.Labels[jobset.ReplicatedJobNameKey]
+		if !ok {
+			continue // Skip if label is missing
+		}
+
+		desired, exists := desiredScaling[rjobName]
+		if !exists {
+			continue
+		}
+
+		needsPatch := false
+		if !ptr.Equal(job.Spec.Parallelism, desired.parallelism) {
+			needsPatch = true
+		}
+		if !ptr.Equal(job.Spec.Completions, desired.completions) {
+			needsPatch = true
+		}
+
+		if needsPatch {
+			patch := client.MergeFrom(job.DeepCopy())
+			job.Spec.Parallelism = desired.parallelism
+			job.Spec.Completions = desired.completions
+
+			log.V(2).Info("Patching child Job for horizontal pod scaling",
+				"job", klog.KObj(job),
+				"parallelism", ptr.Deref(desired.parallelism, 0),
+				"completions", ptr.Deref(desired.completions, 0))
+
+			if err := r.Patch(ctx, job, patch); err != nil {
+				finalErrs = append(finalErrs, fmt.Errorf("failed to patch job %q for scaling: %w", job.Name, err))
+			}
+		}
+	}
+
+	return errors.Join(finalErrs...)
 }
 
 // executeSuccessPolicy checks the completed jobs against the jobset success policy
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -831,6 +831,11 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js 
 // syncJobScaling applies in-place updates to the parallelism and completions of existing active child jobs
 // to match the desired state in the JobSet template (Elastic Indexed Jobs).
 func (r *JobSetReconciler) syncJobScaling(ctx context.Context, js *jobset.JobSet, activeJobs []*batchv1.Job) error {
+
+	if !features.Enabled(features.ElasticJobSet) {
+		return nil
+	}
+
 	log := ctrl.LoggerFrom(ctx)
 	var finalErrs []error
 
@@ -859,15 +864,7 @@ func (r *JobSetReconciler) syncJobScaling(ctx context.Context, js *jobset.JobSet
 			continue
 		}
 
-		needsPatch := false
-		if !ptr.Equal(job.Spec.Parallelism, desired.parallelism) {
-			needsPatch = true
-		}
-		if !ptr.Equal(job.Spec.Completions, desired.completions) {
-			needsPatch = true
-		}
-
-		if needsPatch {
+		if jobScalingNeedsPatch(job, desired.parallelism, desired.completions) {
 			patch := client.MergeFrom(job.DeepCopy())
 			job.Spec.Parallelism = desired.parallelism
 			job.Spec.Completions = desired.completions
@@ -884,6 +881,23 @@ func (r *JobSetReconciler) syncJobScaling(ctx context.Context, js *jobset.JobSet
 	}
 
 	return errors.Join(finalErrs...)
+}
+
+// jobScalingNeedsPatch evaluates whether a job's parallelism or completions
+// are out of sync with the desired state. It also guarantees we only trigger
+// a patch if the desired state is valid for Elastic Indexed Jobs (P == C).
+func jobScalingNeedsPatch(job *batchv1.Job, desiredParallelism, desiredCompletions *int32) bool {
+	// Parallelism and Completions should be equal for elastic scaling.
+	// If the desired state is out of sync with itself, it's an invalid configuration.
+	if !ptr.Equal(desiredParallelism, desiredCompletions) {
+		return false
+	}
+
+	// 2. Check if the Job has drifted from the desired state.
+	parallelismChanged := !ptr.Equal(job.Spec.Parallelism, desiredParallelism)
+	completionsChanged := !ptr.Equal(job.Spec.Completions, desiredCompletions)
+
+	return parallelismChanged || completionsChanged
 }
 
 // executeSuccessPolicy checks the completed jobs against the jobset success policy
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -213,9 +213,11 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 	}
 
 	// Opportunistically patch existing active jobs for pod-level scaling
-	if err := r.syncJobScaling(ctx, js, ownedJobs.active); err != nil {
-		log.Error(err, "syncing job scaling")
-		return ctrl.Result{}, err
+	if features.Enabled(features.ElasticJobSet) {
+		if err := r.syncJobScaling(ctx, js, ownedJobs.active); err != nil {
+			log.Error(err, "syncing job scaling")
+			return ctrl.Result{}, err
+		}
 	}
 
 	// If job has not failed or succeeded, reconcile the state of the replicatedJobs.
--- a/pkg/controllers/jobset_controller_test.go
+++ b/pkg/controllers/jobset_controller_test.go
@@ -2102,3 +2102,93 @@ func TestGroupReplicas(t *testing.T) {
 		})
 	}
 }
+
+func TestSyncJobScaling(t *testing.T) {
+	var (
+		jobSetName        = "test-jobset"
+		replicatedJobName = "replicated-job"
+		jobName           = "test-job"
+		ns                = "default"
+	)
+
+	tests := []struct {
+		name        string
+		jobSet      *jobset.JobSet
+		activeJobs  []*batchv1.Job
+		expectPatch bool
+	}{
+		{
+			name: "jobs in sync, no patch needed",
+			jobSet: testutils.MakeJobSet(jobSetName, ns).
+				ReplicatedJob(testutils.MakeReplicatedJob(replicatedJobName).
+					Job(testutils.MakeJobTemplate(jobName, ns).Parallelism(2).Completions(2).Obj()).
+					Replicas(1).
+					Obj()).Obj(),
+			activeJobs: []*batchv1.Job{
+				makeJob(&makeJobArgs{
+					jobSetName:        jobSetName,
+					replicatedJobName: replicatedJobName,
+					jobName:           "test-job-0",
+					ns:                ns,
+				}).Parallelism(2).Completions(2).Obj(),
+			},
+			expectPatch: false,
+		},
+		{
+			name: "jobs out of sync, patch needed",
+			jobSet: testutils.MakeJobSet(jobSetName, ns).
+				ReplicatedJob(testutils.MakeReplicatedJob(replicatedJobName).
+					Job(testutils.MakeJobTemplate(jobName, ns).Parallelism(8).Completions(8).Obj()).
+					Replicas(1).
+					Obj()).Obj(),
+			activeJobs: []*batchv1.Job{
+				makeJob(&makeJobArgs{
+					jobSetName:        jobSetName,
+					replicatedJobName: replicatedJobName,
+					jobName:           "test-job-0",
+					ns:                ns,
+				}).Parallelism(2).Completions(2).Obj(),
+			},
+			expectPatch: true,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var patchesSent int
+			_, ctx := ktesting.NewTestContext(t)
+			scheme := runtime.NewScheme()
+			utilruntime.Must(jobset.AddToScheme(scheme))
+			utilruntime.Must(corev1.AddToScheme(scheme))
+			utilruntime.Must(batchv1.AddToScheme(scheme))
+
+			fakeClientBuilder := fake.NewClientBuilder().WithScheme(scheme).WithInterceptorFuncs(interceptor.Funcs{
+				Patch: func(ctx context.Context, client client.WithWatch, obj client.Object, patch client.Patch, opts ...client.PatchOption) error {
+					patchesSent++
+					patchedJob := obj.(*batchv1.Job)
+					if *patchedJob.Spec.Parallelism != 8 || *patchedJob.Spec.Completions != 8 {
+						t.Errorf("Expected job to be patched to Parallelism/Completions 8, got %d and %d", *patchedJob.Spec.Parallelism, *patchedJob.Spec.Completions)
+					}
+					return nil
+				},
+			})
+
+			r := &JobSetReconciler{
+				Client: fakeClientBuilder.Build(),
+				Scheme: scheme,
+				Record: events.NewFakeRecorder(32),
+			}
+
+			err := r.syncJobScaling(ctx, tc.jobSet, tc.activeJobs)
+			if err != nil {
+				t.Fatalf("Unexpected error: %v", err)
+			}
+			if tc.expectPatch && patchesSent == 0 {
+				t.Errorf("Expected patch to be sent, but none were")
+			}
+			if !tc.expectPatch && patchesSent > 0 {
+				t.Errorf("Expected no patches to be sent, but got %d", patchesSent)
+			}
+		})
+	}
+}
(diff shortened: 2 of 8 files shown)


kannon92 (commented):
Please add integration tests (test/integration).

GiuseppeTT:
Please rebase to get this [fix](https://github.com/kubernetes-sigs/jobset/pull/1180)

aniket2405:
> Please add integration tests (test/integration).

Added integration tests.


> Please rebase to get this [fix](https://github.com/kubernetes-sigs/jobset/pull/1180)

Rebased.

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -753,5 +758,43 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+							Completions(2).Obj()).
+						Obj())
+			},
+			updateJobSet: func(js *jobset.JobSet) {
+				// Ensure the feature gate is OFF (testing the default behavior)
+				_ = features.SetEnable(features.ElasticJobSet, false)
+				js.Spec.ReplicatedJobs[0].Template.Spec.Parallelism = ptr.To[int32](4)
+				js.Spec.ReplicatedJobs[0].Template.Spec.Completions = ptr.To[int32](4)
+			},
+			updateShouldFail: true,
```
kannon92: Please check the errors you get to confirm its what you expect.

aniket2405: Thanks. Good catch. I've added this now.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -3303,13 +3304,255 @@ func TestValidateUpdate(t *testing.T) {
 				field.Invalid(field.NewPath("spec").Child("replicatedJobs"), "", "field is immutable"),
 			}.ToAggregate(),
 		},
+		{
```
kannon92, line 3453: Can you add a test case for negative case of feature gate?

ie I want to confirm that existing behavior occurs without feature gate.

pkg/webhooks/jobset_webhook.go
```diff
@@ -304,11 +305,61 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
...
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Parallelism = oldRJob.Template.Spec.Parallelism
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Completions = oldRJob.Template.Spec.Completions
+		}
+	}
 
 	// Allow pod template to be mutated for suspended JobSets, or JobSets getting suspended.
 	// This is needed for integration with Kueue/DWS.
 	if ptr.Deref(oldJS.Spec.Suspend, false) || ptr.Deref(js.Spec.Suspend, false) {
 		for index := range js.Spec.ReplicatedJobs {
+			// Skip if the index is out of bounds
```
kannon92: This seems like a bug that should be done as a separate PR.

pkg/webhooks/jobset_webhook.go
```diff
@@ -320,10 +371,14 @@ func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, oldJS, js *jobset.Jo
 		}
 	}
 
-	// Note that SucccessPolicy and failurePolicy are made immutable via CEL.
-	errs := apivalidation.ValidateImmutableField(mungedSpec.ReplicatedJobs, oldJS.Spec.ReplicatedJobs, field.NewPath("spec").Child("replicatedJobs"))
+	// SuccessPolicy and failurePolicy are made immutable
```
kannon92: Please revert unrelated changes or open up a PR to fix it separately?

pkg/webhooks/jobset_webhook.go
aniket2405: Good call. I noticed the existing loop would panic if the array lengths mismatched, so I added the bounds check, but I agree this is a separate issue from Elastic scaling. I've reverted this addition to keep the PR scoped strictly to ElasticJobSet.

Should I open a ticket to track this separate issue?

pkg/webhooks/jobset_webhook_test.go
aniket2405: Makes sense. Done.

pkg/webhooks/jobset_webhook.go
aniket2405: Note: I did include a bounds check strictly inside the new ElasticJobSet evaluation block to defensively prevent panics during parallelism evaluation

pkg/webhooks/jobset_webhook.go
aniket2405: Done.

pkg/controllers/jobset_controller.go
```diff
@@ -823,6 +831,64 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js
...
+
+	// Lookup map for the desired scaling configurations
+	type scalingConfig struct {
+		parallelism *int32
+		completions *int32
+	}
+	desiredScaling := make(map[string]scalingConfig)
+	for _, rjob := range js.Spec.ReplicatedJobs {
+		desiredScaling[rjob.Name] = scalingConfig{
+			parallelism: rjob.Template.Spec.Parallelism,
```
kannon92, line 850: Elastic IndexedJob requires that both paralleism/completion are updated together.

Do we have anything to enforce that here?

ref: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/validation/validation.go#L920

pkg/webhooks/jobset_webhook.go
```diff
@@ -308,25 +309,76 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
...
+			// Safe bounds check strictly for Elastic scaling evaluation
+			if index >= len(oldJs.Spec.ReplicatedJobs) {
+				continue
+			}
+
+			oldRJob := oldJs.Spec.ReplicatedJobs[index]
+			newRJob := newJs.Spec.ReplicatedJobs[index]
+			rJobPath := field.NewPath("spec", "replicatedJobs").Index(index).Child("template", "spec")
+
+			// Check if parallelism and completions have changed
```
kannon92: https://github.com/kubernetes-sigs/jobset/pull/1179/changes#r2942706917

It would be good to catch updates like this without resorting to failures in the Job controller.

Can you do some verification on your end to see what would happen if Job Template rejects the resize and what kind of information we would get on JobSet?

pkg/controllers/jobset_controller.go
aniket2405: Yes, we apply both values to the local job object memory before we execute the single `r.Patch()` call

`r.Patch()` applies them atomically in a single API call, so we do enforce that here.

Because it happens in a single patch request, the API server receives and processes them simultaneously.

```
patch := client.MergeFrom(job.DeepCopy())
job.Spec.Parallelism = desired.parallelism
job.Spec.Completions = desired.completions

if err := r.Patch(ctx, job, patch); err != nil { ... }
```

Ref: https://github.com/kubernetes-sigs/jobset/pull/1179/changes#diff-68677f8530f00fbbb399ac36268f928ee584948dd09091644a8fad81a8658153R883

pkg/webhooks/jobset_webhook.go
aniket2405: I added a check in `ValidateUpdate` and a unit test that explicitly verifies `Parallelism == Completions` if they are being mutated.

pkg/controllers/jobset_controller.go
kannon92: They also have to be updated to be the same right?

The main issue is that they must be updated together and be the same value.

aniket2405: Yes, absolutely.

I'm about to push a commit (was running local integration test) that guarantees both conditions:

1. Enforcing the same value - If a user tries to mutate Parallelism and Completions to different values, the webhook immediately rejects the request with a `field.Invalid` error
2. Updating them together through Atomic Patch - Because the webhook guarantees the JobSet spec has matching values, the controller's syncJobScaling function safely reads those matching values. It applies both values to the local job object memory and calls a single r.Patch(ctx, job, patch).

aniket2405:
@kannon92 
Gentle reminder to review the latest changes. Thank you!

test/integration/controller/suite_test.go
```diff
@@ -57,6 +58,10 @@ func TestAPIs(t *testing.T) {
 var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
+	var err error
+	err = features.SetEnable(features.ElasticJobSet, true)
```
kannon92: Please set the feature gate in the test case and set it back.

Setting this to true for all tests is an antipattern.

test/integration/controller/jobset_controller_test.go
```diff
@@ -2854,6 +2854,74 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 				},
 			},
 		}),
+		ginkgo.Entry("dynamic horizontal pod scaling propagates parallelism and completions to child jobs", &testCase{
```
kannon92: I want to see a test with this feature gate off and a test with this feature gate on.

I also want to see a way to toggler this feature on/off in the test suite and not turn it on for all integration tests.

pkg/webhooks/jobset_webhook.go
```diff
@@ -308,25 +309,79 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
...
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
 func (j *jobSetWebhook) ValidateUpdate(ctx context.Context, oldJs, newJs *jobset.JobSet) (admission.Warnings, error) {
+	mungedSpec := newJs.Spec.DeepCopy()
+	var errs field.ErrorList
+
+	// Elastic JobSet Validation
+	if features.Enabled(features.ElasticJobSet) {
+		// Check if the JobSet is in a terminal state (Completed or Failed)
+		isTerminal := false
+		for _, cond := range oldJs.Status.Conditions {
```
kannon92: Can you use IsStatusConditionTrue from apimachinery here?

kannon92:
@aniket2405 are you able to address the conflicts and finish out this implementation?

aniket2405:
Yes, @kannon92 
Was facing a couple of challenges with the test cases, but I'm able to get over them.
I was away for a few days, will be pushing my changes today.

test/integration/controller/jobset_controller_test.go
aniket2405: I pulled the Elastic tests out of the main DescribeTable and wrapped them in their own `ginkgo.When` contexts. Used `ginkgo.BeforeEach` and `AfterEach` to toggle the feature gate on and off just for these specific tests.

pkg/webhooks/jobset_webhook.go
aniket2405: Ack.

test/integration/controller/suite_test.go
aniket2405: Ack. Isolated the tests inside a `ginkgo.When` block using `BeforeEach` and `AfterEach `to toggle the gate just for your tests.

kannon92:
Looks like this is close. Just need to fix lint.

aniket2405:
Ack. Fixed lint.

aniket2405:
Seems we're good to merge this
@kannon92

kannon92:
can you squash your commits to be more functional?

Maybe the follows up can be squashed.

pkg/webhooks/jobset_webhook.go
```diff
@@ -309,27 +310,74 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
...
+					}
+					if newRJob.Template.Spec.Parallelism != nil && newRJob.Template.Spec.Completions != nil && *newRJob.Template.Spec.Parallelism != *newRJob.Template.Spec.Completions {
+						errs = append(errs, field.Invalid(rJobPath.Child("completions"), *newRJob.Template.Spec.Completions, "completions must be equal to parallelism for Elastic Indexed Jobs"))
+					}
+				}
+			}
+
+			// Mask parallelism and completions in mungedSpec to bypass the strict immutability check
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Parallelism = oldRJob.Template.Spec.Parallelism
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Completions = oldRJob.Template.Spec.Completions
```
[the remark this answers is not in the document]
kannon92: This is a good suggestion. Scaling an Job is only possible if its a IndexedJob.

pkg/controllers/jobset_controller.go
```diff
@@ -820,6 +828,64 @@ func (r *JobSetReconciler) createHeadlessSvcIfNecessary(ctx context.Context, js
...
+		if !ok {
+			continue // Skip if label is missing
+		}
+
+		desired, exists := desiredScaling[rjobName]
+		if !exists {
+			continue
+		}
+
+		needsPatch := false
```
kannon92: There is a bug here that if one of these conditions is satisfied then the needsPatch = true.

We need to check that both parallelism and completitions change together also.

I'd suggest maybe make a simple function to check all these conditions together.

pkg/controllers/jobset_controller_test.go
```diff
@@ -2102,3 +2102,94 @@ func TestGroupReplicas(t *testing.T) {
...
+	)
+
+	tests := []struct {
+		name        string
+		jobSet      *jobset.JobSet
+		activeJobs  []*batchv1.Job
+		expectPatch bool
+	}{
+		{
+			name: "jobs in sync, no patch needed",
```
kannon92, line 2122: can you add the following test cases:

Parallelism kept same, completions is changed.
Parallelismm is changed, completions kept the same.

pkg/controllers/jobset_controller_test.go
```diff
@@ -2102,3 +2102,94 @@ func TestGroupReplicas(t *testing.T) {
...
+				},
+			})
+
+			r := &JobSetReconciler{
+				Client: fakeClientBuilder.Build(),
+				Scheme: scheme,
+				Record: events.NewFakeRecorder(32),
+			}
+
+			features.SetFeatureGateDuringTest(t, features.ElasticJobSet, true)
```
kannon92: The feature gate should be an argument in the test case.

This way you can test negative feature gate also. Right now you are only testing the position aspect of the feature. Please don't forget the negative.

pkg/webhooks/jobset_webhook.go
```diff
@@ -309,27 +310,74 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
...
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Annotations = oldJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Annotations
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Labels = oldJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Labels
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector = oldJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.NodeSelector
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.Tolerations = oldJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.Tolerations
 
 			// Pod Scheduling Gates can be updated for batch/v1 Job: https://github.com/kubernetes/kubernetes/blob/ceb58a4dbc671b9d0a2de6d73a1616bc0c299863/pkg/apis/batch/validation/validation.go#L662
-			newJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates = oldJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates
+			mungedSpec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates = oldJs.Spec.ReplicatedJobs[index].Template.Spec.Template.Spec.SchedulingGates
```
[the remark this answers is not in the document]
kannon92: I think copilot missed the code you added on line 324.

It would be good to add a test case verifying that only changes of parallelism/completiions are allowed and one can not change the replicatedJob shape. 

Even if length is kept the same, if there is a reordering of the spec I think that could cause an issue.

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -753,5 +763,44 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 			},
 			jobSetCreationShouldFail: true,
 		}),
+		ginkgo.Entry("validate updating parallelism and completions is ALLOWED when ElasticJobSet is enabled", &testCase{
```
kannon92, line 767: I think you are missing more test cases here.

kannon92:
@aniket2405 

I'd like to include this for 0.12 release. Can you address the comments?

aniket2405:
@kannon92 
Acknowledged. Will expedite this at the earliest.

pkg/webhooks/jobset_webhook.go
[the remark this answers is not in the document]
aniket2405: Done. I've refactored the validation logic to map the old ReplicatedJobs by their Name. Does that look okay?
Also, added required test cases.

pkg/webhooks/jobset_webhook.go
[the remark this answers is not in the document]
aniket2405: Scoped the elastic scaling validation + immutability bypass to IndexedCompletion only.

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -45,6 +46,10 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 	var ns *corev1.Namespace
 
 	ginkgo.BeforeEach(func() {
+
+		// Ensure feature gate is off by default for all standard tests
+		_ = features.SetEnable(features.ElasticJobSet, false)
```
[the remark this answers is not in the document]
aniket2405: Done.

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -753,5 +763,44 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+							CompletionMode(batchv1.IndexedCompletion).
+							Parallelism(2).
+							Completions(2).Obj()).
+						Obj())
+			},
+			updateJobSet: func(js *jobset.JobSet) {
+				// Turn the feature gate ON just before we hit the update webhook
+				_ = features.SetEnable(features.ElasticJobSet, true)
+				js.Spec.ReplicatedJobs[0].Template.Spec.Parallelism = ptr.To[int32](4)
+				js.Spec.ReplicatedJobs[0].Template.Spec.Completions = ptr.To[int32](4)
```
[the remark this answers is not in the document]
aniket2405: Done

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -753,5 +763,44 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+							CompletionMode(batchv1.IndexedCompletion).
+							Parallelism(2).
+							Completions(2).Obj()).
+						Obj())
+			},
+			updateJobSet: func(js *jobset.JobSet) {
+				// Ensure the feature gate is OFF (testing the default behavior)
+				_ = features.SetEnable(features.ElasticJobSet, false)
+				js.Spec.ReplicatedJobs[0].Template.Spec.Parallelism = ptr.To[int32](4)
+				js.Spec.ReplicatedJobs[0].Template.Spec.Completions = ptr.To[int32](4)
```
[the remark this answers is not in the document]
aniket2405: Done

pkg/controllers/jobset_controller.go
aniket2405: Extracted the condition into a separate helper function jobScalingNeedsPatch. Ensuring both parallelism and completitions change together also.

pkg/controllers/jobset_controller_test.go
aniket2405: Added, thanks for the suggestion.

pkg/controllers/jobset_controller_test.go
aniket2405: Good catch. Added feature gate as argument here.

test/integration/webhook/jobset_webhook_test.go
aniket2405: Added negative test cases to integration tests

aniket2405:
@kannon92
Thanks for reviewing in-depth.
Latest patch has more extensive test cases.

kannon92:
/approve
/lgtm


It would be good to update our website with this feature.

aniket2405:
Thanks for the approval and the countless reviews, @kannon92 
Learnt a lot!

> 
> It would be good to update our website with this feature.

Would you like me to add the feature to the website?

kannon92:
> Would you like me to add the feature to the website?

Yes please!

andreyvelich:
Thanks for your effort to support elastic JobSet @aniket2405!
Looking forward to see how we can enable elasticity with PyTorch on Kubernetes: https://github.com/kubeflow/trainer/issues/2903 🚀

aniket2405:
Thanks for the help, @andreyvelich 
Certainly, looking forward to complete the story for enabling elasticity with PyTorch on Kubernetes!

aniket2405:
Hi @kannon92 

What details should we add in the website about the presence of this feature?

kannon92:
> Hi @kannon92
> 
> What details should we add in the website about the presence of this feature?

Add a new task for Elastic JobSets. And add information about feature gate and the limitations.

aniket2405:
Apologies for the delay, PR for update on website with the feature: https://github.com/kubeflow/website/pull/4405
cc: @kannon92

PR #1185, opened 2026-03-12
chore(api): Add patchStrategy markers to JobSet API for SMP

andreyvelich:
#### What type of PR is this?

/kind feature

While working on RuntimePatches API in TrainJob, we found out that ReplicatedJob API in JobSet doesn't have `PatchStrategy` markers: https://github.com/kubeflow/trainer/pull/3309
That doesn't allow us to use single Strategic Merge Patch call in the TrainJob controller, since Kubernetes [API Machinery requires it](https://github.com/kubernetes/apimachinery/blob/65082b63930356c3d5e0e9bc4e6dbcfe9d8dac57/pkg/util/strategicpatch/meta.go#L31)

I added markers to the appropriate API fields in JobSet. Same markers in Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/apis/kueue/v1beta2/multikueue_types.go#L101-L104

/assign @kannon92 @GiuseppeTT 

cc @astefanutti @tenzen-y 

Fixes #

```release-note
Add patchStrategy markers to JobSet API
```

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -49,7 +49,7 @@ linters:
                 description: "A field with a default value cannot be required"
             conditions:
               isFirstField: Warn # Require conditions to be the first field in the status struct.
-              usePatchStrategy: Forbid # Forbid patchStrategy markers on the Conditions field.
+              usePatchStrategy: Ignore # Ignore patchStrategy markers on the Conditions field.
               useProtobuf: Forbid # We don't use protobuf, so protobuf tags are not required.
             #optionalfields:
             #  pointers:
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -118,9 +118,11 @@ const (
 // +kubebuilder:validation:XValidation:rule="!(has(self.startupPolicy) && self.startupPolicy.startupPolicyOrder == 'InOrder' && self.replicatedJobs.exists(x, has(x.dependsOn)))",message="StartupPolicy and DependsOn APIs are mutually exclusive"
 type JobSetSpec struct {
 	// replicatedJobs is the group of jobs that will form the set.
+	// +patchMergeKey=name
+	// +patchStrategy=merge
 	// +listType=map
 	// +listMapKey=name
-	ReplicatedJobs []ReplicatedJob `json:"replicatedJobs,omitempty"`
+	ReplicatedJobs []ReplicatedJob `json:"replicatedJobs,omitempty" patchStrategy:"merge" patchMergeKey:"name"`
 
 	// network defines the networking options for the jobset.
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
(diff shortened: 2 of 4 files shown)


GiuseppeTT:
Try enabling the API linter for these fields and running the linter. it might detect other improvements and suggest alternative ways to change them. Just make sure the changes are backward compatibility (so not maxLength for legacy array fields).

https://github.com/kubernetes-sigs/jobset/blob/3021d626a6898289f2b5e0cf71448120056dceac/.golangci-kal.yml#L72-L76

cc @kannon92

kannon92:
Can you fix the linter?

andreyvelich:
@kannon92 @GiuseppeTT I applied ignore setting for patch strategy in Condition API, like in Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/.golangci-kal.yaml#L52

GiuseppeTT:
> @kannon92 @GiuseppeTT I applied ignore setting for patch strategy in Condition API, like in Kueue: https://github.com/kubernetes-sigs/kueue/blob/main/.golangci-kal.yaml#L52

Nice. Is this PR ready for review?

andreyvelich:
Yes, I think we should be good to move this forward.

GiuseppeTT:
Code-wise the change looks good, but design-wise @andreyvelich what do you think about [copilot's comment](https://github.com/kubernetes-sigs/jobset/pull/1185#discussion_r2927633739)? I guess in your case it's not a problem because you are not adding / removing replicatedJob items. I guess it's also not a problem because the webhook would block such additions / removals.

andreyvelich:
> Code-wise the change looks good, but design-wise @andreyvelich what do you think about [copilot's comment](https://github.com/kubernetes-sigs/jobset/pull/1185#discussion_r2927633739)? I guess in your case it's not a problem because you are not adding / removing replicatedJob items. I guess it's also not a problem because the webhook would block such additions / removals.

Hmm, this is good point. Do we know if `kubectl` can updated the ordering of the list after we do: `StrategicMergePatch()`? If yes, that might mess up with DependsOn logic validation.
cc @tenzen-y @astefanutti

andreyvelich:
Yeah, I think this should be sufficient to apply SMP, I removed the listType map from rJob list:
```
// +patchMergeKey=name
// +patchStrategy=merge
```

andreyvelich:
Ok, explore a bit on this, we should probably keep the `listType=map` for rJob, and users should preserve ordering while applying patch. Additionally, this line should validate that order is preserved on JobSet update:
https://github.com/kubernetes-sigs/jobset/blob/main/pkg/webhooks/jobset_webhook.go#L395 

On the Trainer side, we just need to ensure we preserve the original ordering when apply the patch:
```go
// 1. Apply SMP to full JobSet spec
  sourceJSON, _ := json.Marshal(originalSpec)
  patchJSON, _ := json.Marshal(patchSpec)
  mergedJSON, err := strategicpatch.StrategicMergePatch(sourceJSON, patchJSON, jobsetv1alpha2.JobSetSpec{})

  // 2. Unmarshal result
  var mergedSpec jobsetv1alpha2.JobSetSpec
  json.Unmarshal(mergedJSON, &mergedSpec)

  // 3. Restore original ordering
  originalOrder := make(map[string]int, len(originalSpec.ReplicatedJobs))
  for i, rj := range originalSpec.ReplicatedJobs {
      originalOrder[rj.Name] = i
  }
  sort.SliceStable(mergedSpec.ReplicatedJobs, func(i, j int) bool {
      return originalOrder[mergedSpec.ReplicatedJobs[i].Name] <
             originalOrder[mergedSpec.ReplicatedJobs[j].Name]
  })
```

GiuseppeTT:
@andreyvelich Just to double-check.

I was reading more about the topic and it seems that the API server does not support strategic merge patches for CRs. I assume you are using `strategicpatch.StrategicMergePatch` purely as a local in-memory utility to merge the RuntimePatches into the base JobSet spec before creating/updating the object. If so, adding these markers makes sense so the Go structs get the correct tags for the local merge.

As an alternative, does server-side apply (e.g., `r.Client.Patch(..., client.Apply)`) handle your use case?

andreyvelich:
> If so, adding these markers makes sense so the Go structs get the correct tags for the local merge.

Yeah, that's right, so markers needed just for local utility function.
Looking at other CRDs in Kueue, they also define those markers.

GiuseppeTT:
Sounds good.

/lgtm
/approve

PR #1188, opened 2026-03-13
Reorder follower pod validation to surface leader scheduling status

### LINKED ISSUE #1187: Pod validation webhook obscures leader scheduling state by validating follower NodeSelector too early
**Description:**
In multi-slice/exclusive placement workloads, if a leader pod (index 0) remains in a `Pending` state (e.g., due to insufficient TPU/GPU resources or node group size limits), the JobSet mutating webhook intentionally skips patching the follower pods with the `NodeSelector`. 

However, the validating webhook (`ValidateCreate` in `pkg/webhooks/pod_webhook.go`) currently checks if the `NodeSelector` is set *before* checking if the leader pod is actually scheduled. 

Because of this ordering, the user receives a hard rejection error:
`admission webhook "vpod.kb.io" denied the request: follower pod node selector not set`

Instead of the much more helpful transient error intended for this exact scenario:
`leader pod not yet scheduled, not creating follower pod. this is an expected, transient error`

**Example:**

Admission error for follower pod (e.g. example-job-0-1):

```
2026-03-01 00:51:00.000 Error creating: admission webhook "vpod.kb.io" denied the request: follower pod node selector not set
```

Root cause:
```
2026-03-01T00:00:00Z	ERROR	admission	finding leader pod for follower	{"webhookGroup": "", "webhookKind": "Pod", "Pod": {"name":"","namespace":"default"}, "namespace": "default", "name": "", "resource": {"group":"","version":"v1","resource":"pods"}, "user": "system:serviceaccount:kube-system:job-controller", "requestID": "<request-id>", "error": "expected 1 leader pod (example-job-0-0), but got 0. this is an expected, transient error"}
```

The current error message acts as a diagnostic mask. It leads users to believe there is a configuration or controller bug regarding node selectors, when the root cause is actually infrastructure capacity preventing the leader pod from scheduling. Moving the check will surface the transient error, instantly directing users to investigate their leader pod's scheduling constraints.

**Proposed Solution:**
PR #1188: In `pkg/webhooks/pod_webhook.go`, move the `NodeSelector` validation block (lines 224-230) to evaluate *after* the `leaderPodScheduled` check (lines 231-238).

https://github.com/kubernetes-sigs/jobset/blob/3021d626a6898289f2b5e0cf71448120056dceac/pkg/webhooks/pod_webhook.go#L224-L238

0xlen:
#### What type of PR is this?
/kind feature

This PR improves the diagnostic output of the `vpod` webhook for JobSets using exclusive placement. 

Previously, `ValidateCreate` checked if a follower pod had a `NodeSelector` before checking if the leader pod was scheduled. If the leader was stuck in `Pending` (e.g., due to quotas), the mutating webhook correctly withheld the selector, but the validating webhook immediately rejected the pod with `"follower pod node selector not set"`. This masked the more accurate `"expected, transient error"` regarding the leader pod's status.

By moving the leader scheduling check first, users will now correctly see the transient error, which guides them to investigate the leader pod's resource constraints rather than assuming a JobSet configuration failure.

Fixes #1187

```release-note
NONE
```

### DIFF
--- a/pkg/webhooks/pod_webhook_test.go
+++ b/pkg/webhooks/pod_webhook_test.go
@@ -535,6 +535,103 @@ func TestPodWebhookValidateCreate(t *testing.T) {
 			},
 			wantErr: "follower pod node selector not set despite leader pod being scheduled",
 		},
+		{
+			name: "follower pod requires topology node selector key after leader is scheduled",
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "js-rjob-0-1-follower",
+					Namespace: "default",
+					Labels: map[string]string{
+						jobset.JobSetNameKey:        "js",
+						jobset.ReplicatedJobNameKey: "rjob",
+						jobset.JobIndexKey:          "0",
+					},
+					Annotations: map[string]string{
+						jobset.JobSetNameKey:                       "js",
+						jobset.ExclusiveKey:                        "topology.kubernetes.io/zone",
+						"batch.kubernetes.io/job-completion-index": "1",
+					},
+					OwnerReferences: []metav1.OwnerReference{
+						{UID: types.UID("job-uid-1"), Kind: "Job", Controller: ptr.To(true)},
+					},
+				},
+				Spec: corev1.PodSpec{
+					NodeSelector: map[string]string{
+						"kubernetes.io/hostname": "node-a",
+					},
+				},
+			},
+			existingObjs: []runtime.Object{
+				&corev1.Pod{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "js-rjob-0-0-leader",
+						Namespace: "default",
+						Labels: map[string]string{
+							jobset.JobSetNameKey:        "js",
+							jobset.ReplicatedJobNameKey: "rjob",
+							jobset.JobIndexKey:          "0",
+						},
+						Annotations: map[string]string{
+							jobset.ExclusiveKey:                        "topology.kubernetes.io/zone",
+							"batch.kubernetes.io/job-completion-index": "0",
+						},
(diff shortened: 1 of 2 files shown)


kannon92 (commented):
/ok-to-test

pkg/webhooks/pod_webhook_test.go
```diff
@@ -441,6 +441,127 @@ func TestDefault(t *testing.T) {
 	}
 }
 
+func TestPodWebhookValidateCreate(t *testing.T) {
```
kannon92, line 444: [AI]
  Suggestions

  - Minor: The tests only cover error paths. A happy-path test case (follower pod with correct node selector + scheduled leader → no error) would strengthen confidence, but this is a gap in the existing test suite and not something this PR introduced.
  - Minor: The test currently doesn't cover the topologyKey missing case (node selector exists but doesn't have the topology key). Again, not a regression

Looks like there are a few missing test cases but I'm happy if you want to address this as a follow up.

kannon92:
/approve

@GiuseppeTT can you take a look and give your LGTM?

I'm curious if we should consider a cherry pick for this.

pkg/webhooks/pod_webhook_test.go
0xlen: Thanks for pointing this out. I’ve added both the happy-path case and the missing-topologyKey test in this PR:
1. the case where the follower has a nodeSelector but is missing the required topology key
2. the happy path where the leader pod is already scheduled and the follower has the expected topology nodeSelector

kannon92:
@GiuseppeTT @imreddy13 could one of you take a look here?

0xlen:
Hi @GiuseppeTT @imreddy13, gentle ping for LGTM when you get a chance. Let me know if cherry pick is required, I can help to create a separate PR. Thanks!

PR #1207, opened 2026-04-09
Fix admission webhook to return Invalid instead of Forbidden

JerryHu1994:
#### What type of PR is this?

Add one of the following kinds:
/kind bug

 When a JobSet is rejected by the validating admission webhook (e.g., invalid failure policy rule name), the API returns HTTP 403 Forbidden instead of the expected 422. This is because ValidateCreate returns a plain error via errors.Join(allErrs...) (line 306 of jobset_webhook.go). Controller-runtime's webhook handler checks if the returned error is an*apierrors.StatusError — if not, it calls admission.Denied() which defaults to 403.
Here is an example:
```
CreateJobSet: admission webhook "validate-jobset-x-k8s-io-v1alpha2-jobset.x-k8s.io" denied the request: invalid failure policy rule name 'fail-non-retryable', a failure policy rule name must start with an alphabetic character, optionally followed by a string of alphanumeric characters or '_,:', and must end with an alphanumeric character or '_' (code=403, reason=Forbidden) service_<truncated 285 symbols, right click to see full line>-v5g-moe-b4c4f: CreateJobSet: admission webhook "validate-jobset-x-k8s-io-v1alpha2-jobset.x-k8s.io" denied the request: invalid failure policy rule name 'fail-non-retryable', a failure policy rule name must start with an alphabetic character, optionally followed by a string of alphanumeric characters or '_,:', and must end with an alphanumeric character or '_' (code=403, reason=Forbidden)
```

#### Test
```
jieh@jieh-mac jobset % go test ./pkg/webhooks/ -count=1
ok      sigs.k8s.io/jobset/pkg/webhooks 0.810s
```

Wrap the validation errors in apierrors.NewInvalid() which produces a *apierrors.StatusError with status code 422 and reason "Invalid".

"NONE"

### DIFF
--- a/pkg/webhooks/jobset_webhook_test.go
+++ b/pkg/webhooks/jobset_webhook_test.go
@@ -9,10 +9,11 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
-	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -3085,7 +3086,16 @@ func TestValidateCreate(t *testing.T) {
 			features.SetFeatureGateDuringTest(t, features.RestartJob, tc.enableRestartJob)
 			_, err := testWebhook.ValidateCreate(context.TODO(), tc.js.DeepCopy())
 			if err != nil && tc.want != nil {
-				assert.Contains(t, err.Error(), tc.want.Error())
+				// Verify it's specifically a 422 StatusError
+				var statusErr *apierrors.StatusError
+				require.ErrorAs(t, err, &statusErr, "expected *apierrors.StatusError")
+				assert.Equal(t, int32(422), statusErr.ErrStatus.Code)
+				assert.Equal(t, metav1.StatusReasonInvalid, statusErr.ErrStatus.Reason)
+
+				// Verify all expected error substrings are present
+				for _, wantErr := range strings.Split(tc.want.Error(), "\n") {
+					assert.Contains(t, err.Error(), wantErr)
+				}
 			} else if err != nil && tc.want == nil {
 				t.Errorf("unexpected error: %v", err)
 			} else if err == nil && tc.want != nil {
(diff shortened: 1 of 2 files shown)


kannon92 (commented):
Thanks for the fix! The logic changes to return the correct 422 Unprocessable Entity via `apierrors.NewInvalid()` look solid.

Regarding the test updates in `pkg/webhooks/jobset_webhook_test.go`, the current approach to verify the `StatusError` introduces some deep nesting (`if assert.True(t, errors.As(...)) { ... }`). 

Here are a couple of alternatives to consider for improving the readability of these assertions:

**Alternative 1: Flattening with `require.ErrorAs` (Recommended)**
Using `require.ErrorAs` from `testify/require` flattens the logic. It immediately fails the test if the type assertion fails, allowing the removal of the `if statusErr != nil` checks. We can also explicitly check against `metav1.StatusReasonInvalid`.

For `ValidateCreate`:
```go
if err != nil && tc.want != nil {
	// Verify it's specifically a 422 StatusError
	var statusErr *apierrors.StatusError
	require.ErrorAs(t, err, &statusErr, "expected *apierrors.StatusError")
	assert.Equal(t, int32(422), statusErr.ErrStatus.Code)
	assert.Equal(t, metav1.StatusReasonInvalid, statusErr.ErrStatus.Reason)

	// Verify all expected error substrings are present
	for _, wantErr := range strings.Split(tc.want.Error(), "\n") {
		assert.Contains(t, err.Error(), wantErr)
	}
}
```

For `ValidateUpdate`:
```go
if err != nil && tc.want != nil {
	var statusErr *apierrors.StatusError
	require.ErrorAs(t, err, &statusErr, "expected *apierrors.StatusError")
	assert.Equal(t, int32(422), statusErr.ErrStatus.Code)
	assert.Equal(t, metav1.StatusReasonInvalid, statusErr.ErrStatus.Reason)

	for _, cause := range statusErr.ErrStatus.Details.Causes {
		assert.Contains(t, tc.want.Error(), cause.Field)
		assert.Contains(t, cause.Message, "field is immutable")
	}
}
```

**Alternative 2: Update the Test Table**
Another option is to update the `tc.want` definitions in the test tables to return an `*apierrors.StatusError` directly. This would allow for a simple `assert.Equal(t, tc.want, err)` or a focused `cmp.Diff`, though it requires updating the expected values across the table cases.

---
*Note: This review comment was assisted by an AI assistant*.

pkg/webhooks/jobset_webhook_test.go
```diff
@@ -3448,9 +3457,21 @@ func TestValidateUpdate(t *testing.T) {
 			newObj := tc.js.DeepCopy()
 			oldObj := tc.oldJs.DeepCopy()
 			_, err := webhook.ValidateUpdate(context.TODO(), oldObj, newObj)
-			// Ignore bad value to keep test cases short and readable.
-			if diff := cmp.Diff(tc.want, err, cmpopts.IgnoreFields(field.Error{}, "BadValue")); diff != "" {
-				t.Errorf("ValidateResources() mismatch (-want +got):\n%s", diff)
+			if err != nil && tc.want != nil {
```
kannon92, line 3461: This test seems to have a lot of branching and is overly complicated to read.

I used AI to propose some ideas to simplify. https://github.com/kubernetes-sigs/jobset/pull/1207#pullrequestreview-4091891359

JerryHu1994: thanks, update the PR according to the recommended refactor

kannon92 (commented):
/lgtm
/approve

PR #1211, opened 2026-04-22
Jobset admission webhook validate each replicated job

JerryHu1994:
#### What type of PR is this?

/kind bug

The JobSet validation webhook only validated JobSet-level semantics (naming, cross-references, feature gates) but did not validate the inner Job/Pod template spec. Invalid specs like missing volumeMounts[0].mountPath were only caught when the controller tried to create the actual Job. As a result, jobset creation is accepted but later on replicated child job failed with invalid spec. It would be good to reject bad spec in the async path on create jobset.

Example failure
```
job "snr-base64-cks-tier2-3p3b-normsp-vidonly-0081d-w-0" creation failed with error: Job.batch "snr-base64-cks-tier2-3p3b-normsp-vidonly-0081d-w-0" is invalid: spec.template.spec.containers[0].volumeMounts[0].mountPath: Required value
```

Added server-side dry-run validation of each ReplicatedJob's Job template during webhook admission.

```
jieh@jieh-mac jobset % go test ./pkg/webhooks/ -count=1
ok      sigs.k8s.io/jobset/pkg/webhooks 0.760s
```

"NONE"

### DIFF
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -28,6 +28,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	apivalidation "k8s.io/apimachinery/pkg/api/validation"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/sets"
@@ -304,6 +305,9 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
 		allErrs = append(allErrs, j.validateVolumeClaimPolicies(ctx, js, js.Spec.VolumeClaimPolicies)...)
 	}
 
+	// Validate each ReplicatedJob's Job template via dry-run creation.
+	allErrs = append(allErrs, j.validateJobTemplates(ctx, js)...)
+
 	return nil, invalidError(js.Name, allErrs)
 }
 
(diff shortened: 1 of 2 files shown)


pkg/webhooks/jobset_webhook.go
```diff
@@ -342,6 +346,38 @@ func (j *jobSetWebhook) ValidateDelete(ctx context.Context, js *jobset.JobSet) (
...
+	for rJobIdx, rJob := range js.Spec.ReplicatedJobs {
+		fieldPath := field.NewPath("spec", "replicatedJobs").Index(rJobIdx).Child("template")
+		job := &batchv1.Job{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      placement.GenJobName(js.Name, rJob.Name, 0),
+				Namespace: js.Namespace,
+			},
+			Spec: *rJob.Template.Spec.DeepCopy(),
+		}
+		if err := j.client.Create(ctx, job, client.DryRunAll); err != nil {
```
kannon92, line 363: I don't think this is a good idea. We don't want to apply any potentially slow operator in the webhook.

Even if this is dry run, the api server could get slow and then this blocks updates.

kannon92: https://github.com/kubernetes-sigs/jobset/issues/422

Maybe we should follow up with this work.

JerryHu1994: can this be guarded under a knob, typically we expect the Create API to catch all user spec errors, if the jobset creation is accepted in this case but failed later, it is extremely hard to surface back error
Or if you have other better suggestions feel free to let me know

kannon92: I was chatting with @everettraven (our openshift API reviewer).

Putting API calls in a webhook is a bad idea for scalability. The webhook contract is supposed to be fast and any network operation can be unreliable. So there is a case that you could block creation of resources on long API calls. 

#422 was an idea to figure out how to validate.

Rather than try to validate, one option could be to emit a status field that the job creation failed. 

WDYT @andreyvelich @GiuseppeTT?

JerryHu1994: thanks for sharing the context, I can close this PR and track on https://github.com/kubernetes-sigs/jobset/issues/422

Huang-Wei: Can we make it a in-memory validation by leverage kubectl-validation's SDK? https://github.com/kubernetes-sigs/jobset/issues/422#issuecomment-2152913289

To clarify a bit, as a start, we don't need to achieve 100% parity with what k8s apiserver validates - that can only be achieved through a dry-run call.

JerryHu1994: let me try out the kubectl-validation option then ? @kannon92 how does it sound

andreyvelich: I agree with @kannon92, since we don't enforce any limits on how many ReplicatedJob can be defined within the single JobSet, this Create call can significantly affect performance. We already do API calls in validation [webhook for VolumeClaimTemplates](https://github.com/JerryHu1994/jobset/blob/0a69e8aaaa402c7312dcc96088df157465555c29/pkg/webhooks/jobset_webhook.go#L541-L545), but this is limited to [50 calls per JobSet](https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L192).


Given this comment from @danielvegamyhre on `kubectl-validate`: https://github.com/kubernetes-sigs/jobset/issues/422#issuecomment-2118528032, it looks like it cannot achieve 1-1 mapping with k8s APIs, right?
How would we translate this to users when they expect JobSet's webhook to throw an error in case some fields set incorrectly?

PR #1212, opened 2026-04-23
update k8s to 0.36.0

kannon92:
#### What type of PR is this?
/kind cleanup

Update k8s to 0.36

Fixes #

```release-note
NONE
```


kannon92:
/hold

This was failing locally. I think we need a release of controller-runtime with 0.36 to finish this out.

GiuseppeTT:
> This was failing locally. I think we need a release of controller-runtime with 0.36 to finish this out.

I agree. Any idea on when this new version of controller-runtime might be released? I am doing some tests with gang scheduling and I had to bump `controller-runtime` to head to make it work.

kannon92:
> > This was failing locally. I think we need a release of controller-runtime with 0.36 to finish this out.
> 
> I agree. Any idea on when this new version of controller-runtime might be released? I am doing some tests with gang scheduling and I had to bump `controller-runtime` to head to make it work.

I don't know but I did see a 1.36 bump yesterday in that repo. I'm sure it will come soon.

kannon92:
Looks like controller-runtime has a v0.24.0 release. Went ahead and bumped this dependency.

kannon92:
## Why `api/config/v1alpha1/doc.go` is needed

`zz_generated.defaults.go` keeps getting deleted on `make generate` because of an intentional change in `defaulter-gen` (part of `k8s.io/code-generator` v0.36.0).

**Root cause:** [kubernetes/kubernetes#135770](https://github.com/kubernetes/kubernetes/pull/135770) changed `defaulter-gen` to only generate defaults when the package-level `+k8s:defaulter-gen` tag is present, making it consistent with `deepcopy-gen`, `validation-gen`, and other code-generator tools. Previously, `defaulter-gen` would always generate a defaults file even without the package-level marker.

The `api/config/v1alpha1` package had `+k8s:defaulter-gen=true` on the `Configuration` struct but no package-level tag. What happens during `make generate`:

1. `kube_codegen.sh`'s `gen_helpers` greps for files containing `+k8s:defaulter-gen=` and finds the package (struct-level tag matches the grep).
2. It deletes any existing `zz_generated.defaults.go` in matched packages.
3. It runs `defaulter-gen` on the package.
4. `defaulter-gen` v0.36.0 now checks for a **package-level** `+k8s:defaulter-gen` comment (`extractTag(pkg.Comments)`), doesn't find one, logs `"did not find required tag"`, and skips the package entirely.
5. Result: file deleted, never regenerated.

**Fix:** Add `doc.go` with `// +k8s:defaulter-gen=TypeMeta` as a package-level comment. This is the standard pattern used by upstream Kubernetes packages (e.g., `k8s.io/api/core/v1/doc.go`). The `TypeMeta` value tells `defaulter-gen` to generate object defaulters for types embedding `TypeMeta`, and the existing struct-level `+k8s:defaulter-gen=true` on `Configuration` acts as an explicit opt-in.

GiuseppeTT:
@kannon92 let me know when this is ready for review.

kannon92:
@GiuseppeTT it is ready for review

GiuseppeTT:
/lgtm
/approve

Thanks for this work : )

kannon92:
Looks like there isn't any kind images for 1.36. I got to revert the test-infra change.

PR #1234, opened 2026-05-15
use kind build node images instead of relying on official images

### LINKED ISSUE #1233: Build our own node images for Kind
***What would you like to be cleaned:***

https://github.com/kubernetes-sigs/kind/issues/4157

Kind would prefer that we build the nodes images on releases without relying on official images.

***Why is this needed:***

This can put us in control of when we do e2e testing and not rely on kind to have an official release.

***What we need to do:***

I think we can switch our CI to do kind build node-image v1.36.0 and that will pull latest release artifacts. And then we can just use that cluster instead of relying on kind to host official images.

kannon92:
#### What type of PR is this?
/kind cleanup

Build kind images from upstream rather than relying on Kind to host official images.

Fixes #1233 

```release-note
NONE
```

### DIFF
--- a/hack/e2e-test.sh
+++ b/hack/e2e-test.sh
@@ -7,6 +7,7 @@ export NAMESPACE="jobset-system"
 export JOBSET_E2E_TESTS_DUMP_NAMESPACE=true
 # E2E config folder to use (defaults to "default")
 export E2E_TARGET_FOLDER="${E2E_TARGET_FOLDER:-default}"
+K8S_VERSION="${E2E_KIND_VERSION##*:}"
 
 # Use a temporary KUBECONFIG so that the script does not mess with the current user's kubeconfig.
 KUBECONFIG=""
@@ -25,6 +26,14 @@ function cleanup {
     fi
     (cd config/components/manager && $KUSTOMIZE edit set image controller=us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset:main)
 }
+
+function build_node_image {
+
+    if [ $USE_EXISTING_CLUSTER == 'false' ] 
+    then
+	   $KIND build node-image $K8S_VERSION --image jobset/kind-node:${K8S_VERSION}
+    fi
+}
 function startup {
     if [ $USE_EXISTING_CLUSTER == 'false' ] 
     then
@@ -34,7 +43,7 @@ function startup {
             exit 1
         fi
         export KUBECONFIG
-        $KIND create cluster --name $KIND_CLUSTER_NAME --image $E2E_KIND_VERSION --wait 1m
+        $KIND create cluster --name $KIND_CLUSTER_NAME --image jobset/kind-node:${K8S_VERSION}  --wait 1m
         kubectl get nodes > $ARTIFACTS/kind-nodes.log || true
         kubectl describe pods -n kube-system > $ARTIFACTS/kube-system-pods.log || true
     fi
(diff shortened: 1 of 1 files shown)


hack/e2e-test.sh
```diff
@@ -7,6 +7,7 @@ export NAMESPACE="jobset-system"
 export JOBSET_E2E_TESTS_DUMP_NAMESPACE=true
 # E2E config folder to use (defaults to "default")
 export E2E_TARGET_FOLDER="${E2E_TARGET_FOLDER:-default}"
+K8S_VERSION="${E2E_KIND_VERSION##*:}"
```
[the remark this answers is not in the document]
kannon92: I did this to avoid breaking the e2e tests.

We can probably refactor once this merges and we update test-infra.

Right now, this matches what we have in test-infra.

hack/e2e-test.sh
```diff
@@ -25,6 +26,14 @@ function cleanup {
     fi
     (cd config/components/manager && $KUSTOMIZE edit set image controller=us-central1-docker.pkg.dev/k8s-staging-images/jobset/jobset:main)
 }
+
+function build_node_image {
+
+    if [ $USE_EXISTING_CLUSTER == 'false' ] 
+    then
+	   $KIND build node-image $K8S_VERSION
+    fi
```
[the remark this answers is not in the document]
kannon92: This is a valid issue but the test-infra does set this.

hack/e2e-test.sh
```diff
@@ -34,7 +43,7 @@ function startup {
             exit 1
         fi
         export KUBECONFIG
-        $KIND create cluster --name $KIND_CLUSTER_NAME --image $E2E_KIND_VERSION --wait 1m
+        $KIND create cluster --name $KIND_CLUSTER_NAME --image kindest/node:latest --wait 1m
```
BenTheElder: you can also explicitly set the image name in the build command and then here, for local development that might make more sense, e.g. `jobset/kind-node:${K8s_VERSION}` just to avoid conflicts between say, another repo using another `kind` version

BenTheElder: The biggest downside of `kind build node-image` vs pulling a pre-built image *locally* is losing the docker cache (in CI you are probably not persisting the docker image cache because it's usually huge).


For local development, you could skip re-building your custom named image if it already exists. 

For extra security something roughly like  `jobset/kind-node:$(kind version)-${K8s_VERSION}`.

mimowo: cc @BenTheElder ptal at the sibling PR in Kueue: https://github.com/kubernetes-sigs/kueue/pull/11251

kannon92: updated.

GiuseppeTT:
> /assign @andreyvelich @GiuseppeTT

Looks good to me. Did you get a chance to test if the test-infra would work with this change?

kannon92:
> Looks good to me. Did you get a chance to test if the test-infra would work with this change?

Well the tests pull-jobset-test-e2e-main-* are exercising this code path.

So yes. I made the change so we don't need to also modify test-infra. We could simply our test-infra scripts but I will leave that as a followup. 

Simplication can be just a k8s tag (1.35, 1.36, ..) and we can remove the kindest image.

I also made this change in Kueue and it seems to have worked fine.

GiuseppeTT:
Gotcha. Thanks for the work!

/lgtm
/approve

PR #1235, opened 2026-05-15
feat: alpha annotation to publish exclusive-topology value as a pod label

abatilo:
#### What type of PR is this?

/kind feature
/kind api-change

Adds a new opt-in alpha annotation `alpha.jobset.sigs.k8s.io/exclusive-topology-label-key`. When set alongside the existing `alpha.jobset.sigs.k8s.io/exclusive-topology` annotation, the JobSet pod mutating webhook copies the resolved topology value (which it already computes for `spec.nodeSelector` on follower pods) onto the follower pod's `metadata.labels` under the user-supplied key. Containers consume it via the standard Downward API:

```yaml
metadata:
  annotations:
    alpha.jobset.sigs.k8s.io/exclusive-topology: topology.example.com/nvlink-domain
    alpha.jobset.sigs.k8s.io/exclusive-topology-label-key: ray.io/gpu-domain
spec:
  replicatedJobs:
  - name: worker
    template:
      spec:
        template:
          spec:
            containers:
            - name: ray-worker
              env:
              - name: RAY_GPU_DOMAIN
                valueFrom:
                  fieldRef:
                    fieldPath: metadata.labels['ray.io/gpu-domain']
```

The motivating use case is Ray's [Label Locality scheduling](https://github.com/ray-project/ray/pull/61442), which expects a Ray-level label `ray.io/gpu-domain` populated with a rack/NVLink-domain identifier. Without this feature, every workload that wants this value has to ship an init container that calls `GET /api/v1/nodes/<name>` with cluster-scoped `get nodes` RBAC. The JobSet pod webhook already resolves this value to write `spec.nodeSelector` for followers — this PR exposes it to the container as well.

Fixes #

This PR is provisional KEP + reference implementation in a single review unit. The KEP at `keps/NNNN-exclusive-topology-label/` carries `kep-number: NNNN`; please assign a real number during review. `reviewers:` and `approvers:` in `kep.yaml` are `- TBD` for the same reason.

- **Alpha-prefixed annotation, opt-in.** Workloads that don't set the new annotation see no behavior change.
- **Follower pods only.** Leader pods are out of scope for alpha because their bound node is not knowable at pod-create admission. A follow-up KEP would add a `pods/binding` admission webhook to cover them. KEP §Non-Goals and §Drawbacks document this explicitly.
- **No feature gate.** The annotation is itself the kill switch. New code paths only run when both `exclusive-topology` and the new annotation are set. `kep.yaml` carries `feature-gates: []` and `disable-supported: false` to make the decision explicit. KEP §"Design Details, point 4" walks through this with a fallback path if maintainers prefer a runtime gate.
- **Commits are split into 5 atomic logical changes**, each independently reviewable:
  1. `docs(keps): propose exclusive-topology-label-key alpha feature` — KEP only
  2. `feat(api): add ExclusiveTopologyLabelKey alpha annotation` — new constant only
  3. `feat(controller): propagate ExclusiveTopologyLabelKey to child Jobs` — mirrors existing `ExclusiveKey`/`NodeSelectorStrategyKey` propagation in `labelAndAnnotateObject`
  4. `feat(webhook): publish exclusive-topology value as a follower pod label` — pod mutating webhook + tests
  5. `feat(webhook): validate ExclusiveTopologyLabelKey annotation combinations` — JobSet `ValidateCreate`/`ValidateUpdate` + unit + integration tests
- **Tests.** Unit + integration. `make test` and `make test-integration` both pass locally. The pod webhook test runner compares error strings rather than `cmp.Diff`-ing typed errors so it can coexist with the new conflict-rejection assertion (`cmp.Diff` panics on `fmt.Errorf`'s unexported `errorString`); this matches the pattern the package's `ValidateCreate` runner already uses.

```release-note
Introduces the `alpha.jobset.sigs.k8s.io/exclusive-topology-label-key` annotation. When set alongside `alpha.jobset.sigs.k8s.io/exclusive-topology` (at the JobSet level or on a ReplicatedJob template), the JobSet pod mutating webhook publishes the resolved exclusive-topology value onto each follower pod as a label under the user-supplied key. Containers can then consume the value via the standard Downward API, eliminating the need for an init container that reads the node's label via the Kubernetes API. Leader pods are not labeled in alpha. The annotation is rejected at admission when paired with `alpha.jobset.sigs.k8s.io/node-selector`.
```

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -51,8 +51,20 @@ const (
 	// ensure exclusive job placement per topology, instead of injecting pod affinity/anti-affinites for this.
 	// The user must add the JobSet name node label to the desired topologies separately.
 	NodeSelectorStrategyKey string = "alpha.jobset.sigs.k8s.io/node-selector"
-	NamespacedJobKey        string = "alpha.jobset.sigs.k8s.io/namespaced-job"
-	NoScheduleTaintKey      string = "alpha.jobset.sigs.k8s.io/no-schedule"
+	// ExclusiveTopologyLabelKey is an annotation that can be set on the JobSet or on a
+	// ReplicatedJob template. When set alongside ExclusiveKey, the JobSet pod mutating
+	// webhook copies the resolved exclusive-topology value (read from the leader pod's
+	// Node label identified by ExclusiveKey) onto follower pods under this label key,
+	// so the container can read it via the Downward API
+	// (e.g. fieldRef: metadata.labels['ray.io/gpu-domain']).
+	//
+	// Applies only to follower pods (job completion index != 0). Leader pods are not
+	// labeled because the topology value is not knowable at pod-create admission time.
+	// Requires alpha.jobset.sigs.k8s.io/exclusive-topology on the same scope.
+	// Incompatible with alpha.jobset.sigs.k8s.io/node-selector.
+	ExclusiveTopologyLabelKey string = "alpha.jobset.sigs.k8s.io/exclusive-topology-label-key"
+	NamespacedJobKey          string = "alpha.jobset.sigs.k8s.io/namespaced-job"
+	NoScheduleTaintKey        string = "alpha.jobset.sigs.k8s.io/no-schedule"
 
 	// JobSetControllerName is the reserved value for the managedBy field for the built-in
 	// JobSet controller.
--- a/keps/NNNN-exclusive-topology-label/README.md
+++ b/keps/NNNN-exclusive-topology-label/README.md
@@ -1,0 +1,373 @@
+# KEP-NNNN: Publish resolved exclusive-topology value as a pod label
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Story](#user-story)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API](#api)
+  - [Implementation](#implementation)
+  - [Validation](#validation)
+  - [Conflict policy](#conflict-policy)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+  - [Init container with <code>get nodes</code> RBAC](#init-container-with--rbac)
+  - [Direct env var injection into containers](#direct-env-var-injection-into-containers)
+  - [Post-scheduling controller patching of leaders](#post-scheduling-controller-patching-of-leaders)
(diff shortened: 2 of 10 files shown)


kannon92:
Can you file a github issue on this?

NNNN corresponds to that github issue number.

kannon92:
cc @ahg-g @GiuseppeTT @imreddy13 @andreyvelich 

I would really like to figure out what the long term goal for exclusive-topology is. Should we continue to add alpha annotations? Or should we eventually move this as an API?

kannon92:
I haven't read the KEP but I am confused on dependency with Ray. Ray has KubeRay.

Are you deploying Ray in JobSet pods?

ahg-g:
> cc @ahg-g @GiuseppeTT @imreddy13 @andreyvelich
> 
> I would really like to figure out what the long term goal for exclusive-topology is. Should we continue to add alpha annotations? Or should we eventually move this as an API?

I recommend to move it to an API.

abatilo:
@kannon92 

1. Ray is just one example, this isn't exclusive to kuberay, but could be used with other things like Monarch
2. Kuberay does not solve this.
3. Yes, I use Ray with JobSet. It's much less work than administering KubeRay

The idea is basically, if the exclusive topology feature already knows the racks being picked, give us a way to forward this to the pod so that the pod can know as well. Even if I don't use Ray, this is still useful for building your world state in things like pytorch.

kannon92:
For 3, maybe you could submit an example to our site. This is separate from this PR.

I’ll try to look at this in two weeks. Conference and vacation so my availability is tough this week.

GiuseppeTT:
> I would really like to figure out what the long term goal for exclusive-topology is. Should we continue to add alpha annotations? Or should we eventually move this as an API?

From my experience, people are happy enough with it or jumped to TAS (topology aware scheduling) from Kueue.

IMO, exclusive topology should be kept at maintenance mode and we should focus on getting gang scheduling right, but I am happy to discuss this in an issue instead of in a PR.

For context:

- The exclusive topology feature is an early feature of JobSet that offered TAS before Kueue and upstream k8s had it
- Gang scheduling is an umbrella term for new upstream APIs (Workload API and PodGroup API) and their features. It will offer native support for TAS and is under active development (we already have feature gates at alpha in 1.36). We have plans to integrate it into JobSet

kannon92:
> > I would really like to figure out what the long term goal for exclusive-topology is. Should we continue to add alpha annotations? Or should we eventually move this as an API?
> 
> From my experience, people are happy enough with it or jumped to TAS (topology aware scheduling) from Kueue.
> 
> IMO, exclusive topology should be kept at maintenance mode and we should focus on getting gang scheduling right, but I am happy to discuss this in an issue instead of in a PR.
> 
> For context:
> 
> * The exclusive topology feature is an early feature of JobSet that offered TAS before Kueue and upstream k8s had it
> * Gang scheduling is an umbrella term for new upstream APIs (Workload API and PodGroup API) and their features. It will offer native support for TAS and is under active development (we already have feature gates at alpha in 1.36). We have plans to integrate it into JobSet

@abatilo 
I agree with this take. I think it would be worth filling this as an issue and we can discuss.

abatilo:
Going to close this as it looks like the aspiration is that more of this type of functionality actually goes to other projects! Thanks all.

PR #1243, opened 2026-06-02
Failing test 1 36

### LINKED ISSUE #1242: 1.36 e2e tests are failing
Looks like 1.36 tests are failing on controller startup. I saw this locally also where on first startup controller is stuck on rotating certs. Bouncing pods seems to be fix it.

kannon92:
#### What type of PR is this?
/kind bug

Fix bug when using self signed certs where we hit a race where the deployment won't start due to certificate rotation failing.

This seems to got a lot worse on 1.36.

I mainly copied existing behavior that Kueue has which doesn't seem to suffer from this issue.

Fixes #1242 

```release-note
Fix bootstrapping self signed certificates failure
```

## Summary by CodeRabbit

* **Refactor**
  * Improved startup sequencing: certificates are bootstrapped synchronously during startup and controller setup no longer waits on an internal readiness gate, resulting in more predictable controller initialization while health/readiness checks remain intact.
  
* **Tests**
  * Added unit tests covering certificate configuration and helper behavior.

### DIFF
--- a/main.go
+++ b/main.go
@@ -186,6 +186,15 @@ func main() {
 		KeyName:        options.Metrics.KeyName,
 	}
 	options.Metrics = metricsServerOptions
+
+	ctx := ctrl.SetupSignalHandler()
+	if cfg.InternalCertManagement != nil && *cfg.InternalCertManagement.Enable {
+		if err = cert.BootstrapCerts(ctx, kubeConfig, cfg); err != nil {
+			setupLog.Error(err, "Unable to bootstrap cert rotation")
+			os.Exit(1)
+		}
+	}
+
 	mgr, err := ctrl.NewManager(kubeConfig, options)
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
@@ -201,8 +210,6 @@ func main() {
 	} else {
 		close(certsReady)
 	}
-
-	ctx := ctrl.SetupSignalHandler()
 	if err := controllers.SetupJobSetIndexes(ctx, mgr.GetFieldIndexer()); err != nil {
 		setupLog.Error(err, "unable to setup jobset reconciler indexes")
 		os.Exit(1)
@@ -212,11 +219,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	// Cert won't be ready until manager starts, so start a goroutine here which
-	// will block until the cert is ready before setting up the controllers.
-	// Controllers who register after manager starts will start directly.
-	go setupControllers(mgr, certsReady)
-
+	setupControllers(mgr)
 	setupHealthzAndReadyzCheck(mgr, certsReady)
 
 	setupLog.Info("starting manager")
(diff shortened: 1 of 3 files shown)


kannon92:
cc @andreyvelich @GiuseppeTT

GiuseppeTT:
cc @jianqiaol

GiuseppeTT:
Quick context: the JobSet controller does some certification setup during [initialization](https://github.com/kubernetes-sigs/jobset/blob/f07add6f0ffc332aded36cfbf52fae6f4da594aa/main.go#L195-L203) (even before any JobSet object reconciliation is done). AFAIK this is only used to serve the JobSet wehbooks securely. The JobSet Pod doesn't become ready until this certification work is finished. It seems that the new k8s version 1.36 is affecting the certification work and this PR fixes it. This is specially important because Kevin just added CI tests for 1.36, which are blocking all PRs.

jianqiaol:
@kannon92 I will review this if you don't mind. I'm still ramping up but targeting to finish the review by EOW if not sooner.

kannon92:
@coderabbitai review

pkg/util/cert/cert.go
```diff
@@ -73,5 +79,71 @@ func CertsManager(mgr ctrl.Manager, cfg config.Configuration, setupFinish chan s
...
+	}
+}
+
+// BootstrapCerts creates a minimal manager to generate webhook certificates and
+// waits for them to be mounted before the main manager starts.
+func BootstrapCerts(ctx context.Context, kubeConfig *rest.Config, cfg config.Configuration) error {
+	namespace := getOperatorNamespace()
+	bootstrapMgr, err := ctrl.NewManager(kubeConfig, ctrl.Options{
+		Metrics:                metricsserver.Options{BindAddress: "0"},
+		HealthProbeBindAddress: cfg.Health.HealthProbeBindAddress,
```
jianqiaol: HealthProbeBindAddress: "0",

Let's set HealthProbeBindAddress: "0" for the bootstrapMgr to disable the health probe server entirely during the bootstrap phase.

Because this manager only runs for a fraction of a second, opening the health probe port here is risky:

It can cause integration tests to randomly flake with bind: address already in use collisions when running in parallel.
In production, if kubelet happens to ping the port exactly as bootstrapMgr is shutting down, controller-runtime's graceful shutdown will wait for that connection to finish, which will artificially delay the start of our actual main manager.

kannon92: Nice idea. done.

main.go
```diff
@@ -186,6 +186,15 @@ func main() {
...
 	}
 	options.Metrics = metricsServerOptions
+
+	ctx := ctrl.SetupSignalHandler()
+	if cfg.InternalCertManagement != nil && *cfg.InternalCertManagement.Enable {
+		if err = cert.BootstrapCerts(ctx, kubeConfig, cfg); err != nil {
+			setupLog.Error(err, "Unable to bootstrap cert rotation")
+			os.Exit(1)
+		}
+	}
```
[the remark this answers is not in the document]
kannon92: This sorta brings up a gap that we don't actually have leader election for JobSet so we should add that as a feature.

jianqiaol (commented):
thanks for the update! /lgtm

GiuseppeTT:
Thanks for the work!

/lgtm
/approve

Issue #796, opened 2025-02-24
Controller panics when reconciling JobSet which is not mutated by webhook


ChenYi015:
**What happened**:

I create a JobSet before installing the jobset controller/webhook, which means the JobSet is not mutated by the webhook, then I install the jobset and the controller panics: 

```text
2025-02-24T08:44:21Z    ERROR    Observed a panic    {"controller": "jobset", "controllerGroup": "jobset.x-k8s.io", "controllerKind": "JobSet", "JobSet": {"name":"demo","n
amespace":"default"}, "namespace": "default", "name": "demo", "reconcileID": "8c487ccd-e84e-4494-adbb-ca3584275adf", "panic": "runtime error: invalid memory address or nil
 pointer dereference", "panicGoValue": "\"invalid memory address or nil pointer dereference\"", "stacktrace": "goroutine 313 [running]:\nk8s.io/apimachinery/pkg/util/runti
me.logPanic({0x1d8b168, 0xc0009219e0}, {0x189e500, 0x2ae5af0})\n\t/go/pkg/mod/k8s.io/apimachinery@v0.31.1/pkg/util/runtime/runtime.go:107 +0xbc\nsigs.k8s.io/controller-run
time/pkg/internal/controller.(*Controller[...]).Reconcile.func1()\n\t/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.19.0/pkg/internal/controller/controller.go:105 +0x114\np
anic({0x189e500?, 0x2ae5af0?})\n\t/usr/local/go/src/runtime/panic.go:785 +0x132\nsigs.k8s.io/jobset/pkg/controllers.dnsHostnamesEnabled(...)\n\t/workspace/pkg/controllers/
jobset_controller.go:848\nsigs.k8s.io/jobset/pkg/controllers.(*JobSetReconciler).createHeadlessSvcIfNecessary(0xc0007bad80, {0x1d8b168, 0xc000921ce0}, 0xc000ba1180)\n\t/wo
rkspace/pkg/controllers/jobset_controller.go:590 +0x66\nsigs.k8s.io/jobset/pkg/controllers.(*JobSetReconciler).reconcile(0xc0007bad80, {0x1d8b168, 0xc0009219e0}, 0xc000ba1
[...]

PR #1266, opened 2026-07-06
fix: avoid nil network panic during JobSet reconciliation

immanuwell:
#### What type of PR is this?
/kind bug

This adds a nil check before reading `spec.network.enableDNSHostnames`.
Without it, an old JobSet that was created before webhook defaulting can blow up the controller with a nil pointer panic. Small fix, nasty bug.

Related to #796

Used an AI coding tool for the initial patch and test draft, then reviewed and verified the final change locally.

Repro:
1. Create a JobSet before installing the JobSet webhook/controller, so `.spec.network` stays unset.
2. Install or start JobSet.
3. Reconcile hits `dnsHostnamesEnabled()` and panics on a nil `spec.network`.

Local checks:
- `go test ./pkg/controllers -run TestCreateHeadlessSvcIfNecessary`
- `make test`
- `make verify` got through vet, ci-lint, lint-api, manifests, codegen and helm generation before I stopped it during the later docs step on first-run tool/image setup.

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented a crash when job network settings are missing, so job reconciliation handles unset network configuration safely.
  * Improved headless service creation logic to avoid creating a service when the job configuration has not been defaulted yet.

* **Tests**
  * Added coverage for jobs created before defaulting to confirm no headless service is created in that case.

### DIFF
--- a/pkg/controllers/jobset_controller.go
+++ b/pkg/controllers/jobset_controller.go
@@ -1152,7 +1152,7 @@ func jobSetMarkedForDeletion(js *jobset.JobSet) bool {
 }
 
 func dnsHostnamesEnabled(js *jobset.JobSet) bool {
-	return js.Spec.Network.EnableDNSHostnames != nil && *js.Spec.Network.EnableDNSHostnames
+	return js.Spec.Network != nil && js.Spec.Network.EnableDNSHostnames != nil && *js.Spec.Network.EnableDNSHostnames
 }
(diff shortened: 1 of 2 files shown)


kannon92:
@immanuwell 

did you see this in production? Trying to gauge how serious this issue is?

PR #1267, opened 2026-07-07
Fix validation for generateName-derived child resource names

immanuwell:
#### What type of PR is this?
/kind bug
/kind regression

Sneaky little gap in `ValidateCreate`: it validated derived Job, Pod, coordinator, and PVC names with `js.Name`, but `js.Name` is still empty on create requests that use `metadata.generateName`.

That means an overlong `generateName` prefix can slip through admission, then blow up later when the controller creates child resources. This patch validates against the effective generated JobSet name instead, using the apiserver's own truncation and 5 char suffix behavior. Added regression tests for child Job and PVC names too.

None

Used AI assistance to help prepare and validate this change.

Repro:

```yaml
kind: JobSet
metadata:
  generateName: aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa-
spec:
  successPolicy:
    operator: All
  replicatedJobs:
  - name: worker
    groupName: default
    replicas: 1
    template:
      spec:
        template:
          spec:
            restartPolicy: Never
            containers:
            - name: test
              image: bash:latest
```

Before this fix, `kubectl apply -f repro.yaml` gets through admission. Then the apiserver adds the suffix, and the controller tries to create an invalid child Job name. Kinda rough.

```release-note
JobSet creation now rejects `metadata.generateName` prefixes that would produce invalid child Job or PVC names after the apiserver assigns the final JobSet name.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation for resources created with generated names, ensuring name-length checks work correctly even when the main name is not set.
  * Added stricter checks to reject configurations that would produce overly long child job or PVC names.
  * Improved consistency in coordinator label and volume claim policy validation when names are derived from generated prefixes.

* **Tests**
  * Added coverage for generated-name validation behavior and overlong derived resource names.

### DIFF
--- a/pkg/webhooks/jobset_webhook.go
+++ b/pkg/webhooks/jobset_webhook.go
@@ -34,6 +34,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	"k8s.io/apiserver/pkg/storage/names"
 	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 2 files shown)


kannon92 (commented):
TY!

/lgtm
/approve

PR #1268, opened 2026-07-09
fix flake for elastic job by using standard feature gate toggling

### LINKED ISSUE #1258: Flaky integration test: "rejects updates to parallelism and completions" (ElasticJobSet feature gate disabled) fails intermittently on release-0.12
## What happened

`periodic-jobset-test-integration-release-0-12` on
https://testgrid.k8s.io/sig-apps-jobset-periodics is flaky: over the last 30
runs, 2 runs failed (~6.7%) with the exact same integration test failure,
while every other test in the suite (65 specs total) passes consistently.

Failing test:

```
Controller Suite.[It] JobSet controller when the ElasticJobSet feature gate is disabled
  Elastic scaling behavior is blocked
    rejects updates to parallelism and completions
```
`test/integration/controller/jobset_controller_test.go:3324` (assertion at line 3361)

## Root cause

Confirmed identical failure output in two separate builds:

- build `2073433869817745408` (2026-07-04)
- build `2070531836869611520` (2026-06-26)

```
[FAILED] API Server should reject the JobSet update when ElasticJobSet feature gate is disabled
  Expected an error to have occurred.  Got:
      <nil>: nil
In [It] at: /home/prow/go/src/sigs.k8s.io/jobset/test/integration/controller/jobset_controller_test.go:3361
```

The test disables the `ElasticJobSet` feature gate in its `BeforeEach`
(`features.SetEnable(features.ElasticJobSet, false)`), then attempts to
mutate `Parallelism`/`Completions` on an existing JobSet via
`k8sClient.Update`, expecting the validating webhook to reject the update
because the feature gate is off. In the failing runs, the update instead
succeeds (no error), meaning the webhook did not reject it as expected.

This test sits right next to a sibling `ginkgo.When("the ElasticJobSet
feature gate is enabled", ...)` block (`jobset_controller_test.go:3133`)
whose `BeforeEach`/`AfterEach` toggle the *same* global, process-wide
feature gate (`utilfeature.DefaultMutableFeatureGate`) true/false directly
via `features.SetEnable`, rather than using the
`SetFeatureGateDuringTest`/`featuregatetesting` helper (which registers an
automatic `t.Cleanup` reset). Since the feature gate is shared, mutable,
global state, and Ginkgo does not guarantee spec run order between sibling
`When` blocks (specs are randomized by default), any residual/async
work still touching this state around spec boundaries is a likely
contributor to an intermittent, hard-to-pin-down race — though it did not
reproduce locally in a quick attempt and needs a maintainer with more
context on the webhook/reconciler's use of this feature gate to confirm
the exact mechanism.

## Evidence / reproduction

- testgrid tab: `periodic-jobset-test-integration-release-0-12`
  (dashboard `sig-apps-jobset-periodics`)
- Failing builds: `2073433869817745408`, `2070531836869611520` — both show
  byte-for-byte identical failure output at the same test/line.
- All other 64 specs pass in both of those runs; only this one test flakes.

## Expected behavior

`rejects updates to parallelism and completions` should deterministically
fail the `Update` call whenever `ElasticJobSet` is disabled, on every run.

## Suggested fix

- Investigate whether the "feature gate enabled" and "feature gate
  disabled" `ginkgo.When` blocks for `ElasticJobSet` scaling
  (`jobset_controller_test.go:3133` and `:3282`) can leave the global
  feature gate in an inconsistent state relative to spec execution order,
  and consider switching to `SetFeatureGateDuringTest`/`ginkgo.Ordered` (or
  otherwise guaranteeing full isolation) for these tests.
- Add a short `gomega.Eventually` (instead of a single point-in-time
  `Update`) or otherwise make the assertion robust to timing so a genuine
  root cause is easier to confirm from a future failure.

This investigation (pulling and analyzing testgrid data and CI logs across
two separate failing runs, and drafting this issue) was assisted by an AI
coding agent, per the [Kubernetes AI Tool Usage Policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance).
The suggested root cause is a hypothesis based on code inspection, not a
confirmed fix — all findings were reviewed before filing, and further
investigation may be needed to pin down the exact mechanism.

kannon92:
#### What type of PR is this?
/kind flake

Fix flake for elastic jobs.

Fixes #1258 

A little boy scouting but the SetEnabled was an odd pattern.

Tests should not have side effects.

```release-note
NONE
```

## Summary by CodeRabbit

* **Tests**
  * Updated integration and webhook tests to configure feature gates per test case using the dedicated test-only gate helper.
  * Ensures “RestartJob” and “ElasticJobSet” scenarios enable/disable the feature at the start of each relevant table entry, improving isolation and reliability for elastic scaling and webhook behavior coverage.

### DIFF
--- a/pkg/features/features.go
+++ b/pkg/features/features.go
@@ -17,7 +17,6 @@ limitations under the License.
 package features
 
 import (
-	"fmt"
 	"testing"
 
 	"k8s.io/apimachinery/pkg/util/runtime"
@@ -82,11 +81,4 @@ func SetFeatureGateDuringTest(tb testing.TB, f featuregate.Feature, value bool) 
 // Enabled is helper for `utilfeature.DefaultFeatureGate.Enabled()`
 func Enabled(f featuregate.Feature) bool {
 	return utilfeature.DefaultFeatureGate.Enabled(f)
-}
-
-// SetEnable helper function that can be used to set the enabled value of a feature gate,
-// it should only be used in integration test pending the merge of
-// https://github.com/kubernetes/kubernetes/pull/118346
-func SetEnable(f featuregate.Feature, v bool) error {
-	return utilfeature.DefaultMutableFeatureGate.Set(fmt.Sprintf("%s=%v", f, v))
 }
--- a/test/integration/controller/jobset_controller_test.go
+++ b/test/integration/controller/jobset_controller_test.go
@@ -791,7 +791,7 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 		}),
 		ginkgo.Entry("[failure policy] jobset restarts with RestartJobSet failure policy action (with RestartJob feature gate).", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
-				gomega.Expect(features.SetEnable(features.RestartJob, true)).To(gomega.Succeed())
+				features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.RestartJob, true)
 				return testJobSet(ns).
 					FailurePolicy(&jobset.FailurePolicy{
 						MaxRestarts: 1,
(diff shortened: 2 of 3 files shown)


aniket2405:
This looks good @kannon92 
I missed `SetFeatureGateDuringTest` while writing the integration tests for this feature, that will keep the test env isolated from rest of the env.

Thanks!

andreyvelich (commented):
Thanks @kannon92!
/lgtm
/approve

PR #1269, opened 2026-07-09
Fix exclusive placement repair for mismatched follower pods

immanuwell:
#### What type of PR is this?
/kind bug
/kind regression

Exclusive placement already tries to repair follower pods that no longer match the leader pod topology.
Right now a plain topology mismatch returns an error from `validatePodPlacements`, so reconcile bails out before `deleteFollowerPods` runs. The bad follower just sticks around.

This change treats a mismatch as a repair case, not a hard failure, and adds a regression test for the reconcile path.

Repro:
1. Create a leader pod for an exclusive-placement job on a node labeled `topologyKey=expected-topology`.
2. Create a follower pod for the same job with `spec.nodeSelector[topologyKey]=wrong-topology`.
3. Reconcile the leader pod.
4. Before this fix, reconcile returns a topology mismatch error and does not delete the follower pod.
5. With this fix, reconcile marks the follower pod with `DisruptionTarget` and deletes it, so the job can heal itself.

NONE

Used minor AI assist (primarily to polish PR's description)

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod placement validation now treats follower/leader topology mismatches as recoverable, enabling automatic correction rather than failing reconciliation.
  * Reconciliation now deletes mismatched follower pods and updates their status when the leader topology is known.

* **Tests**
  * Added coverage for reconciliation behavior that removes mismatched follower pods.
  * Updated existing validation tests to reflect the new non-fatal handling of follower topology mismatch scenarios.

### DIFF
--- a/pkg/controllers/pod_controller.go
+++ b/pkg/controllers/pod_controller.go
@@ -190,7 +190,10 @@ func (r *PodReconciler) listPodsForJob(ctx context.Context, ns, jobKey string, o
 }
 
 // validatePodPlacements returns true if the topology value specified in follower pods' nodeSelectors
-// matches that of their leader pod, and the leader pod exists. Otherwise, it returns false.
+// matches that of their leader pod, and the leader pod exists. It returns false without an error
+// when follower placement data is malformed or mismatched so the reconciler can repair the job by
+// deleting follower pods. If the leader topology is temporarily unknown, it returns an error so
+// reconciliation retries instead of deleting followers based on incomplete information.
 func (r *PodReconciler) validatePodPlacements(ctx context.Context, leaderPod *corev1.Pod, podList *corev1.PodList) (bool, error) {
 	// We know exclusive key is set since we have an event filter for this.
 	topologyKey := leaderPod.Annotations[jobset.ExclusiveKey]
@@ -205,11 +208,11 @@ func (r *PodReconciler) validatePodPlacements(ctx context.Context, leaderPod *co
 			continue
 		}
 		followerTopology, err := followerPodTopology(&pod, topologyKey)
-		if err != nil {
-			return false, err
-		}
-		if followerTopology != leaderTopology {
-			return false, fmt.Errorf("follower topology %q != leader topology %q", followerTopology, leaderTopology)
+		if err != nil || followerTopology != leaderTopology {
+			if leaderTopology == "" {
+				return false, fmt.Errorf("leader topology is unknown (node not found)")
+			}
+			return false, nil
 		}
 	}
 	return true, nil
--- a/pkg/controllers/pod_controller_test.go
+++ b/pkg/controllers/pod_controller_test.go
@@ -21,13 +21,17 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


GiuseppeTT:
@jianqiaol

pkg/controllers/pod_controller.go
```diff
@@ -209,7 +210,7 @@ func (r *PodReconciler) validatePodPlacements(ctx context.Context, leaderPod *co
 			return false, err
 		}
 		if followerTopology != leaderTopology {
```
jianqiaol: Great catch fixing the unreachable repair logic! However, this change introduces a regression risk during transient node unavailability (e.g., node scale-downs, preemptions, or API cache delays).

In leaderPodTopology, if the Node is not found, it intentionally ignores the error and returns ("", nil). With this PR applied, validatePodPlacements will evaluate followerTopology != "" and return (false, nil), which causes the reconciler to aggressively delete all follower pods due to a transient cache miss.

We should only delete followers when we are certain there is a mismatch. I recommend updating this block to return an error if the leader topology is unknown so it safely backs off:
```
		if followerTopology != leaderTopology {
			if leaderTopology == "" {
				// The leader pod's node was not found. 
				// Return an error to retry instead of aggressively deleting follower pods.
				return false, fmt.Errorf("leader topology is unknown (node not found)")
			}
			return false, nil
		}

```

pkg/controllers/pod_controller.go
jianqiaol, line 210: If a follower pod is entirely missing its NodeSelector or topologyKey, followerPodTopology returns an error here. While this is a rare edge case (likely only caused by a race condition or an interfering mutating webhook), returning an error bubbles up to the reconciler and permanently bricks the JobSet in an infinite retry loop—which is the exact bug this PR is fixing.

We should treat this error as a mismatch so the reconciler deletes the malformed pod and allows the system to self-heal. We can do this by combining the error check with the mismatch check:
```
followerTopology, err := followerPodTopology(&pod, topologyKey)
		if err != nil || followerTopology != leaderTopology {
			if leaderTopology == "" {
				return false, fmt.Errorf("leader topology is unknown (node not found)")
			}
			return false, nil
		}
```

pkg/controllers/pod_controller.go
immanuwell: done in the latest push

pkg/controllers/pod_controller.go
immanuwell: fixed in the latest push

immanuwell:
@jianqiaol addressed both your suggestions, PTAL

immanuwell:
@jianqiaol WDYT?

GiuseppeTT:
> /lgtm Thanks for the update!

/lgtm
/approve

GiuseppeTT:
/retest

Might be flake

GiuseppeTT:
/hold

In case it's not a flake

kannon92:
> /hold
> 
> In case it's not a flake

I've been chasing this one for a while. https://github.com/kubernetes-sigs/jobset/pull/1278

Issue #1150, opened 2026-02-12
Enable Mutable JobSet requests and limits on suspended state


kannon92:
**What would you like to be added**:

I would like JobSet to be able to modify PodTemplates for requests/limits while the JobSet is suspended.

**Why is this needed**:

https://github.com/kubernetes/enhancements/issues/5440

In the above KEP, users are able to modify requests/limits while a Job is suspended. This enables easier admission integration with Kueue (modifying resources to "fit" workloads or enable right sizing by an external controller).

There is interest in AI/ML workloads being able to right size resources and this would be required for Trainer.

This enhancement requires the following artifacts:

- [x] Design doc
- [ ] API change
- [ ] Docs update

The artifacts should be linked in subsequent comments.

kannon92:
The above feature will be beta in 1.36 but it is alpha in 1.35.

We probably don't need to rely on this feature though. If we want to update the existing suspended jobs (JobSet suspends Jobs), then this would be needed.

We could consider deleting the Jobs on suspended state and just allow the mutation for templates.

kannon92:
@aniket2405 
I'm going to take this. I want to prove out this work for KEP-5440 for Kubernetes.

PR #1273, opened 2026-07-09
KEP-1150: Enable MutablePod resources with JobSet

kannon92:
#### What type of PR is this?
/kind feature

#### What this PR does / why we need it:
Allow the use of KEP-5440 with JobSet.
#### Which issue(s) this PR fixes:

Fixes #

#### Does this PR introduce a user-facing change?

```release-note
Allow for mutating pod resources from the jobset api while jobsets are suspended.
```

## Summary by CodeRabbit

* **New Features**
  * Added support for updating resource requests and limits on suspended or suspending JobSets when enabled.
  * Propagated matching container and init-container resources to child workloads when a JobSet resumes.
  * Added execution-attempt tracking and promoted TLS options to generally available.

* **Bug Fixes**
  * Preserved other container settings and resource claims during updates.
  * Added warning events when resource propagation fails.

* **Tests**
  * Expanded unit, integration, webhook, and end-to-end coverage.

PR #1274, opened 2026-07-10
add a KEP for mutable suspended job resources

kannon92:
#### What type of PR is this?
/kind feature

KEP for #1150.

Allow for use of JobSet and MutableJob feature.

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Documentation**
  * Added KEP-1150 documentation describing editable container and init-container resource requests and limits while JobSets are suspended.
  * Documented that resource changes are applied to child Jobs when the JobSet resumes.
  * Defined the alpha feature-gate behavior, validation expectations, and milestone criteria.
  * Added metadata identifying KEP-1150 as an alpha-stage, implementable proposal.

### DIFF
--- a/keps/1150-mutable-suspended-job-resources/README.md
+++ b/keps/1150-mutable-suspended-job-resources/README.md
@@ -1,0 +1,273 @@
+# KEP-1150: Mutable JobSet Requests and Limits on Suspended State
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1: Kueue right-sizing before admission](#story-1-kueue-right-sizing-before-admission)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Feature Gate](#feature-gate)
+  - [Webhook Changes](#webhook-changes)
+  - [Controller Changes](#controller-changes)
+  - [Test Plan](#test-plan)
+    - [Unit Tests](#unit-tests)
+    - [Integration Tests](#integration-tests)
+    - [E2E Tests](#e2e-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This KEP proposes allowing container and init-container resource
+requests/limits within a JobSet's `ReplicatedJob` Pod templates to be
+mutated while the JobSet is suspended. This mirrors capabilities JobSet
+already provides for other Pod template fields (annotations, labels,
+`nodeSelector`, `tolerations`, `schedulingGates`) and builds on the
+upstream Kubernetes [KEP-5440], which allows a suspended Job's Pod
+template resources to be mutated via the
+`MutablePodResourcesForSuspendedJobs` feature gate.
+
+[KEP-5440]: https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/5440-mutable-job-pod-resource-updates/README.md
+
+## Motivation
+
+JobSet suspends its child Jobs to integrate with external
+admission/scheduling systems such as Kueue. These integrations often need
+to adjust a workload's requested resources *after* submission but *before*
+it starts — for example, Kueue right-sizing requests to fit available
+quota or a chosen `ResourceFlavor`.
+
+Today, `spec.replicatedJobs` is immutable after creation, except for a
+hand-picked set of Pod template fields allowed to change while suspended.
+This KEP adds container/init-container `resources` to that allow-list.
+
+### Goals
+
+- Allow only the `resources.requests` and `resources.limits` fields of
+  `spec.replicatedJobs[].template.spec.template.spec.containers[].resources`
+  and `initContainers[].resources` to be mutated while a JobSet is
+  suspended (mirroring existing mutable-while-suspended fields). Existing
+  `resources.claims` entries in the Pod template are preserved and are not
+  treated as mutable.
+- Propagate resource mutations to child `Job` objects on resume.
+- Gate this behind a new `MutablePodResourcesForSuspendedJobSets` feature gate,
+  disabled by default at alpha.
+
+### Non-Goals
+
+- Introducing new API fields. This KEP only relaxes an existing
+  immutability constraint.
+- Resource mutation for running (non-suspended) JobSets or active Jobs.
+- Mutating any container field other than `resources` while suspended.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Kueue right-sizing before admission
+
+A user submits a suspended JobSet pending Kueue admission:
+
+```yaml
+apiVersion: jobset.x-k8s.io/v1alpha2
+kind: JobSet
+metadata:
+  name: training-job
+spec:
+  suspend: true
+  replicatedJobs:
+  - name: workers
+    replicas: 4
+    template:
+      spec:
+        template:
+          spec:
+            containers:
+            - name: trainer
+              image: my-trainer:latest
+              resources:
+                requests:
+                  cpu: "4"
+                  nvidia.com/gpu: "1"
+```
+
+Kueue patches the container resources while suspended, then unsuspends.
+The patch targets `/resources/requests` and `/resources/limits`
+specifically, rather than replacing the whole `resources` object, so any
+existing `limits` or `claims` on the container are preserved:
+
+```bash
+kubectl patch jobset training-job --type json --patch '[
+  {"op": "replace",
+   "path": "/spec/replicatedJobs/0/template/spec/template/spec/containers/0/resources/requests",
+   "value": {"cpu": "2", "nvidia.com/gpu": "1"}}
+]'
+```
+
+On resume, the updated resources propagate to child Jobs and their Pods.
+
+### Notes/Constraints/Caveats
+
+- **Upstream dependency:** This feature requires the Kubernetes
+  `MutablePodResourcesForSuspendedJobs` feature gate ([KEP-5440]),
+  introduced as alpha in Kubernetes 1.35. Without it, JobSet's webhook
+  accepts the update but kube-apiserver rejects the child Job update.
+  Cluster operators must enable both feature gates.
+- **Propagation timing:** As with other mutable-while-suspended fields,
+  changes propagate to child Jobs only on resume, not immediately.
+
+### Risks and Mitigations
+
+- **Risk:** Enabling `MutablePodResourcesForSuspendedJobSets` without the upstream
+  Kubernetes feature gate causes confusing failures (webhook accepts,
+  controller rejected by kube-apiserver).
+  **Mitigation:** Documented in feature gate description. Additionally,
+  rather than retrying silently, the controller records a `Warning` Event
+  on the `JobSet` when a child Job resource update is rejected on resume
+  (e.g. `Reason: FailedResourcePropagation`, `Message: Failed to propagate
+  mutated resources to child Job: <error>. Ensure
+  MutablePodResourcesForSuspendedJobs is enabled on the API server.`), so
+  operators have a visible signal rather than a silent backoff loop. A
+  worked example demonstrating correct setup is planned for the
+  implementation PR.
+
+## Design Details
+
+### API Changes
+
+None. This KEP relaxes an existing immutability check, gated by a feature
+gate. No fields are added, removed, or changed.
+
+### Feature Gate
+
+A new `MutablePodResourcesForSuspendedJobSets` feature gate is added in
+`pkg/features/features.go`.
+
+| Stage | Default | Since |
+|---|---|---|
+| Alpha | `false` | v0.13 |
+
+### Webhook Changes
+
+`jobSetWebhook.ValidateUpdate` already builds a "munged" copy of the new
+spec, overwriting mutable Pod template fields with old values before the
+immutability comparison.
+
+When `MutablePodResourcesForSuspendedJobSets` is enabled, a new
+`mungeContainerResources` helper copies each container's `Resources` from
+the old spec (matched by name), leaving all other container fields subject
+to the normal immutability check. When disabled, container resources remain
+fully immutable while suspended.
+
+### Controller Changes
+
+`JobSetReconciler.resumeJob` already propagates mutable Pod template
+fields to child Jobs on resume. This KEP extends it to also propagate
+container/init-container resources (matched by container name) when
+`MutablePodResourcesForSuspendedJobSets` is enabled. If the child Job update is
+rejected by kube-apiserver (e.g. because the upstream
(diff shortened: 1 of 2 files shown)


keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,229 @@
...
+   "value": {"requests": {"cpu": "2", "nvidia.com/gpu": "1"}}}
+]'
+```
+
+On resume, the updated resources propagate to child Jobs and their Pods.
+
+### Notes/Constraints/Caveats
+
+- **Upstream dependency:** This feature requires the Kubernetes
+  `MutablePodResourcesForSuspendedJobs` feature gate ([KEP-5440]),
```
VassilisVassiliadis, line 121: Is it worth pointing to the KEP here?

```suggestion
  `MutablePodResourcesForSuspendedJobs` feature gate ([KEP-5440](https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/5440-mutable-job-pod-resource-updates/README.md)),
```

As a quick sanity check, I had a look at the implementation of KEP-5440 on the latest commit to Kubernetes: in https://github.com/kubernetes/kubernetes/blob/5f91251764f106ba06c25a612d646ff612392e35/pkg/apis/batch/validation/validation.go#L768-L779

Indeed, the `MutablePodResourcesForSuspendedJobs` featuregate enables the mutation of the resources of a Job that was suspended prior to the patch. Thus, it is perfectly valid to delay mutating the podTemplate resources till we're about to resume the child Job objects.

VassilisVassiliadis (approved):
Overall this looks excellent to me, I only had a couple of minor comments. I'm really keen on having support for mutable suspended job resources. We want to enable the automatic configuration of resources for TrainJob Objects (https://github.com/kubeflow/trainer/issues/3328) and having this new functionality for Jobs is going to be a huge enabler for us.

kannon92:
/assign @jianqiaol @andreyvelich 

Can one of you take a look on this?

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+`jobSetWebhook.ValidateUpdate` already builds a "munged" copy of the new
+spec, overwriting mutable Pod template fields with old values before the
+immutability comparison.
+
+When `SuspendedJobResourceMutation` is enabled, a new
+`mungeContainerResources` helper copies each container's `Resources` from
+the old spec (matched by name), leaving all other container fields subject
+to the normal immutability check. When disabled, container resources remain
+fully immutable while suspended.
```
jianqiaol: In `Webhook Changes`, the text states that `mungeContainerResources` copies each container's entire `Resources` from the old spec.

Because `corev1.ResourceRequirements` contains `Claims` (for Dynamic Resource Allocation) alongside `Requests` and `Limits`, copying the entire `Resources` struct erases any user modifications to `Claims` during munging before `DeepEqual` runs. This causes illegal mutations to `resources.claims` to silently pass webhook validation. Later, when the controller attempts to resume the child `Job` with modified claims, `kube-apiserver` will reject the update because upstream KEP-5440 strictly forbids mutating `claims`.

**Action**: Please clarify that `mungeContainerResources` must perform **field-level munging exclusively on `Requests` and `Limits`**, leaving `Claims` untouched so modifications are caught and rejected by `DeepEqual`.

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+
```
jianqiaol, line 170: Relaxing immutability allows new resource values to be submitted. The webhook must ensure that mutated `Requests` and `Limits` are semantically valid before admitting the update:
- `requests[r] <= limits[r]` for standard compute resources (CPU, Memory).
- Extended resources (e.g. `nvidia.com/gpu`) have matching requests and limits.
- Valid resource names and non-negative quantities.
- Container list length, names, and ordering remain invariant (no container additions/removals).
Without explicit validation during `ValidateUpdate`, invalid updates (e.g. `requests > limits`) would succeed at admission and only fail asynchronously in the reconciler when resuming child Jobs.
**Action**: Please add a note under `Webhook Changes` specifying that mutated container resources undergo standard resource validation (`corevalidation.ValidateResourceRequirements` or equivalent).

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+  Cluster operators must enable both feature gates.
+- **Propagation timing:** As with other mutable-while-suspended fields,
+  changes propagate to child Jobs only on resume, not immediately.
+
+### Risks and Mitigations
+
+- **Risk:** Enabling `SuspendedJobResourceMutation` without the upstream
+  Kubernetes feature gate causes confusing failures (webhook accepts,
+  controller rejected by kube-apiserver).
+  **Mitigation:** Documented in feature gate description. A worked
```
jianqiaol: If `SuspendedJobResourceMutation` is enabled in JobSet but the upstream `MutablePodResourcesForSuspendedJobs` gate is disabled on `kube-apiserver`, child Job updates will be rejected with 422 upon resume. Rather than retrying silently in reconcile backoff, the controller should record a Kubernetes `Warning` Event on the `JobSet` (e.g., `Reason: FailedResourcePropagation`, `Message: Failed to propagate mutated resources to child Job: <error>. Ensure MutablePodResourcesForSuspendedJobs is enabled on the API server.`).

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+container/init-container resources (matched by container name) when
+`SuspendedJobResourceMutation` is enabled.
+
+### Test Plan
+
+[x] I/we understand the owners of the involved components may require updates to
+existing tests to make this code solid enough prior to committing the changes necessary
+to implement this enhancement.
+
+#### Unit Tests
```
jianqiaol, line 188: Please expand the **Test Plan** to explicitly list negative test cases:
1. **Claims Immutability**: Assert that attempting to mutate `resources.claims` while suspended is rejected by webhook validation.
2. **Container Invariance**: Assert that adding, removing, or reordering containers during resource mutation is rejected.
3. **Invalid Resource Semantics**: Assert that invalid resource configurations (e.g. `requests > limits`) are rejected synchronously at admission.
4. **Rejection Handling**: Integration test verifying Warning Event emission when child Job updates fail on resume.

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+                  nvidia.com/gpu: "1"
+```
+
+Kueue patches the container resources while suspended, then unsuspends:
+
+```bash
+kubectl patch jobset training-job --type json --patch '[
+  {"op": "replace",
+   "path": "/spec/replicatedJobs/0/template/spec/template/spec/containers/0/resources",
+   "value": {"requests": {"cpu": "2", "nvidia.com/gpu": "1"}}}
```
jianqiaol: +1 to CodeRabbit's suggestion. Replacing `/resources` directly will wipe out existing `limits`/`claims` and sets GPU `requests` without matching `limits`. Please patch `/resources/requests` and `/resources/limits` specifically.

jianqiaol (commented):
Thanks for working on this KEP, @kannon92!  I’ve left a few inline comments regarding:
- Enforcing field-level munging strictly for `Requests`/`Limits` so `Claims` immutability is preserved.
- Adding synchronous semantic resource validation at webhook admission time.
- Emitting warning events in the controller if child Job updates fail on resume.
- Expanding the test plan with negative scenarios and updating `kep.yaml` metadata.

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+
+On resume, the updated resources propagate to child Jobs and their Pods.
+
+### Notes/Constraints/Caveats
+
+- **Upstream dependency:** This feature requires the Kubernetes
+  `MutablePodResourcesForSuspendedJobs` feature gate ([KEP-5440](https://github.com/kubernetes/enhancements/blob/master/keps/sig-apps/5440-mutable-job-pod-resource-updates/README.md)),
+  introduced as alpha in Kubernetes 1.35. Without it, JobSet's webhook
+  accepts the update but kube-apiserver rejects the child Job update.
+  Cluster operators must enable both feature gates.
```
andreyvelich, line 124: Do we plan to promote this feature to GA in 1.37?

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+- Introducing new API fields. This KEP only relaxes an existing
+  immutability constraint.
+- Resource mutation for running (non-suspended) JobSets or active Jobs.
+- Mutating any container field other than `resources` while suspended.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Kueue right-sizing before admission
```
andreyvelich, line 76: You can also add another user story from @VassilisVassiliadis's KEP: https://github.com/kubeflow/trainer/pull/3559
This feature is required to ensure our intelligent algorithm can automatically update TrainJob resources.

andreyvelich: Without this feature, we have hard blocker to control TrainJob resources automatically.

keps/1150-mutable-suspended-job-resources/README.md
```diff
@@ -0,0 +1,232 @@
...
+### Goals
+
+- Allow only the `resources.requests` and `resources.limits` fields of
+  `spec.replicatedJobs[].template.spec.template.spec.containers[].resources`
+  and `initContainers[].resources` to be mutated while a JobSet is
+  suspended (mirroring existing mutable-while-suspended fields). Existing
+  `resources.claims` entries in the Pod template are preserved and are not
+  treated as mutable.
+- Propagate resource mutations to child `Job` objects on resume.
+- Gate this behind a new `SuspendedJobResourceMutation` feature gate,
```
andreyvelich: Can we make it consistent with Job feature?
```suggestion
- Gate this behind a new `MutablePodResourcesForSuspendedJobSets` feature gate,
```

keps/1150-mutable-suspended-job-resources/README.md
andreyvelich: We are going to allow to modify resourceClaims too, right @kannon92 ?

andreyvelich (commented):
Thanks @kannon92!
Overall looks great, and it will unblock @VassilisVassiliadis's work too: https://github.com/kubeflow/trainer/pull/3559!

cc @robert-bell @astefanutti @tenzen-y @astefanutti in case you want to check this too.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: I think story 1 covers this. That was my intention for this story.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: In Kueue we found some problems when we use feature gates that match upstream. It would probably be fine but that is why I have a different feature gate.

For Kueue, the issue is that the scheduler-library imports feature gates and we hit conflicts if the same feature gate is defined in both jobset and kubernetes.

So we make sure to prepend features gates with project so we don't conflict with Kube.
I prefer to leave this feature gate as is.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: No, the beta criteria asks for integration enablement so I want this feature enabled to verify this feature before we GA.

We had 1 bug in 1.37 I believe so I didn't promote the feature to GA in 1.37. I hope we can in 1.38 if we are happy with this integration. Maybe @VassilisVassiliadis could verify that this works for Kubeflow and we can get the upstream feature promoted.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: :( resourceClaims are immutable and we will not be able to change that.

The CDI spec is immutable and this means that resourceClaims cannot be changed once they are allocated.

https://github.com/kubernetes/enhancements/tree/master/keps/sig-apps/5440-mutable-job-pod-resource-updates#dra-support

@jianqiaol is correct that we probably should not change Claims.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: I don't think this should be in scope. We have a feature to figure out how to handle better validation.

But as of now, the job / pod controller would do this validation. I don't think we should fold that into JobSet as part of this work.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: done.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: done.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: updated.

keps/1150-mutable-suspended-job-resources/README.md
andreyvelich: But `MutablePodResourcesForSuspendedJobSets` is different from `MutablePodResourcesForSuspendedJob`, isn't?

kannon92: This is whatever we want to name. Its the JobSet feature gate name. I defined it in kep.yaml.

Are you asking to change the name of the feature gate?

andreyvelich: Yeah, I suggest we use `MutablePodResourcesForSuspendedJobSets` feature gate name

kannon92: okay.

kannon92: done.

keps/1150-mutable-suspended-job-resources/README.md
kannon92: updated.

andreyvelich (commented):
Thanks @kannon92!
/lgtm
/hold in case @jianqiaol has any additional comments.

jianqiaol:
Thanks @kannon92! One minor detail to align on between the discussion and the test plan:
In our discussion earlier, you mentioned that semantic resource validation (e.g. `requests <= limits`) should be left out of scope for JobSet and deferred to downstream child Job / Pod validation on `kube-apiserver`. However, under **Test Plan -> Unit Tests** (L191–193), the KEP currently lists:
> *- **Invalid resource semantics:** requests/limits that fail validation (e.g. `requests > limits`) are rejected synchronously at admission.*

I am fine with either approach:
1. **Synchronous Validation (Fail-Fast)**: Add lightweight validation in `jobSetWebhook.ValidateUpdate` (e.g. `requests <= limits`) and keep the test case as-is.
2. **Deferred Validation**: Defer validation to child Job resume (relying on `kube-apiserver` rejection + Warning event) and update the Test Plan item to remove the claim of synchronous rejection at admission.

Once we align the text between Webhook Changes and the Test Plan on this, I think this is good to go. Thanks!

kannon92:
@jianqiaol I updated the KEP to reflect the tests in #1273

jianqiaol:
Thanks @kannon92! 
/lgtm
/hold cancel

PR #1276, opened 2026-07-13
fix: ignore terminal leader pods in webhook to prevent deadlock on retry

jianqiaol:
#### What type of PR is this?
/kind bug

**Problem**:
In an Indexed Job managed by JobSet, when a Pod fails, the built-in Kubernetes Job controller replaces it by spinning up a new pod with the same index. By default, Kubernetes preserves the old failed pod in the Failed phase (without a DeletionTimestamp set) to allow developers to inspect logs.

During the creation of the retried follower pods, the JobSet pod admission webhook validates the pod by counting the leader pods (index 0) for the Job. If the webhook counts anything other than exactly 1 leader pod, it rejects the follower pod.

Previously, the webhook only ignored leader pods that were actively being deleted (DeletionTimestamp != nil). Consequently, if the leader pod failed and was replaced, the webhook counted both the old Failed leader pod and the new active Pending/Running leader pod, resulting in a count of 2.

This caused the webhook to permanently reject all subsequent follower pod creation requests. Since the API requests were rejected at the admission gate, the Job controller got stuck in an infinite retry loop, leaving the Job in an Active state. Because the Job never transitioned to Failed, the JobSet controller was never notified of a failure, bypassing the JobSet failurePolicy.maxRestarts limit and leaving the workload permanently deadlocked.

**Solution**:
Updated the pod filtering logic in leaderPodForFollower (pkg/webhooks/pod_webhook.go) to explicitly ignore pods that are in a terminal phase (corev1.PodFailed or corev1.PodSucceeded), in addition to checking DeletionTimestamp. This ensures that the webhook only counts active (non-terminal) leader pods.
Also added comprehensive unit and integration test coverage for the webhook to prevent regressions.

Fixes #

Added new test cases to TestLeaderPodForFollower in pkg/webhooks/pod_webhook_test.go to cover scenarios where active leader pods co-exist with failed or succeeded leader pods, verifying that the webhook now correctly filters them out and validates successfully.

```release-note
Fixes a bug where follower pods would fail admission (causing a deadlock) when a leader pod is recreated by the Job controller, by properly ignoring failed and succeeded leader pods during validation.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved leader pod selection to ignore pods that are being deleted and pods already in terminal states (failed or succeeded).
  * Webhook validation now succeeds when an active leader exists alongside an older failed/completed leader.
  * Duplicate leader pod scenarios are handled correctly by focusing on the active leader.
* **Tests**
  * Expanded webhook test coverage for duplicate leader pods and terminal-phase leader candidates.

### DIFF
--- a/pkg/webhooks/pod_webhook.go
+++ b/pkg/webhooks/pod_webhook.go
@@ -274,7 +274,7 @@ func (p *podWebhook) leaderPodForFollower(ctx context.Context, pod *corev1.Pod) 
 		return nil, err
 	}
 	podList.Items = slices.DeleteFunc(podList.Items, func(p corev1.Pod) bool {
-		return !p.DeletionTimestamp.IsZero()
+		return !p.DeletionTimestamp.IsZero() || p.Status.Phase == corev1.PodFailed || p.Status.Phase == corev1.PodSucceeded
 	})
 
 	// Validate there is only 1 leader pod for this job.
--- a/pkg/webhooks/pod_webhook_test.go
+++ b/pkg/webhooks/pod_webhook_test.go
@@ -599,6 +599,34 @@ func TestLeaderPodForFollower(t *testing.T) {
 			wantErr: "expected 1 leader pod (js-rjob-0-0), but got 0",
 		},
 		{
+			name:        "active leader alongside failed leader pod",
+			followerPod: followerPod,
+			existingPods: []runtime.Object{
+				leaderPod,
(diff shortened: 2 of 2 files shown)


kannon92:
/ok-to-test

Please sign the CLA.

kannon92 (commented):
/lgtm
/approve

PR #1278, opened 2026-07-16
fix e2e flake from cert bootstrap causing container restart

kannon92:
The cert-controller bootstrap manager can occasionally take longer than the liveness probe's initial delay (15s) to close its readiness channel, even after certificates have been successfully generated. When that happens, the liveness probe kills the container. The controller recovers on the second attempt because the certificates already exist on disk, but the e2e test harness treated any restart as a fatal error via gomega.StopTrying, aborting the entire suite.

Two changes address this:

1. Increase the liveness probe initialDelaySeconds from 15s to 30s in both the kustomize manager manifest and the Helm chart. This gives BootstrapCerts more headroom before the first health check.

2. Relax the restart check in test/util/util.go to tolerate a single container restart (RestartCount <= 1) instead of aborting on any restart. Only repeated restarts now trigger StopTrying, which correctly distinguishes a transient cert-bootstrap race from a real crash loop.

#### What type of PR is this?
/kind flake

Fixes #

```release-note
NONE
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved controller startup reliability by allowing additional time before liveness checks begin.
  * Reduced premature startup failures when certificate initialization temporarily causes a container restart.
  * Updated readiness testing to tolerate a transient restart during controller initialization.

### DIFF
--- a/charts/jobset/templates/controller/deployment.yaml
+++ b/charts/jobset/templates/controller/deployment.yaml
@@ -89,7 +89,7 @@ spec:
             port: 8081
             scheme: HTTP
             path: /healthz
-          initialDelaySeconds: 15
+          initialDelaySeconds: 30
           periodSeconds: 20
         readinessProbe:
           httpGet:
--- a/config/components/manager/manager.yaml
+++ b/config/components/manager/manager.yaml
@@ -80,7 +80,7 @@ spec:
           httpGet:
             path: /healthz
             port: 8081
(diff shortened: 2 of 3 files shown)


GiuseppeTT:
> /assign @GiuseppeTT @andreyvelich

The changes look good to me. Did you get a chance to run the test a few times to make sure it doesn't flake? : )

PR #1279, opened 2026-07-18
fix: reject negative replicated job replicas

immanuwell:
#### What type of PR is this?

/kind bug

Rejects negative replicated job replicas at admission. Negative replicas create no Jobs, so the JobSet can stay stuck. Zero replicas remain supported.

None.

Repro: apply a JobSet with `replicas: -1`. Before this fix it is admitted but creates no Jobs.

small AI assistance was used for this PR

```release-note
Reject JobSets with negative replicated job replicas.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Replicated job replica counts can no longer be set to negative values.
  * Replica count validation now consistently rejects invalid configurations while allowing zero replicas.

### DIFF
--- a/test/integration/webhook/jobset_webhook_test.go
+++ b/test/integration/webhook/jobset_webhook_test.go
@@ -279,6 +279,31 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
 			},
 			jobSetCreationShouldFail: true,
 		}),
+		ginkgo.Entry("negative replicated job replicas are rejected", &testCase{
+			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
+				return testing.MakeJobSet("negative-replicas", ns.Name).
+					EnableDNSHostnames(true).
+					ReplicatedJob(testing.MakeReplicatedJob("negative-replicas").
+						Replicas(-1).
+						Job(testing.MakeJobTemplate("job", ns.Name).
+							PodSpec(testing.TestPodSpec).
+							CompletionMode(batchv1.IndexedCompletion).Obj()).
+						Obj())
+			},
+			jobSetCreationShouldFail: true,
+		}),
+		ginkgo.Entry("zero replicated job replicas are accepted", &testCase{
+			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
+				return testing.MakeJobSet("zero-replicas", ns.Name).
+					EnableDNSHostnames(true).
+					ReplicatedJob(testing.MakeReplicatedJob("zero-replicas").
+						Replicas(0).
+						Job(testing.MakeJobTemplate("job", ns.Name).
+							PodSpec(testing.TestPodSpec).
+							CompletionMode(batchv1.IndexedCompletion).Obj()).
+						Obj())
+			},
+		}),
 		ginkgo.Entry("suspend jobset", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
 				return testing.MakeJobSet("js-hostnames-non-indexed", ns.Name).
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -332,6 +332,7 @@ type ReplicatedJob struct {
 	// replicas is the number of jobs that will be created from this ReplicatedJob's template.
 	// Jobs names will be in the format: <jobSet.name>-<spec.replicatedJob.name>-<job-index>
 	// +kubebuilder:default=1
+	// +kubebuilder:validation:Minimum=0
 	Replicas int32 `json:"replicas,omitempty"`
 
 	// dependsOn is an optional list that specifies the preceding ReplicatedJobs upon which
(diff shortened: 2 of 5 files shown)


pkg/webhooks/jobset_webhook.go
```diff
@@ -217,6 +217,9 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
 	for rJobIdx, rJob := range js.Spec.ReplicatedJobs {
 		fieldPath := field.NewPath("spec", "replicatedJobs").Index(rJobIdx)
 		rJobNames.Insert(rJob.Name)
+		if rJob.Replicas < 1 {
+			allErrs = append(allErrs, field.Invalid(fieldPath.Child("replicas"), rJob.Replicas, "must be greater than or equal to 1"))
+		}
```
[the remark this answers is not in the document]
immanuwell: done in the latest commit

test/integration/webhook/jobset_webhook_test.go
```diff
@@ -279,6 +279,19 @@ var _ = ginkgo.Describe("jobset webhook defaulting", func() {
...
+					EnableDNSHostnames(true).
+					ReplicatedJob(testing.MakeReplicatedJob("negative-replicas").
+						Replicas(-1).
+						Job(testing.MakeJobTemplate("job", ns.Name).
+							PodSpec(testing.TestPodSpec).
+							CompletionMode(batchv1.IndexedCompletion).Obj()).
+						Obj())
+			},
+			jobSetCreationShouldFail: true,
+		}),
```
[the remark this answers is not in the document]
immanuwell: done in the latest commit

kannon92 (changes requested):
I'd rather put this change on the api itself. Use kube builder to enforce the positive values.

Adding validation is usually a breaking change so curious what @andreyvelich and @GiuseppeTT think here.

pkg/webhooks/jobset_webhook.go
```diff
@@ -217,6 +217,10 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
 	for rJobIdx, rJob := range js.Spec.ReplicatedJobs {
 		fieldPath := field.NewPath("spec", "replicatedJobs").Index(rJobIdx)
 		rJobNames.Insert(rJob.Name)
+		validReplicas := rJob.Replicas >= 1
```
kannon92: Probably should be greater than equal to zero as we do have use cases for replicas of 0.

andreyvelich: @kannon92 @immanuwell Shall we also introduce upper bound for Replicas in JobSet, like we have for parallelism and completions in Jobs.
Could we do that before v1 graduation?
https://github.com/kubernetes-sigs/jobset/issues/380

I mentioned it briefly as part of OptimizationJob KEP: https://github.com/kubeflow/trainer/pull/3565#discussion_r3606122043

immanuwell: done in the latest push here

immanuwell:
@kannon92 addressed your suggestion, please take a look

pkg/webhooks/jobset_webhook.go
```diff
@@ -233,28 +233,30 @@ func (j *jobSetWebhook) ValidateCreate(ctx context.Context, js *jobset.JobSet) (
...
-		}
-		// Check that the generated pod names for the replicated job is DNS 1035 compliant.
-		isIndexedJob := rJob.Template.Spec.CompletionMode != nil && *rJob.Template.Spec.CompletionMode == batchv1.IndexedCompletion
-		if isIndexedJob && rJob.Template.Spec.Completions != nil {
-			maxJobIndex := strconv.Itoa(int(rJob.Replicas - 1))
-			maxPodIndex := strconv.Itoa(int(*rJob.Template.Spec.Completions - 1))
-			// Add 5 char suffix to the deterministic part of the pod name to validate the full pod name is compliant.
-			longestPodName := placement.GenPodName(jobSetNameForValidation, rJob.Name, maxJobIndex, maxPodIndex) + "-abcde"
-			for _, errMessage := range validation.IsDNS1035Label(longestPodName) {
+		if rJob.Replicas >= 0 {
```
kannon92: Is there any reason to have this diff now?

This looks unrelated to goal here.

pkg/webhooks/jobset_webhook.go
andreyvelich: cc @GiuseppeTT @jianqiaol

kannon92: > Shall we also introduce upper bound for Replicas in JobSet, like we have for parallelism and completions in Jobs.

I'd rather do this as a separate issue. For negative values I think its fine to include this as there is no real impact to user. negative never worked.

Forbidding certain replicas above a certain value could break users so that is one I am more concerned on.

andreyvelich: Sure, we can discuss this in a separate issue. I just more asking whether we are ok to introduce this breaking change before moving JobSet APIs to V1?

kannon92: I've heard that introducing validation that protects something that never worked is fine. So dissallowing negative replicas I think is fine. Forbidding replicas that could've worked is a breaking change. 

So we could introduce a maximum value that would not block anyone then I guess that would be fine.

pkg/webhooks/jobset_webhook.go
immanuwell: agreed, fixed in the latest push here

immanuwell:
@kannon92 fixed your most recent suggestion

kannon92 (commented):
/lgtm
/approve

PR #1281, opened 2026-07-21
fix: skip invalid negative priority label on JobSet pods

### LINKED ISSUE #1280: Pods with negative-priority PriorityClass can never be created: injected jobset.sigs.k8s.io/priority label value "-N" is invalid
**What happened**:

When a JobSet's pod template references a PriorityClass with a **negative** value (which is valid in Kubernetes — e.g. the cluster-autoscaler overprovisioning pattern uses `value: -1`), child Pods can never be created.

The pod mutating webhook unconditionally copies `pod.Spec.Priority` into the `jobset.sigs.k8s.io/priority` label (`Default()` in `pkg/webhooks/pod_webhook.go`: `pod.Labels[constants.PriorityKey] = fmt.Sprint(*pod.Spec.Priority)`).
For a negative priority the resulting label value (e.g. `"-1"`) is **not a valid Kubernetes label value** — label values must match `(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?`, i.e. start and end with an alphanumeric character. After the mutation, the API server rejects the Pod:

```
Pod "negative-priority-repro-workers-0-0-xxxxx" is invalid: metadata.labels: Invalid value: "-1": a valid label must be an empty string or consist of alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character (e.g. 'MyValue',  or 'my_value',  or '12345', regex used for validation is '(([A-Za-z0-9][-A-Za-z0-9_.]*)?[A-Za-z0-9])?')
```

The Job controller keeps retrying, so the JobSet remains stuck with zero running Pods.

Note: this is not specific to exclusive placement. Since v0.11.0 the label is injected into **every** JobSet Pod (whenever `pod.Spec.Priority` is set), so any JobSet using a negative-value PriorityClass is affected — even with no `alpha.jobset.sigs.k8s.io/exclusive-topology` annotation at all.

JobSet Pods using a negative-value PriorityClass are created successfully.

1. Create a PriorityClass with a negative value:

```yaml
apiVersion: scheduling.k8s.io/v1
kind: PriorityClass
metadata:
  name: low-priority
value: -1
globalDefault: false
description: "negative priority class"
```

2. Create a JobSet whose pod template uses it:

```yaml
kind: JobSet
metadata:
  name: negative-priority-repro
spec:
  replicatedJobs:
  - name: workers
    replicas: 1
    template:
      spec:
        completions: 1
        parallelism: 1
        backoffLimit: 0
        template:
          spec:
            priorityClassName: low-priority
            containers:
            - name: sleep
              image: busybox
              command: ["sleep", "60"]
```

3. The child Job is created but no Pod ever appears. `kubectl describe job negative-priority-repro-workers-0` shows `Error creating: Pod ... is invalid: metadata.labels: Invalid value: "-1" ...`.

- Affected versions: the label has been injected since v0.7.0 (#687) for Pods using exclusive placement; since v0.11.0 (#1077) it is injected into **all** JobSet Pods, so on v0.11.0+ any JobSet Pod with a negative priority hits this.

**Environment**:
- Kubernetes version (use `kubectl version`): v1.35.2
- JobSet version (use `git describe --tags --dirty --always`): v0.12.0
- Install tools: helm
- Others:

dpacgdm:
#### What type of PR is this?

/kind bug

The pod mutating webhook copies `pod.Spec.Priority` into the `jobset.sigs.k8s.io/priority` label via `fmt.Sprint`. When Priority is negative (valid in Kubernetes, e.g. overprovisioning PriorityClasses with `value: -1`), the resulting label value (e.g. `"-1"`) is rejected by the API server because label values cannot start with `-`. Child Pods never get created and the JobSet stays stuck.

This change only sets the priority label when priority is non-negative, which is a valid Kubernetes label value.

Fixes #1280

Added a unit test covering negative priority so the label is not injected.

```release-note
Fix JobSet pods failing admission when using a negative-value PriorityClass by not injecting an invalid jobset.sigs.k8s.io/priority label.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed handling of pods with negative priority values.
  - Negative priorities are now preserved without adding an invalid priority label.
  - Existing annotations and pod configuration remain unchanged.

### DIFF
--- a/pkg/webhooks/pod_webhook.go
+++ b/pkg/webhooks/pod_webhook.go
@@ -64,7 +64,9 @@ func (p *podWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
 	// Add the priority label to make sure the Pod is compatible with other pods using the exclusive
 	// placement feature.
 	// See https://github.com/kubernetes-sigs/jobset/issues/1057 for more details.
-	if pod.Spec.Priority != nil {
+	// Negative priorities are valid in Kubernetes (e.g. overprovisioning PriorityClasses),
+	// but label values cannot start with '-'. Only set the label for non-negative priorities.
+	if pod.Spec.Priority != nil && *pod.Spec.Priority >= 0 {
 		if pod.Labels == nil {
 			pod.Labels = make(map[string]string)
 		}
--- a/pkg/webhooks/pod_webhook_test.go
+++ b/pkg/webhooks/pod_webhook_test.go
@@ -126,6 +126,31 @@ func TestDefault(t *testing.T) {
 			},
 		},
 		{
+			name: "jobset pod, no exclusive placement, with negative priority",
+			pod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "p",
+					Annotations: map[string]string{
+						"jobset.sigs.k8s.io/jobset-name": "js",
+					},
+				},
+				Spec: corev1.PodSpec{
+					Priority: ptr.To(int32(-1)),
+				},
+			},
+			wantPod: &corev1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "p",
+					Annotations: map[string]string{
+						"jobset.sigs.k8s.io/jobset-name": "js",
+					},
+				},
+				Spec: corev1.PodSpec{
+					Priority: ptr.To(int32(-1)),
+				},
+			},
+		},
+		{
 			name: "leader pod with exclusive placement",
 			pod: &corev1.Pod{
 				ObjectMeta: metav1.ObjectMeta{


pkg/webhooks/pod_webhook.go
```diff
@@ -64,7 +64,9 @@ func (p *podWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
 	// Add the priority label to make sure the Pod is compatible with other pods using the exclusive
 	// placement feature.
 	// See https://github.com/kubernetes-sigs/jobset/issues/1057 for more details.
-	if pod.Spec.Priority != nil {
+	// Negative priorities are valid in Kubernetes (e.g. overprovisioning PriorityClasses),
```
kannon92, line 67: @GiuseppeTT 

Curious if you have seen this come up at all?

GiuseppeTT: > Curious if you have seen this come up at all?

I have never seen it come up.

@jianqiaol

kannon92: I'm asking here because I want to know if there is any reason to keep the negative priority for jobset pods.

kannon92: WAS podgroup priority could probably be one way to fix this.

jianqiaol:
Thanks for the fix! While this resolves the admission crash, skipping the label entirely will break the **exclusive placement** feature for negative-priority JobSets. 
The injected pod anti-affinity selector (in `setExclusiveAffinities`) relies on `jobset.sigs.k8s.io/priority` to enforce exclusivity within the same priority band. If the label is skipped, the selector will look for `Values: [""]` and fail to match, allowing negative-priority JobSets to be co-located.
**Suggested Alternative:**
Instead of skipping the label, we can format negative priorities with a `neg-` prefix (e.g., `-1` becomes `neg-1`). This is a valid label value and ensures exclusive placement still works:
```go
priorityVal := *pod.Spec.Priority
if priorityVal < 0 {
    pod.Labels[constants.PriorityKey] = fmt.Sprintf("neg-%d", -priorityVal)
} else {
    pod.Labels[constants.PriorityKey] = fmt.Sprint(priorityVal)
}

pkg/webhooks/pod_webhook.go
Viswalahiri: @kannon92 in my review of the code I see the apiserver resolving `priorityClassName` into `pod.Spec.Priority` before the webhook runs. 

The repro only sets `priorityClassName`, and the webhook would still read `-1`. So normalizing it here would be JobSet overriding a user requested  setting.

So overall, I believe whatever we encode here would drop out cleanly downstream.

Viswalahiri:
@jianqiaol I think #1291 fixes what you were suggesting, just spelled `n1` rather than `neg-1`. It also widens to int64 before negating. Do you care about which spelling it is? Happy to pick this up and carry to merge if needed.

kannon92:
based on https://github.com/kubernetes-sigs/jobset/issues/1280#issuecomment-5443431803, I am going to close this PR.

Thank you for the discussion but I think we'd like to fix negative priority actually.

PR #1283, opened 2026-07-23
feat: Add ExecutionAttempts monotonic counter to JobSet

### LINKED ISSUE #1282: Add Execution Attempt Annotation
cc @kannon92 @andreyvelich @GiuseppeTT 

## What would you like to be added

Introduce a new pod annotation (e.g., `jobset.sigs.k8s.io/execute-attempt`) to the JobSet API to track the total number of execution iterations a JobSet has undergone. 

Unlike the existing `jobset.sigs.k8s.io/restart-attempt` annotation, which only increments during failure-triggered restarts, this proposed annotation will increment on **both failure-triggered restarts and suspension-driven resumptions.**

## Why is this needed

Logging deployment streams typically rely on the `restart-attempt` annotation to tag logs per execution cycle. However, when queueing workloads with systems like Kueue, JobSets are frequently evicted via suspension (terminating pods) and later resumed (spawning new pods). 

Because JobSet resumptions do not increment the `restart-attempt` counter, logs from multiple distinct execution phases currently share an identical identifier. Consequently, users cannot properly isolate log streams by their scheduling lifecycle, which severely impedes debugging efforts (e.g., when needing to isolate logs to investigate node-level hardware failures during a specific historical execution pass).

I got this request from a big user of JobSet.

## Example

1. **Initial Scheduled Run:** A JobSet starts. Pods land on Nodepools A and B. Both `restart-attempt` and `execute-attempt` evaluate to `"0"`.
2. **Preemption (Eviction):** Kueue evicts the JobSet (`suspend: true`). Active pods are deleted.
3. **Resumed Execution:** Kueue readmits the JobSet (`suspend: false`). New pods land on Nodepools C and D. 
    * Current state: `restart-attempt` remains `"0"`.
    * Proposed state: `execute-attempt` increments to `"1"`.

The new annotation provides a distinct monotonic execution token to filter telemetry data accurately.

### 1. Increment on restart
When a child Job failure triggers a JobSet restart (matching the existing `restart-attempt` conditions), the JobSet controller will increment a new `.status.executeAttempts` field.

### 2. Increment on resume
When the JobSet controller detects the `jobset.spec.suspend` configuration transitioning from `true` to `false`, it will similarly increment the `.status.executeAttempts` field before the new child Jobs (and their pods) are created.

### 3. Data propagation
During child Job creation, the JobSet controller will inject the current `.status.executeAttempts` value into the child Job and corresponding Pod templates as the `jobset.sigs.k8s.io/execute-attempt` annotation.

jianqiaol:
#### What type of PR is this?

/kind feature
/kind api-change

This PR introduces the `ExecutionAttempts` monotonic counter to the JobSet status, tracking the number of times a JobSet has been executed (reconstructed due to failure policy restarts or resumed from a suspended state).

**Key Changes:**
- **API Update:** Added `.status.executionAttempts` to `JobSetStatus` (defaults to `nil`). 
- **Data Propagation:** The current `ExecutionAttempts` value is propagated as the `jobset.sigs.k8s.io/execution-attempt` annotation to child Jobs and Pod templates during creation and updated on existing child Jobs during resumption.
- **Robustness:** Introduced a synchronous status update to etcd in `resumeJobsIfNecessary` before modifying child Jobs, preventing state loss (counter reset) if the controller crashes during the resumption process.

**Why we need it:**
Telemetry and logging systems rely on execution identifiers to segment logs per run. Currently, suspension-driven resumptions (common when using queueing systems like Kueue) do not increment the existing `restart-attempt` counter. This PR provides a distinct monotonic execution token to accurately segment log streams by scheduling lifecycle.

Fixes #1282

- **Upgrade Migration Logic:** If the controller is upgraded while JobSets are already active in the cluster, they will initially have `status.executionAttempts` set to `nil`.
  - For **running workloads** (already started executing, or having `restarts > 0`), the controller initializes `status.executionAttempts` to the current `status.restarts` value upon reconciliation. This preserves history and maintains the invariant `executionAttempts >= restarts`.
  - For **new workloads** (or those resumed for the first time without prior execution), it initializes to `0`.
- **Performance Trade-off:** Resuming a JobSet now triggers a synchronous status write to etcd before updating child Jobs, adding one blocking API write call. Given resumption is a relatively low-frequency event, this overhead is justified by the robustness guarantees against controller crashes.
- **Pod Template Mutability:** Mutating `job.Spec.Template.Annotations` is allowed for suspended Jobs (standard since K8s 1.27 with `AllowMutableSchedulingDirectives`). Since resumption updates occur while child Jobs are suspended, this mutation is safe and accepted by the API server.

Yes.

```release-note
Added `ExecutionAttempts` field to JobSet status and corresponding `jobset.sigs.k8s.io/execution-attempt` annotation on child Jobs and Pod templates to track execution attempts across restarts and resumptions.

### CODE REVIEW

File: api/jobset/v1alpha2/jobset_types.go [hunk 1]
```diff
@@ -217,6 +217,16 @@ type JobSetStatus struct {
...
 
+	// executeAttempts tracks the number of execution lifecycles.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:default=0
+	ExecuteAttempts *int32 `json:"executeAttempts,omitempty"`
+
+	// hasRun tracks whether the JobSet has started executing.
+	// +optional
+	HasRun *bool `json:"hasRun,omitempty"`
```
kannon92: can't we guess this from something like startTime?

kannon92: I guess we only have that field for Job but we should look at how to surmise that the jobset has started by looking at conditions versus adding another API field.

jianqiaol: Thanks for the feedback! We looked into this but it seems we cannot rely on `startTime` or conditions alone:

*   **Job `startTime` is reset to `nil`:** In recent K8s versions, a Job's `startTime` is reset to `nil` when suspended. This behavior was introduced via the `MutableSchedulingDirectivesForSuspendedJobs` feature gate:
    *   **Kubernetes 1.35:** Introduced as an Alpha feature (disabled by default).
    *   **Kubernetes 1.36:** Graduated to Beta and is enabled by default.
    
    This makes it impossible to distinguish a JobSet suspended before running from one suspended after running.
*   **Conditions lack history:** They only show the current state (e.g., `Suspended=True`), not transition history, so we cannot distinguish the first resume from subsequent ones.

### Current Solution

To avoid API bloat, we **removed the `HasRun` field**. Instead, we use the `ExecuteAttempts` field itself:

*   `nil` = Never run.
*   Initialized to `0` on the first execution (starts unsuspended or first resume).
*   Incremented on subsequent resumptions.

This keeps the API clean while ensuring correct behavior.

kannon92: Should a suspend reset it entirely though? 

One idea we had for suspend was you can change the shape of the jobset or the replicas when its suspended. So you could in theory change the shape of your workload and then `hasRun` is sorta meaningless.

jianqiaol: Thanks for the feedback! We actually ended up removing the HasRun field to keep the API changes minimal and clean.

Regarding the counter: we think it's safest to keep ExecuteAttempts monotonically increasing, even if the JobSet shape changes during suspension. This ensures that log streams from different execution runs are always cleanly separated and don't overlap. If a user really needs to start over with a fresh counter, recreating the JobSet is always an option.

Let me know if that makes sense to you!

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -217,6 +217,11 @@ type JobSetStatus struct {
 	// +optional
 	RestartsCountTowardsMax int32 `json:"restartsCountTowardsMax,omitempty"`
 
+	// executeAttempts tracks the number of execution lifecycles.
+	// +optional
+	// +kubebuilder:validation:Minimum=0
```
kannon92, line 225: Do we have an upper limit on how many executions we can have?

design_doc.md
```diff
@@ -0,0 +1,46 @@
+# \[External\] JobSet \- Add Execution Attempt Annotation
```
kannon92: Can you move this into a KEP?

api/jobset/v1alpha2/jobset_types.go
jianqiaol: We don't enforce an upper limit on ExecuteAttempts in JobSet. Since it is just a status-tracking field (*int32), it is only bounded by the max integer value. We believe lifecycle and resumption limits are best managed by higher-level orchestrators like Kueue, rather than JobSet itself.

design_doc.md
jianqiaol: Thanks for the review! I've moved the design doc to KEP-1282 as requested.

pkg/constants/constants.go
```diff
@@ -38,6 +38,9 @@ const (
 	// That is, it should match `jobSet.spec.replicatedJobs[replicatedJobName].jobRestarts[jobIndex]`
 	JobRestartAttemptKey = "jobset.sigs.k8s.io/job-restart-attempt"
 
+	// ExecuteAttemptsKey is a pod annotation which defines the current execution attempt of the jobset.
+	ExecuteAttemptsKey = "jobset.sigs.k8s.io/execute-attempt"
+
```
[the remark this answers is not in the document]
kannon92: Check again here. I'm not sure this is correct anymore.

get_status.py
```diff
@@ -0,0 +1 @@
+import sys, json; data=json.load(sys.stdin); [print(f"{s[\"context\"]}: {s[\"state\"]} - {s[\"target_url\"]}") for s in data.get("statuses", []) if s["state"] != "success"]
```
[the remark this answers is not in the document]
kannon92: Please do not review auto generated python code.

keps/1282-execution-attempts/README.md
```diff
@@ -0,0 +1,114 @@
+# KEP-1282: Add Execution Attempts Tracking
```
kannon92, line 1: nit: can you open up a PR with just this design?

jianqiaol: Yes, just added: https://github.com/kubernetes-sigs/jobset/pull/1292

jianqiaol:
Hi @kannon92 @andreyvelich,
Now that the design in [KEP-1282](https://github.com/kubernetes-sigs/jobset/pull/1292) has been approved and merged, could you please take a look at this implementation PR to help move it forward?
To recap, this PR implements KEP-1282 with all the finalized design decisions:
- Adds `Status.ExecutionAttempts` (`*int32`) and propagates the `jobset.sigs.k8s.io/execution-attempt` annotation down to child Jobs and Pod templates.
- Guarded behind the `ExecutionAttemptsTracking` Alpha feature gate.
- Implements upgrade migration semantics (`nil` -> `Restarts` for existing running workloads; `0` for new workloads) preserving `ExecutionAttempts >= Restarts`.
- Synchronously commits status updates on resume before mutating child job templates.
- Includes unit tests, Ginkgo integration tests (covering multiple suspend/resume cycles, upgrade migration, and feature-gate behavior), regenerated CRDs, and Python SDK updates.

All CI presubmits are passing and the PR is ready for your review. Thanks!

hack/python-sdk/gen-sdk.sh
```diff
@@ -56,13 +56,25 @@ if [[ "$CONTAINER_ENGINE" == podman ]]; then
...
 fi
 
-${CONTAINER_ENGINE} run --user "${user}" --rm \
-  -v "${volume_mapping}" docker.io/openapitools/openapi-generator-cli:v7.11.0 generate \
-  -i /local/hack/python-sdk/swagger.json \
-  -g python \
-  -o /local/sdk/python \
-  -c local/hack/python-sdk/swagger_config.json \
-  --global-property modelDocs=false
+if command -v "${CONTAINER_ENGINE}" >/dev/null 2>&1; then
```
kannon92: This doesn't seem related to this PR. Can you create a separate PR to fix this?

jianqiaol: Thanks for pointing out this! I've reverted the changes in hack/python-sdk/gen-sdk.sh and will open a separate PR for that.

pkg/controllers/jobset_controller.go
```diff
@@ -593,6 +593,26 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
...
+	// This prevents a race condition where the controller crashes after wiping child Job StartTimes,
+	// but before persisting the JobSet status, resulting in irreversible state loss.
+	if updateStatusOpts.shouldUpdate {
+		if err := r.updateJobSetStatus(ctx, js, updateStatusOpts); err != nil {
+			return err
+		}
+		updateStatusOpts.shouldUpdate = false
+		updateStatusOpts.events = nil
+	}
+
```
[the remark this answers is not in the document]
kannon92: This seems very wrong.

pkg/controllers/jobset_controller.go
```diff
@@ -674,6 +695,18 @@ func (r *JobSetReconciler) resumeJob(ctx context.Context, job *batchv1.Job, repl
 			job.Spec.Template.Spec.SchedulingGates,
 			replicatedJobPodTemplate.Spec.SchedulingGates,
 		)
+
+		if features.Enabled(features.ExecutionAttemptsTracking) {
+			if job.Spec.Template.Annotations == nil {
+				job.Spec.Template.Annotations = make(map[string]string)
+			}
+			job.Spec.Template.Annotations[constants.ExecutionAttemptsKey] = strconv.Itoa(int(ptr.Deref(js.Status.ExecutionAttempts, 0)))
```
kannon92: nit: can you convert this once and reuse that variable

kannon92 (changes requested):
can you add some e2e tests with this feature also?

pkg/controllers/jobset_controller.go
jianqiaol: done

pkg/controllers/jobset_controller.go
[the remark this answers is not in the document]
jianqiaol: Yeah, CodeRabbit was confused by a transient intermediate commit from earlier. DeepCopyObject() is already generated in api/jobset/v1alpha2/zz_generated.deepcopy.go and all generated artifacts are up to date. Verified with golangci-lint and go test ./pkg/... (all passing with 0 errors).

jianqiaol:
Thanks for the reviews @kannon92! I've updated the PR with all feedback addressed:

1. **E2E Tests**: Added comprehensive E2E tests in `test/e2e/customconfigs/executionattempts_test.go` covering monotonic attempt tracking across suspend/resume cycles and failure-triggered restarts (`RecreateAll`).
2. **Annotation Conversion**: Cached the string conversion in `resumeJob` to avoid redundant calls.

PTAL when you get a chance!

kannon92:
/assign

If I don't get this by EOW please ping again.

pkg/controllers/jobset_controller.go
```diff
@@ -593,6 +593,30 @@ func (r *JobSetReconciler) resumeJobsIfNecessary(ctx context.Context, js *jobset
 	// Map where key is the Job name and value is the Job replicas.
 	rJobsReplicas := map[string]int32{}
 
+	// Set the suspended condition on the JobSet to false to indicate
```
kannon92: Can you make sure this function is will gauaded by feature gates?

It seems setJobSetResumedCondition was not called here but you moved it here without the feature gate.

If this is a bug outside of this PR, please open up a PR for that.

jianqiaol: thanks for catching this, just fixed it.

kannon92:
### Code review: ExecutionAttempts monotonic counter

Reviewed the resume-path counting logic. Overall the core counting is sound and well-tested — the mid-reconcile commit correctly prevents double-increment across reconciles, and the recreate-all increment in `failure_policy.go` doesn't double-count since restarts don't flow through suspend/resume. Build passes. A few things worth a look:

**Genuine behavior issues**

1. **Stale attempt annotation on resume-recreated Jobs** (`pkg/controllers/jobset_controller.go:606`)
   `reconcileReplicatedJobs` stamps newly-created Jobs with the *pre-increment* `Status.ExecutionAttempts`, but it runs before `resumeJobsIfNecessary`, which stamps existing Jobs with the *post-increment* value. If a child Job is deleted while suspended, on resume the recreated Job carries attempt `N-1` while its siblings carry `N` — inconsistent annotations within a single execution.

2. **`Resumed` condition/event fires prematurely with in-order / `dependsOn` startup** (`pkg/controllers/jobset_controller.go:600`)
   `setJobSetResumedCondition` was previously called only after the last ReplicatedJob started (the in-order branch returns before the old call site). Moving it to the top of `resumeJobsIfNecessary` publishes the condition/event on the *first* resume reconcile, before jobs actually resume in order. Consumers watching that condition may act early.

**Lower severity**

3. **Migration off-by-one** (`pkg/controllers/jobset_controller.go:603`)
   A JobSet suspended with `Restarts > 0` and nil `ExecutionAttempts` won't count the upgrade resume (guarded by `ExecutionAttempts != nil`), then init sets it to `Restarts`. A natively-tracked JobSet from the same state would reach `Restarts + 1`.

4. **Broader status-write trigger than intended** (`pkg/controllers/jobset_controller.go:613`)
   The mid-reconcile `updateJobSetStatus` fires whenever `shouldUpdate` is true (commonly set by earlier reconcile steps), not only when `ExecutionAttempts` changed — an extra API write and mid-reconcile event flush on those resume reconciles. Justified by crash-safety, but wider than the stated intent.

5. **Counter logic split / duplicated condition scan** (`pkg/controllers/jobset_controller.go:1337`)
   `setJobSetResumedCondition` re-scans conditions separately from `setCondition`/`updateCondition` and embeds the increment inside a condition setter, splitting counter logic across two functions and making it harder to follow.

---
_Disclosure: this review was generated with AI assistance (Claude Code) and reviewed by me before posting, in line with the [Kubernetes AI tool usage policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance)._

pkg/controllers/jobset_controller.go
```diff
@@ -572,6 +577,65 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
...
+		return r.Status().Update(ctx, js)
+	}
+
+	// Standard resume: increment on the first reconcile of the resume transition.
+	if allJobsSuspended {
+		js.Status.ExecutionAttempts = ptr.To(*js.Status.ExecutionAttempts + 1)
+		return r.Status().Update(ctx, js)
+	}
+
+	return nil
```
kannon92, line 676:
```suggestion
	return r.Status().Update(ctx,js)
```

And remove the returns on lines 627 and 633.

test/integration/controller/jobset_controller_test.go
```diff
@@ -718,6 +719,7 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 		}),
 		ginkgo.Entry("[failure policy] jobset restarts with RestartJobSet failure policy action (without RestartJob feature gate).", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
+				features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.ExecutionAttemptsTracking, true)
```
kannon92: Please don't enable an alpha feature in these tests.

kannon92: We should test the enablement of this feature separately from the existing tests for now.

test/integration/controller/jobset_controller_test.go
```diff
@@ -1520,8 +1527,225 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+							{
+								Name:      "replicated-job-a",
+								Succeeded: 1,
+							},
+						})
+					},
+				},
+			},
+		}),
+		ginkgo.Entry("suspend and resume a running jobset", &testCase{
```
kannon92:
```suggestion
		ginkgo.Entry("suspend and resume a running jobset with ExecutionAttempsTracking enabled", &testCase{
```

test/integration/controller/jobset_controller_test.go
```diff
@@ -1520,8 +1527,225 @@ var _ = ginkgo.Describe("JobSet controller", func() {
...
+					checkJobSetState: func(js *jobset.JobSet) {
+						gomega.Eventually(matchJobsSuspendState, timeout, interval).WithArguments(js, false).Should(gomega.Equal(true))
+						matchJobSetExecutionAttempts(js, 1)
+						matchChildJobsExecutionAttempts(js, 1)
+					},
+					checkJobSetCondition: testutil.JobSetResumed,
+				},
+			},
+		}),
+		ginkgo.Entry("multiple suspend and resume cycles", &testCase{
```
kannon92:
```suggestion
		ginkgo.Entry("multiple suspend and resume cycles with ExecutionAttempsTracking enabled", &testCase{
```

pkg/controllers/jobset_controller.go
```diff
@@ -572,6 +577,78 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
 	return nil
 }
 
+// syncExecutionAttempts handles initial setup and resume increments for ExecutionAttempts.
+func (r *JobSetReconciler) syncExecutionAttempts(ctx context.Context, js *jobset.JobSet, ownedJobs *childJobs, updateStatusOpts *statusUpdateOpts) error {
+	if !features.Enabled(features.ExecutionAttemptsTracking) {
+		return nil
+	}
+
+	wasSuspended := false
```
kannon92: I read this and wondered if `wasSuspended` and `jobSetSuspended` were redundant. 

- `jobSetSuspended(js)` checks the current desired state
- `wasSuspended` infers the previously observed state from status conditions

That distinction looks necessary here to detect the `suspended -> active` transition and increment `ExecutionAttempts` only on resume, while still treating newly created active JobSets as initialization-only.

It may be worth renaming `wasSuspended` to something like `previouslySuspended` (or adding a short comment) to make that transition-detection role more obvious.

pkg/controllers/jobset_controller.go
```diff
@@ -572,6 +577,78 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
...
+	for _, c := range js.Status.Conditions {
```
kannon92: nit: you could remove this entire block (587 to 589) with:

```
wasSuspended := apimeta.IsStatusConditionTrue(js.Status.Conditions, string(jobset.JobSetSuspended))
``

File: test/integration/controller/jobset_controller_test.go [hunk 17]
```diff
@@ -1453,6 +1454,7 @@ var _ = ginkgo.Describe("JobSet controller", func() {
 		}),
 		ginkgo.Entry("resume a suspended jobset", &testCase{
 			makeJobSet: func(ns *corev1.Namespace) *testing.JobSetWrapper {
+				features.SetFeatureGateDuringTest(ginkgo.GinkgoTB(), features.ExecutionAttemptsTracking, true)
```
kannon92: can you revert this change? We should only enable the feature gates for tests that need it.

test/integration/controller/jobset_controller_test.go
```diff
@@ -3815,6 +4088,81 @@ func matchJobSetRestarts(js *jobset.JobSet, expectedCount int32) {
...
+			return 0, err
+		}
+
+		return ptr.Deref(newJs.Status.ExecutionAttempts, 0), nil
+	}, timeout, interval).Should(gomega.BeComparableTo(expectedCount))
+}
+
+// matchChildJobsExecutionAttempts checks that all child Jobs of a JobSet have expectedCount
+// as the value of their execution-attempt annotation.
+func matchChildJobsExecutionAttempts(js *jobset.JobSet, expectedCount int32) {
```
kannon92, line 4232: `matchChildJobsExecutionAttempts` looks prone to a false positive right now: it returns success as long as it doesn’t find a mismatched annotation, even if there are **zero** matching non-deleting child Jobs. That seems risky for cases like “child job deleted while suspended is recreated on resume”, because the assertion could pass before the replacement Job is actually recreated and annotated.

Could we make this helper count matching, non-deleting child Jobs and only return success once that count matches the expected number of child Jobs for the JobSet?

test/e2e/customconfigs/executionattempts_test.go
```diff
@@ -0,0 +1,243 @@
...
+				g.Expect(jobList.Items[0].Annotations[constants.ExecutionAttemptsKey]).To(gomega.Equal("1"))
+				g.Expect(jobList.Items[0].Spec.Template.Annotations[constants.ExecutionAttemptsKey]).To(gomega.Equal("1"))
+			}, timeout, interval).Should(gomega.Succeed())
+
+			ginkgo.By("checking that jobset succeeds eventually")
+			util.JobSetCompleted(ctx, k8sClient, js, timeout)
+		})
+	})
+
+	ginkgo.When("JobSet fails and restarts via failure policy", func() {
```
kannon92, line 155: Nice coverage on the failure-policy restart path. One gap I still see is that this e2e only checks the final `JobSet.Status.ExecutionAttempts`/`Restarts` values, but not whether the restarted child Job (and its pod template) actually got the updated `jobset.sigs.k8s.io/execution-attempt` annotation.

Could we add a post-restart assertion similar to the suspend/resume scenario that verifies the child Job annotation and `job.Spec.Template.Annotations` are both updated to `"1"` after the restart?

pkg/controllers/jobset_controller.go
```diff
@@ -572,6 +577,78 @@ func (r *JobSetReconciler) suspendJobs(ctx context.Context, js *jobset.JobSet, a
...
+			attempts++
+		}
+		js.Status.ExecutionAttempts = ptr.To(attempts)
+		updated = true
+	} else {
+		// Determine whether ExecutionAttempts has already been incremented for this resume transition.
+		// If any child job is already unsuspended, or if Status.ExecutionAttempts is already greater
+		// than the attempt recorded on suspended child jobs, it was already incremented (e.g. on a reconcile retry).
+		shouldIncrement := true
+		for _, job := range ownedJobs.active {
```
kannon92: during a suspend -> resume transition, there could be no active child jobs. So then the counter would be increasemented.

kannon92 (commented):
cc @andreyvelich @GiuseppeTT 

/approve

I'll leave lgtm for the above folks.

GiuseppeTT:
/lgtm

Thanks for the work Jianqiao!

Also, thanks for the review of the KEP and its implementation Kevin and Andrey!

PR #1291, opened 2026-07-29
fix: render negative Pod priorities as valid label values

### LINKED ISSUE #1280: Pods with negative-priority PriorityClass can never be created: injected jobset.sigs.k8s.io/priority label value "-N" is invalid

SrikarChittemsetty:
#### What type of PR is this?

/kind bug

The Pod mutating webhook copies `pod.Spec.Priority` into the `jobset.sigs.k8s.io/priority` label verbatim:

```go
pod.Labels[constants.PriorityKey] = fmt.Sprint(*pod.Spec.Priority)
```

Kubernetes label values must start and end with an alphanumeric character. A PriorityClass with a negative value is legal — the cluster-autoscaler overprovisioning pattern uses `value: -1` — but produces a label value of `"-1"`, which the API server rejects:

```
metadata.labels: Invalid value: "-1": a valid label must be an empty string or consist of
alphanumeric characters, '-', '_' or '.', and must start and end with an alphanumeric character
```

Every child Pod is refused at admission and the Job controller retries forever, so the JobSet sits with zero Pods. Since v0.11.0 the label is injected into every JobSet Pod whenever `pod.Spec.Priority` is set, so this affects any JobSet using a negative PriorityClass — not only those using exclusive placement.

Negative priorities are now rendered as their absolute value with an `n` prefix (`-1` → `n1`). Non-negative priorities keep their existing decimal form, which matters for upgrades: the value is used in the exclusive-placement anti-affinity term (`setExclusiveAffinities`), so Pods admitted before this change must continue to match Pods admitted after it. Since only negatives change — and no negative value could previously exist, because such Pods were rejected — there is no mixed-encoding case to worry about.

The label is only ever set and compared for equality; nothing parses it back into a number (`grep -rn PriorityKey` shows the constant, the write, and the selector).

Fixes #1280

**AI disclosure:** this change was prepared with the assistance of an AI coding agent, per the [Kubernetes AI Tool Usage Policy](https://www.kubernetes.dev/docs/guide/pull-requests/#ai-guidance) and this repo's `AGENTS.md`. The analysis, fix, and tests were verified against the source and the existing suite before submitting. No AI authorship trailers are present on the commits.

**Testing.** Three unit tests in `pkg/webhooks`:

- every priority — including `math.MinInt32`, whose negation overflows `int32` and is handled via `int64` — produces a value accepted by `validation.IsValidLabelValue`
- the encoding is injective, so two priorities can never collapse onto one label value and silently merge anti-affinity groups
- `Default()` end-to-end for the reported case asserts the label is exactly `n1`

Against the previous behaviour they fail with exactly the reported symptom:

```
priorityLabelValue(-1) = "-1" is not a valid label value: [a valid label must be an empty
string or consist of alphanumeric characters ...]
```

The full `pkg/webhooks` package passes and `gofmt` is clean.

**Review feedback addressed.** Both automated review comments are resolved: the end-to-end test now pins the exact `n1` encoding rather than only checking label validity, and the assertions use `testify` `assert`/`require`, matching `pkg/webhooks/jobset_webhook_test.go`.

This is still waiting on `/ok-to-test` from a kubernetes-sigs member before CI will run.

```release-note
Fixed a bug where JobSets using a negative-value PriorityClass produced an invalid `jobset.sigs.k8s.io/priority` label, causing every child Pod to be rejected at admission. Negative priorities are now encoded as `n<absolute value>`.
```

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed priority labels for pods with negative priority values, ensuring they use valid Kubernetes label formats.
  - Preserved correct label values for zero and positive priorities.
  - Ensured pod priority fields remain unchanged while labels accurately reflect their configured values.

- **Tests**
  - Added coverage for negative, zero, and positive priority values, including the minimum supported priority.


SrikarChittemsetty:
Both review comments are addressed: the end-to-end webhook test now asserts the label is exactly `n1` instead of only checking that it's a valid label value (a wrong-but-valid encoding would have passed before), and the assertions moved to testify `assert`/`require` to match `pkg/webhooks/jobset_webhook_test.go`. I also reformatted the description to follow the PR template and added the release note.

This is still sitting behind `needs-ok-to-test`, so CI hasn't run on it. Could a kubernetes-sigs member drop an `/ok-to-test` when convenient? `pkg/webhooks` passes locally.

cc @andreyvelich

pkg/webhooks/pod_webhook_priority_test.go
```diff
@@ -0,0 +1,95 @@
...
+
+import (
+	"context"
+	"math"
+	"testing"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/validation"
+	"k8s.io/utils/ptr"
```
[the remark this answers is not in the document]
SrikarChittemsetty: Addressed in 4e2a1e5 — the priority tests now use testify `assert` for the non-fatal checks and `require` for the fatal one, matching `pkg/webhooks/jobset_webhook_test.go`. The exact-encoding point from the other thread is covered in 6640a78: the end-to-end webhook test asserts the label is exactly `n1` rather than only that it is a valid label value.

tenzen-y:
@SrikarChittemsetty, are you still working on this PR? If yes, can you fix formatting errors?

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_jobset/1291/pull-jobset-verify-main/2093031736161079296

pkg/webhooks/pod_webhook.go
```diff
@@ -68,7 +68,7 @@ func (p *podWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
 		if pod.Labels == nil {
 			pod.Labels = make(map[string]string)
 		}
-		pod.Labels[constants.PriorityKey] = fmt.Sprint(*pod.Spec.Priority)
+		pod.Labels[constants.PriorityKey] = priorityLabelValue(*pod.Spec.Priority)
```
kannon92, line 71: can you add an integration test for the webhook also?

pkg/webhooks/pod_webhook_priority_test.go
```diff
@@ -0,0 +1,89 @@
+/*
```
kannon92: can you move this into pkg/webhooks/pod_webhook_test.go?

I don't think we need to introduce a new file for this feature?

SrikarChittemsetty:
@tenzen-y yes, still on it — thanks for the nudge, and thanks @kannon92 for the `/ok-to-test`.

I dug into that job before pushing, and I don't think formatting was the failure. `verify` runs `vet fmt-verify ci-lint lint-api manifests generate ...`, and the log shows all of those completing: `go vet ./...`, `golangci-lint run`, the kube-api-linter, and `controller-gen`. It then died two steps later in `code-generator`:

```
go: k8s.io/code-generator/cmd/client-gen@v0.36.4: loading deprecation for k8s.io/code-generator:
    module k8s.io/code-generator: read "https://proxy.golang.org/k8s.io/code-generator/@v/list":
    stream error: stream ID 29; INTERNAL_ERROR; received from peer
make: *** [Makefile:387: code-generator] Error 1
```

That's a module-proxy read failing mid-stream, so `fmt-verify` had already passed. I confirmed locally anyway — `gofmt -l` over every non-vendor `.go` file returns nothing, and `go vet` is clean.

I've rebased onto current `main` (the branch was 10 commits behind) and pushed, which should give it a fresh run. The change itself is unchanged and still scoped to two files:

- `pkg/webhooks/pod_webhook.go` — render negative priorities as `n<abs>` so the label is a valid Kubernetes label value
- `pkg/webhooks/pod_webhook_priority_test.go` — the three tests, still passing

/retest

kannon92:
> Both review comments are addressed: the end-to-end webhook test now asserts the label is exactly `n1` instead of only checking that it's a valid label value (a wrong-but-valid encoding would have passed before), and the assertions moved to testify `assert`/`require` to match `pkg/webhooks/jobset_webhook_test.go`. I also reformatted the description to follow the PR template and added the release note.
> 
> This is still sitting behind `needs-ok-to-test`, so CI hasn't run on it. Could a kubernetes-sigs member drop an `/ok-to-test` when convenient? `pkg/webhooks` passes locally.
> 
> cc @andreyvelich

I'm sorry I missed this. Thank you for following up.

pkg/webhooks/pod_webhook.go
```diff
@@ -77,6 +77,23 @@ func (p *podWebhook) Default(ctx context.Context, pod *corev1.Pod) error {
...
+// a label value must start and end with an alphanumeric character, so a leading
+// "-" makes the Pod rejected by the API server. Negative priorities are written
+// as their absolute value with an "n" prefix. Non-negative priorities keep their
+// plain decimal form so that Pods admitted before this change continue to match
+// the exclusive-placement anti-affinity term.
+//
+// The value is only ever compared for equality, never parsed back into a number.
+func priorityLabelValue(priority int32) string {
+	if priority < 0 {
+		return fmt.Sprintf("n%d", -int64(priority))
```
tenzen-y, line 95:
```suggestion
		return fmt.Sprintf("n%d", -int32(priority))
```

Why do we need to use `int64`? IIUC, Pod priority field is int32 :https://github.com/kubernetes/kubernetes/blob/7bf53c7b96b2a155f7812ae93a226069085d08d2/pkg/apis/core/types.go#L4031

pkg/webhooks/pod_webhook_priority_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+package webhooks
+
+import (
+	"context"
+	"math"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
tenzen-y:
```suggestion
```

IIUC, the JobSet repository doesn't use these libraries in almost any places.

pkg/webhooks/pod_webhook_priority_test.go
tenzen-y, line 1: I was not sure these new tests were valid since some objectives are mixed in UT. Instead of those UTs, I would recommend (1 adding new test cases to https://github.com/kubernetes-sigs/jobset/blob/370f93690d193aec233e3e44c39baf3a532438cc/pkg/webhooks/pod_webhook_test.go#L39 and removing all of these test functions.

Additionally, as @kannon92 mentioned in another thread, (2 we can add new test cases to https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/webhook/jobset_webhook_test.go for webhook behaviors and (3 also add new test cases to https://github.com/kubernetes-sigs/jobset/blob/main/test/integration/controller/jobset_controller_test.go to verify if the underlying Pods can be created.

tenzen-y:
@SrikarChittemsetty Hello, what's your progress on this PR?

tenzen-y:
@SrikarChittemsetty Hi, sorry for the periodic messages to you. This is currently tracked by the new release candidate, but it is not a blocker for the next release. So, slipping this PR from the next release indicates a delay in releasing this fix for a few months.

I hope to include it in the next one. Please let me know if you need any help.

cc @kannon92

SrikarChittemsetty:
Nothing outstanding on my end. This is ready.

All checks pass on the current head: verify, unit, integration, and all four e2e jobs. Tide is pending because it needs lgtm and approve, not because anything is failing.

The change is still scoped to two files. `pod_webhook.go` renders negative priorities as `n` plus the absolute value so the label is a valid label value, and `pod_webhook_priority_test.go` covers the encoding, injectivity, and the webhook end to end.

The branch is 7 commits behind main. It merges cleanly as is, but I can rebase if you would rather review it on top of current main. Say the word and I will push it today.

Thanks for the heads up on the release. I would like it in this one. Tell me what you need from me to get it there.

SrikarChittemsetty:
Correction to what I just posted. I was looking at CI, not at your review comments, and I missed the five from @kannon92 and @tenzen-y on 28 August. That is on me, and it explains why you were chasing progress.

Working through them now. I will reply on each thread rather than in one lump.

pkg/webhooks/pod_webhook.go
SrikarChittemsetty: Added in 33924b6, at test/integration/webhook/pod_webhook_test.go.

Worth flagging what it turned up. spec.priority cannot be set directly: the priority admission plugin rejects a Pod carrying an integer priority and derives it from a PriorityClass instead. So each case creates a PriorityClass and references it by name, which is also how an operator actually reaches this.

That makes it a real end to end check. With the encoding reverted, the two negative cases fail because the API server rejects the label, which is the original symptom. The unit tests alone never exercised that.

pkg/webhooks/pod_webhook_priority_test.go
SrikarChittemsetty: Done in 33924b6. pod_webhook_priority_test.go is gone and the cases live in the TestDefault table alongside the existing priority case.

pkg/webhooks/pod_webhook_priority_test.go
SrikarChittemsetty: Removed. The testify usage went away with the file.

For context on how it got there: CodeRabbit asked for it earlier in this PR, and pkg/webhooks/jobset_webhook_test.go does use it, so I took that as the local convention. Your read is the one that matters here, and the folded cases use the plain table style.

pkg/webhooks/pod_webhook_priority_test.go
SrikarChittemsetty: Agreed, and done in 33924b6. The three test functions are gone and the cases are now entries in the TestDefault table, plus the integration test kannon92 asked for.

pkg/webhooks/pod_webhook.go
SrikarChittemsetty: You are right that the field is int32. The reason for int64 is the negation, not the field.

-int32(math.MinInt32) overflows and wraps back to MinInt32, so the sign never goes away:

```
priority=-1           int32-> n1             int64-> n1
priority=-1000        int32-> n1000          int64-> n1000
priority=-2147483647  int32-> n2147483647    int64-> n2147483647
priority=-2147483648  int32-> n-2147483648   int64-> n2147483648
```

I want to be accurate about how bad that is. I checked, and n-2147483648 still passes validation.IsValidLabelValue, so the suggestion would not reintroduce the original bug. The Pod is still admitted.

What it does break is the encoding this PR documents. n is meant to be followed by the absolute value, and for that one input it is followed by a negative number instead. It also stops being obviously injective at a glance.

Widening to int64 for the negation costs nothing and keeps the one input consistent with the other four billion. That is the whole argument. If you would rather take the simpler expression and accept MinInt32 rendering as n-2147483648, say so and I will apply your suggestion.

Either way I will add a short comment on the line, since the int64 is not obvious without knowing about the overflow.

SrikarChittemsetty:
All five review comments are addressed, pushed in 33924b6 and 3064c18. Replied on each thread rather than here, but in short:

- Integration test added at `test/integration/webhook/pod_webhook_test.go`. It creates a PriorityClass and references it by name, because the priority admission plugin rejects a Pod that sets `spec.priority` directly. With the encoding reverted, the negative cases fail there because the API server rejects the label, so it reproduces the original symptom end to end.
- `pod_webhook_priority_test.go` is deleted. The cases are now entries in the existing `TestDefault` table.
- testify is gone with that file.
- The one I pushed back on is `int64`. The field is int32, but the negation is what needs the width: `-int32(math.MinInt32)` wraps back to MinInt32. I checked, and the wrapped value still passes label validation, so your suggestion would not reintroduce the bug. It just breaks the `n` plus absolute value encoding for that one input. Happy to take the simpler expression if you would rather. Added a comment on the line either way.

Ran locally against envtest: `pkg/webhooks` and the full webhook integration suite pass, 42 specs.

@tenzen-y this should be ready for the release now. @kannon92 thanks for the integration test push, it caught something the unit tests could not.

tenzen-y (commented):
@SrikarChittemsetty Thank you for updates!
LGTM!

/lgtm

/assign @kannon92

kannon92 (commented):
/lgtm
/approve

We don't really do a lot of jobset patch releases. Our hope is to release a new minor in the next few weeks. I hope that is okay.

tenzen-y:
> /lgtm /approve
> 
> We don't really do a lot of jobset patch releases. Our hope is to release a new minor in the next few weeks. I hope that is okay.

I'm fine w/o patch release.

PR #1292, opened 2026-07-30
KEP-1282: Add Execution Attempts Tracking

jianqiaol:
#### What type of PR is this?
/kind documentation

This PR introduces **KEP-1282: Add Execution Attempts Tracking**. 

The KEP proposes adding a monotonic `ExecutionAttempts` counter to `JobSet.Status` and propagating it as a pod annotation (`jobset.sigs.k8s.io/execution-attempt`). 

**Why we need it:**
Currently, `jobset.sigs.k8s.io/restart-attempt` only tracks failure-triggered restarts. When queueing workloads (e.g., via Kueue), JobSets are frequently suspended (terminating pods) and later resumed (spawning new pods). These suspension/resumption cycles do not increment the restart counter, causing logs from different execution phases to share the same identifier. 

The proposed `execution-attempt` annotation will allow external logging and telemetry systems to cleanly identify and isolate log streams for different execution cycles of a JobSet.

Issue [#1282](https://github.com/kubernetes-sigs/jobset/issues/1282)

This PR only contains the KEP design document. The implementation changes have been split into a separate PR ([#1283](https://github.com/kubernetes-sigs/jobset/pull/1283)).

**Key Design Updates in KEP:**
- **Upgrade Migration Logic**: The design has been updated to handle controller upgrades gracefully. For active JobSets with `nil` attempts, the controller will initialize `status.executionAttempts` to the current `status.restarts` value (instead of `0`) to preserve history and maintain the `executionAttempts >= restarts` invariant. New JobSets will initialize to `0`.

```release-note
NONE

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

* **New Features**
  * Added a proposal to track JobSet execution attempts through status information.
  * Jobs and Pods will expose their execution attempt through annotations.
  * Defines behavior for retries, suspended execution resumes, upgrades, and feature-gate-disabled environments.
* **Documentation**
  * Added lifecycle, rollout, migration, and testing details for the proposed feature.


<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/keps/1282-execution-attempts/README.md
+++ b/keps/1282-execution-attempts/README.md
@@ -1,0 +1,114 @@
+# KEP-1282: Add Execution Attempts Tracking
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [Example](#example)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Controller Logic](#controller-logic)
+    - [1. Initialization](#1-initialization)
+    - [2. Increment on resume](#2-increment-on-resume)
+    - [3. Increment on restart](#3-increment-on-restart)
+    - [4. Data propagation](#4-data-propagation)
+- [Test Plan](#test-plan)
+  - [Unit Tests](#unit-tests)
+  - [Integration Tests](#integration-tests)
+<!-- /toc -->
+
+## Summary
+
+This KEP proposes adding an `ExecuteAttempts` counter to the JobSet status and propagating it as a pod annotation (`jobset.sigs.k8s.io/execute-attempt`). This allows external logging and telemetry systems to identify and isolate log streams for different execution cycles of a JobSet, especially when it is repeatedly suspended and resumed (e.g., by Kueue).
+
+## Motivation
+
+Logging deployment streams currently rely on the `jobset.sigs.k8s.io/restart-attempt` annotation to tag logs per execution cycle. However, when queueing workloads (e.g., via Kueue), JobSets are frequently evicted via suspension (terminating pods) and later resumed (spawning new pods).
+
+Because JobSet resumptions do not increment the `restart-attempt` counter (which only increments during failure policy evaluations), logs from multiple distinct execution phases share an identical identifier. Consequently, users cannot properly isolate log streams by their scheduling lifecycle, severely impeding debugging efforts (e.g., isolating logs to investigate node-level hardware failures during a specific execution pass).
+
+### Goals
+
+- Provide a persistent, monotonic counter that increments on **both** failure-triggered restarts and suspension-driven resumptions.  
+- Inject this new metric as a pod annotation to allow external logging systems (like FluentBit) to cleanly slice log iterations.
+
+### Non-Goals
+
+- Alter the lifecycle definitions for `jobset.sigs.k8s.io/restart-attempt`. The existing restart tracking acts strictly on failure conditions and will remain unchanged.
+
+## Proposal
+
+Introduce a new pod annotation (`jobset.sigs.k8s.io/execute-attempt`) to the JobSet API to track the total number of execution iterations a JobSet has undergone.
+
+To track this robustly across controller restarts, a new state field `executeAttempts` will be introduced within `JobSet.status`.
+
+### Example
+
+1. **Initial Scheduled Run:** A JobSet starts. Pods land on Nodepools A and B. Both `restart-attempt` and `execute-attempt` evaluate to `"0"`.  
+2. **Preemption (Eviction):** Kueue evicts the JobSet (`suspend: true`). Active pods are deleted.  
+3. **Resumed Execution:** Kueue readmits the JobSet (`suspend: false`). New pods land on Nodepools C and D.  
+   * Current state: `restart-attempt` remains `"0"`.  
+   * Proposed state: `execute-attempt` increments to `"1"`.
+
+The new annotation provides a distinct monotonic execution token to filter telemetry data accurately.
+
+## Design Details
+
+### API Changes
+
+Add `ExecuteAttempts` to `JobSetStatus` struct:
+
+```go
+type JobSetStatus struct {
+    ...
+	// executeAttempts tracks the number of execution lifecycles.
+	// +optional
+	ExecuteAttempts *int32 `json:"executeAttempts,omitempty"`
+    ...
+}
+```
+
+We use a pointer `*int32` to allow the field to be `nil` initially.
+*   `nil` indicates the JobSet has never run.
+*   `0` indicates the first run.
+*   `>0` indicates subsequent resumptions or restarts.
+
+Define a new constant for the annotation key:
+```go
+const (
+    ...
+	// ExecuteAttemptsKey is the annotation key for the execution attempt count.
+	ExecuteAttemptsKey = "jobset.sigs.k8s.io/execute-attempt"
+)
+```
+
+### Controller Logic
+
+#### 1. Initialization
+When the JobSet starts executing for the first time (either because it was created unsuspended, or when it is first resumed from an initially suspended state), the controller initializes `Status.ExecuteAttempts` to `0`.
+
+#### 2. Increment on resume
+When the JobSet transitions from suspended to unsuspended:
+*   If `Status.ExecuteAttempts` is not `nil`, it is incremented by 1.
+*   The suspended condition is updated to `false`.
+
+#### 3. Increment on restart
+When a failure policy triggers a recreate of all jobs (which is a restart):
+*   `Status.ExecuteAttempts` is incremented by 1.
+
+#### 4. Data propagation
+During child Job creation (in `constructJob`) and during Job resumption (in `resumeJob`), the controller injects the current `Status.ExecuteAttempts` value (defaulting to `0` if nil) into the Job's annotations and its Pod template annotations using the `jobset.sigs.k8s.io/execute-attempt` key.
+This ensures both newly created jobs and existing resumed jobs (which might have updated templates from Kueue) receive the correct annotation.
+
+## Test Plan
+
+### Unit Tests
+*   Verify `ExecuteAttempts` initialization and increment logic in controller unit tests (`pkg/controllers/jobset_controller_test.go`).
+*   Verify annotation propagation in job construction tests.
+
+### Integration Tests
+*   Verify suspend/resume cycles increment `ExecuteAttempts` and propagate annotations to child Jobs and Pods (`test/integration/controller/jobset_controller_test.go`).
+*   Verify failure-triggered restarts increment `ExecuteAttempts`.
+*   Verify that initially suspended JobSet starts with `ExecuteAttempts` unset and initializes to `0` on first resume.
--- a/keps/1282-execution-attempts/README.md
+++ b/keps/1282-execution-attempts/README.md
@@ -10,10 +10,11 @@
 - [Design Details](#design-details)
   - [API Changes](#api-changes)
   - [Controller Logic](#controller-logic)
-    - [1. Initialization](#1-initialization)
-    - [2. Increment on resume](#2-increment-on-resume)
-    - [3. Increment on restart](#3-increment-on-restart)
-    - [4. Data propagation](#4-data-propagation)
+    - [1. Initialization and Upgrade Migration](#1-initialization-and-upgrade-migration)
+    - [2. Increment on resume (with Idempotency)](#2-increment-on-resume-with-idempotency)
+    - [3. Increment on restart (with Idempotency)](#3-increment-on-restart-with-idempotency)
+    - [4. Atomicity of Updates on Resume and Data Propagation](#4-atomicity-of-updates-on-resume-and-data-propagation)
+    - [5. Feature Gate Disabled Behavior](#5-feature-gate-disabled-behavior)
 - [Test Plan](#test-plan)
   - [Unit Tests](#unit-tests)
   - [Integration Tests](#integration-tests)
@@ -21,7 +22,7 @@
 
 ## Summary
 
-This KEP proposes adding an `ExecuteAttempts` counter to the JobSet status and propagating it as a pod annotation (`jobset.sigs.k8s.io/execute-attempt`). This allows external logging and telemetry systems to identify and isolate log streams for different execution cycles of a JobSet, especially when it is repeatedly suspended and resumed (e.g., by Kueue).
+This KEP proposes adding an `ExecutionAttempts` counter to the JobSet status and propagating it as a pod annotation (`jobset.sigs.k8s.io/execution-attempt`). This allows external logging and telemetry systems to identify and isolate log streams for different execution cycles of a JobSet, especially when it is repeatedly suspended and resumed (e.g., by Kueue).
 
 ## Motivation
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: keps/1282-execution-attempts/README.md [hunk 1]
```diff
@@ -0,0 +1,114 @@
...
+- Provide a persistent, monotonic counter that increments on **both** failure-triggered restarts and suspension-driven resumptions.  
+- Inject this new metric as a pod annotation to allow external logging systems (like FluentBit) to cleanly slice log iterations.
+
+### Non-Goals
+
+- Alter the lifecycle definitions for `jobset.sigs.k8s.io/restart-attempt`. The existing restart tracking acts strictly on failure conditions and will remain unchanged.
+
+## Proposal
+
+Introduce a new pod annotation (`jobset.sigs.k8s.io/execute-attempt`) to the JobSet API to track the total number of execution iterations a JobSet has undergone.
```
kannon92: What do you mean by a pod annotation?

Will this be on the pods that JobSet creates?

kannon92: Farther down I think this is correct.

keps/1282-execution-attempts/README.md
```diff
@@ -0,0 +1,114 @@
...
+Define a new constant for the annotation key:
+```go
+const (
+    ...
+	// ExecuteAttemptsKey is the annotation key for the execution attempt count.
+	ExecuteAttemptsKey = "jobset.sigs.k8s.io/execute-attempt"
+)
+```
+
+### Controller Logic
```
kannon92, line 90: What about in place pod restarts? Would that count as a execution attempt?

keps/1282-execution-attempts/kep.yaml
```diff
@@ -0,0 +1,18 @@
...
+creation-date: 2026-07-27
+reviewers:
+  - "@kannon92"
+approvers:
+  - "@andreyvelich"
+see-also:
+  - "NA"
+replaces:
+  - "NA"
+stage: alpha
```
kannon92, line 15: Will you introduce a feature gate? 

What is the graduation criteria if this feature is alpha?

kannon92:
Minor nits but proposal looks good to me.

/cc @andreyvelich

keps/1282-execution-attempts/README.md
jianqiaol: In-place pod restarts (e.g., container restarts within an existing Pod due to crash loops, or in-place pod resource resizing) will **not** count as an execution attempt.

`ExecuteAttempts` specifically tracks JobSet-level scheduling and execution lifecycles:
1. **Resumptions**: Transitions from `suspend: true` -> `suspend: false` (e.g., Kueue evicting and re-admitting the workload).
2. **Controller Restarts**: JobSet failure policy evaluations that trigger recreation of child Jobs.

Container-level restarts or in-place updates keep the same Pod UID and lifecycle within an ongoing execution pass. I added an explicit note in the **Non-Goals** section of the KEP to make this explicit.

keps/1282-execution-attempts/kep.yaml
jianqiaol: Good catch! Yes, we will introduce a feature gate `ExecuteAttemptsTracking` in `pkg/features` (disabled by default in Alpha).

I added a **Graduation Criteria** section to the KEP with the following milestones:
- **Alpha (v0.13.0)**:
  - Introduce `ExecuteAttemptsTracking` feature gate (disabled by default).
  - Add `ExecuteAttempts` to `JobSetStatus` and propagate `jobset.sigs.k8s.io/execute-attempt` to Jobs and Pod templates.
  - Complete unit and integration test coverage for initialization, resume increments, restart increments, and idempotency.
- **Beta**:
  - Enable `ExecuteAttemptsTracking` by default.
  - Validate integration with Kueue suspend/resume workflows and external logging pipelines (e.g., FluentBit).
  - No major bugs or breaking behavior reported over a release cycle.
- **GA**:
  - Lock feature gate to default-enabled / unconditional.
  - At least one release cycle in Beta with demonstrated production adoption and stability.

kannon92: Looks great.

keps/1282-execution-attempts/README.md
```diff
@@ -0,0 +1,143 @@
...
+### API Changes
+
+Add `ExecuteAttempts` to `JobSetStatus` struct:
+
+```go
+type JobSetStatus struct {
+    ...
+	// executeAttempts tracks the number of execution lifecycles.
+	// +optional
+	ExecuteAttempts *int32 `json:"executeAttempts,omitempty"`
```
kannon92: I don't love this name. It sounds a bit awkard.

What about the following:

- executionAttempts
- runAttempts
- lifecycleAttempts
- executionGeneration
- executionSequence

andreyvelich: If this API controller by `suspend` why not calling it: `suspendAttempts`?

andreyvelich: So your total counter will be:
restart-attempts + suspend-attempts

keps/1282-execution-attempts/README.md
```diff
@@ -0,0 +1,143 @@
...
+### Non-Goals
+
+- Alter the lifecycle definitions for `jobset.sigs.k8s.io/restart-attempt`. The existing restart tracking acts strictly on failure conditions and will remain unchanged.
+- Track container-level restarts, crash loops, or in-place pod restarts/resizes. Container restarts within an existing running Pod do not recreate the Pod or Job, and therefore do not increment `ExecuteAttempts`.
+
+## Proposal
+
+Introduce a new pod annotation (`jobset.sigs.k8s.io/execute-attempt`) to the JobSet API to track the total number of execution iterations a JobSet has undergone.
+
+To track this robustly across controller restarts, a new state field `executeAttempts` will be introduced within `JobSet.status`.
```
andreyvelich: Why we didn't add new `JobSet.status` field for restart attempts?

keps/1282-execution-attempts/README.md
```diff
@@ -0,0 +1,143 @@
...
+    ...
+	// executeAttempts tracks the number of execution lifecycles.
+	// +optional
+	ExecuteAttempts *int32 `json:"executeAttempts,omitempty"`
+    ...
+}
+```
+
+We use a pointer `*int32` to allow the field to be `nil` initially.
+*   `nil` indicates the JobSet has never run.
```
andreyvelich: Do we really need `nil` value here?
For in-place restarts we used 0 as the first value: https://github.com/kubernetes-sigs/jobset/blob/main/api/jobset/v1alpha2/jobset_types.go#L248

keps/1282-execution-attempts/README.md
```diff
@@ -0,0 +1,143 @@
...
+```
+
+### Feature Gate
+
+This enhancement will be guarded by a new feature gate in `pkg/features`:
+
+```go
+// ExecuteAttemptsTracking enables tracking and propagating JobSet execution attempts.
+ExecuteAttemptsTracking featuregate.Feature = "ExecuteAttemptsTracking"
+```
```
andreyvelich: @kannon92 Do want to control every new JobSet feature under feature gate?
I am wondering if this is really needed, tho.

andreyvelich (commented):
Thanks for this @jianqiaol!
I left a few thoughts.

keps/1282-execution-attempts/README.md
kannon92: I think its a good practice. I'm fine with this going to beta on but what happens if this feature causes side effects for users and they want to disable it?

That is the point of feature gates. Otherwise we have to do a painful hotfix.

keps/1282-execution-attempts/README.md
kannon92: I think its suspend or restarts. Resuming the JobSet does not reset the `executeAttemps`.

keps/1282-execution-attempts/README.md
andreyvelich: > users and they want to disable it?

Are feature gates good mechanism to allow users to opt out from the capability? I thought, it is usually managed by [manager config](https://github.com/kubernetes-sigs/jobset/blob/main/api/config/v1alpha1/configuration_types.go) or dedicated API.

keps/1282-execution-attempts/README.md
andreyvelich: Yeah, but why not just iterating the `SuspendAttempts` value when JobSet was suspended?
Or I am missing something here @jianqiaol?

keps/1282-execution-attempts/README.md
kannon92: Well the feature is unproven and if there is a bug the feature gate is a good mechanism to not impact users.

But if a user wants to permanetly disable this feature than a configuration is necessary.

keps/1282-execution-attempts/README.md
jianqiaol: Thanks for the feedback on the naming and API structure.

I agree that executionAttempts is a better fit than executeAttempts. I'll update both the KEP and the implementation to use executionAttempts (JSON: executionAttempts) and the annotation jobset.sigs.k8s.io/execution-attempt.

Regarding the suggestion to use suspendAttempts or keep the counters separate: Since this counter is designed to track both suspension resumptions AND failure-triggered restarts, suspendAttempts would be too narrow.

Additionally, having a single monotonic counter is highly preferred for logging and telemetry pipelines (like FluentBit/Fluentd). Requiring these systems to calculate restarts + suspendAttempts dynamically is complex, often requiring custom scripts (like Lua), and introduces query-time overhead in logging backends.

Extra context: This unified design (total execution attempts = restarts + resumptions) was actually requested by and aligned with the customer team, who are the primary users of this telemetry data. Given this is a direct request from our customers to simplify their logging pipelines, I would prefer to keep it as a single unified counter.

keps/1282-execution-attempts/README.md
jianqiaol: We actually already have [status.restarts](https://github.com/kubernetes-sigs/jobset/blob/9147d8d61eaaf5f76f13c419da6f58e8b2a1b81c/api/jobset/v1alpha2/jobset_types.go#L213) in the JobSetStatus struct, which tracks the number of restarts. The jobset.sigs.k8s.io/restart-attempt annotation on the pods is derived from this status.restarts field.

keps/1282-execution-attempts/README.md
jianqiaol: Unlike in-place restarts (which only apply to pods that are already active and running), a JobSet can be created in a suspended state. 

Because of this, the controller needs a way to distinguish between two scenarios when a JobSet is resumed:
1.  **First Run (Start)**: It was created suspended and is now starting for the first time (should stay `0`).
2.  **Resumption**: It was already running, got suspended, and is now resuming (should increment to `1`).

By using a pointer (`*int32`) that allows `nil`, we can handle this cleanly:
*   **`nil`** explicitly represents the **"never executed"** state.
*   **If we resume and the counter is `nil`**: We know it is the first run, so we initialize it to `0` (no increment).
*   **If we resume and the counter is already `0` (or more)**: We know it has run before, so we increment it.

This design is critical because it **allows us to avoid adding an extra `HasRun` status field** to the API. If we defaulted to `0` (non-pointer), we would have no way to distinguish a first-time start from a subsequent resumption without introducing another tracking field.

#### Bonus: Upgrade Compatibility
The `nil` state also helps us handle upgrade compatibility safely. It allows the controller to detect running JobSets from a pre-upgrade version (which won't have the field in `etcd` and will thus deserialize as `nil`) and migrate them without resetting their history mid-run.

keps/1282-execution-attempts/README.md
jianqiaol: Thanks for the great discussion!

I agree with Kannon that a feature gate (ExecuteAttemptsTracking) is the better fit here. Since this telemetry automatically applies to all JobSets, having a global off-switch is a great safety net in case of any unexpected bugs during Alpha.

Unlike features that change runtime behavior, this is just metadata, so a permanent configuration option in the API seems like overkill. A temporary feature gate (eventually graduating to GA/always-on) is consistent with how JobSet handles similar rollouts.

I'll update the KEP to clearly define the behavior when the gate is disabled and when it gets re-enabled. Thanks again!

keps/1282-execution-attempts/README.md
andreyvelich: I see, thanks for sharing.
What if JobSet is created with `suspend: true`? The `executeAttempts` will be 0 in that case, right?

keps/1282-execution-attempts/README.md
andreyvelich: Don't we prefer to name it `JobSet.status.executions` to be consistent with Restarts?

keps/1282-execution-attempts/README.md
andreyvelich: Yeah, this is good point. Makes sense to me to distinguish JobSets that are created with `suspend: true`.

keps/1282-execution-attempts/README.md
andreyvelich: Found an answer here, thanks: https://github.com/kubernetes-sigs/jobset/pull/1292#discussion_r3723025057

keps/1282-execution-attempts/README.md
jianqiaol: I considered `executions` for consistency, but there's a semantic difference in how the numbers are interpreted:
* **`restarts`** is an event counter (`restarts: 0` means zero restarts have happened).
* **`executionAttempts`** is a 0-indexed identifier for the current run. 
If we use `executions` (and keep it 0-indexed), the first run would show `status.executions: 0`, which sounds like it hasn't executed yet. `executionAttempts: 0` clearly communicates "attempt index 0." 
If we strongly prefer `executions`, we should probably make it 1-indexed so the first run is `executions: 1`. I lean towards keeping `executionAttempts` (0-indexed), but I'm happy to change it if you prefer!

andreyvelich: Yeah, this is good point. @kannon92 Any thoughts on the naming here?

kannon92: `executionAttempts` sounds good to me.

jianqiaol:
Hi @kannon92 @andreyvelich,

All previous comments and design discussions on both the KEP (PR #1292) and the implementation (PR #1283) have been addressed:

1. **KEP-1282 (PR #1292)**:
   - Updated the API field name to `executionAttempts` and annotation to `jobset.sigs.k8s.io/execution-attempt`.
   - Documented the upgrade baseline semantics for `nil` values (`executionAttempts >= restarts`).
   - Defined the `ExecutionAttemptsTracking` feature gate (Alpha, disabled by default) and graduation criteria.
   - Clarified idempotency and atomicity guarantees during resume/restart reconcile cycles.

2. **Implementation (PR #1283)**:
   - Aligned code, CRDs, and Python SDK with the finalized KEP naming and feature gate.
   - Added integration test coverage for all lifecycle scenarios (fresh unsuspended, create-suspended-then-resume, multiple eviction/resume passes, failure restarts, upgrade migration, and feature gate disabled/re-enabled states).
   - All CI presubmit checks are green.

Could you please take a look when you have a moment and help move these forward? Thank you!

kannon92:
can you check over the coderabbit threads? They could be legit.

From a review standpoint, I am happy with this KEP.

/lgtm

@andreyvelich is marked as a approver so ask him to add his approve label.

andreyvelich (commented):
That looks good to me, thanks @jianqiaol!
/lgtm

kannon92 (commented):
/lgtm
/approve

jianqiaol:
Thanks @kannon92 and @andreyvelich!

PR #1305, opened 2026-08-26
Fix nil deref of whenDeleted in volume claim retention checks

### LINKED ISSUE #1304: Half-guarded RetentionPolicy dereference in VolumeClaimPolicies handling
**What happened**:
Two call sites check the outer `RetentionPolicy` pointer for nil and then dereference the
inner `WhenDeleted` pointer unconditionally:

- `pkg/controllers/volume_claim_policy.go:54`
  ```go
  if volumeClaimPolicy.RetentionPolicy != nil && *volumeClaimPolicy.RetentionPolicy.WhenDeleted == jobset.RetentionPolicyDelete {
  ```
- `pkg/webhooks/jobset_webhook.go:595`
  ```go
  if policy.RetentionPolicy != nil && *policy.RetentionPolicy.WhenDeleted != jobset.RetentionPolicyRetain {
  ```

`WhenDeleted` is a `*RetentionPolicyType` (`api/jobset/v1alpha2/jobset_types.go:607`).

To be clear, I do not believe this is reachable today. `WhenDeleted` has
`+kubebuilder:default=Delete`, which lands in the CRD as `default: Delete`, so structural
defaulting populates it whenever `retentionPolicy` is present (including an explicit
`whenDeleted: null`, since the field is not nullable). That defaulting runs before
validating admission, so the webhook sees it set as well, and the mutating webhook
(`jobset_webhook.go:152-159`) separately fills in `RetentionPolicy` when it is nil.

A guard that checks the outer pointer reads as defensive, so a reader reasonably assumes it
is safe against a nil inner pointer too. Right now it is half-defensive: it relies on a
CRD-level default that is invisible at the call site, and any future in-process caller that
builds a JobSet in Go (a unit test, or another controller constructing the object before
submitting it) will panic rather than take the default branch.

Proposed fix: make the intent explicit and drop the reliance on defaulting, e.g.

```go
if volumeClaimPolicy.RetentionPolicy != nil &&
    ptr.Deref(volumeClaimPolicy.RetentionPolicy.WhenDeleted, jobset.RetentionPolicyDelete) == jobset.RetentionPolicyDelete {
```

k8s.io/utils/ptr is already a dependency (used in pkg/webhooks/jobset_webhook.go); the controller file would need the import added.

Not reproducible through the API server, for the reasons above. It is reachable from
in-package Go, e.g. a unit test in `pkg/controllers`:

```go
js := &jobset.JobSet{
    Spec: jobset.JobSetSpec{
        VolumeClaimPolicies: []jobset.VolumeClaimPolicy{{
            Templates:       []corev1.PersistentVolumeClaim{{ObjectMeta: metav1.ObjectMeta{Name: "cache"}}},
            RetentionPolicy: &jobset.VolumeRetentionPolicy{}, // WhenDeleted left nil
        }},
    },
}
// r.createPVCsIfNecessary(ctx, js, js.Spec.VolumeClaimPolicies[0]) -> panic
```

Both lines were introduced in #1098 (KEP-572, VolumeClaimPolicies). I found this by code inspection, not by running into panic on a running cluster. Happy to send a PR if maintainers agree this is worth tightening. 

**Environment**:
- Kubernetes version (use `kubectl version`):  N/A
- JobSet version (use `git describe --tags --dirty --always`): v0.13.0-devel-48-g370f936
- Cloud provider or hardware configuration: N/A
- Install tools: N/A
- Others: go1.27.0 darwin/arm64

__________
Disclaimer : AI Assistance was used in drafting this issue.

Viswalahiri:
#### What type of PR is this?
/kind bug

The controller and the validating webhook both checked the outer RetentionPolicy pointer for nil, then dereferenced WhenDeleted. At first glance although it appears defensive, but is not. It leans on the CRD default having been applied, which we can't see from the call site.

This isn't reachable through the API server today. Structural defaulting populates WhenDeleted whenever retentionPolicy is set, and the mutating webhook fills in RetentionPolicy when it's nil.  Use ptr.Deref with Delete as the fallback so the intent is explicit.

Fixes #1304 

```release-note
NONE
```

____
Disclosre : AI Assistance was used in part in this issue

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented errors when volume retention policies omit the deletion behavior.
  * PVCs now safely default to the Delete policy when no deletion policy is specified.
  * Improved validation for existing PVCs to ensure incompatible retention settings are rejected.
  * Correctly assigns ownership to newly created PVCs under the default policy.


Viswalahiri:
@andreyvelich please let me know if this PR can be improved in any way, thank you!

andreyvelich (commented):
Good catch, thanks @Viswalahiri!
/lgtm
/approve

Issue #1186, opened 2026-03-13
[Feature] Add active deadline Seconds for JobSet


kannon92:
**What would you like to be added**:

Add a way to limit the execution time of a JobSet.

**Why is this needed**:

Similar to Job, we should have the concept of an Active Deadline Seconds on the top level jobset. When the timer exceeds the jobset should transition to failed and the underlying objects are removed.

This enhancement requires the following artifacts:

- [x] Design doc
- [x] API change
- [x] Docs update

The artifacts should be linked in subsequent comments.

gzb1128:
Before writing a KEP for this feature, I'd like to align on three semantic decisions for `ActiveDeadlineSeconds` so the implementation direction doesn't need to change later. Here are my proposals and rationale, but I'm happy to defer to maintainers.

### 1. When the timer starts

Options:
- **(a)** From `metadata.creationTimestamp` (matches upstream `batch/v1.Job.ActiveDeadlineSeconds`).
- **(b)** From when the JobSet first becomes active (i.e., after being unsuspended).
- **(c)** From the most recent run (i.e., reset on each `failurePolicy` restart).

**Recommendation: (a)** — it mirrors K8s Job behavior, is unambiguous, and needs no new status field. (b)/(c) force a new field in `.status` to track "active-since" time and add coupling with the suspend/restart logic. If the team wants (b)/(c), that's fine, but it's worth acknowledging the extra complexity up front.

### 2. Interaction with `suspend`

When a JobSet is `suspend: true`, should the deadline pause?

- **(a)** No pause — the timer runs on wall-clock time from creation regardless of suspend.
- **(b)** Pause — accumulate "active duration" and exclude suspended intervals.

**Recommendation: (a) for alpha**, documented as a known limitation. (b) requires tracking suspend/resume transitions and time-in-state fields, which is heavy for alpha. A middle ground, if we want stricter semantics, is to reject the combination of `activeDeadlineSeconds` + `suspend: true` in the webhook (CEL or webhook-level).

### 3. Interaction with `failurePolicy` restarts

When the JobSet restarts (actions `RestartJobSet` / `RestartJob`), does the deadline reset?

- **(a)** Reset — the deadline applies to one complete run.
- **(b)** Don't reset — the deadline applies to the whole JobSet lifetime across restarts.

[...]

PR #1306, opened 2026-08-26
Add KEP-1186 for JobSet activeDeadlineSeconds

yindia:
#### What type of PR is this?
/kind feature

Adds the KEP for a top-level `activeDeadlineSeconds` on JobSet. It bounds how long a JobSet may be continuously active before the controller marks it Failed and deletes its active Jobs. Semantics follow batch/v1 Job: the timer is measured from `status.startTime`, does not run while suspended, and resets on resume. This PR is the design only; no API or controller code is included yet.

Related to #1186 (design artifact; the issue stays open until implementation).

Updated per review feedback to align with upstream `batch/v1.Job`:
- `activeDeadlineSeconds` is mutable on a running JobSet (no immutability CEL rules); the controller recomputes remaining time from `status.startTime` each reconcile.
- `.status.startTime` is a general active-start timestamp, populated whenever the JobSet is active regardless of `activeDeadlineSeconds`.
- On expiry the controller deletes active child Jobs and returns early in the same reconcile to free accelerators promptly.

Remaining open item for maintainer input: the target milestones in kep.yaml (marked TODO).

This PR was drafted with AI assistance.

NONE
```release-note
NONE
```

### DIFF
--- a/keps/1186-active-deadline-seconds/README.md
+++ b/keps/1186-active-deadline-seconds/README.md
@@ -1,0 +1,530 @@
+# KEP-1186: Add ActiveDeadlineSeconds for JobSet
+
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals](#non-goals)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Story 1](#story-1)
+    - [Story 2](#story-2)
+  - [Notes/Constraints/Caveats](#notesconstraintscaveats)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+  - [API Changes](#api-changes)
+  - [Semantics](#semantics)
+    - [1. When the timer starts](#1-when-the-timer-starts)
+    - [2. Interaction with suspend](#2-interaction-with-suspend)
+    - [3. Interaction with failurePolicy restarts](#3-interaction-with-failurepolicy-restarts)
+      - [What Job does for activeDeadlineSeconds and backoffLimit](#what-job-does-for-activedeadlineseconds-and-backofflimit)
+  - [Controller Logic](#controller-logic)
+  - [Feature Gate Disabled Behavior](#feature-gate-disabled-behavior)
+  - [Test Plan](#test-plan)
+    - [Prerequisite testing updates](#prerequisite-testing-updates)
+    - [Unit Tests](#unit-tests)
+    - [Integration tests](#integration-tests)
+  - [Graduation Criteria](#graduation-criteria)
+- [Implementation History](#implementation-history)
+- [Drawbacks](#drawbacks)
+- [Alternatives](#alternatives)
+<!-- /toc -->
+
+## Summary
+
+This KEP proposes adding an optional `spec.activeDeadlineSeconds` field to the
+JobSet API. It bounds the total time a JobSet may be continuously active. When
+the deadline is exceeded, the JobSet controller marks the JobSet as `Failed`
+with a dedicated reason and deletes the active child Jobs, matching
+`batch/v1.Job.spec.activeDeadlineSeconds`.
+
+The timer is measured from a new `.status.startTime`, does not run while the
+JobSet is suspended, and resets on resume.
+
+## Motivation
+
+JobSet has no way to bound total execution time. A distributed training or HPC
+run that hangs (deadlocked collective, stuck rendezvous, wedged data loader)
+can hold expensive accelerators indefinitely until a human notices. `batch/v1`
+Jobs solve this with `activeDeadlineSeconds`; JobSet users expect the same
+guardrail at the JobSet level.
+
+`ttlSecondsAfterFinished` already exists, but it only cleans up after a JobSet
+finishes. It cannot force an unfinished, wedged JobSet to stop. This KEP covers
+that case.
+
+The deadline targets a JobSet that is continuously active but making no
+progress (a hang). Crash-loops are already bounded by `maxRestarts`, and each
+global restart or resume starts a fresh run. So the deadline and `maxRestarts`
+cover different failure modes: the deadline catches a single wedged attempt,
+`maxRestarts` caps how many times the JobSet restarts.
+
+### Goals
+
+- Add `spec.activeDeadlineSeconds` to bound the continuous active runtime of a
+  JobSet.
+- On expiry, transition the JobSet to a terminal `Failed` state with a distinct
+  condition reason and delete active child Jobs.
+- Do not accrue the deadline while the JobSet is suspended.
+- Ship behind a feature gate, disabled by default in alpha.
+
+### Non-Goals
+
+- Per-`ReplicatedJob` or per-Job deadlines. Users can already set
+  `activeDeadlineSeconds` on the Job template inside a `ReplicatedJob` for
+  per-Job bounds; this KEP is the JobSet-level aggregate deadline.
+- Changing `ttlSecondsAfterFinished` semantics or the failure policy restart
+  counters.
+- A "deadline exceeded → restart" action. Expiry is terminal `Failed`.
+
+## Proposal
+
+Introduce `spec.activeDeadlineSeconds *int64` and a `.status.startTime
+*metav1.Time`. The controller measures elapsed active time from `startTime`; on
+each reconcile it either requeues for the remaining duration or, if the deadline
+has passed, fails the JobSet and deletes its active Jobs.
+
+### User Stories
+
+#### Story 1
+
+As an ML platform operator, I run large training JobSets on scarce GPUs. If a
+run wedges (NCCL deadlock, stuck checkpoint restore), I want it to be
+force-failed after N hours so the accelerators are freed and my queue
+(e.g. Kueue) can schedule the next workload, without manual intervention.
+
+#### Story 2
+
+As a batch/HPC user, I submit JobSets with a known worst-case runtime per
+attempt. I set `activeDeadlineSeconds` so a single continuously-active run that
+overruns its budget is failed instead of hanging. The bound applies to one
+active attempt, not to total wall-clock across suspends and restarts (see
+Notes).
+
+### Notes/Constraints/Caveats
+
+- The deadline is continuous active time, not wall-clock since creation:
+  suspended intervals do not count. A Kueue-managed JobSet is frequently
+  suspended and resumed.
+- `activeDeadlineSeconds` is independent of `maxRestarts`. A JobSet can be
+  failed by exhausting restarts or by exceeding the deadline. When a child-Job
+  failure and an elapsed deadline are observed in the same reconcile, the
+  deadline is evaluated first, so the JobSet fails with `DeadlineExceeded`. This
+  is deliberate: the failure policy may restart the JobSet (`RestartJobSet`
+  resets `startTime`; `RestartJob` returns early), which would let a wedged or
+  over-budget workload bypass the deadline. Checking the deadline before the
+  failure policy closes that bypass. A genuinely completed JobSet still wins over
+  the deadline (success policy runs first).
+- The bound is per continuous-active attempt, not a JobSet lifetime cap.
+  Resume and global restart (`RestartJobSet`) each begin a fresh run and reset
+  the timer. `maxRestarts` bounds the number of global restart attempts.
+- `RestartJobSetAndIgnoreMaxRestarts` bypasses `maxRestarts`, so a JobSet driven
+  only by that action can restart without limit and receive a fresh per-attempt
+  deadline each time.
+- The timer starts at first unsuspend, before pods are ready. Time spent waiting
+  for startup ordering (`DependsOn`, coordinator, leader readiness) counts
+  against the deadline, so a JobSet stuck in startup is failed once the deadline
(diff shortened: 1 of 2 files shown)


keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,494 @@
...
+    // status.startTime, that the JobSet may be continuously active before the
+    // JobSet controller marks it Failed and deletes its active Jobs.
+    //
+    // The timer does not accrue while the JobSet is suspended: startTime is
+    // cleared on suspend and reset when the JobSet resumes. The value must be a
+    // positive integer. The field is immutable once set.
+    // +kubebuilder:validation:Minimum=1
+    // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="activeDeadlineSeconds is immutable"
+    // +optional
+    ActiveDeadlineSeconds *int32 `json:"activeDeadlineSeconds,omitempty"`
```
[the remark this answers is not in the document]
kannon92: Good bot. Yea this should be fixed.

kannon92:
I did a skim and I think this is a very good feature.

Maybe @jianqiaol can do a deeper review?

kannon92:
/retest

github seems to really hate this PR..

kannon92:
/assign @jianqiaol 

based on +1. Thank you for taking this on.

yindia:
@jianqiaol do you have any feedback on this ?

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,497 @@
...
+    // status.startTime, that the JobSet may be continuously active before the
+    // JobSet controller marks it Failed and deletes its active Jobs.
+    //
+    // The timer does not accrue while the JobSet is suspended: startTime is
+    // cleared on suspend and reset when the JobSet resumes. The value must be a
+    // positive integer. The field is immutable once set.
+    // +kubebuilder:validation:Minimum=1
+    // +kubebuilder:validation:XValidation:rule="self == oldSelf",message="activeDeadlineSeconds is immutable"
+    // +optional
+    ActiveDeadlineSeconds *int64 `json:"activeDeadlineSeconds,omitempty"`
```
jianqiaol, line 165: Thanks for detailing the validation rules!

Could we consider making `activeDeadlineSeconds` mutable on a running JobSet, even in alpha?

In upstream `batch/v1.Job`, `activeDeadlineSeconds` is mutable (see [`ValidateJobSpecUpdate`](https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/batch/validation/validation.go#L638-L652)). This allows operators or platform tools to extend the deadline for long-running jobs without restarting from scratch. Since the controller computes remaining time dynamically on each reconcile, supporting updates shouldn't require extra state.
@kannon92, what are your thoughts on allowing updates to this field?

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,497 @@
...
+
+```go
+// startTime is the timestamp from which activeDeadlineSeconds is measured.
+// It is set when the JobSet first becomes active (unsuspended), cleared when
+// the JobSet is suspended, and reset to the current time when the JobSet
+// resumes. It is only populated when activeDeadlineSeconds is set and the
+// feature gate is enabled.
+// +optional
+StartTime *metav1.Time `json:"startTime,omitempty"`
+```
```
jianqiaol, line 180: Nice catch noting this in the open questions! In upstream `batch/v1.Job`, [`.status.startTime`](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/job_controller.go#L1051-L1055) is populated whenever any Job becomes active, even if `activeDeadlineSeconds` is unset.
Since JobSet currently lacks a standard start timestamp in status, making `startTime` general would be valuable for ecosystem tools (CLI printers, dashboards) while staying consistent with upstream Kubernetes.

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,497 @@
...
+if !features.Enabled(JobSetActiveDeadlineSeconds) { return 0 }
+if js.Spec.ActiveDeadlineSeconds == nil { return 0 }         // unset
+if jobSetSuspended(js) { return 0 }                          // paused while suspended
+if js.Status.StartTime == nil { return 0 }                   // not active yet
+
+deadline  := js.Status.StartTime.Add(activeDeadline)
+remaining := deadline.Sub(clock.Now())
+if remaining <= 0 {
+    setJobSetFailedCondition(js, DeadlineExceededReason,
+        fmt.Sprintf("JobSet was active for %s, exceeding the %ds deadline",
```
jianqiaol: When the deadline expires, it might be safer to delete running child jobs (`r.deleteJobs(ctx, js, ownedJobs.active)`) immediately within this reconcile and return early (`return ctrl.Result{}, nil`), rather than waiting for the next reconcile loop.

This frees GPU/TPU resources right away (matching [upstream Job cleanup](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/job/job_controller.go#L1124-L1135)) and prevents subsequent steps in `Reconcile` from attempting to sync or recreate child jobs after timeout.

jianqiaol (commented):
Thanks for this great KEP, @yindia! The design is very thorough.
I left a few inline comments on the KEP for discussion:
- **Field Mutability:** Allowing `spec.activeDeadlineSeconds` to be updated on running JobSets, matching `batch/v1.Job`.
- **`status.startTime` Scope:** Making `startTime` a standard timestamp whenever a JobSet is active.
- **Immediate Cleanup:** Deleting active child jobs in the same reconcile upon deadline expiry to free GPU/TPU resources promptly.

keps/1186-active-deadline-seconds/README.md
yindia: Done, made `activeDeadlineSeconds` mutable and dropped both immutability CEL rules. Controller recomputes remaining time from `status.startTime` each reconcile, so no extra state.

keps/1186-active-deadline-seconds/README.md
yindia: Good point, made `startTime` general now, set whenever the JobSet is active regardless of `activeDeadlineSeconds`, matching upstream Job. Updated the KEP.

keps/1186-active-deadline-seconds/README.md
yindia: Agreed, updated it to delete the active Jobs and return early in the same reconcile so resources free up right away instead of waiting a loop.

yindia:
@jianqiaol Thanks for the review. I've addressed all the comments, please take another look.

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,507 @@
...
+(`pkg/controllers/jobset_controller.go`): `startTime` maintenance and the
+deadline check.
+
+**Placement in the reconcile loop.** The current order is: `jobSetFinished`
+early-exit (cleanup + TTL requeue) → failure policy (early return) → success
+policy (early return on completion) → `syncExecutionAttempts` →
+`reconcileReplicatedJobs` → suspend/resume handling. The deadline check runs
+after the failure and success policies, so a JobSet that completed or failed
+in the same reconcile wins over the deadline. Because
+`jobSetFinished` short-circuits at the top of the loop, an already-finished
```
jianqiaol: How about refine the order in the reconcile loop into:  success
policy (early return on completion) -> deadline check -> failure policy? If a child job fails while the deadline has already expired, ActiveDeadlinePolicy must evaluate before FailurePolicy. Otherwise, FailurePolicy would trigger a retry restart (e.g. RestartJobSet resetting startTime to now, or RestartJob returning early), allowing a wedged or over-budget workload to bypass the deadline.

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,507 @@
...
+// populated whenever the JobSet is active, independent of activeDeadlineSeconds,
+// matching batch/v1.Job.status.startTime.
+// +optional
+StartTime *metav1.Time `json:"startTime,omitempty"`
+```
+
+For observability, `.status.startTime` is exposed as a printer column:
+
+```go
+// +kubebuilder:printcolumn:name="StartTime",type=string,JSONPath=`.status.startTime`,description="Time the active-deadline timer started"
```
jianqiaol: Since `.status.startTime` is now a general active-start timestamp, let's update the printer column description to something like "Time the JobSet became active"

jianqiaol (commented):
Thanks for the update! Just left some minor feedbacks.

keps/1186-active-deadline-seconds/README.md
yindia: Done

keps/1186-active-deadline-seconds/README.md
yindia: Nice catch, that was a real bypass. I reordered it to `success → deadline → failure` so an expired JobSet fails with `DeadlineExceeded` instead of getting restarted and resetting `startTime`. I also called this out in the KEP as a deliberate departure from upstream ordering: in JobSet a restart resets the timer, whereas upstream `backoffLimit` is terminal, so we have to check the deadline first.

yindia:
@jianqiaol Done. Please let me know if anything is remaining

keps/1186-active-deadline-seconds/kep.yaml
```diff
@@ -0,0 +1,27 @@
...
+  - "/keps/1150-mutable-suspended-job-resources"
+replaces:
+  - "NA"
+stage: alpha
+# TODO(@yindia): confirm target milestones with maintainers before merging.
+latest-milestone: "v0.13.0"
+milestone:
+  alpha: "v0.13.0"
+  beta: "v0.14.0"
+  stable: "v0.16.0"
```
yindia, line 21: @jianqiaol Do you want me to update any of these milestones?

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,527 @@
...
+at the top of the loop, an already-finished JobSet is never re-failed by the
+deadline.
+
+**`startTime` maintenance** is general and not gated by `activeDeadlineSeconds`
+or the feature gate: `.status.startTime` tracks the JobSet's active-start time
+whenever the JobSet is active, matching upstream `batch/v1.Job`. Maintenance
+reuses the same suspend/resume transition detector as
+`syncExecutionAttempts`: `previouslySuspended =
+apimeta.IsStatusConditionTrue(js.Status.Conditions, JobSetSuspended)` compared
+with `isSuspended = jobSetSuspended(js)`.
```
jianqiaol: We probably don't need `previouslySuspended` transition tracking from `syncExecutionAttempts`. `status.startTime` is an idempotent timestamp, so we can simplify it to match upstream Job: clear on suspend, and set to `now` whenever `StartTime == nil && !isSuspended`.
Also, please make sure `startTime` initialization runs at the top of `reconcile` (before the deadline and failure policy checks) so `StartTime` is populated on the first reconcile when evaluating the deadline.

jianqiaol (commented):
Thanks @yindia, just one comment on the startTime maintenance. I'll let @kannon92 double-check if anything is missing for his use cases before final approval.

keps/1186-active-deadline-seconds/README.md
yindia: Done in c94bd5b. Dropped the `previouslySuspended` transition tracking; `startTime` maintenance now derives from the current suspend state only (clear on suspend, set to `now` when `StartTime == nil && !isSuspended`), matching upstream Job. Also moved `startTime` maintenance to the top of the reconcile loop (right after `jobSetFinished`) so `StartTime` is populated before the deadline and failure policy checks on the first reconcile.

Disclosure: this reply and the change were drafted with AI assistance.

yindia:
PTAL @kannon92

keps/1186-active-deadline-seconds/README.md
kannon92: Makes sense to me.

keps/1186-active-deadline-seconds/README.md
```diff
@@ -0,0 +1,530 @@
...
+  deadline on a running JobSet without restarting it, and may add it to a JobSet
+  that did not set it. Because the controller computes the remaining time
+  dynamically from `.status.startTime` on each reconcile, an update takes effect
+  on the next reconcile with no extra state to track: the new deadline is
+  `startTime + activeDeadlineSeconds` evaluated against the current time.
+  Lowering the value below the already-elapsed active time fails the JobSet on
+  the next reconcile; removing the field stops enforcement. No CEL transition
+  rules are needed.
+
+### Semantics
```
kannon92, line 204: One thing to check in implementation is how this interacts with ManagedBy.

kannon92 (commented):
/lgtm
/approve

I have one item but I don't see this blocking the KEP.

https://github.com/kubernetes-sigs/jobset/pull/1306/changes#r3936453554

keps/1186-active-deadline-seconds/README.md
yindia: @kannon92 Good callout. I missed how this feature affects managedBy. Want me to handle it in a separate PR?

kannon92: yea!

PR #1315, opened 2026-09-03
Apply TTL cleanup to externally managed JobSets

### LINKED ISSUE #1310: Support TTL cleanup for externally managed JobSets

Support `spec.ttlSecondsAfterFinished` cleanup for JobSets that set `spec.managedBy` to an external controller.

The JobSet reconciler currently returns immediately when `managedByExternalController` reports an external controller, before reaching `executeTTLAfterFinishedPolicy`:

https://github.com/kubernetes-sigs/jobset/blob/53640df8b4f90434f348c7778490770679af7481/pkg/controllers/jobset_controller.go#L147-L173

It would be useful to keep skipping normal JobSet reconciliation for externally managed objects while still reconciling the TTL policy. When the external controller has written a terminal `Completed` or `Failed` condition and the configured TTL expires, JobSet should delete the JobSet CR. A JobSet without a TTL should retain the current behavior.

This could be implemented as a cleanup path before the external-controller early return or as a separate TTL controller, so it does not cause the built-in controller to manage child Jobs or other status fields.

**Why is this needed**:

MultiKueue uses `spec.managedBy` on the manager-cluster JobSet and synchronizes terminal status from a worker-cluster JobSet. To prevent the worker JobSet from being deleted before its terminal status is synchronized, the remote copy should not own the lifecycle TTL. However, the manager JobSet is also never deleted because the built-in JobSet reconciler skips every externally managed JobSet, including its TTL policy.

Since `spec.managedBy` is immutable, an external controller cannot hand the completed JobSet back to the built-in controller. Supporting TTL cleanup independently would preserve external ownership while making `ttlSecondsAfterFinished` effective and avoiding completed manager JobSets being retained indefinitely.

Related Kueue work: https://github.com/kubernetes-sigs/kueue/pull/14952

- [ ] Design doc
- [ ] API change
- [ ] Docs update

AI usage disclosure: OpenAI Codex was used to help investigate the behavior and draft this issue.

yindia:
#### What type of PR is this?

/kind feature

The built-in JobSet controller skips reconciling any JobSet that sets a custom `managedBy`, and that early return also skipped the `ttlSecondsAfterFinished` policy. As a result, MultiKueue manager-cluster JobSets were never deleted after they finished, even when a TTL was configured.

This change runs the TTL policy in the external-controller path once a terminal `Completed` or `Failed` condition is present. The manager JobSet is deleted after its TTL expires, while child Jobs and status remain the responsibility of the external controller. A JobSet without a TTL keeps the current behavior and is retained indefinitely.

One point for reviewers: this changes behavior for externally managed JobSets that were previously never deleted. Setting a TTL is an explicit request to clean the object up, so it is left unguarded here, but I am happy to put it behind a feature gate if that is preferred.

Fixes #1310

Added integration coverage for both paths: an externally managed JobSet is deleted after its TTL expires once a terminal condition is written, and a terminal externally managed JobSet with no TTL is retained.

```release-note
Externally managed JobSets (with a custom `managedBy`) are now cleaned up by `ttlSecondsAfterFinished` once the managing controller records a terminal condition.
```

## Summary by CodeRabbit

- **New Features**
  - Added TTL cleanup support for externally managed JobSets after the external controller records a terminal condition.
  - Externally managed JobSets with `Completed` or `Failed` status are now deleted when their configured TTL expires.
  - Externally managed JobSets without a TTL continue to be retained.

- **Documentation**
  - Clarified TTL cleanup behavior for externally managed JobSets across API, CRD, SDK, and reference documentation.

- **Tests**
  - Added coverage for TTL-based deletion and retention behavior.


kannon92:
This seems really good.

@kevin85421

can you take a quick look and make sure this solves the case you want?

kevin85421 (commented):
Thanks @yindia! The test looks good to me and is the behavior I am asking for.

PR #1319, opened 2026-09-04
docs: document activeDeadlineSeconds interaction with managedBy in KEP-1186

yindia:
#### What type of PR is this?

/kind documentation

#### What this PR does / why we need it:

Follow-up to the #1306 KEP review, where @kannon92 asked how `activeDeadlineSeconds` interacts with `spec.managedBy`.

Externally managed JobSets are skipped by the built-in controller before any `.status.startTime` maintenance or `activeDeadlineSeconds` enforcement, since both sit below the `managedByExternalController` early-return in `Reconcile`. The managing controller owns deadline enforcement, exactly as it already owns child-Job and status management; no extra gating code is needed.

Documents this in KEP-1186 under Notes/Constraints/Caveats, the design walkthrough, and the test plan. Docs-only; controller implementation and tests follow in a separate PR.

#### Which issue(s) this PR fixes:

Related to #1186

#### Special notes for your reviewer:

Docs-only KEP change.

#### Does this PR introduce a user-facing change?

```release-note
NONE
```

## Summary by CodeRabbit

- **Documentation**
  - Clarified how `activeDeadlineSeconds` interacts with externally managed JobSets.
  - Documented that external controllers are responsible for maintaining job start time and enforcing deadlines when they manage the JobSet.
  - Added guidance describing the behavior and limitations when a controller other than the built-in JobSet controller manages the JobSet.

- **Tests**
  - Added integration coverage for JobSets managed by external controllers.

PR #1320, opened 2026-09-05
Add activeDeadlineSeconds to bound JobSet active runtime

yindia:
#### What type of PR is this?

/kind feature
/kind api-change

Implements KEP-1186. Adds an optional `spec.activeDeadlineSeconds` that bounds how long a JobSet may be continuously active. When the deadline is exceeded, the controller marks the JobSet `Failed` with reason `DeadlineExceeded` and deletes its active child Jobs, matching `batch/v1.Job.spec.activeDeadlineSeconds`. This gives operators a hard upper bound so a wedged distributed run (stuck rendezvous, deadlocked collective, hung data loader) frees its accelerators instead of holding them until a human intervenes.

The timer is measured from a new `.status.startTime`, does not accrue while the JobSet is suspended, and resets on resume and on a global `RestartJobSet` restart. A single-Job `RestartJob` leaves it unchanged. The success policy is evaluated before the deadline, so a JobSet that completes is never failed by it; the deadline is evaluated before the failure policy, so an expired JobSet fails with `DeadlineExceeded` rather than being restarted.

Enforcement is behind the `JobSetActiveDeadlineSeconds` feature gate, alpha and off by default. `.status.startTime` is only maintained while the gate is enabled; when the gate is off the controller does not set or update it.

Part of #1186

AI disclosure: AI assistance was used to generate code documentation (godoc comments). The implementation and design decisions are my own and I have reviewed all changes.

The interaction with `spec.managedBy` was documented separately in #1319 and is validated here by an integration test: an externally managed JobSet gets neither `startTime` nor deadline enforcement, since the `managedByExternalController` early-return sits above both.

```release-note
Added an alpha `spec.activeDeadlineSeconds` field, gated by the `JobSetActiveDeadlineSeconds` feature gate (off by default), that fails a JobSet with reason `DeadlineExceeded` and deletes its active Jobs once it has been continuously active beyond the deadline. Also added `.status.startTime`, the JobSet active-start timestamp, maintained only while the gate is enabled.
```

## Summary by CodeRabbit

- **New Features**
  - Added optional active deadlines for JobSets to limit continuous active runtime.
  - JobSets exceeding their deadline are marked failed with `DeadlineExceeded`, and active child Jobs are removed.
  - Added active-period tracking with a visible `StartTime` column.
  - Deadlines pause while suspended and reset when resumed or globally restarted.
  - Added Python SDK support and Prometheus metrics.
- **Documentation**
  - Documented active deadline behavior and configuration.
- **Feature Gate**
  - Added an alpha feature gate, disabled by default.

### DIFF
--- a/api/jobset/v1alpha2/jobset_types.go
+++ b/api/jobset/v1alpha2/jobset_types.go
@@ -191,6 +191,20 @@ type JobSetSpec struct {
 	// +optional
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
 
+	// activeDeadlineSeconds is the maximum duration in seconds, relative to
+	// status.startTime, that the JobSet may be continuously active before the
+	// JobSet controller marks it Failed and deletes its active Jobs.
+	//
+	// The timer does not accrue while the JobSet is suspended: startTime is
+	// cleared on suspend and reset when the JobSet resumes. The value must be a
+	// positive integer. The field is mutable: operators may raise or lower the
+	// deadline on a running JobSet, matching batch/v1.Job. The controller
+	// recomputes the remaining time from status.startTime on the next reconcile,
+	// so no timer state is reset on update.
+	// +kubebuilder:validation:Minimum=1
+	// +optional
+	ActiveDeadlineSeconds *int64 `json:"activeDeadlineSeconds,omitempty"`
+
 	// volumeClaimPolicies is a list of policies for persistent volume claims that pods are allowed
 	// to reference. JobSet controller automatically adds the required volume claims to the
 	// pod template. Every claim in this list must have at least one matching (by name)
@@ -211,6 +225,16 @@ type JobSetStatus struct {
 	// +listType=map
 	// +listMapKey=type
 	Conditions []metav1.Condition `json:"conditions,omitempty" patchStrategy:"merge" patchMergeKey:"type"`
+
+	// startTime is the timestamp from which activeDeadlineSeconds is measured, and
+	// also serves as the JobSet's active-start time. It is set when the JobSet
+	// first becomes active (unsuspended), cleared when the JobSet is suspended, and
+	// reset to the current time when the JobSet resumes, matching
+	// batch/v1.Job.status.startTime. It is maintained only while the
+	// JobSetActiveDeadlineSeconds feature gate is enabled, independent of whether
+	// activeDeadlineSeconds is set.
+	// +optional
+	StartTime *metav1.Time `json:"startTime,omitempty"`
 
 	// restarts tracks the number of times the JobSet has been globally restarted.
 	// That is, restarts is the number of times the restart action RestartJobSet or RestartJobSetAndIgnoreMaxRestarts have been executed and led to the recreation of all Jobs.
@@ -309,6 +333,7 @@ type ReplicatedJobStatus struct {
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
 // +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="JobSet suspended"
+// +kubebuilder:printcolumn:name="StartTime",type=date,JSONPath=`.status.startTime`,description="Time the JobSet became active"
 // +kubebuilder:printcolumn:name="Age",JSONPath=".metadata.creationTimestamp",type=date,description="Time this JobSet was created"
 
 // JobSet is the Schema for the jobsets API
--- a/pkg/controllers/active_deadline_test.go
+++ b/pkg/controllers/active_deadline_test.go
@@ -1,0 +1,451 @@
+/*
+Copyright The Kubernetes Authors.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"fmt"
+	"math"
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus/testutil"
+	"github.com/stretchr/testify/require"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/klog/v2/ktesting"
+	clocktesting "k8s.io/utils/clock/testing"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+	"sigs.k8s.io/jobset/pkg/constants"
+	"sigs.k8s.io/jobset/pkg/features"
+	"sigs.k8s.io/jobset/pkg/metrics"
+	testutils "sigs.k8s.io/jobset/pkg/util/testing"
+)
+
+func TestSyncActiveDeadlineStartTime(t *testing.T) {
+	const (
+		jobSetName = "test-jobset"
+		ns         = "default"
+	)
+	now := metav1.Now()
+	earlier := metav1.NewTime(now.Add(-time.Hour))
+
+	tests := []struct {
+		name           string
+		jobset         *jobset.JobSet
+		expectStartSet bool // startTime should be non-nil after sync
+		expectCleared  bool // startTime was non-nil, expect nil after sync
+		expectChanged  bool // shouldUpdate expected
+		expectPreserve bool // startTime should keep its original value
+	}{
+		{
+			name:           "active jobset with no startTime gets one",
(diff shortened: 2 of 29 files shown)


pkg/controllers/active_deadline_test.go
```diff
@@ -0,0 +1,255 @@
...
+	"github.com/stretchr/testify/require"
+	apimeta "k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog/v2/ktesting"
+	clocktesting "k8s.io/utils/clock/testing"
+
+	jobset "sigs.k8s.io/jobset/api/jobset/v1alpha2"
+	"sigs.k8s.io/jobset/pkg/constants"
+	"sigs.k8s.io/jobset/pkg/features"
+	testutils "sigs.k8s.io/jobset/pkg/util/testing"
```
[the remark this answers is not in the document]
yindia: Deepcopy already generated and committed, `zz_generated.deepcopy.go` has `DeepCopyObject` for `JobSet/JobSetList` and handles `StartTime`. Build, vet, and tests pass, make generate gives no Go diff. Looks like a false positive

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -309,6 +332,7 @@ type ReplicatedJobStatus struct {
 // +kubebuilder:printcolumn:name="Restarts",JSONPath=".status.restarts",type=string,description="Number of restarts"
 // +kubebuilder:printcolumn:name="Completed",type="string",priority=0,JSONPath=".status.conditions[?(@.type==\"Completed\")].status"
 // +kubebuilder:printcolumn:name="Suspended",type="string",JSONPath=".spec.suspend",description="JobSet suspended"
+// +kubebuilder:printcolumn:name="StartTime",type=string,JSONPath=`.status.startTime`,description="Time the JobSet became active"
```
kannon92: vodersbbit had a suggestion worth checking out here.

https://github.com/kubernetes-sigs/jobset/pull/1320#pullrequestreview-5121370871

kannon92:
/assign

I'll take a look this week.

api/jobset/v1alpha2/jobset_types.go
```diff
@@ -191,6 +191,20 @@ type JobSetSpec struct {
 	// +optional
 	TTLSecondsAfterFinished *int32 `json:"ttlSecondsAfterFinished,omitempty"`
 
+	// activeDeadlineSeconds is the maximum duration in seconds, relative to
```
kannon92, line 194: can you add validation so that if someone sets this field when the feature gate is off the jobset is rejeccted?

pkg/controllers/jobset_controller.go
```diff
@@ -204,20 +205,67 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
 		setRestartingConditionFalse(js, constants.RestartingJobSetReasonJobsReady, constants.RestartingJobSetReasonJobsReadyMessage, updateStatusOpts)
 	}
 
-	// If any jobs have failed, execute the JobSet failure policy (if any).
-	if len(ownedJobs.failed) > 0 {
-		if err := executeFailurePolicy(ctx, js, ownedJobs, updateStatusOpts); err != nil {
-			log.Error(err, "executing failure policy")
+	// Maintain .status.startTime (the JobSet active-start time). This is general
```
kannon92: Please make sure that there are no sideeffects with this feature when the gate is disabled.

I see a lot of code no guarded under a feature gate.

AI review also calls this out:

>  ● 217 [correctness]   The JobSetActiveDeadlineSeconds gate reorders success-vs-failure policy evaluation cluster-wide, changing terminal outcomes for JobSets that never set activeDeadlineSeconds.

kannon92:
This looks good to me but I'd love one of your reviews also.

/cc @andreyvelich @jianqiaol

yindia:
@andreyvelich @jianqiaol do you have feedback ?

pkg/controllers/jobset_controller.go
```diff
@@ -204,20 +205,68 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
...
 	}
 
-	// If any jobs have failed, execute the JobSet failure policy (if any).
-	if len(ownedJobs.failed) > 0 {
-		if err := executeFailurePolicy(ctx, js, ownedJobs, updateStatusOpts); err != nil {
-			log.Error(err, "executing failure policy")
+	// Maintain .status.startTime (the JobSet active-start time and the anchor for
+	// spec.activeDeadlineSeconds). Self-gated: a no-op while the feature gate is off,
+	// so a JobSet carries no active-deadline status side effects when disabled.
+	r.syncActiveDeadlineStartTime(js, updateStatusOpts)
```
andreyvelich: Would it be better to check feature flag here before executing the `syncActiveDeadlineStartTime` function?

yindia: Moved the gate check to the call site so it's explicit: `syncActiveDeadlineStartTime` now only runs when `JobSetActiveDeadlineSeconds` is enabled, and dropped the redundant internal guard. `resetStartTimeOnGlobalRestart` stays self-gated since its no-deadline/else-branch caller is still reachable with the gate off.

pkg/controllers/jobset_controller.go
```diff
@@ -204,20 +205,71 @@ func (r *JobSetReconciler) reconcile(ctx context.Context, js *jobset.JobSet, upd
...
+	// The success -> deadline -> failure ordering applies only to JobSets that opt in
+	// via spec.activeDeadlineSeconds while the gate is on. Every other JobSet keeps the
+	// original failure -> success ordering, so enabling the cluster-wide gate never
+	// changes the terminal outcome of a JobSet that does not use the feature.
+	if activeDeadlineGateEnabled && js.Spec.ActiveDeadlineSeconds != nil {
+		// Success runs first so a completed JobSet always wins; the deadline runs
+		// before the failure policy so an expired JobSet fails with DeadlineExceeded
+		// instead of being restarted (which would reset or bypass the deadline).
+		if len(ownedJobs.successful) > 0 {
+			if completed := executeSuccessPolicy(js, ownedJobs, updateStatusOpts); completed {
```
andreyvelich, line 229: **Setting `activeDeadlineSeconds` flips a `Failed` JobSet to `Completed`, even when the deadline is nowhere near expiry.**

Hoisting `executeSuccessPolicy` above the failure policy applies unconditionally, not just on expiry. I ran this through your own `adlsReconcile` helper (2 replicas, `Any` success policy, one `JobComplete` child + one `JobFailed` child, `startTime` 1s ago):

| gate | `activeDeadlineSeconds` | terminal state |
|---|---|---|
| on | `3600` (not expired) | **`Completed`** |
| on | unset | `Failed` / `FailedJobs` |
| off | `3600` | `Failed` / `FailedJobs` |

So merely setting a 1-hour deadline converts a JobSet that would be `Failed` into one marked `Completed` — a user-visible semantic change with no relation to deadlines.

The comment just above correctly claims JobSets that don't use the feature are unaffected, and `TestReconcileNoDeadlineKeepsOriginalOrdering` pins that half. The half that *does* change behavior is untested.

Suggested fix: split expiry detection from the action, so nothing reorders until the deadline has actually expired. This keeps both KEP guarantees (success beats an expired deadline; the deadline beats the failure policy) without touching the ordering in the common case:

```go
armed, remaining := r.activeDeadlineRemaining(js)   // pure, no side effects
if armed && remaining <= 0 {
    // success still beats an expired deadline
    if len(ownedJobs.successful) > 0 && executeSuccessPolicy(...) { return }
    // ... fail with DeadlineExceeded, delete active Jobs
    return
}
deadlineRequeueAfter = remaining
// original failure -> success ordering, unchanged for everyone
```

Whichever way you resolve it, please add the `gate on / deadline live / both complete+failed children` row to a table test.

andreyvelich (commented):
Thanks for the thorough work on this — the KEP, the self-gating per function, the injected clock, and the integration coverage of suspend/resume and restart interactions are all in good shape.

One inline comment below. Disclosure: this review was performed with AI assistance.

pkg/controllers/jobset_controller.go
andreyvelich: Good point, let's ensure deadline and policies are working correctly together.

jianqiaol: +1 — and since this reorder came from my KEP review, to clarify the intent: it was scoped to the **already-expired** case ([#1306](https://github.com/kubernetes-sigs/jobset/pull/1306#discussion_r3921968500): *"If a child job fails while the deadline has already expired, ActiveDeadlinePolicy must evaluate before FailurePolicy"*).

pkg/features/features.go
```diff
@@ -59,6 +59,14 @@ const (
 	//
 	// Enables tracking and propagating JobSet execution attempts as a monotonic counter.
 	ExecutionAttemptsTracking featuregate.Feature = "ExecutionAttemptsTracking"
+
+	// owner: @yindia
+	// kep: https://github.com/kubernetes-sigs/jobset/blob/main/keps/1186-active-deadline-seconds/README.md
+	//
+	// Enables enforcement of the JobSet-level spec.activeDeadlineSeconds, which bounds the
+	// continuous active runtime of a JobSet before the controller marks it Failed and deletes
+	// its active Jobs. The .status.startTime field is maintained regardless of this gate.
```
jianqiaol, line 68: I think we planned to make startTime to be general in the KEP. I'm fine with the reversal and gate it with alpha gate, but let's update the descriptions relate to this. For example, now this field is not maintained when the feature gate is off.

site/content/en/docs/concepts/_index.md
```diff
@@ -176,6 +176,24 @@ to automatically restart the JobSet. A restart is done by recreating all child j
...
+active attempt, while `maxRestarts` bounds how many times the JobSet restarts.
+
+The timer is measured from `.status.startTime` and does not accrue while the
+JobSet is suspended: `startTime` is cleared on suspend and reset on resume and on
+a global restart (`RestartJobSet`), so a Kueue-managed JobSet is not penalized for
+queued time. The field is mutable; the controller recomputes the remaining time on
+each reconcile.
+
+This feature is alpha and gated by the `JobSetActiveDeadlineSeconds` feature gate
+(disabled by default). `.status.startTime` is maintained regardless of the gate.
```
jianqiaol: Here and the PR description (including the release-note block) both still say startTime is maintained regardless of the gate. Let's update these to make them consistent with the implementation.

jianqiaol (commented):
Thanks for the update! Just suggested a minor document update for startTime. Overall this looks good to me.

yindia:
@jianqiaol i think we can ignore the netlify checks

PR #1326, opened 2026-09-08
docs: link to k8s-CronJobSet for cron scheduling

### LINKED ISSUE #1203: CronJobSet

It would be great if there was an equivalent to a Kubernetes CronJob. So JobSets can be scheduled regularly, with the same advantages that JobSets have over Jobs. 

**Why is this needed**:

Maybe I'm barking up the wrong tree, but more expressive cron jobs would be nice. Specifically I want to run regular backups on a cron schedule (like backing up gmail, or 1password every night), and then I'd like to snapshot the volumes, and maybe do other things. This seems like the perfect use-case for JobSets, as it allows the tasks to be defined in one place, run in parallel, and have some dependency ordering. Without a CronJobSet abstraction, this extra work needs to be built into the running container itself with scripts.

- [ ] Design doc
- [ ] API change
- [ ] Docs update

yindia:
#### What type of PR is this?

/kind documentation

Adds a "Related projects" note on the overview page pointing to the community k8s-CronJobSet project, for users who want to run JobSets on a cron schedule. This follows the discussion in the issue, where maintainers preferred linking to the existing implementation rather than building CronJobSet into the repo.

Fixes #1203

Happy to move the link to a different page if there is a preferred spot.

```release-note
NONE
```

### DIFF
--- a/site/content/en/docs/overview/_index.md
+++ b/site/content/en/docs/overview/_index.md
@@ -74,6 +74,10 @@ See our github project for our [roadmap](https://github.com/orgs/kubernetes-sigs
 See the [troubleshooting](https://jobset.sigs.k8s.io/docs/troubleshooting/) guide for help resolving common issues.
 
 
+## Related projects
+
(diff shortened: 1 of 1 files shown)


kannon92 (commented):
/lgtm
/approve

/ok-to-test

Thank you!

